PR #508, opened 2014-07-11
Do not hook exit, use PTRACE_O_TRACEEXIT.

batrick:
This is also the first step to handling execve correctly.


toiletrollen:
EULA 508

PR #641, opened 2014-12-10
Work Queue Container Support

charleszheng44:
add --run-in-docker option for work_queue_worker, it enable to run work_queue_worker process inside a docker container


dthain:
@charleszheng44 let's get this cleaned up now so that other people can use it.

Please take a pass through and remove al lthe extraneous comments and printfs.

Also, I see that work_queue_process_execute and work_queue_process_execute_container share a lot of similar code.  Find a way to avoid that duplication.

charleszheng44:
@dthain  Several arguments, like container_mode and image_name need to be passed into the 
work_queue_process_execute_container function. 

What about a generic function work_queue_process_execute(struct work_queue_process *p, ...) using argument list?

dthain:
Right, pass those things as args and leave them null if irrelevant to the container mode.

dthain:
This pull request seems to be mixing two features.
Best thing to do is restart from a clean branch, apply code changes and make pull request.

btovar:
Closing, given last comment.

PR #658, opened 2015-01-12
Add `/tmp /tmp` into common-mountlist

hmeng-19:
Allow the filesystem created by `parrot_package_run` to use /tmp from the host machine.

### DIFF
--- a/parrot/src/parrot_package_create.c
+++ b/parrot/src/parrot_package_create.c
@@ -534,6 +534,7 @@ int post_process( ) {
 		(fputs("/proc /proc\n", file) == EOF) ||
 		(fputs("/sys /sys\n", file) == EOF) ||
 		(fputs("/var /var\n", file) == EOF) ||
+		(fputs("/tmp /tmp\n", file) == EOF) ||
 		(fputs("/selinux /selinux\n", file) == EOF)) {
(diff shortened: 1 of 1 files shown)


dthain:
Hmm, are you sure about this -- do you really want the preserved package to see whatever mess was left  behind by other applications in /tmp?

hmeng-19:
As for the file dependencies from /tmp dir, currently parrot_package_create
just ignores them (I assume the files from /tmp is not important). However,
the application itself may require /tmp exists in the root filesystem of
the sandbox, for example,  /cvmfs/
grid.cern.ch/3.2.11-1/etc/profile.d/grid-env.csh has a statement:
set mycshtmpfile=`mktemp /tmp/gridenv.XXXXXX`, which needs to access /tmp
dir.
Adding `/tmp /tmp` into common-mountlist is guarantee the application can
always access /tmp dir, and the package can get rid of /tmp dir.

dthain:
Of course, I understand that the application must have a /tmp in its root dir.
The question is, do you want /tmp shared with everyone else on the machine?

hmeng-19:
You are right. Sharing /tmp is not a good idea. Creating a tmp directory in
the package manually is a better choice.

PR #661, opened 2015-01-13
Use the interpreter as the logical_exe.

### LINKED ISSUE #660: Parrot does not correctly handle interpreted executables with the -l option
When Parrot is used with an interpreted executable and the loader redirection option `-l`, it passes the interpreted executable directly to the loader as its first argument. This is not correct. The loader complains it cannot load the file:

````
$ cat abc.sh 
#!/bin/sh
echo 'haiyan'
$ ./parrot_run -l /usr/lib/ld-linux-x86-64.so.2 -- ./abc.sh 
2015/01/13 12:09:00.93 parrot_run[2482] notice: parrot_run 4.4.0 has not been tested on Linux 3.17.6-1-ARCH yet, this may not work
./abc.sh: error while loading shared libraries: ./abc.sh: file too short
````

Instead, Parrot should pass the interpreter (`/bin/sh`) as the first argument and the interpreted executable (`./abc.sh`) as the second argument.

Credit: @hmeng-19 found and isolated the problem.

batrick:
The loader must be given a real executable to load, not an interpreted one.

Fixes #660.

### DIFF
--- a/parrot/src/pfs_dispatch.cc
+++ b/parrot/src/pfs_dispatch.cc
@@ -1023,10 +1023,10 @@ static void decode_execve( struct pfs_process *p, int entering, INT64_T syscall,
 				goto failure;
 
 			if (strlen(interp_arg)) {
-				if (fix_execve(p, old_user_argv, physical_name, logical_name, interp_exe, interp_arg, logical_name) == -1)
+				if (fix_execve(p, old_user_argv, physical_name, interp_exe, interp_exe, interp_arg, logical_name) == -1)
 					goto failure;
 			} else {
-				if (fix_execve(p, old_user_argv, physical_name, logical_name, interp_exe, logical_name, NULL) == -1)
+				if (fix_execve(p, old_user_argv, physical_name, interp_exe, interp_exe, logical_name, NULL) == -1)
 					goto failure;
 			}
 		} else {
--- a/parrot/src/pfs_dispatch64.cc
+++ b/parrot/src/pfs_dispatch64.cc
@@ -826,10 +826,10 @@ static void decode_execve( struct pfs_process *p, int entering, INT64_T syscall,
 				goto failure;
 
 			if (strlen(interp_arg)) {
-				if (fix_execve(p, old_user_argv, physical_name, logical_name, interp_exe, interp_arg, logical_name) == -1)
+				if (fix_execve(p, old_user_argv, physical_name, interp_exe, interp_exe, interp_arg, logical_name) == -1)
 					goto failure;
 			} else {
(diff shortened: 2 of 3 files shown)


dthain:
This seems to invert the sense of the previous commit 881a58917ff38a349dc50f17d52608b19b4b7882, is that correct?

batrick:
No, that commit was solving a different problem (and still does solve it). The issue there was we need ld-linux to load the logical exe rather than the physical exe, otherwise it does not set `argv[0]` correctly.

Unfortunately, 881a589 introduces a different problem where ld-linux can't load an interpreted exe (because we now give it the logical exe, the shell script path). Here's an example that should clear it up:

````
$ ./parrot_run --work-dir=/chirp/localhost:10000 /bin/sh -c 'ls -l; cat a.py'
total 792
-rwx------ 1 batrick users     71 Jan 13 15:33 a.py
-rwx------ 1 batrick users  10400 Jan 13 15:49 python
-rwx------ 1 batrick users 787584 Jan 13 16:09 sh
#!/chirp/localhost:10000/python

import sys

print(' '.join(sys.argv))
````

Note that `a.py` references a Python on Chirp. So that must be copied to the Parrot cache:

For the parent commit (851b34) of 881a589:

````
$ ./parrot_run -l /usr/lib/ld-linux-x86-64.so.2 -- /chirp/localhost:10000/a.py 1 2
/chirp/localhost:10000/a.py 1 2
$ ./parrot_run -l /usr/lib/ld-linux-x86-64.so.2 -- /chirp/localhost:10000/sh -c 'echo "$0" "$@"'
/tmp/parrot.1000/a9/a9af73e481fddfdb5ffa04429f8d0e00
````

For 881a589:

````
$ ./parrot_run -l /usr/lib/ld-linux-x86-64.so.2 -- /chirp/localhost:10000/a.py 1 2
/chirp/localhost:10000/a.py: error while loading shared libraries: /chirp/localhost:10000/a.py: invalid ELF header
$ ./parrot_run -l /usr/lib/ld-linux-x86-64.so.2 -- /chirp/localhost:10000/sh -c 'echo "$0" "$@"'
/chirp/localhost:10000/sh
````

For parrot-ld-interp:

````
$ ./parrot_run -l /usr/lib/ld-linux-x86-64.so.2 -- /chirp/localhost:10000/a.py 1 2
/chirp/localhost:10000/a.py 1 2
$ ./parrot_run -l /usr/lib/ld-linux-x86-64.so.2 -- /chirp/localhost:10000/sh -c 'echo "$0" "$@"'
/chirp/localhost:10000/sh
````

dthain:
Ok, just making sure you are on top of it.
Please add a suitable test case.

PR #667, opened 2015-01-28
PRUNE

pivie:
The basic PRUNE commands working (PUT, GET, USE, EVAL, GO are working plus a few more STATUS, LOCATE, and RUN). Umbrella integration is not working in this version.


dthain:
Ok, a good start.

A couple of items for simplification:

- Can you replace your custom readline routine with GNU readline or something similar?
- I see that you have a background thread to do the processing, can you do it synchronously within the GO command instead?
- Please include the complete text for some simple functions (e.g. sort) within the README.
- Join all of "prunespace" into a single directory and require that it be given on the command line.  e.g. "prune /home/pivie/prune" such that /home/pivie/prune contains a directory for the database and a directory for the file blobs.  That way, a single prunespace has a single name and can easily be be named, compared, copied, etc.

pivie:
@dthain 
Okay I switched to a readline library and eliminated the processing thread.
I added files in the test folder to support an example in the readme for how to do part of a merge sort.
Instead of specifying a folder as a location to store everything, I add a config file to the working directory. The config file specifies where each type of data gets stored, database, data files, and sandbox directories. The user gets prompted to fill in this data the first time prune is run in a given directory. I hope this satisfies the need expressed.

PR #670, opened 2015-02-17
make_cache_name called one time per file/directory

btovar:
@dthain @nhazekam

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -896,13 +896,23 @@ including a hash and a name where one is known, or another
 unique identifier where no name is available.
 */
 
-char * make_cached_name( struct work_queue_task *t, struct work_queue_file *f )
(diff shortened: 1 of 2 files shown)


dthain:
Looks like you are changing it from a stateless function to one that adds state to the file object.
In that case, be sure to make the function idempotent by checking to see if it has already been generated, and then to free the state when the file object goes away.

btovar:
make_cache_name is now only called inside work_queue_file_create.

dthain:
Unless I am missing something, cached_name isn't freed.
I think it should be cleaned up in work_queue_task_delete

btovar:
Correct, I'll fix it.

dthain:
Or, to make things even clearer, there should be a work_queue_task_file_delete to mirror w_q_t_f_create

btovar:
Testing still, do not merge.

btovar:
Done. Added wq_file_delete too.

PR #673, opened 2015-02-22
Added cytoscape output option for makeflow_viz

vhawley:
If display mode is set to "cytoscape",  makeflow_viz will generate a
file (cytoscape.xgmml) of the dag in memory that can be visualized in
Cytoscape (tested on version 3.2)


dthain:
@btovar please review.

PR #674, opened 2015-02-28
wq task status

btovar:
Keep status of a task explicitly.

Adds q.task_status(taskid) (python) or q->task_status(taskid) (perl).

Status is one of WORK_QUEUE_TASK(UNKNOWN|READY|RUNNING|RESULTS|RETRIEVED|DONE|CANCELED).

Status change is now consolidated in one function. In the future we may want to reduce the number of structures keeping state.


btovar:
Debugging. Do not merge yet.

btovar:
Renamed status to state, so we do not confuse it with exit status.

dthain:
I like how you have the task state change in one functin, that really clarifies how the data structures are used.  Even better if you can synchronize the names of the data structures with the states.

CANCELLED has two Ls, I think.

btovar:
It seems it is one of those words.. cancelled is British, canceled is American. Does wq wear a red, or a blue coat?

dthain:
I didn't realize this was going to get political.  :)

PR #685, opened 2015-03-13
Disk broken promise

btovar:
Workers disconnect when their workspace uses more disk than the specified --disk option.


dthain:
Sorry it took so long to look at this.  I think this is a step in the right direction, and I would like to see things develop towards tracking usage of each task independently.

PR #686, opened 2015-03-16
Link against uuid if libcvmfs is used

jblomer:
As of the upcoming cvmfs 2.1.20, cvmfs requires libuuid.  This patch adds the linker flag to the parrot build if it is configured with cvmfs support.

### DIFF
--- a/configure
+++ b/configure
@@ -644,6 +644,14 @@ then
 		echo "*** Couldn't find $library in ${cvmfs_path}"
 		exit 1
 	fi
+  
+	if library_search uuid "/usr"
+	then
+		cvmfs_ldflags="${cvmfs_ldflags} ${library_search_result}"
+	else
+		echo "*** Couldn't find libuuid"
+		exit 1
+	fi
 else
 	if [ $config_cvmfs_path = yes ]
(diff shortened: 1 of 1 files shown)


batrick:
@btovar is this needed for for 4.3.4? I think this just needs a check for older versions of cvmfs. Unfortunately they don't specify a version number in their header files...

batrick:
@btovar I meant to say that the only problem with this PR is that configure will fail for older versions of cvmfs, because libuuid may not be available (and is not needed). I think we need a check to see if cvmfs actually needs libuuid?

btovar:
It is, thanks. I was looking for it.

btovar:
Yes, checking for a way to do that now.

jblomer:
@batrick better versioning is actually something we can do.  Do you mean something like `#define LIBCVMFS_FULL_VERSION 2.1.20` in libcvmfs.h?

btovar:
Yes, that would work. Right now I am using  LIBCVMFS_VERSION >= 2 to link against uuid, even cases when it might not be necessary ( < 2.1.20).

PR #694, opened 2015-03-26
Change cvmfs symlink error to ENOENT.

btovar:
As reported by Brian Bockelman. EXDEV is only valid for errors regarding
hard links.

### DIFF
--- a/parrot/src/pfs_service_cvmfs.cc
+++ b/parrot/src/pfs_service_cvmfs.cc
@@ -942,7 +942,7 @@ static bool path_expand_symlink(struct pfs_name *path, struct pfs_name *xpath)
 
 			if(sscanf(link_target, "/cvmfs/%[^/]%[^\n]", xpath->host, path_head) < 1)
 			{
-				errno = EXDEV;
+				errno = ENOENT;
 				return false;
 			}
 


batrick:
What is this code checking for?

btovar:
What do you mean?

batrick:
Nevermind, I think I understand.

dthain:
I am concerned that this change will make it harder for the user to debug a problem.

It is true that Unix only uses EXDEV for hard links, but the message is there to indicate that a link cannot be used in the desired way.  We are doing the same thing here, to clearly indicate that the link traversal is not supported outside of the current driver.  When this happens, the user then clearly knows that there is a problem with a link.

If it is necessary to revert this to a more benign errno, then we need to emit a detailed message to the user to explain what's going on.

batrick:
Another option is EIO which virtually all IO syscalls allow. That would probably be about as disruptive (i.e. the application cannot know how to handle that and should bail) and the user will be even more likely to look at the Parrot debug log. I think adding a debug message is a good idea too.

btovar:
I like the option of ENOENT and a message. An application may then be able to recover, as ENOENT is an expected error code.

Brian did mention that in one of their applications, EXDEV did cause a problem.

bbockelm:
In particular, the code that causes us heartache is boost::filesystem:

https://github.com/boostorg/filesystem/blob/cb24d1cdb52bd26bc2d02a57e07ab920f2911d0e/src/operations.cpp#L401

We iterate through directories listed in an environment behavior which behaves like $PATH using boost::filesystem; it's OK / ignorable if the stat results in ENOENT or ENOTDIR (not_found_error above returns true).  EIO would cause a C++ exception to be thrown.

batrick:
I added a debug comment in cff760037c6100b6aea260375854c4abcebd7a0e. I'm considering this resolved unless there is further comment.

PR #697, opened 2015-03-30
Fix segfault from trying to split failed mapping.

batrick:


batrick:
@btovar this needs backported to the next stable release. I think this fixes #716.

btovar:
On it.

btovar:
Included now in release candidate 4.4.0rc3

PR #699, opened 2015-03-31
Add Mac OS X installation note.

pivie:
See issue #691


batrick:
@pivie, our configure script also needs to be updated to detect this problem. You can test for the header using `optional_include` and for the operating system using the configure variable `$build_sys`.

pivie:
I must confess I am not familiar with how optional_include works in the configure file. It appears to look for specific files in /usr/include/ and attach compiler flags based on the existence of those files. I will point out that the problem is not that the one of the /usr/include/ is not there, but that the entire folder is missing. Will an optional_include work with "./", and if so, what flag would be appropriate?

I have added a PR to give a more informative error message. I have tested it as far as I can without a mac that doesn't have the command line tools installed yet.

p.s. I may have messed up the merge before the this last commit. Let me know if I need to do something to fix it.

batrick:
@pivie stop by 222 and we can talk about it. Bring your laptop.

batrick:
Superseded by #711.

PR #723, opened 2015-04-23
Link cleanup

batrick:
Various cleanup of the link library.


dthain:
I don't recall seeing LINK_TYPE_{STANDARD|FILE} before.
Any idea what's going on there?

batrick:
Seems that was added in 0b8cd0b0d9b9640b38e595ab408428ccefe92e2e.

PR #728, opened 2015-04-27
Use process_vm_* if available.

batrick:


dthain:
What platforms can we effectively test this?

batrick:
process_vm_readv/writev were introduced in Linux 3.2. So RHEL7 and its derivatives.

Issue #672, opened 2015-02-17
tests succeed when makeflow segfaults


btovar:
on make test, if makeflow segfaults, the test report success.

dthain:
This would be a good one for vhawley to look at, since he is already poking at Makeflow.

dthain:
Pull #732

PR #732, opened 2015-05-01
Test Framework Should Detect Failures

dthain:
The test framework was not detecting failures in the tests!
Now it does. And we have a few latent failures. (To be addressed in later pulls.)


dthain:
Bug #672

PR #734, opened 2015-05-01
optimization of Allpairs in the symmetric case

### LINKED ISSUE #731: optimization of Allpairs in the symmetric case
Currently, `Allpairs(setA,setB, Func)` applies `Func` on each pair of A[i] and B[j]. If the length of A is m, and the length of B is n, `Func` will execute `m x n` times totally.

Problem:
In the case of A=B (i.e., m=n, A[i] = B[i]), `Func(A[i], B[j])` and `Func(A[j], B[i])` are duplicate. Totally nearly half of the executions of Func are unnecessary. 
We need to optimize Allpairs to avoid doing the same computing twice.

Solution:
If i<=j, do Func(A[i], B[j]); if i>j, do nothing.

Potential Problem:
`allpairs_master` can set the width and height of one work unit. The modification of `allpairs_master` should try to avoid splitting work units unevenly.

hmeng-19:
add --symmetric option to allparis_multicore and allpairs_master
	this option allows allpairs compute only half of a symmetric matrix

Add --index option into allpairs_multicore
	this option allows the user to specify the indexes of a matrix

This branch tries to fix #731

### DIFF
--- a/allpairs/src/allpairs_master.c
+++ b/allpairs/src/allpairs_master.c
@@ -9,6 +9,7 @@ See the file COPYING for details.
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
+#include <limits.h>
 
 #include "cctools.h"
 #include "debug.h"
@@ -47,6 +48,11 @@ static int xblock = 0;
 static int yblock = 0;
 static int xstop = 0;
 static int ystop = 0;
+static int is_symmetric = 0;
+
+enum {
+	LONG_OPT_SYMMETRIC=UCHAR_MAX+1,
+};
 
 static void show_help(const char *cmd)
 {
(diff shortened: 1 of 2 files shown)


dthain:
Looks pretty good, except one more piece needed:

In the master, skip generating a sub-task if it is completely on the "wrong" side of the diagonal.  As written, the sub task will be generated and do no work (which is ok) but it will still result in sending all the input files to the empty task.

hmeng-19:
Good point.

hmeng-19:
@dthain , I added a new commit to fix the problem you pointed out. 
Please recheck it.

dthain:
Overall, looks good.
You could simplify the option parsing by using something like `n=sscanf(optarg,"%d,%d",&x,&y)` in main.

hmeng-19:
The only drawback of using `sscanf` is that I have no way to judge whether the option is valid or not. If the user set --index to be `3 6 8`, I want to tell him that his option is wrong.

hmeng-19:
BTW, `string_split` may be a little more expensive. But I think it is a good fit here. We only call it once.

dthain:
sscanf returns the number of fields matched, so that you can check.  Performance is not my concern, I am simply showing you how to make your C code even more compact and clear.

hmeng-19:
You are right. I even forgot the return value of `sscanf` (shameful...). It seems I really need to take your compiler class to review C.

hmeng-19:
@dthain , it is there. Hope no more bug. :)

dthain:
Beautiful.  :)  Merge it.

allpairs/src/allpairs_master.c
```diff
@@ -64,6 +71,7 @@ static void show_help(const char *cmd)
 	fprintf(stdout, " %-30s Set the project name to <project>\n", "-N,--project-name=<project>");
 	fprintf(stdout, " %-30s Priority. Higher the value, higher the priority.\n", "-P,--priority=<integer>");
 	fprintf(stdout, " %-30s Enable debugging for this subsystem.  (Try -d all to start.)\n", "-d,--debug=<flag>");
+	fprintf(stdout, " %-30s Compute half of a symmetric matrix.\n", "   --symmetric");
```
btovar: Is it a matrix? 
If not, I rather have --commutative, to mean that F(a,b) = F(b,a).

hmeng-19:
@btovar , this is a symmetric matrix and F(a, b) = F(b, a).
I think `--symmetric` is a good fit here.
Thanks for your suggestion. :)

PR #736, opened 2015-05-02
Change the structure of the cctools binary dir

hmeng-19:
The changes are:
mv cctools/doc to cctools/share/doc/cctools
mv config.mk to cctools/share/doc/cctools/config.mk
remove cctools/etc
mv cctools/lib/* to cctools/lib/cctools/* (except for cctools/lib/perl5 and cctools/lib/python*)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,11 +32,10 @@ $(INSTALL_PACKAGES): $(CCTOOLS_PACKAGES)
 install: $(INSTALL_PACKAGES)
 	mkdir -p $(CCTOOLS_INSTALL_DIR)/bin
 	for file in $(CYGWINLIB) ; do if [ -f /bin/$$file ] ; then cp /bin/$$file $(CCTOOLS_INSTALL_DIR)/bin/ ; fi ; done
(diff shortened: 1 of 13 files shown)


dthain:
The Makefile is the wrong place to make these changes.

If a user does a simple `configure --prefix X` then we want the installed package to have the conventional directories of X/bin, X/doc, X/etc and so forth.  

If RPMs require these directories to be different, then the changes should be made by the RPM build process *after* `make install`

hmeng-19:
@dthain , once we agreed that the structure of our rpm follows the unix filesystem hierarchy standard, I did not see why not change the binary code of cctools?

dthain:
As we discussed today, please get together with @btovar and @pivie to discuss this.

The main idea is that `make install` should do a simple cross-platform installation, since we expect it to work on Linux, Windows, MacOS, Solaris, or whatever comes next.  If there are locations that are specific to Linux RPMs, OSX portfiles, etc, then those locations should be arranged by the RPM, portfile, or other platform specific tool.

btovar:
We'll keep discussing on our list.

PR #740, opened 2015-05-05
WQ Viz and Amazon AMI

rboccabe:
Adding the work queue visualization tool. Some things from the Amazon AWS code at the beginning of the semester is present too.


dthain:
@btovar is going to split the pull request

PR #743, opened 2015-05-06
Parrot exec etxtbsy

batrick:


batrick:
@btovar please take a look.

btovar:
Great! Thanks Patrick.

Issue #590, opened 2014-11-03
Add extra arguments to workers in submission scripts.


btovar:
For example, to add --run-in-docker:

condor_submit_workers  -E "--run-in-docker" ....

PR #745, opened 2015-05-06
Worker flags

btovar:
Some additions to the submission scripts.

In particular, the -E flag, to directly pass arguments to a worker.


btovar:
Fixes #590.

dthain:
Looks good.

PR #748, opened 2015-05-11
Makeflow Container Support

charleszheng44:
For makeflow CONTAINER_DOCKER_MODE, generate a global wrapper script (docker.wrapper.sh) when the makeflow starts, and each task use this general wrapper to execute the job with a docker 
container.

The general process is
~~~
1. makeflow_run( struct dag *d ) //launch makeflow
2. makeflow_create_docker_sh() //create docker.wrapper.sh
3. list_push_tail(wrapper_input_files, dag_file_create(CONTAINER_SH)) //put global script into the list of the wrapper input files
3. makeflow_node_submit(struct dag *d, struct dag_node *n) //start task
4. task_sh = makeflow_create_task_sh(const char* command, int nodeid) //create task command shell script
5. input_files = string_combine(input_files, task_sh) // put task_sh into the list of input files
6. string_wrap_command(task_sh, wrapper_command) //combine global wrapper and local scripts
~~~


dthain:
First thing -- your code is failing in autobuild, please use `configure --strict` to enable all compiler warnings before you check in and push.

makeflow/src/makeflow.c
```diff
@@ -144,6 +144,11 @@ static void makeflow_wrapper_add_command( const char *cmd )
 	}
 }
 
+/* XXX 
+ * makeflow_wrapper_add_input_file() and
+ * makeflow_wrapper_add_output_file() do not work.
+ * replace "optarg" by "file" ?
+ */
```
dthain: What exactly doesn't work?

dthain: Ah, there is a bug in `wrapper_add_output_file`, it should be manipulating `wrapper_input_files` instead of `wrapper_output_files`, you can fix that.

dthain: Oh, you are right -- use `file` instead of `optarg`.

makeflow/src/makeflow.c
```diff
@@ -615,6 +620,56 @@ static batch_job_id_t makeflow_node_submit_retry( struct batch_queue *queue, con
...
+curr_dir=`pwd`\n\
+default_dir=/root/worker\n\
+flock /tmp/lockfile /usr/bin/docker pull %s\n\
+docker run --rm -m 1g -v $curr_dir:$default_dir -w $default_dir %s $default_dir/$@\n", container_image, container_image);
+
+  	fclose(wrapper_fn);
+
+	chmod(CONTAINER_SH, 0755);   
+
+    if(!wrapper_input_files) wrapper_input_files = list_create();
```
dthain: Use `wrapper_add_input_file` here.

makeflow/src/makeflow.c
```diff
@@ -978,6 +1007,19 @@ static void makeflow_run( struct dag *d )
...
+   
+    if (container_mode == CONTAINER_MODE_DOCKER) {
+    /* XXX 1. for the workflow if docker mode is on 
+     * 1) create a global script for running docker container
+     * 2) add this script to the global wrapper list
+     */
+        makeflow_create_docker_sh();
+        
+        //makeflow_wrapper_add_input_file(CONTAINER_SH);
+        char global_cmd[64];
```
dthain: Why use sprintf here?  Just pass CONTAINER_SH as the argument.

makeflow/src/makeflow.c
```diff
@@ -1702,11 +1749,11 @@ int main(int argc, char *argv[])
 
 	/* XXX better to write created files to log, then delete those listed in log. */
 
-	if (container_mode == CONTAINER_MODE_DOCKER) {
+	/*if (container_mode == CONTAINER_MODE_DOCKER) {
```
dthain: Don't commit commented-out code, just delete it.

dthain:
The code is getting better, but why is it necessary to create a helper script with each task that is submitted?
Can you figure out a way for the master wrapper script to do the job itself?

charleszheng44:
The problem is that if there are special symbols in the command (like, ">>", ">"), docker run command cannot read it. 

For example 
~~~
docker run ubuntu command >output_file 2>error_file 
~~~
does not work.

To solve this problem, 
"command >output_file 2>error_file " has to be written into a shell script, say tmp.sh
then run
~~~
docker run ubuntu ./tmp.sh
~~~

btovar:
@charleszheng44, it is better if we use sh -c, like this:

 docker run someimg sh -c "ls > out; cat out"

charleszheng44:
I am going to try this out.

dthain:
FYI, the wrapper bug were my dumb mistake, please pull master to get the fix.

charleszheng44:
The autobuild failed on centos5 and centos7. I do not understand what is the cause of the failure. 

Following is log files

centos5
http://ccl.cse.nd.edu/a/a6aff215/a6aff215.x86_64-centos5.log

centos7
http://ccl.cse.nd.edu/a/a6aff215/a6aff215.x86_64-centos7.log

Do you have any thoughts?

btovar:
Docker images were not being updated. It should work now.

dthain:
- Better, but still not right, I want you to add the wrapper *once* in `main` and have no special code in `makeflow_submit_node`.
- `load_from_tar` should be `docker_tar` so that it is clear the option relates to docker.
- The options must be documented briefly in `makeflow.m4` and `makeflow.html`

charleszheng44:
It should work as expected now.

dthain:
What is the most complex makeflow that you have tested this with?

dthain:
Don't you need an `sh -c {}` in there somewhere in order to handle the case of shell redirections, as you pointed out above?

charleszheng44:
I tried to put "sh -c" in the global shell script. It does not work as expected. But without "sh -c", both BWA workflow and makeflow example program works well.

Another thing I tried is make "sh -c" a part of global wrapper script, something like "docker.wrapper.sh "sh -c {}"", this one works as well, but I am not sure is it still necessary.

dthain:
Ok, I'll take this one as is.  Keep chipping away at your other outstanding pull requests, please.

PR #756, opened 2015-05-14
Master now checks if there is space to receive file.

nhazekam:
This functionality existed in the fact that if a file was to big it would transfer until it failed and then realize that the lengths were not the same. It would report that the files were different length and not the cause of the issue. This preemptively checks and fails if won't fit.

### DIFF
--- a/dttools/src/disk_info.c
+++ b/dttools/src/disk_info.c
@@ -6,6 +6,8 @@ See the file COPYING for details.
 */
 
 #include "disk_info.h"
+#include "debug.h"
+#include "macros.h"
 
 #include <sys/types.h>
 #include <sys/param.h>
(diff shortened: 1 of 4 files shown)


nhazekam:
Two questions: 
Should this attempt to check somewhere other than the current directory, aka "."? I suppose the path could resolve outside of the current area, which would make this space calculation erroneous.

The function used to determine space now exists in both work_queue.c and work_queue_worker.c. Should this be moved to a central location? Ben had suggested work_queue_internal.h, but I did not move it as it is static and uses a global variable. Opinions/suggestions on this.

dthain:
1 - Check the space available in the directory where the file is to be placed.

2 - Common general purpose code should go into libdttools.a  (e.g. get_disk_info)

dthain:
@btovar what do you think of the appropriate failure status for this case?  Is this worker failure, app failure, resource exhaustion, or something else?

nhazekam:
Personally I think it fits under resource exhaustion best. However, I chose
APP_FAILURE over WORKER_FAILURE because, to my understanding, an app
failure does not retry the task, while a worker failure does. I did not see
a resource exhaustion failure option.

btovar:
As we have been talking about resources, I would mark it as something else. The worker did its job correctly, the app logic worked correctly, and resource exhaustion (as it is now) refers only to running task on a worker.

It would be nice to have the option for the task to be in a holding pattern, until there is space at the master (e.g., makeflow's gc).

nhazekam:
I moved two auxiliary functions to dttools, and made the debug messages D_DEBUG. I also updated the function so the directory is passed.

dthain:
Perhaps I wasn't clear in my comment before.

If I understand correctly, `cwd_disk_info_get` traverses the filesystem to check for available space, which is a very expensive operation.  This will result in drastic consequences if we do it every time that a file comes back.  Instead, just use `disk_info_get` which will do a `statfs` to determine the space available in the volume corresponding to the given directory.

nhazekam:
The function check_disk_space_for_filesize, which is used most of the time
in work_queue_worker.c and the only one used in work_queue.c uses
disk_info_get, not cwd_disk_info_get. cwd_disk_info_get was moved in case
it is used in the future (at the request of @btovar). I use
check_disk_space_for_filesize because it has a debug status and also uses a
threshold to leave some extra space.

dthain:
Ok, I see now.

The code looks good to me, but  I will wait for @btovar to approve.

btovar:
Nick, looks good. Is this ready to merge?

nhazekam:
Ya, should be good to go.

btovar:
Merged in 71619bd925a238c1e3cc2bd0d807560db0c4805c after fixing whitespace.

Issue #722, opened 2015-04-23
make macports for cctools


dthain:
Create a macports distribution of cctools.
First figure out how to write a portfile and test it locally.
Then, get it integrated into the macports distribution.

https://guide.macports.org/chunked/development.html

pivie:
While I haven't done this before, this is probably up my alley if you want
me to work on it.

dthain:
Go for it

pivie:
I almost have this working. I have two questions:

1- It seems that "cctools" is already taken. I propose "ndcctools" as an
alternative package name. Any thoughts?

2- Macports doesn't like things in the ${prefix}/doc folder, it "violates
the layout of the ports-filesystems!". I am having a hard time fixing the
problem in the portfile, but it is probably possible with some more effort.
Is the doc folder a common approach, rather than using
${prefix}/share/doc/${package}? Or do we just do that because we don't
expect to integrate with other systems.

btovar:
For 2, it seems in line with @hmeng-19 modifications for rpms.

dthain:
No problem with `ndcctools` but do a little investigation (with @hmeng-19) and find a package name that is unused across rpm, macports, deb, etc so that we can pick one name and be done with it.

`share/doc` seems reasonable to me, but do it by changing the macports build procedure, not our Makefile.

pivie:
I finally got this working. Now I just need to submit it to the official MacPorts repository. I am posting it below as a reference for anyone interested and as a backup since the file doesn't really belong in the repository. @dthain let me know if the descriptions are satisfactory.

# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
# $Id$

PortSystem              1.0
PortGroup               github 1.0

set description_common  {Notre Dame Cooperative Computing Software}

name                    ndcctools
version                 release-4.3.4
categories              sysutils
platforms               darwin
license                 MIT
 
maintainers             nd.edu:pivie \
                        openmaintainer
                        
description             ${description_common} (cooperative-computing-lab)
long_description        ${description_common} \
                        Enable collaborators to easily harness large scale \
                        distributed systems such as clusters, clouds, and grids.
 
homepage                http://ccl.cse.nd.edu/
master_sites            https://github.com/cooperative-computing-lab/cctools/archive/
distfiles               ${version}.zip
[...]

PR #757, opened 2015-05-15
Include Portfile for installing cctools with MacPorts

pivie:
This should satisfy issue #722.

### DIFF
--- a/packaging/macports/Portfile
+++ b/packaging/macports/Portfile
@@ -2,12 +2,11 @@
 # $Id$
 
 PortSystem              1.0
-PortGroup               github 1.0
 
 set description_common  {Notre Dame Cooperative Computing Software}
 
 name                    ndcctools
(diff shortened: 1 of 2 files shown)


dthain:
So, what is the workflow for getting new versions of cctools distributed through macports?
Do we need to update version numbers in the portfile, and then submit an update to macports?
Or something else?

pivie:
Yes we submit an update in the form of a patch with new version numbers. It
appears that anyone can submit an update so I don't know the exact approval
process yet. I hope we can automate it and maybe secure it somehow, but I
was going to wait for the initial approval to go through before exploring
that possibility. There may even be a way to automatically pull the version
from somewhere which would be ideal, but I haven't seen how yet.

dthain:
This is not the last word on macports, but let's get it in the repo and iterate as needed.

PR #762, opened 2015-05-15
Wq single structure

btovar:
Use a single itable to keep tasks' state. This simplifies the code, and also ensures that a task is in a single state. q->ready_list simply lists the task order.


btovar:
@dthain, do not merge yet. Still cleaning some things.

dthain:
I wholeheartedly approve of the general approach.

PR #766, opened 2015-05-18
Add `make rpm` into Makefile 

hmeng-19:
Add `make rpm` into Makefile
Put RPM stuff into cctools/packaging/rpm

The full feature of ndcctools depends on 9 packages: fuse, python, globus, swig, libuuid, readline, zlib, perl, and cvmfs.
To create RPMs for ndcctools with all package support mentioned above, there are two options.
Option 1: Use `./configure.afs` inside the `%build` section and create RPMs on a machine which supports the access of /afs/crc.nd.edu/group/ccl/software/cctools-dependencies.
This option is currently supported by the `make rpm` rule of cctools.

Option 2: Specify each package dependency as a `BuildRequires` attribute, and create RPMs on a machine with all these dependencies installed or you have root access to install these dependencies.
This option is currently not supported by the `make rpm` rule of cctools.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,10 @@ include rules.mk
 
 CYGWINLIB = cygwin1.dll cyggcc_s-1.dll cygintl-8.dll cygreadline7.dll cygncursesw-10.dll cygiconv-2.dll cygattr-1.dll sh.exe
 
+#set the default values for RPM_VERSION and RPM_RELEASE
+RPM_VERSION=4.4.1
+RPM_RELEASE=1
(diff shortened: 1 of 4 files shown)


dthain:
Let's do option 2, thanks.

hmeng-19:
@dthain , please give it a new look.

dthain:
This isn't the last word on rpms, but it's a good start and will help us to start playing around with it.

hmeng-19:
Sounds good.

PR #769, opened 2015-05-20
Use the default value of IFS inside configure

hmeng-19:
Adding `=` into IFS may cause the tilde expansion being ignored when configure is executed like this `./configure --prefix=~/cctools`.
To avoid the trouble caused by this, we should keep IFS as its default value, which includes `<space><tab><newline>`.

### DIFF
--- a/configure
+++ b/configure
@@ -219,9 +219,6 @@ then
 	globus_path=${GLOBUS_LOCATION}
 fi
 
-IFS=' ='
-set -- $*
-
 while [ $# -gt 0 ]
 do
 	case $1 in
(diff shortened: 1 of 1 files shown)


dthain:
Er, but the point of having = in IFS is that it is what allows equals to be used as a separator when shifting arguments in configure.  Taking this out will break current users and examples that are doing something like `--with-cvmfs-path=/some/cvmfs/path`

hmeng-19:
The commit which added the support for `=` was this May, I suppose this feature is still not public shared with all our collaborators, right?

    commit 2ed241214a58207175958e47efc3a34aebfdf16f
    Author: Benjamin Tovar <btovar@nd.edu>
    Date:   Mon May 4 15:05:18 2015 -0400
 
        Fix to allow --prefix=... (note =)

hmeng-19:
BTW, if we really want to keep this `=` feature being supported, then we need to process it elegantly, so it will not by accident make the whole HOME dir of our users ruined. Otherwise, I think our user would rather a narrowed-version of the feature without `=` being supported.

dthain:
@btovar you can decide this one

btovar:
I'll check it when I come back.

PR #784, opened 2015-05-27
Fix eq args

btovar:


hmeng-19:
@btovar , it looks good to me. Using `eval echo` is very clever way to force tilde expansion. Cool.

Just one tiny problem, in your branch, the line 415 and 416 of `configure` file should also be removed, right?

    IFS=" "
    export IFS

btovar:
Indeed. Done.

hmeng-19:
Looks good to me.

PR #788, opened 2015-06-01
Whitespace fixes

batrick:


batrick:
@btovar please look over.

batrick:
Any objections to merging this?

btovar:
Yes, why? Does not it make most of our history close to useless, or if not, harder to decipher? I understand that it is nice to have all spaces standardized, but changing 410 files at the same time sounds a bit much. 

Do you have a strong reason why do you want to do this? If anything, isn't it better to change the files little by little as needed?

batrick:
> Does not it make most of our history close to useless, or if not, harder to decipher?

Well, it **may** require an extra invocation of `git blame` to look at the parent of this PR, i.e. `git blame abcdef^ foo.c`.

> I understand that it is nice to have all spaces standardized, but changing 410 files at the same time sounds a bit much.
>
> Do you have a strong reason why do you want to do this? 

Few reasons for this PR:

* I want to set the `.gitattribute` for CCTools so Autobuild can complain if the PR makes bad whitespace changes.

* I want to fix all the current whitespace problems in one go because:

>If anything, isn't it better to change the files little by little as needed?

This is instinctively more practical but is actually more obstructive to history (IMO). Rather than having one large commit that fixes everything, we have many small commits that do so. Also, we end up having extraneous whitespace fixes attached to pull requests that are distracting and make the PR harder to analyze.

I'd like to point out that most of the changes are small for any particular file. Only a few files required changes on nearly every line. It is acceptable to me to limit this PR to files requiring only a few (<100 lines) whitespace fixes.

btovar:
Ok then, go ahead.

PR #790, opened 2015-06-01
Makeflow GC and Clean

nhazekam:
Branch that aims to move Garbage collection and cleaning the log to be more precise with the state of files through out execution. @btovar @dthain Please take a look and let me know where I should make modifications, or what needs to be fixed.


batrick:
Is this something we should add to 5.0?

Nick, you need to remove the merges from this branch.

nhazekam:
This would be great to have in 5.0. I will remove the merges and try and
finish up testing to make sure it's working as expected.

btovar:
@nhazekam, before this can be merged, please address a couple of comments still not fixed from @dthain.

btovar:
@nhazekam, the issues have been fixed, thanks!

dthain:
I approve of the general approach here, but I would also like @btovar to take a look, since he is more familiar with the code.

btovar:
Will do.

makeflow/src/dag_file.c
```diff
@@ -37,4 +61,24 @@ int dag_file_is_sink( const struct dag_file *f )
 		return 1;
 }
 
+int dag_file_exists( const struct dag_file *f )
```
btovar: Just a clarification. This function is more a 'should exist', rather than an existence predicate, right?

makeflow/src/dag_file.h
```diff
@@ -16,16 +16,32 @@ See the file COPYING for details.
  * unique node, if any, that produces the file.
  */
 
+typedef enum {
+	DAG_FILE_STATE_UNKNOWN = 0,
+	DAG_FILE_STATE_EXPECT = 1,
```
btovar: You can do without the = x. It is easier to maintain given that the numbers are in sequence anyway.

makeflow/src/dag_file.h
```diff
@@ -16,16 +16,32 @@ See the file COPYING for details.
...
+	DAG_FILE_STATE_UP = 6
+} dag_file_state_t;
+
+
 struct dag_file {
 	const char *filename;
 	struct list     *needed_by;              /* List of nodes that have this file as a source */
 	struct dag_node *created_by;             /* The node (if any) that created the file */
 	int    ref_count;                        /* How many nodes still to run need this file */
+	time_t creation_logged;                  /* Time that file creation is logged */
```
btovar: creation_time_logged?

makeflow/src/dag_visitors.c
```diff
@@ -521,10 +521,10 @@ void dag_to_cyto(struct dag *d, int condense_display, int change_size)
 	fprintf(cytograph, "\t<att name = \"layoutAlgorithm\" value = \"Grid Layout\" type = \"string\" cy:hidden = \"1\"/>\n");
 
 	if(change_size) {
-		hash_table_firstkey(d->completed_files);
-		while(hash_table_nextkey(d->completed_files, &label, (void **) &name)) {
-			stat(label, &st);
-			average += ((double) st.st_size) / ((double) hash_table_size(d->completed_files));
+		hash_table_firstkey(d->files);
+		while(hash_table_nextkey(d->files, &name, (void **) &f) && dag_file_exists(f)) {
+			stat(name, &st);
```
btovar: This stat may fail, right? I think dag_file_exists does not actually check the file is there.

makeflow/src/makeflow.c
```diff
@@ -857,6 +804,17 @@ static void makeflow_node_complete(struct dag *d, struct dag_node *n, struct bat
 
 	if(job_failed) {
 		makeflow_log_state_change(d, n, DAG_NODE_STATE_FAILED);
+
+		/* Record which target files have been generated by this node. */
```
btovar: This comment does not seem to match the code below.

makeflow/src/dag_file.c
nhazekam: Yes, a file is either an input which is checked for existence when we start, or is logged as existing from a check after the node has completed. There is no actual checking within the function. Should we rename the function?

makeflow/src/dag_file.h
nhazekam: I will make this change and include it in my next commit.

makeflow/src/dag_file.h
nhazekam: This value is there so that when we log its creation we know the time stamp, which is used when recovering the log to determine is a file was modified between runs.

makeflow/src/dag_visitors.c
nhazekam: I will take a look at this as I believe you are correct and this may need some clarification/modification.

makeflow/src/makeflow.c
nhazekam: I will update this to the correct comment and push it with the next commit.

makeflow/src/dag_visitors.c
nhazekam: I agree that this stat may fail, though I don't believe this is different than what we were previously doing. The completed files structure was removed as we now rely on file state to determine if it was completed. Let me know if we should update this to an actual file check before the stat.

btovar: @nhazekam. Let's rename the function to something like dag_file_should_exist. Also, I would like the check on the stat since it can silently mess up the average variable.

Note that on line 689 we do check for the stat to be succeed, so it is not a big deal.

nhazekam:
@btovar take a look at the updates. Also, @dthain I have added more expressive clean options. I was wondering if there are more options than "intermediates" and "outputs" that we wanted to include.

btovar:
Thanks Nick, testing it.

nhazekam:
@btovar Let me know if I need to make the manpage and -h options more clear.

btovar:
Looks good @nhazekam. Tell me if you are ready to merge.

(We should merge manually, as there are whitespace complaints.)

nhazekam:
I think this is ready to be merged. As the clean and garbage collection options evolve I will make sure to updated the manpage and -h output for clarity. Let me know if you need anything else from me. Also, where should I update the ref_count test so we don't introduce a consistent test failure.

btovar:
What do you mean?

nhazekam:
Since the ref_count test fails, where do I update the expected output? I don't want to cause our test system to post a failure when it is behaving as expected. This is in reference to our conversation that the test may be out of date with the frequency at which files are cleaned.

nhazekam:
I located it. I will step through the test to make sure I am correct and then update the test to reflect the correct values.

btovar:
You mean makeflow/test/TR_makeflow_030_gc_ref_count.sh? The prepare() function in that file writes the values desired.

btovar:
Ha, photo-finish.

btovar:
Nice job Nick.

Issue #781, opened 2015-05-26
'bad ELF interpreter: No such file or directory' when using Parrot + CVMFS


annawoodard:
When running some workflows which use Parrot to access CVMFS, we see the following error:

    bash: /cvmfs/cms.cern.ch/slc6_amd64_gcc491/cms/cmssw/CMSSW_7_3_5/bin/slc6_amd64_gcc491/cmsRun: /cvmfs/cms.cern.ch/slc6_amd64_gcc491/external/glibc/2.12-2/lib64/ld.so: bad ELF interpreter: No such file or directory

@matz-e has made a script which reproduces the problem here: ```/afs/crc.nd.edu/user/m/mwolf3/parrot_test.sh```.

I've put a log with -d all enabled here: ```/afs/crc.nd.edu/user/a/awoodard/Public/elf_error.log``` (using cctools from ```/afs/crc.nd.edu/group/ccl/software/cctools-a078e9c9d-cvmfs-f50d781c0```.) Bypassing Parrot and running on a machine with cvmfs mounted (earth.crc.nd.edu, for example) works fine.

matz-e:
See also [this gist](https://gist.github.com/matz-e/c20ed81601013685604e).

btovar:
@batrick, sounds like 32bit libraries are missing?

batrick:
Historically, we would see bad elf interpreter because of problems with the channel. I'm taking a look now.

batrick:
Oh, I see the problem:

````
$ parrot_run ldd /cvmfs/cms.cern.ch/slc6_amd64_gcc491/cms/cmssw/CMSSW_7_3_5/bin/slc6_amd64_gcc491/cmsRun
[...]
        /cvmfs/cms.cern.ch/slc6_amd64_gcc491/external/glibc/2.12-2/lib64/ld.so => /lib64/ld-linux-x86-64.so.2 (0x00000030ede00000)
````

The loader is in /cvmfs which the kernel can't find. This is something that would be difficult for Parrot to fix as we would need to rewrite the executable's ELF headers to use a local (i.e. not /cvmfs) ld-linux.

I wonder if whoever compiled cmsRun needed to bundle their own ld.so?

[This feels like #678 is haunting me, demanding to be fixed.]

batrick:
@annawoodard @matz-e you can use the `-l` option for Parrot as a temporary workaround:

`parrot_run -l /lib64/ld-linux-x86-64.so.2`

[...]

PR #793, opened 2015-06-02
Fix static executable handling broken by #786

batrick:
Reported by @jblomer in #781.


dthain:
@batrick please add a test case for this.  Can be as simple as a static build of hello world.

PR #794, opened 2015-06-03
Test execve of static executables.

batrick:
Continuation of #793.


batrick:
This test is a little silly because some Autobuild platforms don't have libc-devel so you can't compile static executables.

dthain:
Amazing to me that `gcc -static` is not widely portable...

PR #796, opened 2015-06-09
Replace the usage of CWD with PWD

hmeng-19:
Replace the usage of CWD with PWD
Adjust umbrella log


hmeng-19:
@dthain , please have a look.

dthain:
No objections here.

hmeng-19:
@dthain , then let us merge it.

PR #800, opened 2015-06-12
Separate CMSSW from CVMFS in umbrella

hmeng-19:
cvmfs now functions as one method to deliver cmssw, not the only method to
deliver cmssw.


hmeng-19:
@dthain , please have a look.

umbrella/src/umbrella.py
```diff
@@ -380,41 +380,47 @@ def dependency_process(env_para_dict, name, id, mountpoint, action, packages_jso
...
+	#table schema: (name text, version text, platform text, store text, store_type text, type text)
+	store = item["source"][0]
+	logging.debug("%s is chosen to deliver %s", store, name)
 
-		#if local_cvmfs is empty, set $HTTP_PROXY, download cctools package, set cvmfs_siteconf_mountpoint, and call parrotize_user_cmd.
-		if not local_cvmfs:
-			logging.debug("Add env variable HTTP_PROXY = http://cache01.hep.wisc.edu:3128 into env_para_dict")
-			env_para_dict["HTTP_PROXY"] = "http://cache01.hep.wisc.edu:3128"
+	if name[:5] == 'cmssw': # another solution: split name based on '-' and check whether the first item is 'cmssw'.
+		is_cms_app = 1
```
dthain: There exist software systems other than CMS that are delivered by cvmfs: atlas, alice, minos, etc.
Will this work for any software, or just for things named "cmssw"?

hmeng-19: This code is just for "cmssw", which corresponds `/cvmfs/cms.cern.ch`.
For atlas, we need add another `if` logic judgement statement like `if name[:5] == 'atlas'`, which corresponds `/cvmfs/atlas.cern.ch`.

dthain:
Well, that's my point: the umbrella code shouldn't know anything about any particular dependency. The mapping from `cmssw` to `cvmfs:/cvmfs/cms.cern.ch` should be expressed only in the metadata catalog.  The umbrella code should know how to access cvmfs in general, not cmssw specifically.  That way, new users will not have to modify the umbrella code..

hmeng-19:
Got it . Your point is Umbrella should be aware of cvmfs (more general, resource type), but should not be aware of cms, or atlas (more general, application type). Right?
This branch is not thorough. I will keep working on it.

hmeng-19:
However, there is one exception.
In case when cvmfs is needed to deliver cmssw and the execution node does not have cvmfs installed by FUSE, Umbrella still needs to know this is an cms application to set the SITECONF: `parrot_run -M /cvmfs/cms.cern.ch/SITECONF/local=<pwd>/SITECONF/local ...`

hmeng-19:
@dthain , please give it a new look.
The key conclusion is:
Umbrella should not be aware of the types of software dependencies according to
the specification. The execution logics for software dependencies only depend
on the resource type like cmvfs, http, or https.

In case when cvmfs is needed to deliver cmssw and the execution node does not
have cvmfs installed by FUSE, Umbrella still needs to know this is an cms
application to set the SITECONF: parrot_run -M
/cvmfs/cms.cern.ch/SITECONF/local=<pwd>/SITECONF/local ...
However, umbrella figures out an application is a cms application through the /cvmfs/cms.cern.ch mountpoint, not by the umbrella specification.

dthain:
You have the right idea -- as a general rule, umbrella should know about cvmfs but not cmssw in particular. I think it's ok to have a little hack in place for the siteconf issue, as long as we all understand it is a hack, and it is documented appropriately.

On a related note, I see that umbrella is hardcoded to use the http proxy at UW-Madison.  That's OK for us to do, but probably not what we want the whole world.  How would a different end user go about specifying the right proxy to use.  Should we require that `HTTP_PROXY` be already set when cvmfs is used?  Require a command-line-argument?  Something else?

hmeng-19:
Relying on the user setting `HTTP_PROXY` may confuse the user. Because the user only knows his application needs cmssw and knows nothing about cvmfs, it is umbrella that determines using cvmfs to deliver cmssw. 

If the user wants to set `HTTP_PROXY` for his application:
The `--env` option of umbrella is one way.
The `environ` section of umbrella specification file is another option.

dthain:
Ok, so if the user wants to run the app at University of X which has its own web proxy, how does he/she configure that?

(Putting it in the environ section of the Umbrella spec isn't correct, because the *app* doesn't need to see it, cvmfs does.)

hmeng-19:
Putting `HTTP_PROXY` into the `environ` section of umbrella  is not so bad. cvmfs is not the only application which benefits from the setting of `HTTP_PROXY`.

Another solution is: the user says nothing about `HTTP_PROXY` in the `environ` of umbrella specification and `--env` option of umbrella. Only when umbrella decides to use cvmfs + parrot to deliver cmssw and fails to find the `HTTP_PROXY` environment variable and then notifies the user of the missing of `HTTP_PROXY`. Then the user sets the environment variable and restarts umbrella.

dthain:
Careful here: HTTP_PROXY is relevant only to the sandbox technology, not to the user's application, so it doesn't belong in the Umbrella spec.  It's only something that should be passed to Umbrella at runtime.

You could simply require that the user *running* umbrella state the proxy whenever cvmfs use is requested.  As an example, see what parrot does if you try to use cvmfs without defining a proxy.

hmeng-19:
After this pull request is merged, umbrella is ready for cctools 5.0.

Issue #801, opened 2015-06-12
Python utilities need to specify python2 as interpreter


batrick:
Starch for example fails if it is run on a system where `python` is Python 3.X.

batrick:
@pbui, do you know the right way to fix this?

pbui:
Unfortunately, there is no good way of specifying python 2.  On most distributions, `python == python2`.  On a handful, like ArchLinux, `python == python3`.  The solution is suppose to be to use `python2` or `python3`... unfortunately not all distributions have a python2 (they will have a specific `python2.6` or `python2.7`, but not `python2`).

Here is [PEP 0394](https://www.python.org/dev/peps/pep-0394/) with more information.

One way forward is to make starch with with both python2 and python3.

batrick:
I think we have at least these options:

* Use the Python 2.X found by configure to hard code the shebang line. (similar to what we do with CCTOOLS_VERSION in our python code).
* Use something like [this](http://stackoverflow.com/a/26056481).

Whatever we end up doing, we should add Python version detection code to all of our Python code. It's unacceptable that our scripts fail once they hit an incompatibility. Instead the code should quit early if the Python version is wrong.

btovar:
I'll take this one.

dthain:
Two questions:

1 - Does there exist a subset of Python language that is valid in both Python 2 and Python 3?

2 - Is there any distribution other than ArchLinux that defines python=python3?  To me, that sounds like a pretty glaring backwards compatibility error.

batrick:
> 1 - Does there exist a subset of Python language that is valid in both Python 2 and Python 3?

Peter seems to have found a subset for Starch at least. We still need to fix our other Python programs.

> 2 - Is there any distribution other than ArchLinux that defines python=python3? To me, that sounds like a pretty glaring backwards compatibility error.

Heh, Arch Linux doesn't believe in backwards compatibility (it's a rolling release distro). However, I don't think it's the only distro to do this.

pbui:
> Peter seems to have found a subset for Starch at least. We still need to fix our other Python programs.

In general, supporting both Python 2.7 and Python 3.x is relatively straighforward.  It is supporting Python 2.4, 2.5, and 2.6 along with modern versions of Python that is difficult.  Eventually, we should cut support for RHEL5 (and thus Python 2.4) rather than doing gymnastics.

[...]

PR #803, opened 2015-06-12
starch: support both python2 and python3

pbui:
This now works with python2 (2.6, 2.7) and python3 (3.X) unmodified.

Support for python2.4 is dropped due to lack of BytesIO and changes in
exception handling syntax.

Addresses #801

### DIFF
--- a/makeflow/src/starch
+++ b/makeflow/src/starch
@@ -8,6 +8,13 @@
 
 from __future__ import print_function
 
+import os
+import sys
+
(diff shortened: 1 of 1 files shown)


pbui:
@btovar I made the changes you requested.

batrick:
@btovar This causes tests to fail for Python 2.4 (RHEL5/Centos5). We need a way to disable those tests somehow.

PR #804, opened 2015-06-12
starch: support python2.4 through python3.x

pbui:
To support python2.4 avoid octal literals, avoid binding exception
variables, and utilize logging for reporting.

Further cleansup #803 and helps with #801


dthain:
Thanks, Peter!

PR #809, opened 2015-06-16
Add new parrot_debug syscall.

### LINKED ISSUE #797: Parrot should have a system call to turn on debugging.
We are increasingly seeing Parrot used as a wrapper for entire application frameworks (now even acting as an "init" process in Docker). It will become more important to be able to get debugging at choice moments. Editing the wrapper may not be feasible and may result log files that are too large.

I recommend we add a utility like `parrot_cp` which executes a new `parrot_debug` system call that sets flags, output file, etc.

batrick:
Fixes #797.

### DIFF
--- a/doc/man/parrot_run.m4
+++ b/doc/man/parrot_run.m4
@@ -48,6 +48,7 @@ OPTION_ITEM(--cvmfs-repo-switching) Allow repository switching with CVMFS.
 OPTION_TRIPLET(-R, root-checksum, cksum)Enforce this root filesystem checksum, where available.
 OPTION_ITEM(`-s, --stream-no-cache')Use streaming protocols without caching.
 OPTION_ITEM(`-S, --session-caching')Enable whole session caching for all protocols.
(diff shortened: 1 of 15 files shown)


batrick:
@dthain look about right?

dthain:
Maybe consider separating the three operations (change file, change detail, change size) since they each have very different security considerations.  (e.g. changing the output file would not be safe if parrot is used as a sandbox.)

batrick:
I'm not sure I see the benefit of separating the operations into different syscalls.

As for the security of changing the output file, I think that's handled well enough via

https://github.com/cooperative-computing-lab/cctools/pull/809/files#diff-c7a7e7c1ffa8f36dc86019560e5798f9R2968

We try to do a normal `pfs_open` and then get the local path from that. This way the usual permission checks are enforced.

Perhaps there should also be a Parrot option to disable the syscall?

batrick:
I added an option to `parrot_run` to disable the syscall.

I think this is ready for merging.

btovar:
Once activated, is there a way to disable debugging? Like parrot_debug none?

batrick:
Yes: `parrot_debug clear`

PR #814, opened 2015-06-18
Rmonitor library

btovar:
Reorganizes the resource monitor code so that it can be used as a library.  The library gathers the resources information given a pid.
Short example:

#include "rmonitor_poll.h"
int main(int argc, char **argv) {
    sleep(2);

    struct rmsummary resources;
    rmonitor_measure_process(&resources, getpid());

    fprintf(stdout, "wall time used (s): %3.0lf, ",
            resources.wall_time/1000000.0);

    fprintf(stdout, "total memory used (MB): %" PRId64 ", ",
            resources.resident_memory + resources.swap_memory);

    fprintf(stdout, "total cores used: %" PRId64 "\n",
            resources.resident_memory + resources.swap_memory);

    return 0;
}

I would like to incorporate this library into work queue to measure the resources used by the master, and into parrot, to give measurements to the CMS folks.


dthain:
I like the overall approach, and it will be nice to see this available in WQ.
I'll trust you to get the code details right.

PR #815, opened 2015-06-19
Wq master monitor resources.

btovar:
Based on #814 (only adds one more commit).

When using work_queue_enable_monitoring, the first summary of the resources report corresponds to the master. Resources are polled every 5 minutes. The summary also includes the master command line and the master project name (if given).

This follows the new convention on the resources archive, in which the first summary of a file is taken as the parent of all the summaries in that file.

### DIFF
--- a/dttools/src/rmonitor_poll_internal.h
+++ b/dttools/src/rmonitor_poll_internal.h
@@ -39,6 +39,8 @@ See the file COPYING for details.
 #include "rmonitor_types.h"
 #include "rmsummary.h"
 
+#include "rmonitor_poll.h"
+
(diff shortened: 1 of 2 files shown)


dthain:
Does this require a change the general linking instructions in the work queue examples and manual?

btovar:
Yes, and I have not changed them because I have not make a decision yet.

As it is now, we have to add -lrmonitor_poll every time we use -lwork_queue.

I was thinking of adding rmonitor_poll to dttools. This simplifies linking, but seems ugly. 

Do you have any preference?

btovar:
No linking changes necessary. Moved monitoring api to dttools.

PR #816, opened 2015-06-22
Python exec

### LINKED ISSUE #801: Python utilities need to specify python2 as interpreter

btovar:
Proposal to resolve #801.
You can check its use with starch.

Note that I did not want to use hard coded values from configure, as that may break for the binary releases.

cctool_python reads the second line of a python script, which
should be of the form:

  # CCTOOLS_PYTHON_VERSION 2.7 3 2.6 42

which lists, in order of preference, the acceptable versions to run the
script. For example:

  # CCTOOLS_PYTHON_VERSION 2.4

means: only run on python 2.4.

  # CCTOOLS_PYTHON_VERSION 2.7 2.6

means: 2.7 and 2.6 are both ok, but prefer 2.7.

  # CCTOOLS_PYTHON_VERSION 3

means: any version 3 is ok.

### DIFF
--- a/apps/makeflow_blasr/makeflow_blasr
+++ b/apps/makeflow_blasr/makeflow_blasr
@@ -1,4 +1,5 @@
-#! /usr/bin/env python
+#!/usr/bin/env cctools_python
+# CCTOOLS_PYTHON_VERSION 2.7 2.6
 
 # Copyright (C) 2014- The University of Notre Dame
 # This software is distributed under the GNU General Public License.
--- a/apps/makeflow_blast/makeflow_blast
+++ b/apps/makeflow_blast/makeflow_blast
@@ -1,4 +1,5 @@
-#! /usr/bin/env python
+#!/usr/bin/env cctools_python
+# CCTOOLS_PYTHON_VERSION 2.7 2.6
 
 # Copyright (C) 2011- The University of Notre Dame
 # This software is distributed under the GNU General Public License.
--- a/apps/makeflow_bwa/makeflow_bwa
+++ b/apps/makeflow_bwa/makeflow_bwa
@@ -1,4 +1,5 @@
-#! /usr/bin/env python
+#!/usr/bin/env cctools_python
+# CCTOOLS_PYTHON_VERSION 2.7 2.6
 #
 #Copyright (C) 2013- The University of Notre Dame
 #This software is distributed under the GNU General Public License.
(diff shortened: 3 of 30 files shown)


batrick:
The main thing that makes me nervous about this is that no one else does it AFAIK. But then again, most people don't have the deployment issues we do.

I'll go through the code by Wednesday.

btovar:
We would like some way to explicitly state the version of python required. Per @pbui comments, this does not seem to be a standard practice, and in fact in doing so, we seem to be going against common python idioms.

I see the following alternatives:

1) Use the proposed cctools_python. This plays nicely with our current thinking of cctools and umbrella, but as you mention, it is not standard. This option solves all our problems, but as any wrapping script, it may cause conflicts with particular setups.

2) Use the comment CCTOOLS_PYTHON_VERSION, and the try: except: blocks that @pbui first proposed with the appropriate version comments. This seems the most idiomatic, but does not solve the problem of /usr/bin/python being python 3 in arch. 

3) Use the comment CCTOOLS_PYTHON_VERSION, and check explicitly for the python version as we do now in starch. This is not idiomatic, and does not solve the arch problem. However, it does have the advantage over 1) and 2) that the script explicitly handles the different python versions.

Is there an easy way we can make 2) or 3) work for arch? Note that rewriting the #! with a python path may not be a solution, as the script may not be executing in the machine ./configure was run (Note that shipping cctools_python with the script and running ./cctools_python script does work.)

batrick:
Sorry, I don't want to suggest I'm against this solution. I was just nervous. I think it's better than the alternatives.

I have a few comments for the code which I'll do in-line then I think this will be good to merge.

batrick:
That's it from me. Looks good otherwise Ben.

btovar:
@hmeng-19, @nhazekam, @pivie 
Please check that your scripts run with the modifications in bbb5836. You need to set your PATH so that cctools_python is found (originally found in dttools/src).

Also, if your scripts run in python 2.4, 2.5 or 3, please update CCTOOLS_PYTHON_VERSION appropriately.

hmeng-19:
@Ben, it works for me. Thanks.

hmeng-19:
@Ben, it works for me. Thanks.

btovar:
Thanks everyone.

PR #820, opened 2015-06-25
App update to standard setup

nhazekam:
These were updates that modify bio-apps to a standard form. Removed hardcoded variable passing to passing the extra arguments through. @btovar


batrick:
@nhazekam does this need to be in 5.0?

nhazekam:
Yes this should be in 5.0.

PR #821, opened 2015-06-25
Prune

pivie:
Prune and umbrella are working together and a 3 stage test case is included for HEP. (HTTP_PROXY is temporarily hard coded to support umbrella and work queue at the same time)


batrick:
@pivie does this need to be in 5.0?

pivie:
I was hoping it could be.

PR #822, opened 2015-06-26
Adds test, example, and documentation to chirp python bindings

btovar:


batrick:
I tried out the test on my machine, works without issue.

PR #825, opened 2015-06-28
Several Umbrella Features

hmeng-19:
1) Add mount_env into the umbrella spec syntax
    
        mount_env allows the user's analysis code to ignore the data location, and
        also makes the relocation of the dependency easier without changing the
        analysis code and just change the spec file.

2) Add a new umbrella example for cern_opendata

        This example comes from the cern opendata website.

3) Check HTTP_PROXY for Umbrella+Docker+Parrot app

4) Use the right version of cctools
    
        If the sandbox mode is parrot, use the cctools version which matches
        the host machine.
        If the sandbox mode is docker or chroot, and parrot is needed inside
        the sandbox to deliver cvmfs, use the cctools version which matches
        the specification os requirement.
    
5)    Create mountpoint for each level of a path, Add fake mounts into parrot mount file
    
        If a mountitem '/a/b/c' is added into the parrot mountfile, and if '/a' does
        not exist or '/a/b' does not exist, `cd` fails because it needs to stat /a and
        stat /a/b.  Solution: create mountpoints for /a and /a/b when they do not
        exist.

6)    Replace the cctools binary in umbrella archive
    
        The new cctools binary is based on the commit:
                Date:   2015-06-26 17:08:45 -0400
                Commit: f2bb681fdf8403632aa3da767983bf566e40007d (master)

### DIFF
--- a/doc/umbrella.html
+++ b/doc/umbrella.html
@@ -238,6 +238,7 @@ cmsDriver.py Hadronizer_MgmMatchTuneZ2star_8TeV_madgraph_cff.py -s GEN \
 	--inputs 'cms_complex.sh=cms_complex.sh' \
 	--localdir /tmp/umbrella_test/ \
 	--output /tmp/umbrella_test/parrot_cms_complex_output \
+	--cvmfs_http_proxy http://cache01.hep.wisc.edu:3128 \
 	run '/bin/sh cms_complex.sh'</code>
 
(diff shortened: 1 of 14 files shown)


hmeng-19:

hmeng-19:

dthain:
Looks good overall, I am especially happy with how the example spec using the open data portal has turned out.

PR #826, opened 2015-06-29
add explicit port to master node

btovar:
@batrick, does this look correct? This is in case the default port is already taken.

With this modifications, I am getting the following error:

2015/06/29 09:09:27.42 makeflow[2714] batch: exited abnormally: ERRORED (Invalid cross-device link)

### DIFF
--- a/doc/confuga.html
+++ b/doc/confuga.html
@@ -142,6 +142,7 @@
     --catalog-update=30s \
     --debug=confuga \
     --jobs \
+    --port=9000 \
     --project-name=`whoami`-test \
     --root='confuga://./confuga.root/?auth=unix&amp;nodes=node:chirp://localhost:9001/,chirp://localhost:9002/'</code>
                 </p>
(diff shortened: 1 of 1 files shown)


batrick:
@btovar I'm looking into this.

btovar:
If it makes any difference, both confuga root and makeflow workdir are on AFS, but different cells

batrick:
@btovar I think the problem is that you didn't modify makeflow as well:

    makeflow --batch-type=chirp --working-dir=chirp://localhost/ makeflow/example/example.makeflow

You need to do:

    makeflow --batch-type=chirp --working-dir=chirp://localhost:9000/ makeflow/example/example.makeflow

As for the error message you found, you can't run Chirp jobs which use LINK bindings (the default) because AFS doesn't let you do cross-directory hard links.

btovar:
Thanks Patrick, I did use localhost:9000. Could you add a comment regarding AFS in "Getting Started with Confuga"?

batrick:
I think I'll add a warning in Chirp as well, like we do for batch job Condor.

btovar:
I can confirm that on /tmp the example works.

batrick:
@btovar please go ahead and merge this after fixing the makeflow line and updating the man page example too.

PR #830, opened 2015-06-29
Fixes to WQ workers started as SGE jobs on a big cluster with a shared file system

andreyto:
This branch addresses two critical problems that manifested themselves under the use case identified in the title of this pull request

Scenario: I use sge_submit_workers to submit a 1000 workers on a big cluster where all file systems are shared, so the worker sandbox space has to be on a shared FS. `CentOS 6; OGS/Grid Engine 2011.11.`

Problems and fixes:

1. When SGE has to kill the worker (e.g. hard wallclock limit in a queue is reached or I do qdel), SGE  sends SIGKILL to the process group that it has created for the worker. Worker, however, runs its task subprocesses under a new process group. Worker cannot trap SIGKILL, so it gets killed, and tasks keep running (they are under different process group) and wreck havoc.

  Fixes: 
 - added wallclock run-time option to worker so that it will exit before its job wallclock runs out
 - worker traps now SIGUSR1 and SIGUSR2 and exits killing tasks, so that it can be submitted with SGE qsub options `-notify` and `-l s_rt XXX` which result in these signals being send to the job before the SIGKILL or SIGSTOP
 - killing tasks as soon as possible after detecting the exit request

 The fixes above are partial solutions. Nothing 100% proof can be done without using cgroups.

2. In the main worker loop, check_disk_space_for_filesize has been called once per 10 ms (before any changes introduced in p.1 above), bringing down the shared file system with statvfs calls when a 1000 workers were running at the same time.

 Fixes:
 - results of that call are now cached and calls only made every 20s

Tests

No unit tests added, but these changes were tested in an actual large scale runs, as well as with temporarily added diagnostic printouts.

### DIFF
--- a/dttools/src/disk_info.c
+++ b/dttools/src/disk_info.c
@@ -58,19 +58,19 @@ int disk_info_get(const char *path, UINT64_T * avail, UINT64_T * total)
 }
 
 /* Slower disk check, poor man's du on workspace */
-int check_disk_workspace(char *workspace, int64_t *workspace_usage, int force, int64_t manual_disk_option, int measure_wd_interval, time_t last_cwd_measure_time, int64_t last_workspace_usage, UINT64_T disk_avail_threshold) {
+int check_disk_workspace(char *workspace, int64_t *workspace_usage, int force, int64_t manual_disk_option, int measure_wd_interval, time_t *last_cwd_measure_time, int64_t *last_workspace_usage, UINT64_T disk_avail_threshold) {
 
 	if(manual_disk_option < 1)
 		return 1;
 
-	if( force || (time(0) - last_cwd_measure_time) >= measure_wd_interval ) {
-		cwd_disk_info_get(workspace, &last_workspace_usage);
-		debug(D_DEBUG, "worker disk usage: %" PRId64 "\n", last_workspace_usage);
-		last_cwd_measure_time = time(0);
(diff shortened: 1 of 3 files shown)


btovar:
Thanks @andreyto!

dthain:
This intersects nicely with a suggestion from @leeping, who mentioned that it would be nice for the worker to advertise is available execution time, and then (labelled) tasks could be matched to workers with a minimum expected runtime.

leeping:
Thanks. :) I've been following the developments (despite my silence) and this sounds like a great improvement.

btovar:
@andreyto, @dthain 
I have incorporated to master the first two commits (1cc4b0d, and 746c56f) without change, and the last one (beb5ac6) with major modifications.

I am not incorporating the third one yet (40cff7c), as I would like to make sure the SIGALARM does not cause issues.

dthain:
Other libraries often use SIGALRM quietly, which could throw us off.
Perhaps just make this another check for absolute time in the main loop..

andreyto:
Good point. There is no advantage in using SIGALRM in this code.

btovar:
Final issue address with #851.

btovar:
@andreyto, all your proposed ideas have been incorporated in 5.0.3. Thanks again!

andreyto:
Great, thanks!
Andrey

PR #834, opened 2015-07-01
Job chirp bindings

btovar:
Python and Perl bindings for the chirp job interface.

Examples:
chirp/src/python/chirp_jobs_python_example.py
and
chirp/src/perl/chirp_jobs_perl_example.pl

Bindings are written so that a user should not need to write json directly, but instead use hashes.


batrick:
Works for me @btovar, great job!

btovar:
Thanks!

PR #838, opened 2015-07-04
Chirp acl misc bindings

btovar:


chirp/src/chirp_swig_wrap.c
```diff
@@ -54,6 +54,18 @@ char *chirp_wrap_whoami(const char *hostname, time_t stoptime)
...
+	unsigned char digest[CHIRP_DIGEST_MAX];
+
+	result = chirp_reli_hash(hostname, path, algorithm, digest, stoptime);
+
+	if(result < 0)
+		return NULL;
+
+	return xxstrdup( (char *) digest);
+}
+
```
batrick: The digest is in binary and variable size (the digest size is result if > 0). So we can't use xxstrdup.

btovar: Thanks @batrick. This should be fixed now.

batrick:
@btovar added a small change to make the code work with new hashes automatically.

PR #839, opened 2015-07-05
Add support for chunked files available in libcvmfs revision >= 18

jblomer:
This patch adds support for libcvmfs revision >= 18, which includes support for reading large files in several chunks on request of the corresponding byte range.  It complements a [pull request for the cvmfs code](https://github.com/cvmfs/cvmfs/pull/1003). Files that are available in multiple chunks will have a libcvmfs file descriptor with bit 30 set.  Since now read requests are channeled through `cvmfs_read`, libcvmfs can handle such file descriptors in a special way and defer the download of chunks from `cvmfs_open()` to `cvmfs_read()`.

I tested this change against the [LHC test applications](https://github.com/jblomer/parrot-test).  An example for a chunked file can be found in /cvmfs/cernvm-prod.cern.ch/cvm3/var/lib/rpm/Packages.

The changes to parrot should be backwards compatible so that a patched parrot works with an older version of libcvmfs.  Note, however, that libcvmfs revision >= 18 won't properly work with a non-patched parrot because of the possibility of virtual file descriptors being returned by libcvmfs.

If you have no objections, I'll merge the current cvmfs devel branch into the libcvmfs-stable branch once this patch is merged.

### DIFF
--- a/parrot/src/pfs_service_cvmfs.cc
+++ b/parrot/src/pfs_service_cvmfs.cc
@@ -11,6 +11,10 @@ See the file COPYING for details.
 #ifndef LIBCVMFS_VERSION
 	#define LIBCVMFS_VERSION 1
 #endif
+#ifndef LIBCVMFS_REVISION
+	#define LIBCVMFS_REVISION 0
+#endif
+
 
 extern "C" {
 #include "buffer.h"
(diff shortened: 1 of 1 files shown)


dthain:
As a general design principle, parrot should not have any problem when cvmfs starts using virtual file descriptors.  From the perspective of the parrot-cvmfs driver, the fd returned by libcvmfs is just stashed away as a secret -- parrot manages its own table of fds for the process.

btovar:
@jblomer, just to be sure, the latest libcvmfs-cctools is on revision 16, correct?

jblomer:
I moved it forward today, it should be at revision 18 now.

PR #840, opened 2015-07-05
remove restriction of maximum number of file descriptors for cvmfs

jblomer:
The `max_open_files` option for cvmfs was originally meant to increase the maximum number of open files.  In the context of non-privileged users, that doesn't make much sense though because users are typically anyway at their limit. The current default has in many cases the opposite effect of restricting the user from the normal 1024 limit. The lhcb test case from the [LHC test applications](https://github.com/jblomer/parrot-test), for instance, does not work with a limit as small as 500. This patch removes the default limit.


dthain:
@btovar I seem to recall that we had a user suffering from an out-of-file-descriptors problem.  Do you recall who that was?  (And if so, we should get this patch to them...)

btovar:
I remember @annawoodard, @hmeng-19 and @pivie.

pivie:
Haiyan seemed to figure out some way to solve, in Umbrella, the problem I
was having.

hmeng-19:
Patrick fixed the problem before, so all the release 5.0.0 rc should work.

Issue #847, opened 2015-07-06
Should cctools_python set PYTHONPATH?


batrick:
Right now to use Weaver and probably other tools we need the user to set `$PYTHONPATH` like so:

    env PYTHONPATH="$(dirname "$(which weaver)")/../lib/python2.6/site-packages/" weaver ...

which kinda stinks. Since we're already wrapping python to execute an appropriate version, we should set `PYTHONPATH` too since the location depends on the version selected. [In fact, the version on the system may be 2.6 but the library is in `lib/python2.7/`, yuck...]

Also, based on my recent experience using Weaver, the main drawback of `cctools_python` so far is that it must be in your path. Doing something like `./weaver` won't work anymore.

batrick:
@dthain, comments?

btovar:
I agree, PYTHONPATH should also be set in cctools_python.

PR #848, opened 2015-07-06
set python path according to script

btovar:
Sets PYTHONPATH in cctools_python.

I am not too happy about setting PYTHONPATH implicitly for the user, as it may break local setups. That said, we cannot guarantee cctools_python to work correctly without this fix.

Let's give these solutions a try, but with our eyes open for something less icky.

(ref #847)

### DIFF
--- a/dttools/src/cctools_python
+++ b/dttools/src/cctools_python
@@ -42,8 +42,8 @@ match_versions () {
 find_python_path () {
 	_script=$1
 	_version=$2
-	_dirname="$(dirname $_script)/../../lib"
-	echo $_dirname/python$_version:$_dirname/python
+	_dirname="$(dirname $_script)/../lib"
+	echo $_dirname/python$_version/site-packages:$_dirname/python/site-packages
 }
 
 if [ -z "$1" ]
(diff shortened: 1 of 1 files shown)


batrick:
I need this patch to get `weaver -h` to work:

````
$ git diff
diff --git a/dttools/src/cctools_python b/dttools/src/cctools_python
index 7a7f730..e6f637d 100755
--- a/dttools/src/cctools_python
+++ b/dttools/src/cctools_python
@@ -42,8 +42,8 @@ match_versions () {
 find_python_path () {
        _script=$1
        _version=$2
-       _dirname="$(dirname $_script)/../../lib"
-       echo $_dirname/python$_version:$_dirname/python
+       _dirname="$(dirname $_script)/../lib/"
+       echo $_dirname/python$_version/site-packages/:$_dirname/python/site-packages/
 }
 
 if [ -z "$1" ]
````

btovar:
Mmmm, it works for me. How are you calling it?

btovar:
Oh, nevermind. I did not push my latest version. Sorry about that.

btovar:
Fixed, thanks @batrick !

batrick:
Works for me @btovar, if you are satisfied go ahead and merge it.

dthain:
I can imagine many ways in which this could accidentally fail if the user ends up moving files here and there.  At a minimum, be sure to emit a detailed error message if the appropriate python lib directory is not found where it is expected..

PR #849, opened 2015-07-07
emmit warning when python modules are not where expected

btovar:


dttools/src/cctools_python
```diff
@@ -43,7 +43,18 @@ find_python_path () {
 	_script=$1
 	_version=$2
 	_dirname="$(dirname $_script)/../lib"
-	echo $_dirname/python$_version/site-packages:$_dirname/python/site-packages
+
+	for dir in $_dirname/python{$_version,}/site-packages; do
+		echo $dir
```
batrick: seems you echo $dir twice?

PR #851, opened 2015-07-08
Worker quits after --wall-time=seconds have passed.

btovar:
@dthain, @andreyto


nhazekam:
Works as expected both locally and in Condor. Looks good to me. Ready to be merged. @btovar

PR #857, opened 2015-07-15
Umbrella git support

hmeng-19:
Umbrella git support + two new examples:
1) umbrella/example/cms_opendata_git: git is used inside the user's application.
2) umbrella/example/git_protocol: git is not used inside the user's application, but used by umbrella to deliver some dependencies from git repos.
    
Other tiny changes:
   1) Check the existence of attributes before using them.
    2) Set the default action for software dependencies to be unpack.


hmeng-19:
@dthain , please have a look.

PR #858, opened 2015-07-15
Batch sys

nhazekam:
Support Torque and SLURM for XSEDE. @btovar please take a look.

### DIFF
--- a/configure
+++ b/configure
@@ -220,12 +220,6 @@ config_xrootd_path=no
 config_cvmfs_path=no
 config_swig_path=auto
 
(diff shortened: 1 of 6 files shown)


dthain:
Update `doc/makeflow.html` and `doc/man/makeflow.m4` while you are at it.

nhazekam:
I added a PBS specific options, but after looking more into the documentation better I realized I was wrong, and both Blacklight and Gordon use Torque. @btovar let me know if you want me to revert to the previous or just leave the PBS code in there.

PR #860, opened 2015-07-16
Illustrate usage of umbrella id attr

hmeng-19:
Illustrate usage of umbrella id attr inside a umbrella specification file;	

add doc for git and root protocol support;


hmeng-19:

In case you are interested, here is the current metadata database:
http://ccl.cse.nd.edu/software/umbrella/database/packages.json

hmeng-19:
@dthain , I think this one is ready to merge.

PR #862, opened 2015-07-17
work_queue_worker export resources

btovar:
Request by @khurtado, for better reporting to the CMS dashboard.

### DIFF
--- a/work_queue/src/batch_job_work_queue.c
+++ b/work_queue/src/batch_job_work_queue.c
@@ -58,7 +58,7 @@ static void specify_envlist( struct work_queue_task *t, struct nvpair *envlist )
 		char *name, *value;
 		nvpair_first_item(envlist);
 		while(nvpair_next_item(envlist,&name,&value)) {
-			work_queue_task_specify_env(t,name,value);
+			work_queue_task_specify_enviroment_variable(t,name,value);
 		}
 	}
 }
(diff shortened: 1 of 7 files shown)


dthain:
Maybe I am reading this incorrectly, but it doesn't seem right to me.

It looks like the intent of the code is to pass to the worker the env vars representing the total amount of  resources managed by the worker (manual_{cores|disk|memory}.

I think what we want is to pass to the worker env vars representing the resources allocated to that particular *task*, e.g. `CORES=task->resources->cores`.

btovar:
You are right, but actually I think we want both. Let me think more about it.

@khurtado wants to know the size of the worker, but maybe they really want to know the size of the task.

annawoodard:
I think what we need is the size of the task. This is for reporting to dashboard, and each task will be doing its own reporting.

btovar:
Great, disregard this commit then.

khurtado:
I agree with @annawoodard, sorry if there was any misunderstanding!

btovar:
@dthain, added changes we talked about this morning.

dthain:
496dc95 appears to be setting the environment at the master side, which isn't quite right.

I think we want the *worker* to set `CORES, MEMORY, DISK` at the moment of execution,
since they describe the size of the resource container in which the task is run.

This is because the task may not specify some resource, or it may specify a range, or something more complex.  The actual allocation is made by the worker and is distinct from the request, so the worker should specify the info.

If that's not clear, let's talk in person.

btovar:
Not quite. We are adding the variables to the environment of the task. They do not get set until work_queue_process does its thing.

btovar:
Oh, I see what you mean now.

btovar:
@dthain ready to merge if you are.

dthain:
Looks good!

PR #864, opened 2015-07-21
Add _path global for weaver scripts.

batrick:
This makes it possible to reference files relative to the location of the
weaver script.


batrick:
@pbui is `_path` a good name for this variable?

batrick:
Peter tells me we can use `CurrentScript().path` instead. Closing.

PR #865, opened 2015-07-22
Add disk_alloc.[ch] and disk_alloc_test.c to dttools/src/

nkremerh:
Initial addition of disk_alloc.h, disk_alloc.c, and disk_alloc_test.c for peer review.


dttools/src/disk_alloc.c
```diff
@@ -0,0 +1,230 @@
...
+#include <unistd.h>
+#include <sys/stat.h>
+#include <sys/mount.h>
+#include <string.h>
+#include <ctype.h>
+#include <errno.h>
+
+#include "disk_alloc.h"
+
+int disk_alloc_create(char *loc, char *space) {
```
dthain: Pass in space simply as the number of bytes in an int64_t, this will make it easier to use the library.
In the test program, use `string_metric_parse` to convert from a string to an int64_t.

dttools/src/disk_alloc.c
```diff
@@ -0,0 +1,230 @@
...
+				size = size << 10;
+			case 'K':
+				break;
+		}
+
+	}
+
+	else {
+
+		printf("\nInvalid device size parameters passed.\n");
```
dthain: A library function should (generally) not print to stdout -- that's the job of main().
Instead, return failure and set errno=EINVAL.

dttools/src/disk_alloc.c
```diff
@@ -0,0 +1,230 @@
...
+		free(mount_args);
+		return -1;
+	}
+
+	//Mount Loop Device
+	result = mount(mount_args, loc, "ext2", 0, "");
+	if(result != 0) {
+		printf("\nFailed to mount/dev/loop%d: %s.\n", j, strerror(errno));
+		char *rm_dir_args = (char *) malloc(sizeof(char) * 400);
+		sprintf(rm_dir_args, "losetup -d /dev/loop%d; rm -r %s", j, loc);
```
dthain: This sequence of rm/free/free/free occurs multiple times in the code.
Can you see a way to avoid repeating it?

dttools/src/disk_alloc.c
```diff
@@ -0,0 +1,230 @@
...
+		return -1;
+	}
+
+	free(device_loc);
+	free(dd_args);
+	free(losetup_args);
+	free(mk_args);
+	free(mount_args);
+
+	printf("\nDisk space %s allocated on device: /dev/loop%d.\n", space, j);
```
dthain: Again, don't printf in a library, move the printf to main().

dttools/src/disk_alloc.c
dthain: Instead, use the debug() function to send a note to the debugging stream.

dttools/src/disk_alloc.c
```diff
@@ -0,0 +1,230 @@
...
+	sprintf(losetup_args, "losetup -d /dev/loop%d", dev_num);
+
+	printf("\nPreparing to clean disk allocation.\n");
+
+	//Loop Device Unmounted
+	result = umount2(loc, MNT_FORCE);
+	if(result != 0) {
+
+		printf("\nFailed to unmount /dev/loop%d: %s.\nStopping disk deallocation.\n", dev_num, strerror(errno));
+		free(losetup_args);
```
dthain: Again, can you find a way to avoid repeating free/free?

dthain:
Also, modify the Makefile to build disk_alloc.c as part of libdttools and disk_alloc_test.c as a test program.

dthain:
The code is looking much better, and simplified.

There is one last anti-pattern that I would like you to fix: using malloc to create a string of a constant size isn't any better than creating a fixed size array, and still leaves you with the burden of freeing it.

So, change code like this:

```
char *device_loc = (char *) malloc(sizeof(char) * 200); 
sprintf(device_loc,...);
free(device_loc);
```

To either this, with some reasonable maximum size:

```
char device_loc[SOME_MACRO];
sprintf(device_loc,...);
```

Or this, which generates the needed string on the fly:

```
char * device_loc;
device_loc = string_format( ... );
free(device_loc);
```

Then, do be consistent with the rest of dttools, call your library modules `disk_alloc.c` and `disk_alloc.h` (which become part of libdttools.a) and your main program `disk_alloc_main.c` which builds an executable called `disk_alloc`

dttools/src/disk_alloc.c
```diff
@@ -0,0 +1,230 @@
...
+#include "disk_alloc.h"
+
+int disk_alloc_create(char *loc, char *space) {
+
+	long long size;
+	char metric;
+	int result;
+
+	//Check for trailing '/'
+	if(loc[strlen(loc) - 1] == 47) {
```
batrick: Use:

https://github.com/cooperative-computing-lab/cctools/blob/master/dttools/src/path.h#L32

dttools/src/disk_alloc.c
```diff
@@ -0,0 +1,230 @@
...
+
+		loc[strlen(loc) - 1] = 0;
+	}
+
+	//Scale disk allocation to correct size
+	if(sscanf(space, "%lli%c", &size, &metric) == 2) {
```
batrick: Use instead:

https://github.com/cooperative-computing-lab/cctools/blob/master/dttools/src/stringtools.c#L222

PR #867, opened 2015-07-22
use enums instead of defines for better doc

btovar:
@annawoodard found that our API doc does not list the possible values of task states and task results. This is because we use define CONSTANT, and we have to manually update all of the documentation.

This pr changes these defines to an enum, so that the documentation is updated automatically.


dthain:
Looks good to me.  Do what want to convert the scheduling and reset flags into enumerations while you are at it?

btovar:
Will do.

PR #868, opened 2015-07-22
fix arg cores comparison

btovar:
While preparing for XSEDE tutorial @nhazekam found a problem with the submission workers. @nhazekam found that in  if [ $cores=0 ], the variable is assigned, rather than compared. @nhazekam also found the solution:  if [ $cores=0 ].

This pr implements the solution that @nhazekam found.

Credit to @nhazekam.

### DIFF
--- a/work_queue/src/pbs_submit_workers
+++ b/work_queue/src/pbs_submit_workers
@@ -21,7 +21,7 @@ pbs_parameters=""
 
 parse_arguments()
 {
-	if [ -z "$cores" -o $cores=0 ]
+	if [ -z "$cores" -o $cores = 0 ]
 	then
 		cores=1
 	fi
--- a/work_queue/src/pbs_submit_workers
+++ b/work_queue/src/pbs_submit_workers
@@ -21,7 +21,7 @@ pbs_parameters=""
 
 parse_arguments()
 {
-	if [ -z "$cores" -o $cores = 0 ]
+	if [ -z "$cores" -o "$cores" = 0 ]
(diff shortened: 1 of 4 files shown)


nhazekam:
As I am testing this, this works if cores is specified, but if cores is not specified nothing is set. I am looking for a more full solution.

work_queue/src/pbs_submit_workers
```diff
@@ -21,7 +21,7 @@ pbs_parameters=""
 
 parse_arguments()
 {
-	if [ -z "$cores" -o $cores=0 ]
+	if [ -z "$cores" -o $cores = 0 ]
```
nhazekam: if [ -z "$cores" -o [ $cores = 0 ]
The added [ allows it to work in both case but throws a shell error [: too many arguments

Reading into a solution for this.

dthain:
This is a classic problem in shell programming: substitution of variables is performed syntactically, not semantically.  That is to say, if you have this code:

```
if [ $cores = 0 ] ...
```

...and cores is not defined, then an empty string is substituted in its place, and the resulting code is:

```
if [ = 0 ] ...
```

...which doesn't parse correctly.  The usual solution is to quote variables wherever they are used:

```
if [ "$cores" = 0 ]
```

...or to prefix them in a silly way:

```
if [ X$cores = X ] ...
```

batrick:
Note that this prefix hack is evil and not equivalent when the variable has
spaces. Please use quotes.

PR #870, opened 2015-07-23
Modify slurm and torque submits for compatibility

nhazekam:
Small compatibility changes for XSEDE15. @btovar please checkout.


btovar:
So srun was not necessary after all?

nhazekam:
srun : I don't believe so. I tested with Stampede. Comet is not working currently, but it worked like this yesterday. 
pwd : The PWD change is to emulate the -d 'pwd'. I also made this change to the batch job cluster implementation of TORQUE and PBS.

PR #871, opened 2015-07-23
send exhaustion result per task when worker runs out of resources

btovar:
When a worker runs a out of resources, kill all the task and per task, send a "result ..." message to the master with result WORK_QUEUE_RESULT_RESOURCE_EXHAUSTION.

Two kind of limits are considered: hard, such as wall time, in which the worker aborts right away and the master does not get information about the tasks. And soft, in which the worker has the chance to send the results it knows, and the it aborts. A worker in this last state will not receive any more tasks from the master.


btovar:
Resubmitting in another pr.

Issue #872, opened 2015-07-24
Task directories not removed at the worker.


btovar:
When a task is no longer active at the master, @klannon reports that sometimes tasks leave their workspaces behind.

btovar:
Maybe fixed by #875.

btovar:
Solved by d946d9e707ffe3dfb060f9d51c65995693e5decf.

PR #873, opened 2015-07-24
Cleanup return codes

btovar:
Removes any 'if(result)' or 'if(result > x)'. Now say 'if(result == SUCCESS)' or alike.
Also, remove return -1, 0, 1 for messages. Now say, e.g.: return MSG_NOT_PROCESSED.

These changes eliminated already two bugs:
1) Sometimes task with app failures were treated like worker failures.
2) Some invalid messages were ignored without an error message.

The use of enumerations also allows better compiler messages in the most recent versions of gcc and clang.

This is motivated from the need to find the bug #872.


dthain:
Looks good to me.

PR #879, opened 2015-08-04
Irods don't ask for passwd with parrot

### LINKED ISSUE #725: Parrot with iRODS no longer asks for password
If we try to use an iRODS server, we expect to see:

````
$ ./parrot_run -d all /bin/ls /irods/serv/
Enter your current iRODS password:
````

This no longer works as, by default, Parrot runs the application (`ls`) in its own **foreground** process group. So when the iRODS library tries to read the password from the terminal, it stops indefinitely due to `SIGTTIN`. (Actually, the library waits on a forked child that stops indefinitely.)

One way to fix this is to use the recently added `--no-set-foreground` option to Parrot. Then we can do:

````
$ ./parrot_run --no-set-foreground /bin/ls /irods/host:port/
Enter your current iRODS password:
tempZone
````

Unfortunately this isn't completely correct. As soon as we execute a shell that makes its own process groups, it fails again.

We need to inform the user that reading the password from the terminal might not work, and to suggest a workaround.

### PR DESCRIPTION
**Pull request by btovar**
Based on #878.

Check if the user password file is created, and if not, advice the user to use iinit.

Fixes #725.

### DIFF
--- a/parrot/src/irods_reli.cc
+++ b/parrot/src/irods_reli.cc
@@ -35,11 +35,14 @@ extern "C" {
 #include "stringtools.h"
 }
 
+#include <string>
+
 extern int pfs_irods_debug_level;
 
 static struct hash_table *connect_cache = 0;
 static rodsEnv irods_env;
 static int got_irods_env = 0;
+static int got_irods_passwd_file = 0;
 static int irods_serial = 0;
 
 /* iRODS considers paths ending in a forward slash to be invalid and throws an error. */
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: parrot/src/irods_reli.cc [hunk 1]
```diff
@@ -117,6 +120,17 @@ static struct irods_server * connect_to_host( const char *hostport )
 		return 0;
 	}
 
+	if(!got_irods_passwd_file) {
+		std::string passwd_file(getenv("HOME"));
```
batrick: I think we need to handle HOME not being set. IIRC, in Condor, if you run Parrot directly without a shell, you don't get HOME.

I looked at iRODS's source and it appears they just concatenate the two though, without checking and no fallback. We should check though.

btovar:
Correct, they do not do anything special.

Good idea, HOME is checked now.

PR #880, opened 2015-08-05
Allow tests to be skipped.

btovar:


btovar:
Making changes, do not merge yet.

chirp/test/TR_chirp_python.sh
```diff
@@ -9,6 +9,14 @@ c="./hostport.$PPID"
 
 ticket=my.ticket
 
+check_needed()
+{
+	[ -f ../src/python/_CChirp.so ] || return 1
+	python=$(cctools_python -n 3 2.7 2.6 2> /dev/null)
+	[ -n "$python" ] || return 1
+	${python} -c "import json; import Chirp" 2> /dev/null
```
batrick: Why not just:

````
cctools_python <<EOF
# CCTOOLS_PYTHON_VERSION 2.7 2.6
import json
import Chirp
EOF
[ $? -ne 0 ] && return 1
````

Comment (batrick) [inline chirp/test/TR_chirp_python.sh]: This would just be so we don't need to change `cctools_python`.

**Comment by btovar** [comment]
Won't work without any modifications. cctools_python needs the name of the script.

Also, I like the new -n option, where we can easily get the location of the interpreter to use somewhere else.

File: run_all_tests.sh [hunk 2]
```diff
@@ -38,23 +39,36 @@ for package in ${CCTOOLS_PACKAGES_TEST}; do
 				printf "%-66s" "--- Testing ${package}/test/${script} ... "
 				TEST_START_TIME=$(date +%s)
 				(
-					echo "======== ${script} PREPARE ========"
-					"./${script}" prepare
-					result=$?
-					if [ $result -ne 0 ]; then
+					"./${script}" check_needed
+				)
```
batrick: This needs to be:

    ) >> "$CCTOOLS_TEST_LOG" 2>&1

chirp/test/TR_chirp_python.sh
batrick: One last change, I think remove the redirection to /dev/null. Otherwise it's not possible to know why the test was skipped.

btovar:
Done, thanks Patrick.

PR #886, opened 2015-08-10
Adds do_link to chirp_tool.c, fixes #885

### LINKED ISSUE #885: Chirp tool is missing a link/symlink command
A `ln [-s]` command should be added to Chirp tool for creating hard/symbolic links.

A temporary workaround is to use Parrot instead.

nkremerh:


dthain:
@batrick pls check

chirp/src/chirp_tool.c
```diff
@@ -446,6 +446,26 @@ static INT64_T do_getacl(int argc, char **argv)
...
 
+static INT64_T do_link(int argc, char **argv)
+{
+	INT64_T result;
+	char old_full_path[CHIRP_PATH_MAX];
+	char new_full_path[CHIRP_PATH_MAX];	
+	int sym = (argc == 4 && strcmp(argv[1], "-s") == 0);
+
+	if(sym) {
+		complete_remote_path(argv[2], old_full_path);
```
batrick: `argv[2]` is the target of the symbolic link and doesn't necessarily have to be a valid path. It should not be "fixed" by `complete_remote_path`. Consider:

````
chirp host:port <<EOF
chdir a/directory
ln -s /an/absolute/path ./foo
EOF
````

This should create a symbolic link `/a/directory/foo` that points to `/an/absolute/path`.

PR #889, opened 2015-08-11
Police task usage

btovar:
Rename functions related to worker resource measuring, to make clear they are about the host.

On the worker side:
- Size of files in the cache is subtracted from the size of the box needed by a wq process.
- If a task exhaust it limits, that task result is RESOURCE_EXHAUSTION. Other running tasks are stopped, with result FORSAKEN.

### DIFF
--- a/allpairs/src/allpairs_multicore.c
+++ b/allpairs/src/allpairs_multicore.c
@@ -23,7 +23,7 @@ See the file COPYING for details.
 #include "xxmalloc.h"
 #include "fast_popen.h"
 #include "text_list.h"
-#include "memory_info.h"
+#include "host_memory_info.h"
 #include "load_average.h"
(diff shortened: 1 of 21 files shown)


btovar:
Let me think about it, but I do not see how to do it for processes that fork children.

btovar:
Refactoring...

dthain:
Is this pull still relevant after re-working the task sandbox measurement?

btovar:
This is part of the re-working.

Perhaps we are thinking of different re-workings?

dthain:
Er, sorry, I was getting confused between pulls.
Is this one ready to be reviewed, or are you still working?

btovar:
Let me give it one more check.

btovar:
@dthain ready now

PR #892, opened 2015-08-13
Handle log rotations for multiprocess debugging.

batrick:
Before we used fstat to test the size of the log and then perform a rotation if
it exceeded the maximum. However, this does not handle the situation where the
log was rotated by another process. This would cause the log to be rotated
multiple times with each process working with its own log, all but one
anonymous files.

This commit performs a (racey) stat on the file path for rotation checks and
for testing whether the inode of the file has changed. If the inode has
changed, we reopen the log as it was rotated by a peer process.

### DIFF
--- a/dttools/src/debug_file.c
+++ b/dttools/src/debug_file.c
@@ -21,6 +21,7 @@ See the file COPYING for details.
 #include <string.h>
 
 static int file_fd = -1;
+static struct stat file_stat;
 static char file_path[PATH_MAX];
 static off_t file_size_max = 1<<20;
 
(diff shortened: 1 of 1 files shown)


btovar:
You are not sharing file descriptors between processes. How is that all of them can append to the same debug file? I would think that every time you want to write to the log, you need to open the file, lock it for writing (even if only advisory), and then close it.

batrick:
The first process opens the debug log and that fd is passed onto children. Each write requires a stat on the file to determine if the file has been rotated, by checking if the inode is different.

Open/Lock/Close would solve the race condition but it also makes it more expensive. I'm not sure it's worth it.

btovar:
Thanks for the explanation. Go ahead and merge it when you are ready.

PR #895, opened 2015-08-17
Proposal for external downloads at a worker

btovar:
Download the files until the task is going to run. This makes it easier to associate errors with a task.
Uploads would be handled in the same way. 

Changed to externals_* to make it easier to add other resources, such as chirp.

If there is an error, the result of the task is an OR or INPUT_MISSING and an appropriate description.

### DIFF
--- a/work_queue/src/work_queue.h
+++ b/work_queue/src/work_queue.h
@@ -62,7 +62,15 @@ typedef enum {
 	WORK_QUEUE_RESULT_SIGNAL         = 8,       /**< The task was terminated with a signal **/
 	WORK_QUEUE_RESULT_RESOURCE_EXHAUSTION = 16, /**< The task used more resources than requested **/
 	WORK_QUEUE_RESULT_TASK_TIMEOUT   = 32,      /**< The task ran after specified end time. **/
-	WORK_QUEUE_RESULT_UNKNOWN        = 64       /**< The task ran successfully **/
+	WORK_QUEUE_RESULT_UNKNOWN        = 64,      /**< The task encountered an unknown error. */
+
+	/* 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536, 131072, 262144 for future use */
+
+	WORK_QUEUE_RESULT_EXTERNAL_NO_RESOLVE    = 524288,    /**< External resource could not be resolved. */
+	WORK_QUEUE_RESULT_EXTERNAL_NO_CONNECT    = 1048576,   /**< Could not connect to external resource. */
+	WORK_QUEUE_RESULT_EXTERNAL_NO_AUTH       = 2097152,   /**< Could not authenticate to external resource. */
+	WORK_QUEUE_RESULT_EXTERNAL_WRITE_ERROR   = 4194304,   /**< Write error on external resource. */
+	WORK_QUEUE_RESULT_EXTERNAL_TIMEOUT       = 8388608    /**< Timeout when connecting to external resource. */
 } work_queue_result_t;
 
 typedef enum {
--- a/work_queue/src/work_queue_process.c
+++ b/work_queue/src/work_queue_process.c
@@ -1,6 +1,7 @@
 
 #include "work_queue_process.h"
-#include "work_queue.h"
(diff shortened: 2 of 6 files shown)


dthain:
I like the general approach, it makes sense for the task to own external downloads.  A few code comments inline.

work_queue/src/work_queue.h
```diff
@@ -62,7 +62,15 @@ typedef enum {
...
-	WORK_QUEUE_RESULT_UNKNOWN        = 64       /**< The task ran successfully **/
+	WORK_QUEUE_RESULT_UNKNOWN        = 64,      /**< The task encountered an unknown error. */
+
+	/* 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536, 131072, 262144 for future use */
+
+	WORK_QUEUE_RESULT_EXTERNAL_NO_RESOLVE    = 524288,    /**< External resource could not be resolved. */
+	WORK_QUEUE_RESULT_EXTERNAL_NO_CONNECT    = 1048576,   /**< Could not connect to external resource. */
+	WORK_QUEUE_RESULT_EXTERNAL_NO_AUTH       = 2097152,   /**< Could not authenticate to external resource. */
+	WORK_QUEUE_RESULT_EXTERNAL_WRITE_ERROR   = 4194304,   /**< Write error on external resource. */
+	WORK_QUEUE_RESULT_EXTERNAL_TIMEOUT       = 8388608    /**< Timeout when connecting to external resource. */
```
dthain: This space of errors seems tightly coupled to what curl can detect, and would likely change depending on the mechanism.  For example, is there really a fundamental difference between NO_CONNECT and TIMEOUT?  And will the master know whether to retry or return the task as done?

Suggest that this be reduced to EXTERNAL_INPUT_UNAVAILABLE and EXTERNAL_OUTPUT_UNAVAILABLE, which will help the user to troubleshoot where the problem lies.

work_queue/src/work_queue_process.c
```diff
@@ -268,4 +269,73 @@ void work_queue_process_kill(struct work_queue_process *p)
 	waitpid(p->pid, NULL, 0);
 }
 
+
+static int download_url(const char *url, const char *cache_name) {
+	debug(D_WQ, "Retrieving %s from (%s)\n", cache_name, url);
+
+	char command[WORK_QUEUE_LINE_MAX];
+	snprintf(command, WORK_QUEUE_LINE_MAX, "curl -f -o \"cache/%s\" \"%s\"", cache_name, url);
+
```
dthain: I predict that sooner or later, we wlil run on a machine where _curl_ is unavailable, and mysterious errors will result.  Suggest that you _debug()_ the command line and check for curl in the path.

batrick: There is now a dttools library "shell" which allows you to catch stdout/stderr. You could then output that to the debug log:

https://github.com/cooperative-computing-lab/cctools/blob/55f3b83d3b81cb51e929657699db83f450b06693/dttools/src/shell.h

Here's an example where I use it:

https://github.com/cooperative-computing-lab/cctools/blob/55f3b83d3b81cb51e929657699db83f450b06693/chirp/src/chirp_client.c#L549-L577

work_queue/src/work_queue.h
batrick: Is there a reason result codes are powers of two?

btovar: Yes. Without EXTERNAL_INPUT_UNAVAILABLE and EXTERNAL_OUTPUT_UNAVAILABLE, you could check better it was the input and output that could not be resolved, for example.

Now, we could report that a task did not send stdout, and output files.

But this is ugly, so we could just report the first error we encounter.

work_queue/src/work_queue_process.c
btovar: I'll use the shell from dttools.

btovar:
Holding on on this one while we rethink the  cache behavior.

dthain:
I think you can proceed on this now that the cache behavior is fixed.

btovar:
To be superseded.

PR #903, opened 2015-09-02
Turn some errors into resource exhaustion signals

btovar:
1) Keep last related error.
2) Capture write with ld preload.
3) Quit on no space and no file descriptors.


dthain:
Maybe I'm reading this incorrectly, but it looks to me like every errno from write is returned to the supervisor process, which then considers any errno at all to signal resource exhaustion.  I suspect this is a little too aggressive, since write does occasionally return EINTR, EAGAIN, or perhaps EBADF if the user does strange things.

I think you want to only consider it resource exhaustion when you explicitly get known errnos such as ENOSPC and EMFILE.  (or is it ENFILE?  I always have to look those two up.)

btovar:
@dthain, you are correct. The last two commits should address your comments.

btovar:
Do not merge yet, working on a bug.

btovar:
@dthain ready now. (autobuild failure is fixed by one of your commits already in master)

PR #904, opened 2015-09-02
Adds disk_alloc to Work Queue

nkremerh:


work_queue/src/work_queue_worker.c
```diff
@@ -869,17 +872,18 @@ static int do_task( struct link *master, int taskid, time_t stoptime )
 			}
 			free(env);
 		} else if(!strcmp(line,"end")) {
-			work_queue_process_compute_disk_needed(p);
+			//work_queue_process_compute_disk_needed(p);
```
dthain: Deliberate?

nkremerh: Our process, p, is not yet defined because we want our task to have all the necessary information (such as disk if specified). This is so the process can in turn use the information from the task to determine if it should use disk_alloc or create_dir.

dthain: Ok, that's fine, but don't leave the dead code in place.
Presumably work_queue_process_compute_disk_needed ought to be called at some later point?

dthain:
Upon further review, two more things:

1 - I think loop_mount should be a member of work_queue_process, not work_queue_task..  The latter is visible to users through the API, and I think this is not an API change.  (Also, are you sure it is getting set and cleared appropiately?)

2 - As a matter of caution, add a --disk-allocation option to the worker, so that allocations are only used when this option is given.  We don't want a bug in this to throw off normal users of the system.  After some testing time, we may turn it back on to be the default.

PR #907, opened 2015-09-03
Makeflow Sandbox Mode

dthain:
First pass at local execution in a sandbox.


dthain:
@nkremerh @nhazekam please take a look and see if this is on the right track.

dthain:
Still thinking about this one.  Is sandbox better off as a batch type, or as a kind of a wrapper?

btovar:
Given the current state of the code, a wrapper will be most useful. We can reuse it in the other batch systems easily, specially LOCAL.

Also, we can have a --strict flag, in which all task should run in sandbox mode. We can then advice the user that --strict is the way to go.

btovar:
This is of course, assuming we will be able to chain wrappers, which I think you want to do in the new json code.

dthain:
Updated to bring into sync with master.

dthain:
Closing this as the wrong approach.

PR #908, opened 2015-09-04
use system mask for group permissions

btovar:
As requested by @matz-e.


dthain:
I believe the general custom in Unix is to indicate mask 0777 in the code, and rely on the user's umask to reduce it.  Any reason not to go all the way to 0777?

PR #909, opened 2015-09-04
Umbrella support for self-contained specs

hmeng-19:
Add the self-contained cms_opendata umbrella spec

Add umbrella support for self-contained specs:
        1) Change the logic of the --packages option of umbrella. When a self-contained
spec is given, abstract metadata info into a separate metadata db.
        2) Move cctools binary and cms_siteconf info into umbrella source code from umbrella metadata database to allow easy shipping of umbrella spec and flexible resetting of cctools binary and cms_siteconf 
        3) reorder the options of umbrella

Modify umbrella.html
Modify umbrella example README commands
Remove umbrella options from working_principle


hmeng-19:
@dthain , this pull request add umbrella support for self-contained specs (here we assume the self-contained specs already be there).

The umbrella support for creating self-contained specs based on metadata db and  non self-contained specs will come up next.

hmeng-19:
@pivie , this pull request changes the logic of `--packages` option. 

You need to add the following option to your umbrella command:
    `--packages http://ccl.cse.nd.edu/software/umbrella/database/packages.json` 
By doing this, umbrella already use the umbrella metadata database pointed by the url, and you do not need to worry about creating your own umbrella metadata database.

If you want to specify your own database with the file path `/tmp/packages.json`, you can set the `--packages` option to be `file:///tmp/packages.json`.

PR #914, opened 2015-09-16
First pass at using hostname

nhazekam:
Due to difference in internal/external Stampede IPs we want to allow for using the hostname that can be resolved correctly from either location. @btovar Please take a look, particularly in Makeflow where I pass as I'm not positive this is the cleanest way to pass this.

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -105,6 +105,7 @@ static int port = 0;
 static int output_len_check = 0;
 
 static int cache_mode = 1;
+static int master_prefer_hostname = 0;
 
 static char *monitor_exe  = "resource_monitor_cctools";
 
@@ -1072,7 +1073,6 @@ int main(int argc, char *argv[])
 	timestamp_t time_completed = 0;
 	const char *work_queue_keepalive_interval = NULL;
 	const char *work_queue_keepalive_timeout = NULL;
-	const char *master_prefer_hostname = NULL;
 	const char *work_queue_master_mode = "standalone";
 	const char *work_queue_port_file = NULL;
 	const char *priority = NULL;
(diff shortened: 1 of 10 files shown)


makeflow/src/makeflow.c
```diff
@@ -1578,6 +1584,7 @@ int main(int argc, char *argv[])
 	batch_queue_set_option(remote_queue, "caching", cache_mode ? "yes" : "no");
 	batch_queue_set_option(remote_queue, "wait-queue-size", wq_wait_queue_size);
 	batch_queue_set_option(remote_queue, "working-dir", working_dir);
+	batch_queue_set_option(remote_queue, "master-prefer-hostname", master_prefer_hostname);
```
batrick: Make `master_prefer_hostname` an `int`. Set the the option to "yes" or "no" like we do for "caching" above.

work_queue/src/work_queue_worker.c
```diff
@@ -1796,7 +1797,8 @@ static int serve_master_by_name( const char *catalog_host, int catalog_port, con
 		last_port = port;
 
 		debug(D_WQ,"selected master with project=%s name=%s addr=%s port=%d",project,name,addr,port);
-
+		if(pref && pref == 1)
```
batrick: `if (pref)` ?

makeflow/src/makeflow.c
nhazekam: Do you suggest I then pass yes or no on to the Catalog or equate those to values (1 or 0)  in batch_job_work_queue.c?

batrick:
> Do you suggest I then pass yes or no on to the Catalog or equate those to values (1 or 0) in batch_job_work_queue.c?

You should do what's consistent with the rest of the code. [Do we use yes/no true/false 1/0 in boolean catalog fields?]

btovar:
Nick,
Looks good. Please also update wavefront, allpairs, the respective show_help's, and the man pages.

btovar:
Nick, did you also update the help and manual of makeflow?

nhazekam:
I updated the help message and man page for makeflow, allpairs, and wavefront. Is there anything I need to add for work queue?

btovar:
Nick,
A small change. IP and HOSTNAME give the idea that the user can write an IP or HOSTNAME.

Perhaps by_ip, by_hostname is better?

btovar:
I've just tried on stampede, and it works.

nhazekam:
@btovar Let me know how this looks. I tested on Stampede, but more testing never hurts.

btovar:
Looks good. Ready to merge if you are.

nhazekam:
I think its good to go.

Issue #925, opened 2015-09-22
Incorrect check for Python.h


btovar:
For swig bindings we ask for python-dev.

umbrella, prune, and weaver use the same check for verifying the python version. If python-dev is not installed, ./configure fails for umbrella, etc., even when these packages do not need it.

Supersedes #924

hmeng-19:
@ben, the failure happens on `vecnet-qa.crc.nd.edu`, (this is a private crc server used by Alex and Rob, two crc folks we recently started to collaborate.)
As you expected, this machine has no python-dev installed.

hmeng-19:
@ben, is it possible to release the checking for `python-dev` if a user only wants to install `umbrella` without `swig`?

hmeng-19:
BTW, it should be `python-devel`, not `python-dev`.

PR #928, opened 2015-09-23
check for python and python_dev separately

btovar:
@hmeng-19, can you give this one a go?

Fix proposal for #925


hmeng-19:
@ben, test it on a Amazon EC2 VM with the following identifer:
RHEL-6.5_GA_HVM-20140929-x86_64-11-Hourly2-GP2 (ami-11125e21)

The VM has no `python-devel` installed.

It works perfectly. After `./configure; make; make install`, I can see `umbrella`, `prune`, and `weaver` in the `bin` dir.

PR #934, opened 2015-09-24
wq tasks left

btovar:
As requested by @klannon, @awoodward, and @matz-e.

Applications such as Lobster submit tasks according to the number of workers available. This creates a chicken-and-egg problem when using work_queue_pool, as the pool does not grow if there are no tasks available. Setting a minimum number of workers in the pool is not a good solution, as many workers may go underutilized with long tails.

This request gives the master the API work_queue_specify_num_tasks_left. work_queue_pool can then compute the number of tasks as: left + running + waiting. Lobster can then inform the catalog how many tasks are left to be created, and the pool can create the workers accordingly.

### DIFF
--- a/work_queue/src/perl/Work_Queue.pm
+++ b/work_queue/src/perl/Work_Queue.pm
@@ -136,6 +136,11 @@ sub specify_priority {
 	return work_queue_specify_priority($self->{_work_queue}, $priority);
 }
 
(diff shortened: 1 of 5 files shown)


dthain:
Looks good, go ahead and merge.

Longer term, I would like to think more carefully about what this means and how to name it in the API.

PR #937, opened 2015-09-25
Make special file descriptors invisible to tracee.

### LINKED ISSUE #927: voms-proxy-init reports that 'ls -al <PATH>' fails under parrot/cvmfs
Hi all,

following #919 and #921 we hit another block with the `voms-proxy-init` command.  In order to reproduce the steps, you can create a throw-away certificate and key with

    openssl req -x509 -newkey rsa:2048 -keyout /tmp/userkey.pem -out /tmp/usercert.pem -days 10 -nodes -subj "/"

With this in place, the following two steps under fuse-mounted cvmfs result in the expected error 'User credential is not valid!'.

    source /cvmfs/grid.cern.ch/emi3ui-latest/etc/profile.d/setup-ui-example.sh
    voms-proxy-init -voms cms -cert /tmp/usercert.pem -key /tmp/userkey.pem -debug

If ran under parrot, the error is 

    Credentials couldn't be loaded [/tmp/userkey.pem, /tmp/usercert.pem]: Cannot list properties for file '/tmp/userkey.pem': error invoking the 'ls -al /tmp/userkey.pem' os command!
    No credentials found!

I tried to create a debug log.  If the debug log is redirected to a file, the problem does not occur.  Perhaps this could be a hint to the cause of the issue.

Thanks for any help,
Jakob

batrick:
Special file descriptors (currently the channel and the Parrot directory) are
allocated from the top of our file descriptor pool. After setting up all
special file descriptors, the root tracee will lower its RLIMIT_NOFILE so that
special file descriptors are outside of its allocation/visibility.

Fixes #927.

### DIFF
--- a/parrot/src/pfs_channel.c
+++ b/parrot/src/pfs_channel.c
@@ -34,6 +34,8 @@ struct entry {
 #define CHANNEL_FMT "`%s':%" PRIx64 ":%zu"
 #define CHANNEL_FMT_ARGS(e) e->name, (uint64_t)e->start, (size_t)e->length
 
+extern int parrot_fd_start;
+
 static struct entry *head=0;
 static int channel_fd=-1;
 static char *channel_base=0;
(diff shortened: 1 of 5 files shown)


batrick:
"invisible" probably isn't the best term for this... What this commit does is move the special file descriptors above the tracee file descriptor limit. I think it's reasonable to expect applications will not reference these file descriptors. If they do try however, then Parrot will issue a warning:

https://github.com/batrick/cctools/commit/cad3c360689dba35591ba9ff36baf5e135719e46#diff-c7a7e7c1ffa8f36dc86019560e5798f9R1494

dthain:
It's like applying segmentation to the fd array...

batrick:
Yes that's more apt!

batrick:
@dthain good to merge?

PR #941, opened 2015-10-02
Blacklist pool

btovar:
As request by @klannon, @matz-e, and @annawoodard.

Calls like:

q.blacklist("babadook.net")
q.blacklist("hello.com")

are converted in work queue pool with condor in the submit file to:

requirements = ((machine != "babadook.net") && (machine != "hello.com"))

This means that once the worker idles out, the pool request no to get that machine again.

@dthain, @nhazekam


dthain:
I think it looks good -- the master is simply reporting what it has decided, and then the work_queue_pool takes action on it.

My only suggestion is that you add a printf or debug to the work_queue_pool so that the user has some idea that it is changing its behavior due to information from the master.

Issue #942, opened 2015-10-02
resource_monitor does not redirect > and < from command line.


btovar:
The following fails:

resource_monitor  --  gzip <  file > file.gz

The redirections are not sent to gzip.

btovar:
@nhazekam fyi, this is why your example does not work. For now, try an example that does not use redirection.

btovar:
Redirections from the command line work fine. It seems they only fail from inside work_queue. @nhazekam, can you confirm that --sh fixes the problem within makeflow/work_queue?

nhazekam:
I'll take a look at this. Thanks @btovar

btovar:
The monitor does work with redirection. It seems the problem is how we are internally constructing the command lines.

PR #945, opened 2015-10-05
adds --sh <str> to resource_monitor

btovar:
@nhazekam, @batrick:

Example:

resource_monitor -dall -Omy_mon_output --sh "date > date.output"

Partial solution to #942: This is the solution we should use inside makeflow. We still need to redirect >, < when the monitor is tested directly on a terminal.

### DIFF
--- a/doc/man/resource_monitor.m4
+++ b/doc/man/resource_monitor.m4
@@ -90,6 +90,7 @@ OPTIONS_BEGIN
 OPTION_TRIPLET(-d,debug,subsystem)Enable debugging for this subsystem.
 OPTION_TRIPLET(-o,debug-file,file)Write debugging output to this file. By default, debugging is sent to stderr (":stderr"). You may specify logs be sent to stdout (":stdout"), to the system syslog (":syslog"), or to the systemd journal (":journal").
 OPTION_TRIPLET(-i,interval,n)Interval between observations, in seconds (default=1).
+OPTION_TRIPLET(-c,sh,str)Read command line from <str>, and execute as '/bin/sh -c <str>'.
 OPTION_TRIPLET(-l,limits-file,file)Use maxfile with list of var: value pairs for resource limits.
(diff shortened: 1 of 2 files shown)


dthain:
IMHO, the underlying issue here is that the I/O redirection really doesn't belong inside of the command.  Suggest that we address this going forward by cleanly separating the specification of those things, at least when it is under our control.  e.g. as a data structure:

```
{ "cmd" = "date"; output = "date.output"; error = "date.error" }
```

Or as a command line:

```
-c "date" -o date.output -e date.error
```

batrick:
This issue is multilayered. We should discuss this in @dthain's office to bring him up to speed?

nhazekam:
As per out discussion earlier there seem to be three suitable options.

The first is to escape all special characters prior to wrapping a command. This would end up with us running through the command string and adding a backwhacks before [ \ ' " $ ] these (possibly more if I missed them) characters. The escapes are then dropped as each command is interpreted as we execute.

The second option is to write the command into a file and replace the command with the execution of the file. In action, this would be writing to a uniquely named file and then setting the command to be sh UNIQUE_FILE or something similar. In Makeflow/WQ the file would also have to be added to input files and programs such as the Resource Monitor that watch the process would have to be aware of the changes.

The third options would be to use base64 encoding. In this case we would encode the command in base64 and send it to the execution site. Upon arriving at the site we would have the command decoded and then executed via "echo $cmd | base64 --decode | sh" or something more succinct for execution. 

@btovar @batrick We should find sometime tomorrow to talk over this. Let me know if something isn't clear.

btovar:
@nhazekam, I talked with @dthain about this option. The plan is to use --sh for command lines we would otherwise need to parse, as in makeflow and work queue, and to think about something more comprenhensive for everything else.

dthain:
FYI, I spoke with @nhazekam  and @batrick after our discussion, and some more issues came up.  I'll let you three talk it over and make a decision.

btovar:
To be superseded.

PR #954, opened 2015-10-07
Shell escape

nhazekam:
@btovar Take a look. Slightly confusing, but there was an older escape_shell_string that I am told is used for hadoop. Naming changes and whatnot can be addressed as needed just wanted to get this to you.

### DIFF
--- a/dttools/src/stringtools.c
+++ b/dttools/src/stringtools.c
@@ -9,6 +9,7 @@ See the file COPYING for details.
 #include "stringtools.h"
 #include "timestamp.h"
 #include "xxmalloc.h"
+#include "buffer.h"
 
 #include <assert.h>
 #include <ctype.h>
@@ -44,6 +45,27 @@ char *escape_shell_string(const char *str)
 	}
 	strcpy(current, "'");
 	return escaped_string;
+}
+
+char *string_escape_shell( const char *str )
+{
+	buffer_t buffer;
+	buffer_init(&buffer);
+
+	char *s;
+	buffer_putlstring(&buffer,"\"",1);
+	for(s=str;*s;s++) {
+		if(*s=='"' || *s=='\\' || *s=='$' || *s=='`')
+			buffer_putlstring(&buffer,"\\",1);
+		buffer_putlstring(&buffer,s,1);
+	}
+	buffer_putlstring(&buffer,"\"",1);
+
+	char *result;
+	buffer_dup(&buffer,&result);
+	buffer_free(&buffer);
+
+	return result;
 }
 
 void string_from_ip_address(const unsigned char *bytes, char *str)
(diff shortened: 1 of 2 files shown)


dthain:
@nhazekam it would be helpful if you can look up some reference for `/bin/sh` so that you can be sure you have identified everything that needs to be escaped.

nhazekam:
As per @batrick 's suggestion I looked at the opengroup documentation of the Shell Command Langauge.

In section 2.2.3 on Double-Quoted Strings, it indicates you only need to escape dollar sign, backtick, and backslash. I also escape double quote as we are adding and exterior double quote around the string.

http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02

I can look at other resources/add more if I missed any.

dttools/src/stringtools.c
```diff
@@ -46,6 +47,27 @@ char *escape_shell_string(const char *str)
 	return escaped_string;
 }
 
+char *string_escape_shell( const char *str )
+{
+	buffer_t buffer;
```
batrick: I find it helpful to use an array here so you don't need to use `&` everywhere:

`buffer_t B[1];`

then you can do:

`buffer_putliteral(B, "foo");`

dttools/src/stringtools.c
```diff
@@ -46,6 +47,27 @@ char *escape_shell_string(const char *str)
...
+	buffer_init(&buffer);
+
+	char *s;
+	buffer_putlstring(&buffer,"\"",1);
```
batrick: Use `buffer_putliteral(&buffer, "\"");`.

dthain:
That's great, pls add that reference to the code..

dttools/src/stringtools.c
```diff
@@ -46,6 +47,27 @@ char *escape_shell_string(const char *str)
 	return escaped_string;
 }
 
+char *string_escape_shell( const char *str )
+{
+	buffer_t buffer;
+	buffer_init(&buffer);
```
batrick: I'm guessing you also want `bufer_abortonfailure` since you're not checking for (memory) errors when using the buffer.

nhazekam:
As per @batrick 's suggestion I looked at the opengroup documentation of
the Shell Command Langauge.

In section 2.2.3 on Double-Quoted Strings, it indicates you only need to
escape dollar sign, backtick, and backslash. I also escape double quote as
we are adding and exterior double quote around the string.

http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02

dttools/src/stringtools.c
```diff
@@ -839,14 +861,24 @@ char * string_wrap_command( const char *command, const char *wrapper_command )
 {
 	if(!wrapper_command) return strdup(command);
 
-	char * result = malloc(strlen(command)+strlen(wrapper_command)+2);
 	char * braces = strstr(wrapper_command,"{}");
+	char * square = strstr(wrapper_command,"[]");
+
+	if(square)
+		command = string_escape_shell(command);
```
batrick: forgot `free(command)`

dthain:
Don't forget to add something to the manual.

btovar:
Superseded by #956

PR #956, opened 2015-10-08
Monitor cleanup

btovar:
@nhazekam, here it is.

It includes #954 and #945.


btovar:
@nhazekam: This should be ready to be tested.

btovar:
@nhazekam: Note that 1ac36c7 is all of #954.

nhazekam:
@btovar Works well for Work Queue and Makeflow, except when Makeflow is run locally. I appears you find the full path aka /afs/nd.edu/... to the resource monitor and then prepend that with ./ causing Makeflow to not find it.

Also, in the documentation for Makeflow we say -M, --monitor=<dir> for using monitor, but -M is already reserved for Master name.

Also, as per @dthain's comment on the #954 we should update the manual for Makeflow showing the allowance for [] to be a shell escaped command for wrapping. @dthain let me know if there should be more documentation for these changes.

batrick:
Make sure to update the man page too.

PR #957, opened 2015-10-08
Parrot Dynamic Mounts

dthain:
Added ability to mount/unmount the namespace at runtime.

@batrick @hmeng-19 please take a look.

Q: Should this be enabled or disabled by default?


parrot/src/parrot_mount.c
```diff
@@ -0,0 +1,84 @@
...
+	printf("Use: parrot_mount [OPTIONS] PATH DEST\n");
+	printf("Where options are:\n");
+	printf("-d --debug <flags>  Enable debugging for this subsystem.\n");
+	printf("-v --version        Show version number.\n");
+	printf("-h --help           Help: Show these options.\n");
+}
+
+int main( int argc, char *argv[] )
+{
+	char c;
```
batrick: should be `int c;`

parrot/src/parrot_mount.c
```diff
@@ -0,0 +1,84 @@
...
+		show_help();
+		return 1;
+	}
+
+	const char *relpath = argv[optind];
+	const char *destination = argv[optind+1];
+	const char *path;
+
+	if(relpath[0]!='/') {
+		const char * cwd = path_getcwd();
```
batrick: `cwd` should be freed.

batrick:
I wonder, since this is just an `MS_BIND` mount, perhaps we should just override the mount syscall?

[This could pave the way for some really crazy stuff like mounting an ext4 file system in Parrot... hint hint @nkremerh]

batrick:
http://serverfault.com/questions/403175/possible-to-mount-an-ext4-partition-image-via-fuse

(Mounting ext4 from userspace.)

dthain:
Combined everything into one tool.

To add a mount:
```
parrot_mount /mydata /chirp/whatever/mydata
```
To remove a mount:
```
parrot_mount --unmount /mydata
```
To fix the mount table as it is:
```
parrot_mount --disable
```
To prevent all dynamic mounting at runtime:
```
parrot_run --no-dynamic-mounts bash
```

dthain:
@jblomer would this be useful to you?

jblomer:
@dthain I understand this would give dynamic access to the -M option?  I think this would be very useful for symbolic links across file systems, for instance in our docker container (/etc/foo --> /cvmfs/bar).

dthain:
I will give #958 a chance to develop before merging this one...

dthain:
Doing a manual merge..

PR #958, opened 2015-10-09
Add permissions to parrot mountlist.

trshaffer:
This feature just sets up some mounts behind the scenes to prevent access to paths not explicitly listed in the whitelist file it's given. I've tested it with user-defined mounts (via mountfile and via command line option) and everything seems to work as expected. The executable being run has to be in one of the allowed paths, and will most probably require some libraries. A minimal whitelist like the following is enough to run simple programs, e.g. dash.

/bin
/lib
/lib64
/usr/bin
/usr/lib
/usr/lib64

### DIFF
--- a/parrot/src/pfs_table.cc
+++ b/parrot/src/pfs_table.cc
@@ -184,7 +184,7 @@ int pfs_table::bind( int fd, char *lpath, size_t len )
 
 	/* Resolve the path... */
 	struct pfs_name pname;
-	if (!resolve_name(1, lpath, &pname, F_OK)) //TODO check this
+	if (!resolve_name(1, lpath, &pname, F_OK))
 		return -1;
 
 	if (!pname.is_local)
@@ -389,7 +389,7 @@ void pfs_table::follow_symlink( struct pfs_name *pname, int depth )
 				name_to_resolve = absolute_link_target;
 			}
 		}
-		if (resolve_name(0, name_to_resolve, &new_pname, X_OK, true, depth + 1)) { //TODO check this
+		if (resolve_name(0, name_to_resolve, &new_pname, X_OK, true, depth + 1)) {
 			*pname = new_pname;
 		}
 	}
@@ -1171,7 +1171,7 @@ int pfs_table::access( const char *n, mode_t mode )
 	pfs_name pname;
 	int result = -1;
 
-	if(resolve_name(0,n,&pname,X_OK)) {
+	if(resolve_name(0,n,&pname,X_OK | mode)) {
 		result = pname.service->access(&pname,mode);
 	}
 
@@ -1183,7 +1183,7 @@ int pfs_table::chmod( const char *n, mode_t mode )
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,n,&pname,X_OK|W_OK)) {
+	if(resolve_name(0,n,&pname,W_OK)) {
 		result = pname.service->chmod(&pname,mode);
 	}
 
@@ -1195,7 +1195,7 @@ int pfs_table::chown( const char *n, uid_t uid, gid_t gid )
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,n,&pname,X_OK|W_OK)) {
+	if(resolve_name(0,n,&pname,W_OK)) {
 		result = pname.service->chown(&pname,uid,gid);
 	}
 
@@ -1216,7 +1216,7 @@ int pfs_table::lchown( const char *n, uid_t uid, gid_t gid )
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,n,&pname,X_OK|W_OK,false)) {
+	if(resolve_name(0,n,&pname,W_OK,false)) {
 		result = pname.service->lchown(&pname,uid,gid);
 	}
 
@@ -1228,7 +1228,7 @@ int pfs_table::truncate( const char *n, pfs_off_t offset )
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(1,n,&pname,X_OK|W_OK)) {
+	if(resolve_name(1,n,&pname,W_OK)) {
 		result = pname.service->truncate(&pname,offset);
 	}
 
@@ -1240,7 +1240,7 @@ ssize_t pfs_table::getxattr (const char *path, const char *name, void *value, si
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,path,&pname,X_OK)) {
+	if(resolve_name(0,path,&pname,R_OK)) {
 		result = pname.service->getxattr(&pname,name,value,size);
 	}
 
@@ -1252,7 +1252,7 @@ ssize_t pfs_table::lgetxattr (const char *path, const char *name, void *value, s
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,path,&pname,X_OK,false)) {
+	if(resolve_name(0,path,&pname,R_OK,false)) {
 		result = pname.service->lgetxattr(&pname,name,value,size);
 	}
 
@@ -1271,7 +1271,7 @@ ssize_t pfs_table::listxattr (const char *path, char *list, size_t size)
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,path,&pname,X_OK)) {
+	if(resolve_name(0,path,&pname,R_OK)) {
 		result = pname.service->listxattr(&pname,list,size);
 	}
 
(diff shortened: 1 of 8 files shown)


dthain:
@trshaffer you are off to a good start, and I am glad you are into the code.

As we discussed today, I would like to push this in a somewhat different direction:

We want to be able to modify both the namespace and the access control separately.  It should be possible to run a process with different permissions, but seeing the same namespace.  This will allow makeflow to generate a simple access control list so that the process can run safely, but without copying any files around.

I think you can accomplish this by merging the mountlist and the whitelist into a single file, where each line gives the logical path, the physical path, and then (optionally) permissions, something like this:

```
/data                                /path/to/some/other/data   R
/home/dthain/input1.txt    /home/dthain/input1.txt     R
/home/dthain/output2.txt  /home/dthain/output2.txt  RW
*                                       *                                       DENY
```

Then, the trick is to modify `pfs_table::open` to respect the access control flags, by matching O_RDONLY, O_WRONLY, and O_RDWR to the third field of the mountlist.

dthain:
Looking better!  As we discussed, please convert the modes to be symbolic (RWX) and add some documentation.  When you think it is ready, please ask @batrick for a code review.

parrot/src/pfs_table.cc
```diff
@@ -1366,8 +1378,11 @@ int pfs_table::unlink( const char *n )
 {
 	pfs_name pname;
 	int result = -1;
+	char dirname[PFS_PATH_MAX];
+
+	path_dirname(n,dirname);
 
-	if(resolve_name(0,n,&pname,false)) {
+	if(resolve_name(0,dirname,&pname,X_OK|W_OK,false) && resolve_name(0,n,&pname,F_OK,false)) {
```
batrick: The two checks here are for these example situations:

````
/a/b R_OK
````

`unlink("/a/b/c") = -1 (EPERM)`

and

````
/a/b/c /etc/services R_OK
````

`unlink("/a/b/c") = 0`

Yes?

File: parrot/src/pfs_table.cc [hunk 2]
```diff
@@ -1479,7 +1513,12 @@ int pfs_table::symlink( const char *n1, const char *n2 )
 	verbatim down to the needed driver.
 	*/
 
-	if(resolve_name(0,n2,&pname,false)) {
+	path_dirname(n2,dirname2);
+
+	// Require write on the target to prevent linking into a RW directory
+	// and bypassing restrictions
+	if(resolve_name(0,n1,&pname,W_OK,false) &&
```
batrick: I'd rather not have this check on the target of the symlink (`n1`). Correctly handling path resolution is the right way to fix this and that's a separate problem.

parrot/src/pfs_table.cc
```diff
@@ -1599,8 +1647,11 @@ int pfs_table::mkalloc( const char *n, pfs_ssize_t size, mode_t mode )
 {
 	pfs_name pname;
 	int result=-1;
+	char dirname[PFS_PATH_MAX];
+
+	path_dirname(n,dirname);
 
-	if(resolve_name(1,n,&pname)) {
+	if(resolve_name(0,dirname,&pname,X_OK|W_OK) && resolve_name(0,n,&pname,F_OK)) { //TODO check this
```
batrick: This seems to be a common check in the code: modifying/creating the directory entry for the given path. I suggest creating a new flag, `E_OK`, which tests this and combine the check in a single call to `resolve_name`.

parrot/src/pfs_table.cc
```diff
@@ -2056,7 +2107,7 @@ int pfs_table::locate( const char *n, char *buf, int length )
 		if(loc) delete(loc);
 		loc = 0;
 
-		if(resolve_name(0, n, &pname)) {
+		if(resolve_name(0, n, &pname, X_OK)) { //TODO check this
```
batrick: Why do we use `X_OK` in some places and `F_OK` in others?

batrick:
@trshaffer the last thing I'll want you to do is make a test in `parrot/test/` that tests a few different possible situations: files which are read-only are actually read-only, unlinking a file in a read-only directory, renaming a file, and anything else you think of.

Good job on the code!

parrot/src/pfs_table.cc
trshaffer: Right, we need to check permissions on the parent if we change that directory entry rather than the item itself. That's done with the E_OK flag as of commit fe8825e13d31c38f64108cbb0bd0e8706df14a2b

parrot/src/pfs_table.cc
trshaffer: Changed in fe8825e13d31c38f64108cbb0bd0e8706df14a2b

parrot/src/pfs_table.cc
trshaffer: Changed in fe8825e13d31c38f64108cbb0bd0e8706df14a2b

parrot/src/pfs_table.cc
trshaffer: Generally if there's an explicit mention in the documentation that no permissions are required on the item itself, I used F_OK (no checks) and something more restrictive on the parent. Also, for some of the Parrot-specific calls, I wasn't sure about the correct permissions, so I marked those parts with TODO and put something reasonable, or maybe a bit lenient

parrot/src/pfs_table.cc
```diff
@@ -1137,7 +1171,7 @@ int pfs_table::access( const char *n, mode_t mode )
 	pfs_name pname;
 	int result = -1;
 
-	if(resolve_name(0,n,&pname)) {
+	if(resolve_name(0,n,&pname,X_OK)) {
 		result = pname.service->access(&pname,mode);
```
batrick: Also, the mode passed to resolve_name will need to be OR'd with the mode passed to `pname.service->access`.

parrot/src/pfs_table.cc
```diff
@@ -1161,7 +1195,7 @@ int pfs_table::chown( const char *n, uid_t uid, gid_t gid )
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,n,&pname)) {
+	if(resolve_name(0,n,&pname,X_OK|W_OK)) {
```
batrick: Just W_OK.

parrot/src/pfs_table.cc
```diff
@@ -1182,7 +1216,7 @@ int pfs_table::lchown( const char *n, uid_t uid, gid_t gid )
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,n,&pname,false)) {
+	if(resolve_name(0,n,&pname,X_OK|W_OK,false)) {
```
batrick: W_OK

parrot/src/pfs_table.cc
```diff
@@ -1194,7 +1228,7 @@ int pfs_table::truncate( const char *n, pfs_off_t offset )
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(1,n,&pname)) {
+	if(resolve_name(1,n,&pname,X_OK|W_OK)) {
```
batrick: W_OK

parrot/src/pfs_table.cc
```diff
@@ -1206,7 +1240,7 @@ ssize_t pfs_table::getxattr (const char *path, const char *name, void *value, si
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,path,&pname)) {
+	if(resolve_name(0,path,&pname,X_OK)) {
```
batrick: Just R_OK

batrick: and the other get/list xattrs

parrot/src/pfs_table.cc
```diff
@@ -1268,7 +1302,7 @@ int pfs_table::setxattr (const char *path, const char *name, const void *value,
 	pfs_name pname;
 	int result=-1;
 
-	if(resolve_name(0,path,&pname)) {
+	if(resolve_name(0,path,&pname,X_OK|W_OK)) {
```
batrick: Just W_OK, and the other setxattrs

parrot/src/pfs_table.cc
batrick: I think generally just use F_OK unless there's a clear need for X_OK. chdir/listdir are the only ones that come to mind for me.

trshaffer:
I've adjusted the permissions to use less X_OK, and added a test script with come common use cases. I'm not aware of any other outstanding issues

dthain:
@batrick What do you think?

batrick:
I've been looking through it. I have a few changes I've added and then will merge.

batrick:
@trshaffer, I added several changes to your branch before merging. Some were necessary to fix test failures. Please have a look when you get a chance.

PR #969, opened 2015-10-15
Updated file logging and deleting for Makeflow

nhazekam:
@btovar let me know. This has some larger changes to wrappers and file tracking.

### DIFF
--- a/dttools/src/stringtools.c
+++ b/dttools/src/stringtools.c
@@ -722,6 +722,71 @@ void string_replace_backslash_codes(const char *a, char *b)
 	*b = 0;
 }
 
+char *string_replace_percents( char *str, const char *replace )
+{
+	/* Common case: do nothing if no percents. */
+	if(!strchr(str,'%')) return str;
+
+	buffer_t buffer;
+	buffer_init(&buffer);
+
+	char *s;
+	for(s=str;*s;s++) {
+		if(*s=='%' && *(s+1)=='%' ) {
+			if( *(s+2)=='%' && *(s+3)=='%') {
+				buffer_putlstring(&buffer,"%%",2);
+				s+=3;
+			} else {
+				buffer_putstring(&buffer,replace);
+				s++;
+			}
+		} else {
+			buffer_putlstring(&buffer,s,1);
+		}
+	}
+
+	char *result;
+	buffer_dup(&buffer,&result);
+	buffer_free(&buffer);
+
+	free(str);
+
+	return result;
+}
+
+
+/*
+char *string_replace_percents( char *str, const char *replace )
+{
+	/* Common case: do nothing if no percents. /
+	if(!strchr(str,'%')) return str;
+
+	char *r = replace;
+	char *s = str;
+	char *b;
+	while(*s) {
+		if(*s=='%' && *(s+1)=='%' ) {
+			if( *(s+2)=='%' && *(s+3)=='%') {
+				*b++ = '%';
+				*b++ = '%';
+				s+=4;
+			} else {
+				*r = replace;
+				while(*r){
+					*b++ = *r++;
+				}
+				s+=2;
+			}
+		} else {
+			*b++ = *s++;
+		}
+	}
+	free(str);
+	*b = 0;
+	return b;
+}
+*/
+
 int strpos(const char *str, char c)
 {
 
--- a/dttools/src/stringtools.c
+++ b/dttools/src/stringtools.c
@@ -754,39 +754,6 @@ char *string_replace_percents( char *str, const char *replace )
 	return result;
 }
 
-
-/*
-char *string_replace_percents( char *str, const char *replace )
-{
-	/* Common case: do nothing if no percents. /
-	if(!strchr(str,'%')) return str;
-
-	char *r = replace;
-	char *s = str;
-	char *b;
-	while(*s) {
-		if(*s=='%' && *(s+1)=='%' ) {
-			if( *(s+2)=='%' && *(s+3)=='%') {
-				*b++ = '%';
-				*b++ = '%';
-				s+=4;
-			} else {
-				*r = replace;
-				while(*r){
-					*b++ = *r++;
-				}
-				s+=2;
-			}
-		} else {
-			*b++ = *s++;
-		}
-	}
-	free(str);
-	*b = 0;
-	return b;
-}
-*/
-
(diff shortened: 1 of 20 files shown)


dthain:
This may seem a bit excessive, but I see many wrapper types in the future.  Perhaps you could organize things even further by splitting into `makeflow_wrapper.[ch]`, `makeflow_wrapper_docker.[ch]`, `makeflow_wrapper_monitor.[ch]` and so forth?

dttools/src/stringtools.c
```diff
@@ -722,6 +722,38 @@ void string_replace_backslash_codes(const char *a, char *b)
...
+		} else {
+			buffer_putlstring(&buffer,s,1);
+		}
+	}
+
+	char *result;
+	buffer_dup(&buffer,&result);
+	buffer_free(&buffer);
+
+	free(str);
```
btovar: You don't want to free str here.

dttools/src/stringtools.c
```diff
@@ -722,6 +722,38 @@ void string_replace_backslash_codes(const char *a, char *b)
...
+				buffer_putstring(&buffer,replace);
+				s++;
+			}
+		} else {
+			buffer_putlstring(&buffer,s,1);
+		}
+	}
+
+	char *result;
+	buffer_dup(&buffer,&result);
```
btovar: I usually do: char *result = xxstrdup(buffer_tostring(&buffer)); @batrick?

dttools/src/stringtools.c
```diff
@@ -722,6 +722,38 @@ void string_replace_backslash_codes(const char *a, char *b)
 	*b = 0;
 }
 
+char *string_replace_percents( char *str, const char *replace )
+{
+	/* Common case: do nothing if no percents. */
+	if(!strchr(str,'%')) return str;
```
batrick: This needs to be `return xxstrdup(str);`. Otherwise the caller may double-free `str` or try to free a stack allocated `str`.

dttools/src/stringtools.c
```diff
@@ -722,6 +722,38 @@ void string_replace_backslash_codes(const char *a, char *b)
 	*b = 0;
 }
 
+char *string_replace_percents( char *str, const char *replace )
```
batrick: `const char *str`

dttools/src/stringtools.c
```diff
@@ -722,6 +722,38 @@ void string_replace_backslash_codes(const char *a, char *b)
...
+{
+	/* Common case: do nothing if no percents. */
+	if(!strchr(str,'%')) return str;
+
+	buffer_t buffer;
+	buffer_init(&buffer);
+
+	char *s;
+	for(s=str;*s;s++) {
+		if(*s=='%' && *(s+1)=='%' ) {
```
btovar: s+n may be out of bounds.

makeflow/src/makeflow.c
```diff
@@ -89,7 +80,8 @@ static int makeflow_retry_flag = 0;
 static int makeflow_retry_max = MAX_REMOTE_JOBS_DEFAULT;
 
 static makeflow_gc_method_t makeflow_gc_method = MAKEFLOW_GC_NONE;
-static int makeflow_gc_param = -1;
+static uint64_t makeflow_gc_size   = 0;
```
btovar: Now that you are looking at this code, please add comments to all these gc related variables.

dttools/src/stringtools.c
batrick: I think `buffer_dup` should be preferred right? It handles embedded NULs and memory allocation failures.

makeflow/src/makeflow.c
```diff
@@ -134,29 +112,42 @@ static int file_creation_patience_wait_time = 0;
...
+ *   * replacement for monitor file names. */
+static struct list *makeflow_generate_input_files( struct dag_node *n, struct makeflow_wrapper *w, struct makeflow_monitor *m )
 {
-	if(!wrapper_command) {
-		wrapper_command = strdup(cmd);
-	} else {
-		wrapper_command = string_wrap_command(wrapper_command,cmd);
-	}
-}
+    struct list *result = list_duplicate(n->source_files);
```
btovar: Why do you need the duplicate here?

makeflow/src/makeflow.c
```diff
@@ -806,19 +645,21 @@ static void makeflow_node_complete(struct dag *d, struct dag_node *n, struct bat
...
-				makeflow_file_clean(d, remote_queue, f, 1);
+				makeflow_clean_file(d, remote_queue, f, 1);
 			} else {
-				makeflow_file_clean(d, remote_queue, f, 0);
+				makeflow_clean_file(d, remote_queue, f, 0);
 			}
 		}
 
-		if(monitor_mode && info->exit_code == 147)
+		if(monitor && info->exit_code == 147)
```
btovar: Ah, I forgot about this ugliness. In the future we want exit_code = 0 if task did not overflow, -1 otherwise, and get the task status from information that the monitor generates.

makeflow/src/makeflow_wrapper.c
```diff
@@ -0,0 +1,274 @@
...
+/*
+ * Creates a wrapper command with the appropriate resource monitor string for a given node.
+ * Returns a newly allocated string that must be freed.
+ * */
+
+char *makeflow_rmonitor_wrapper_command( struct makeflow_monitor *m, struct dag_node *n )
+{
+	char *executable;
+	if(m->exe_remote && !n->local_job){
+		executable = string_format("./%s", m->exe_remote);
```
btovar: I wonder if it is better to just have "%s", and add the "./" to exe_remote. Right now this function changes the namespace of exe_remote, which might be undesirable.

dttools/src/stringtools.c
btovar: Ok, I should be using buffer_dup too, then.

makeflow/src/makeflow.c
nhazekam: @btovar I was creating a duplicate because I was adding to the list and we don't want the original list on the node to be modified. What would be preferable to duplicate?

btovar:
@nhazekam: look good. I ran tests with it all this morning. Ready to merge if you are.

nhazekam:
Thanks @btovar. I think it's ready to go.

PR #978, opened 2015-10-21
Fix maximum_end_time handling.

matz-e:
Fix overflow.  The altered conditional always results to true, I think this may be due to the subtraction of two `uint64_t`s?

Without this, tasks with a specified end time terminate immediately.


btovar:
Thanks @matz-e!

PR #982, opened 2015-10-22
adds stats for workers lost, fast aborted, or idled out

btovar:
As requested by @klannon to replace grep-fu with the debug output.

Adds as to the queue stats:

total_workers_lost                 (e.g., evicted)
total_workers_fast_aborted
total_workers_idled_out


dthain:
Looks good.
This will be a prime place to use JSON instead.
(Working on that now.)

PR #987, opened 2015-10-23
Makeflow Local Resource Management

NeilB879:


makeflow/src/makeflow.c
```diff
@@ -1152,6 +1216,9 @@ int main(int argc, char *argv[])
 		{"help", no_argument, 0, 'h'},
 		{"makeflow-log", required_argument, 0, 'l'},
 		{"max-local", required_argument, 0, 'j'},
+		{"-max_local_disk", required_argument, 0, '1'},
```
nhazekam: You will want to make this correspond to a LONG_OPT instead of 1,2,3 to avoid confusion. You can also drop the '-' at the beginning of each as all long options have double hyphen leading automatically.

makeflow/src/makeflow.c
```diff
@@ -1541,7 +1617,18 @@ int main(int argc, char *argv[])
 	if(batch_queue_type == BATCH_QUEUE_TYPE_LOCAL) {
 		explicit_remote_jobs_max = explicit_local_jobs_max;
 	}
-
+	if(!loc_info.local_disk)
```
nhazekam: Due to the fact that you already initialize these values at line 1148, I don't think these will ever be executed. You will also want to set cores to load_average_get_cpus() initially as now it will be 1 by default.

makeflow/src/makeflow.c
```diff
@@ -76,6 +76,11 @@ an example.
 
 #define MAX_REMOTE_JOBS_DEFAULT 100
 
+struct {
```
nhazekam: @btovar suggestion, you could actually use an rmsummary struct in place of this newly created struct as it has fields for these values.

makeflow/src/makeflow.c
```diff
@@ -587,6 +593,54 @@ docker run --rm -m 1g -v $curr_dir:$default_dir -w $default_dir %s \"$@\"\n", im
...
+{
+	return (batch_queue_type == BATCH_QUEUE_TYPE_LOCAL || (n->local_job && local_queue));
+}
+static int makeflow_can_alloc_local(struct dag_node *n)
+{
+	int mem_ok = 0;
+	int disk_ok = 0;
+	int cores_ok = 0;
+	mem_ok = (loc_info.local_mem == -1 || !n->resources || n->resources->resident_memory >= loc_info.local_mem);
+	disk_ok = (loc_info.local_disk == -1 || !n->resources || n->resources->swap_memory >= loc_info.local_disk);
```
nhazekam: Instead of n->resource->swap_memory you will want to use n->resource->workdir_footprint as this is the value used to describe disk space in rmsummary.

makeflow/src/makeflow.c
```diff
@@ -692,6 +749,12 @@ static int makeflow_node_ready(struct dag *d, struct dag_node *n)
 	if(n->state != DAG_NODE_STATE_WAITING)
 		return 0;
 
+		 if(makeflow_is_local_node(n)){
```
nhazekam: Quick fix on indentation.

dthain:
Neil, the changes to makeflow for local resource management and the bug fix for work queue are two different features.  Please keep those two things separate by creating two distinct branches and pull requests.

dthain:
Neil, a small syntactic thing: please call the options `--local-cores` `--local-disk` and so forth.  Mixing dashes with underscores is frustrating to type and debug.

Also, please pay attention to the commit messages and the name of the pull request.  Make them (briefly) describe the intent of your changes, so that someone going through the history of commits can see what you did.  ";' and "fixed bugs" doesn't communicate any useful information.

dthain:
What you need to do here is `git pull --rebase origin master` to pull in changes from the repo.  This will result in conflicts, so you have to solve them, and continue.   Suggest that you sit down with @batrick, @nhazekam, or @btovar the first time you go through this.

batrick:
@NeilB879 is remaking this with a non-master branch.

PR #988, opened 2015-10-23
JX - JSON Expressions Library

dthain:
First pass at json implementation.  Don't merge yet, still playing around with it.

### DIFF
--- a/chirp/src/chirp_server.c
+++ b/chirp/src/chirp_server.c
@@ -223,9 +223,9 @@ static int update_all_catalogs(const char *url)
 	jx_insert_string (j,"backend",url);
 	jx_insert_string (j,"cpu",name.machine);
 	jx_insert_integer(j,"cpus", cpus);
-	jx_insert_float  (j,"load1",avg[0]);
-	jx_insert_float  (j,"load5",avg[1]);
-	jx_insert_float  (j,"load15",avg[2]);
+	jx_insert_double  (j,"load1",avg[0]);
+	jx_insert_double  (j,"load5",avg[1]);
+	jx_insert_double  (j,"load15",avg[2]);
 	jx_insert_integer(j,"memory_avail",memory_avail);
 	jx_insert_integer(j,"memory_total",memory_total);
 	jx_insert_integer(j,"minfree",minimum_space_free);
--- a/chirp/src/chirp_server.c
+++ b/chirp/src/chirp_server.c
@@ -243,7 +243,7 @@ static int update_all_catalogs(const char *url)
 
 	jx_insert(j,
 		jx_string("url"),
-		jx_format("chirp://%s:%d\n", hostname, chirp_port));
+		jx_format("chirp://%s:%d", hostname, chirp_port));
 
 	jx_insert(j,
 		jx_string("version"),
--- a/chirp/src/chirp_server.c
+++ b/chirp/src/chirp_server.c
@@ -31,6 +31,8 @@
 #include "host_disk_info.h"
 #include "host_memory_info.h"
 #include "json.h"
+#include "jx.h"
+#include "jx_print.h"
 #include "link.h"
 #include "list.h"
 #include "load_average.h"
@@ -192,7 +194,6 @@ static int gc_tickets(const char *url)
 
 static int update_all_catalogs(const char *url)
 {
-	buffer_t B[1];
 	struct chirp_statfs info;
 	struct utsname name;
 	int cpus;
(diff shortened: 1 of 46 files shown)


dttools/src/jx.c
```diff
@@ -0,0 +1,320 @@
...
+See the file COPYING for details.
+*/
+
+#include "jx.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+struct jx_pair * jx_pair( struct jx *key, struct jx *value, struct jx_pair *next )
```
batrick: An object's key must be a string. Probably the function should just take a `const char *`?

dttools/src/jx.c
```diff
@@ -0,0 +1,320 @@
...
+static struct jx * jx_create( jx_type_t type )
+{
+	struct jx *j = malloc(sizeof(*j));
+	j->type = type;
+	return j;
+}
+
+struct jx * jx_null()
+{
+	return jx_create(JX_NULL);
```
batrick: This could be a static const null struct to save some memory.

batrick: Nevermind that makes memory management much harder.

dttools/src/jx.c
dthain: For strict JSON, you are correct.  However, I am thinking of using this library to represent a superset of JSON, so I made things a bit more general.  For example, a non-string key would be handy for an external representation of an itable.  And, the symbol value type will allow for interesting substitutions.  It might be wise to add some options which control whether strict JSON or the superset is accepted...

dthain:
@btovar I converted work_queue_pool from json to jx in the latest commit.  As far as I can tell it works, but please check it for me.  Is there a way to simplify the use of macros there?

btovar:
The macro is simply for type checking. What about adding:

error = jx_object_lookup_int(struct jx *J, char *field, int *result);

that returns 0 if field is an int, and record it in result?

We do the same for double, string (with a max len), etc.

dthain:
I like that approach.

batrick:
Ah, seeing this I just realized that the JX library uses an `int` to hold integers. This really needs to be an `int64_t` so it can represent file sizes.

dthain:
Good point!

work_queue/src/work_queue_worker.c
```diff
@@ -1768,19 +1768,19 @@ static int serve_master_by_name( const char *catalog_host, int catalog_port, con
...
+		const char *pref = jx_lookup_string(jx,"master_preferred_connection");
+		int port = jx_lookup_integer(jx,"port");
 
 		/* Do not connect to the same master after idle disconnection. */
 		if(last_addr) {
 			if( time(0) > idle_stoptime && strcmp(addr, last_addr) == 0 && port == last_port) {
 				if(list_size(masters_list) < 2) {
 					free(last_addr);
-					/* convert idle_stoptime into connect_stoptime (e.g., time already served). */
+					/* cojxert idle_stoptime into connect_stoptime (e.g., time already served). */
```
pbui: This appears to be a search and replace fail.

work_queue/src/work_queue_worker.c
```diff
@@ -1975,7 +1975,7 @@ int main(int argc, char *argv[])
 			break;
 		case 'C':
 			if(!work_queue_catalog_parse(optarg, &catalog_host, &catalog_port)) {
-				fprintf(stderr, "The provided catalog server is invalid. The format of the '-C' option is '-C HOSTNAME:PORT'.\n");
+				fprintf(stderr, "The provided catalog server is ijxalid. The format of the '-C' option is '-C HOSTNAME:PORT'.\n");
```
pbui: Another search and replace fail.

dttools/src/jx.c
```diff
@@ -0,0 +1,410 @@
...
+}
+
+struct jx * jx_string( const char *string_value )
+{
+	struct jx *j = jx_create(JX_STRING);
+	j->string_value = strdup(string_value);
+	return j;
+}
+
+struct jx * jx_string_nodup( char *string_value )
```
batrick: Why not just `jx_stringf`? i.e. take a format + varargs.

work_queue/src/work_queue_worker.c
dthain: Good catch!

dttools/src/jx.c
dthain: The macro jx_format does exactly that.  (I would prefer jx_format to be a function but was running into some odd problems in passing a va_list down through the stack.  Maybe you can figure out how to make it a proper function.)

jx_string takes a plain string and we want to keep that option, otherwise someone will eventually get burned by trying to put a literal percent into a constant string.

chirp/src/chirp_server.c
```diff
@@ -215,37 +216,47 @@ static int update_all_catalogs(const char *url)
...
+
+	if (strlen(chirp_project_name)) {
+		jx_insert_string(j,"project",chirp_project_name);
+	}
 
-	buffer_free(B);
+	jx_insert(j,
+		jx_string("url"),
+		jx_format("chirp://%s:%d\n", hostname, chirp_port));
+
```
dthain: Example use of jx_format here.

dthain:
Rebased to latest master.

I believe this is as far as I want to go with this branch at the moment:

- Most uses of nvpair have been replaced with jx.
- `chirp_server`, `work_queue_worker` and `work_queue_pool` all report to the catalog via json.
- `work_queue_status` and `chirp_status` query json records from the catalog.
- Direct queries to `work_queue_worker` are done with json.
- The catalog server can accept and produce both json and nvpairs for backwards compatibility.

There is still some conversion to be done in the future:
- The catalog server should probably use json internally, but this would require porting the nvpair database and the deltadb query system.  Since we have several years of high value data in this format, this will require some greater care.

- The simpler json parser (instead of jx) is still used within `batch_job_chirp`, `chirp_job`, and `configua`. While this duplicates the jx library, it is currently working, and I don't want to disrupt @batrick's work unnecessarily.

And this should enable some new things going forward:
- JX objects to represent jobs within `makeflow` and `batch_job`
- JSON as a serialization between the WQ master and worker.
- JSON as an external intermediate language for makeflow.

Does anyone see any big problems that would prevent a merge?

batrick:
I think during your rebase you somehow lost the jx_format PR I sent you.

dthain:
Ok, checking on it...

dthain:
Couldn't quite figure out how to resurrect your PR, but here is the code.

dttools/src/jx.c
```diff
@@ -0,0 +1,410 @@
...
+	}
+}
+
+jx_int_t jx_lookup_integer( struct jx *object, const char *key )
+{
+	struct jx *j = jx_lookup(object,key);
+	if(j && jx_istype(j,JX_INTEGER)) {
+		return j->integer_value;
+	} else {
+		return 0;
```
btovar: 0 means:
- integer 0
- wrong type
- missing field

Let's have:
enum jx_status_t { JX_SUCCESS, ...

status = jx_lookup_integer(jx, key, &myint);

and so forth, for double, string, etc.

dttools/src/jx.c
```diff
@@ -0,0 +1,410 @@
...
+		case JX_INTEGER:
+			return j->integer_value==k->integer_value;
+		case JX_SYMBOL:
+			return !strcmp(j->symbol_name,k->symbol_name);
+		case JX_STRING:
+			return !strcmp(j->string_value,k->string_value);
+		case JX_ARRAY:
+			return jx_item_equals(j->items,k->items);
+		case JX_OBJECT:
+			return jx_pair_equals(j->pairs,k->pairs);
```
btovar: Compilers may complain about a missing default:

dttools/src/jx.c
```diff
@@ -0,0 +1,410 @@
...
+		case JX_INTEGER:
+			return jx_integer(j->integer_value);
+		case JX_SYMBOL:
+			return jx_symbol(j->symbol_name);
+		case JX_STRING:
+			return jx_string(j->string_value);
+		case JX_ARRAY:
+			return jx_array(jx_item_copy(j->items));
+		case JX_OBJECT:
+			return jx_object(jx_pair_copy(j->pairs));
```
btovar: Missing default, as above.

dttools/src/jx.c
btovar: Mmm, that would make the code in work_queue really ugly. What about:
jx_int_t jx_lookup_integer( struct jx *object, const char *key, ...);
use as:
myint = jx_lookup_integer(jx, key);
or:
myint = jx_lookup_integer(jx, key, &status);

dthain: Keep in mind that `jx_lookup_integer` is already a common-case wrapper around `jx_lookup` for brevity.  If you care about the alternate cases, you can do this:

```
j = jx_lookup(j,key);
if(!j) { //no such key } 
else if(!jx_istype(JX_STRING) { // unexpected type }
else { // expected case }
```

batrick: Inevitably there is some effort to do by the caller. You can see I had to some nasty macros for type checking/lookup for the other json module:

https://github.com/cooperative-computing-lab/cctools/blob/c776ac301a4597e297a0577861a2d0f67e792f7f/chirp/src/chirp_job.c#L244-L298

Notable example uses:

https://github.com/cooperative-computing-lab/cctools/blob/c776ac301a4597e297a0577861a2d0f67e792f7f/chirp/src/chirp_job.c#L330

https://github.com/cooperative-computing-lab/cctools/blob/c776ac301a4597e297a0577861a2d0f67e792f7f/chirp/src/chirp_job.c#L395-L405

Maybe that will inspire ideas for improving things in the jx module... perhaps in a helper module.

dttools/src/jx.c
```diff
@@ -0,0 +1,425 @@
...
+{
+	va_list args;
+	va_start(args,value);
+
+	struct jx *array = jx_array(0);
+	while(value) {
+		jx_array_append(array,value);
+		value = va_arg(args,struct jx *);
+	}
+
```
batrick: Missing `va_end`.

dttools/src/jx.c
```diff
@@ -0,0 +1,425 @@
...
+	if(!array->items) {
+		array->items = jx_item( value, 0 );
+	} else {
+		struct jx_item *i;
+		for(i=array->items;i->next;i=i->next) { }
+		i->next = jx_item(value,0);
+	}
+}
+
+void jx_pair_delete( struct jx_pair *pair )
```
batrick: Should this and `jx_item` be static/private? I don't think there is a sensible external use.

dttools/src/jx.h
```diff
@@ -0,0 +1,191 @@
...
+	union {
+		int boolean_value;      /**< value of @ref JX_BOOLEAN */
+		jx_int_t integer_value; /**< value of @ref JX_INTEGER */
+		double double_value;   /**< value of @ref JX_DOUBLE */
+		char * string_value;  /**< value of @ref JX_STRING */
+		char * symbol_name;   /**< value of @ref JX_SYMBOL */
+		struct jx_item *items;  /**< value of @ref JX_ARRAY */
+		struct jx_pair *pairs;  /**< value of @ref JX_OBJECT */
+	};
+};
```
batrick: These struct definitions should be private I think?

dthain: They could be, but that would prevent code from traversing the data structure.  For example, `catalog_query` loads an entire object and then iterates over the pairs one by one.

batrick: Oh, the `jx_lookup_*` functions got me thinking that we had a bunch of getters for the various types. Nevermind...

PR #994, opened 2015-10-28
Take into account shared memory when computing resident memory for the resource monitor

btovar:
Accumulate the maps we just found per file. First, we merge together all the maps segment that overlap. With this, we do not overcount private segments, but do consider that segments are shared as little as possible.

After this merging, we determine upper bounds, such as: 
virtual >= referenced >= private >= referenced - private >= shared.

If swap size is zero, then virtual and private have the exact values, and we have worst case counts for referenced and shared.

When we accumulate resident for the result, we do it from private + shared, rather than the resident reported originally as this would overcount shared.

There could be a way to use Pss (proportional resident) to improve these bounds.

This should mitigate some obvious overcounting, as reported by @annawoodard.


btovar:
Do not merge yet... Seems correct, but I would like to test it some more.

Issue #981, opened 2015-10-22
docker+ucernvm+parrot+condor


tommasoboccali:
Ciao, referring here to a htcondor thread (where I was adviced about opening an issue here: https://www-auth.cs.wisc.edu/lists/htcondor-users/2015-October/msg00067.shtml )

So this is the problem: I an trying to run condor inside a uCERNVM under docker, which gets /usr, /bin etc from CVMFS (the usual uno fro @jblomer ). Also condor is not in a local FS, but taken from cvmfs. There is nothing outisde parrot, since before parrot the system is not functional (no /usr, no /bin).

i can start condor, connect to a pool, be visible to external users, and accept jobs.
The problem (if we debugged the logs to a good level), is that at that point condor does not allow to run the payload as root, but wants to move to the user"nobody" (see logs below), using setuid().

In my understanding, that is not allowed by parrot, which traps it and the user remains root. at the next second, condor screams it is still root and closes the job.

I can reproduce the thing in a simplified way:

bash-4.1# su nobody
2015/10/22 13:10:45.68 parrot_run[1] <child:10616> notice: cannot execute the program /bin/su because it is setuid.
bash: /bin/su: Permission denied

even if I copy the /bin/su command to a non cvmfs disk (local), it still fails. I could not really test via the -M parrot option since /bin/su does not exist before parrot (and not even /bin), but I understood the effect is the same as my local test:

bash-4.1# cp /bin/su /root
bash-4.1# /root/su nobody
2015/10/22 13:21:51.96 parrot_run[1] <child:10664> notice: warning: system call 250 (keyctl) not supported for program /root/su

bash-4.1# chmod +s /root/su 
bash-4.1# /root/su nobody
2015/10/22 13:22:03.67 parrot_run[1] <child:10668> notice: cannot execute the program /root/su because it is setuid.
bash: /root/su: Permission denied

I think the basic question is: is there a way to have setuid () (and hence all the command using it, like su and sudo) working inside parrot?

Here is the log which shows how Condor gets the payload and tries to change user:

[...]

PR #997, opened 2015-10-29
Fake setuid

trshaffer:
This option lets programs *think* they're changing their uid/gid without actually doing anything, besides keeping track of changes. The various `getuid`, `setresgid`, etc. variations are supported, but `setfsuid`/`setfsgid` are slightly noop-ed. The new tests are passing with this option, everything seems to be behaving well. I tested that root can successfully use `su` and `sudo` through Parrot with this option, but running as root seems ill advised...

This option might be relevant to #981


batrick:
@dthain looks okay to me. What do you think?

dthain:
This is the right approach, but there are still some corner cases to clean up:

If `--fake-setuid` is not enabled, it seems to me that we either need to pass the `set*id` operations through to the kernel, or cause them to fail with an appropriate errno.  Right now, they are diverted to a false success, which is likely to be misleading to the user.

Should uid-changing be allowed if the fake-uid is not zero?

I don't think the special case of a setuid program is handled correctly.  If you simply remove the check for setuid programs, what happens is that ptrace will allow the execve, but the program will continue to run as the original user.  I think that we want to detect that the program is setuid, and if the execve succeeds, then change the fake uid to the uid of the file.

Does that make sense?

trshaffer:
This batch of commits should cover those changes. There could be some weirdness if a program uses fake setuid and services that don't support file ownership. It looked like the changes to fix that might be serious

parrot/src/pfs_dispatch.cc
```diff
@@ -973,6 +977,11 @@ static void decode_execve( struct pfs_process *p, int entering, INT64_T syscall,
 failure:
 		divert_to_dummy(p, -errno);
 done:
+		p->euid = new_uid;
+		p->suid = new_uid;
+		p->egid = new_gid;
+		p->sgid = new_gid;
```
batrick: These uid/gid changes need to be made only if execve succeeds. Otherwise, a failed execve of a setuid executable would raise privileges even though the setuid executable is not running.

Fixing this require saving the changes in the pfs_process struct. The `execve` success is here:

````
      if (actual == 0) {
            p->table->complete_at_path(AT_FDCWD, p->new_logical_name, p->name);
            debug(D_PROCESS, "execve: %s (%s) succeeded in 64-bit mode", p->new_logical_name, p->name);
````

File: parrot/src/pfs_dispatch.cc [hunk 2]
```diff
@@ -113,9 +113,11 @@ extern int parrot_dir_fd;
 extern int *pfs_syscall_totals32;
 
 int pfs_dispatch_prepexe (struct pfs_process *p, char exe[PATH_MAX], const char *physical_name);
-int pfs_dispatch_isexe( const char *path );
+int pfs_dispatch_isexe( const char *path, uid_t *uid, gid_t *gid );
 
 #define POINTER( i ) ((void *)(uintptr_t)(i))
+#define PRIVILEGED_ID(r,e,s) (((r) == 0) || ((e) == 0) || ((s) == 0))
+#define ALLOWED_ID(n,r,e,s) ((n) == (uid_t) -1 || ((n) == (r)) || ((n) == (e)) || ((n) == (s)))
```
batrick: I think these macros could be simplified to:

`#define PRIVILEGED_ID(p) (((p->ruid) == 0) || ((p->euid) == 0) || ((p->suid) == 0))`

ditto for `ALLOWED_ID`. Let's also move the two macros to `pfs_dispatch.h`.

dthain: While macros can be handy, I would prefer to see functions used where possible, so that we get consistent typechecking, debugging, and so forth.  (They can be made inline if performance is a concern.)

trshaffer:
OK, made those changes. Are there any other outstanding issues?

batrick:
@trshaffer I tried executing `/bin/sudo ls /` using Parrot and got this error:

````
$ ./parrot_run -d all -o debug --fake-setuid -- /bin/sudo ls /
sudo: unable to change to root gid: Operation not permitted
sudo: unable to initialize policy plugin
````

I think this may be a bug in the handling for `setresgid`?

Also, `pfs_dispatch.cc` and `pfs_dispatch64.cc` share the same logic for the permission checking. Can you think of a way to move the setuid/setgid handling to `pfs_process.cc` so the dispatch code can simply do something like `pfs_process_setresuid`/`pfs_process_setresgid`? The fake_uid check would also go in `pfs_process_setres[gu]id`.

trshaffer:
Were you running that test as root? Even if we fake the uid/gid, `sudo` requires *actual* root for authentication, so it'll fail at some point. If I run that command as root, it works as expected. I also just confirmed that `sudo -u somebody` tricks the process into thinking that it's running as another user, but really still retains root privileges.

It looks like `pfs_dispatch.cc` and `pfs_dispatch64.cc` contain a lot of duplicated syscall code, so I followed the existing layout. The different `set*id` flavors behave just differently enough to make refactoring tricky. Would it be cleaner to make one big id change function that includes logic for syscall variations?

batrick:
> Were you running that test as root? Even if we fake the uid/gid, sudo requires actual root for authentication, so it'll fail at some point. If I run that command as root, it works as expected. I also just confirmed that sudo -u somebody tricks the process into thinking that it's running as another user, but really still retains root privileges.

Heh, I tried to trick sudo but it's very careful:

````
$ env PARROT_UID=0 PARROT_GID=0 parrot/src/parrot_run --fake-setuid -- /bin/sudo ls /
sudo: unable to open /etc/sudoers: Permission denied
sudo: no valid sudoers sources found, quitting
sudo: unable to initialize policy plugin
$ cat > sudoers
ALL ALL=(ALL) NOPASSWD: ALL
$ env PARROT_UID=0 PARROT_GID=0 parrot/src/parrot_run --fake-setuid -M /etc/sudoers=sudoers -- /bin/sudo ls /
sudo: /etc/sudoers is owned by uid 1000, should be 0
sudo: no valid sudoers sources found, quitting
sudo: unable to initialize policy plugin
$ sudo chown root:root sudoers 
$ sudo chmod 444 sudoers 
$ env PARROT_UID=0 PARROT_GID=0 parrot/src/parrot_run --fake-setuid -M /etc/sudoers=sudoers -- /bin/sudo ls /
sudo: unable to set runas group vector: Operation not permitted
sudo: unable to set runas group vector: Operation not permitted
root is not in the sudoers file.  This incident will be reported.
````

> It looks like pfs_dispatch.cc and pfs_dispatch64.cc contain a lot of duplicated syscall code, so I followed the existing layout. 

The duplicated code is usually due to an unfortunate consequence of the different word sizes. Where possible, it's best to move common code elsewhere.

> The different set*id flavors behave just differently enough to make refactoring tricky. Would it be cleaner to make one big id change function that includes logic for syscall variations?

Hmm, I thought it might be as simple as just adding a `setres[ug]id` but I can see that `setreuid` and `setuid` have some weird logic. I still think it's a good idea to move the code to pfs_process, it will just require several functions (one for each syscall).

trshaffer:
I moved the meat of of those syscalls into `pfs_process`, and things look cleaner now.

batrick:
Nice work @trshaffer.

PR #998, opened 2015-10-29
change default cvmfs uri from /opt/$VO to /cvmfs/$FQRN

jblomer:
Original convention for web servers providing cvmfs repositories was to use 

    http://<HOSTNAME>/opt/<VIRTUAL ORGANIZATION>

like http://cvmfs-stratum-one.cern.ch/opt/atlas.  These URLs are still active on WLCG web servers but considered deprecated.  The URI scheme changed to

    http://<HOST NAME>/cvmfs/<FULLY QUALIFIED REPOSITORY NAME>

like http://cvmfs-stratum-one.cern.ch/cvmfs/atlas.cern.ch

### DIFF
--- a/apps/parrot_atlas/parrot.atlas.sh
+++ b/apps/parrot_atlas/parrot.atlas.sh
@@ -2,7 +2,7 @@
 
 export PARROT_ALLOW_SWITCHING_CVMFS_REPOSITORIES="yes"
 export HTTP_PROXY=http://cache01.hep.wisc.edu:3128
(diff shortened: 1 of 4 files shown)


btovar:
@jblomer, do I understand correctly that current users are not forced to modify their repositories option? That is, we are simply moving parrot from the deprecated convention?

jblomer:
That's correct, this is not yet a problem.  At some point in the future, the /opt/... URLs will be turned off.  Only then parrot users are affected.

PR #1008, opened 2015-11-06
Wq pool keep workers stderr

### LINKED ISSUE #1000: keep workers debug log in work queue pool
work queue pool overwrites the error output of the workers as they exit. This makes debugging more difficult. I propose to have a single error stream, in which all the worker debug outputs are appended.

btovar:
Fixes #1000.


btovar:
Complication without too much payoff.

PR #1010, opened 2015-11-09
Makeflow Local Resource Management

NeilB879:
Added options for local machines -max-local-cores -max-local-disk -max-local-ram to limit the resource usage of a make flow job.


nhazekam:
Remove the work done to work_queue_pool.c. This pull request should only apply changes to makeflow.

makeflow/src/makeflow.c
```diff
@@ -952,8 +1037,9 @@ int main(int argc, char *argv[])
 	char *work_queue_preferred_connection = NULL;
 	char *write_summary_to = NULL;
 	char *s;
-	char *log_dir = NULL;
-	char *log_format = NULL;
```
nhazekam: These need to be added back in. I mistakenly suggested removing them.

makeflow/src/makeflow.c
```diff
@@ -451,6 +457,72 @@ static batch_job_id_t makeflow_node_submit_retry( struct batch_queue *queue, con
 	return 0;
 }
 
+static int makeflow_is_local_node(struct dag_node *n)
+{
+	return (batch_queue_type == BATCH_QUEUE_TYPE_LOCAL || (n->local_job && local_queue));
+}
```
nhazekam: Add newline breaks between the functions for readability.

nhazekam:
After you have made the above changes, perform the white space rebase.

nhazekam:
We also need to add the options to the help message and the man page.

nhazekam:
This is looking good and it does limit the number of submitted tasks based on resources. The only issue right now is that when a task cannot fit with in the specified space it silently ignores the task. For example if --local-max-cores = 4 and a task has CORES=8, the task is never run and nothing is mentioned to notify the user. 

I propose that you add a check if the task resources are larger than the specified limit, and if so issue a fatal error and exit. This tells the user they need to update their resource limits.

makeflow/src/makeflow.c
```diff
@@ -450,6 +456,88 @@ static batch_job_id_t makeflow_node_submit_retry( struct batch_queue *queue, con
...
+ /*
+ * 
+ * Takes the allocates the local machine resources 
+ * of the job given as a parameter.
+ */
+static void makeflow_alloc_local(struct dag_node *n)
+{
+
+	if( (loc_info.local_disk == -1 && loc_info.local_mem == -1)) {
```
nhazekam: What is gained by having this if else statement? It looks as if we always need to make sure n->resources is defined and the first if statement is handled in the else if either way. This applies to both alloc and dealloc local. 

Also if we rely on n->resources being defined and there is chance it isn't at this point we either need to return an int for success or failure, or create it. As it stands if we have defined local_mem or local_disk and n->resources is null we don't alloc or track anything. I believe this will never happen so these should be cleanup to reflect that we already know it exists.

makeflow/src/makeflow.c
```diff
@@ -539,6 +631,12 @@ static int makeflow_node_ready(struct dag *d, struct dag_node *n)
 	if(n->state != DAG_NODE_STATE_WAITING)
 		return 0;
 
+		 if(makeflow_is_local_node(n)){
```
nhazekam: Correct the indentation here. It confuses that it is a separate check from checking if the node is waiting.

makeflow/src/makeflow.c
```diff
@@ -1424,6 +1517,15 @@ int main(int argc, char *argv[])
 			remote_jobs_max = MIN(local_jobs_max, n);
 		}
 	}
+	
+	struct dag_node *n;
+	// check to ensure that all jobs can be ran with resources given
+	for(n = d->nodes; n; n = n->next) {
```
nhazekam: Move this loop down after the calls to makeflow_check and makeflow_check_batch_consistency. Then surround it with an if statement that checks if we are in clean mode, because we only want to fail out if we are going to run the makeflow and not prevent cleaning.

makeflow/src/makeflow.c
```diff
@@ -1424,6 +1517,15 @@ int main(int argc, char *argv[])
...
+		if(makeflow_is_local_node(n) && !makeflow_can_alloc_local(n)) {	
+			fprintf(stderr, "Critical error, not enough resources to run job");
```
nhazekam: Use the node number in the print statement so that you can print out every node that can not run under the resource limit. Also use a flag to indicate an error occurred and find every instance where a node won't fit and then exit.

makeflow/src/makeflow.c
```diff
@@ -1159,7 +1244,8 @@ int main(int argc, char *argv[])
 						makeflow_gc_count = 1 << 14;	/* Inode threshold of 2^14. */
 				} else {
 					fprintf(stderr, "makeflow: invalid garbage collection method: %s\n", optarg);
-					exit(1);
+					/c
```
nhazekam: I think this may have been a mistake?

makeflow/src/makeflow.c
```diff
@@ -1425,6 +1519,21 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	// check to ensure that all jobs can be ran with resources given
+	if(clean_mode == MAKEFLOW_CLEAN_ALL)	
+	{	
+		int tooBig;
```
nhazekam: tooBig should be declared outside of this if statement. Also, loop over all of the nodes regardless of it we are in clean mode. Then check tooBig and clean mode before exiting.

NeilB879:
@nhazekam @btovar Anything else I need to fix on this?

PR #1013, opened 2015-11-09
Makeflow/Parrot Enforcement

trshaffer:
This is a first draft of the sandbox feature, using the new parrot mountlist stuff. I'll still need to test it more and see if this is the best way to approach things. I also have some questions about how some of the remote features play together.

At the moment, the `--sandbox` flag causes makeflow to
- make sure there's a copy of `parrot_run` in the working directory and add it to the input files
- generate and add to input files a wrapper script,
- which generates a mountlist on the worker and calls the given command through parrot


dthain:
Great, you are on the right track here, and I'm glad to see that this could be done using the wrapper framework.

Let's think of a slightly different name for this feature though, since the "sandbox" feature will be something different: basically creating a new directory for each task.  This is something different: enforcement, protection, ...  Can you think of some alternatives?

Also, see `copy_stream.h` for some existing file copy functions.

nhazekam:
Another interesting question that comes up is the order in which these
wrappers should be applied. Do we want the monitor to monitor only the
lower task or the environment enforcement as well. I like the order as we
have it now, but different wrappers are created this needs to be addressed
so we have a consistent understanding of what enabling several wrappers at
once means.

I also agree that sandbox may not be the clearest name as we are likely
going to create a sandbox wrapper for outside of parrot_run.Unfortunately I
have no good names to add to the options.

btovar:
Regardless of the order, there is a complication to consider. The monitor binary writes a shared library to a random name, which is then used with LD_PRELOAD. This library should be white-listed by the parrot enforcer. 

The problem is the random name. We can always use the same name, with a possible (but perhaps unlikely) name collision with the user task. We can also add an option to the monitor to specify a name to the library, which makeflow can then compute and white-list.

nhazekam:
For that matter, this could cause issues with other programs that use randomly named intermediate files. I am thinking of files created and used that are not inputs or outputs so the normal user never knows, and if they did couldn't predict the full name anyway.

Would it be possible to add an options that allows file access if it meets a pattern. This would go partially to fixing the monitor issue. Have it go something like --allow_file_pattern=monitor_* . Thoughts? It just occurred to me after talking with Ben other programs such as GATK may fail as they put out intermediate reads to a temp file.

trshaffer:
To avoid stepping on the monitor's toes, maybe the Parrot wrapper should be applied after any user-supplied wrapper but before the monitor.

Adding wildcards to Parrot path resolution could be involved, but there are some other options that might work. As I understand it, programs that use temp files would need to explicitly list the directory where they'll live in the makefile. Specifying a directory as a dependency should allow whatever changes in that directory. In the case of system-wide stuff like `/tmp`, we have the option of using Parrot to rebind that into the current directory. @dthain, is that what you had in mind?

As to a name, how about `--contain`?

btovar:
I think the monitor should be the outermost wrapper, even from those given by a user. The reasoning is that the measurement will be suspect if we only know part of the story. 

The monitor should also record the chain of wrappers.

dthain:
Per our discussion yesterday, let's call this feature `enforcement`

trshaffer:
These commits cover all the changes I can remember. Also, using `makeflow_log_file_expectation` lets `makeflow -c` clean up temp files. Is that something that can help clean up after the monitor?

dthain:
Please discuss with @nhazekam to make sure you are on the same page with logging and cleanup.

makeflow/src/makeflow_wrapper_enforcement.c
```diff
@@ -0,0 +1,140 @@
...
+		fprintf(enforcer, "$PWD/%s\trwx\n", f->filename);
+	}
+	fprintf(enforcer, "EOF\n\n");
+	fprintf(enforcer, "./parrot_run -m \"$MOUNTFILE\" -- \"$@\"\n");
+	fprintf(enforcer, "exit $?\n");
+	fclose(enforcer);
+
+	list_push_tail(enforcer_paths, dag_file_lookup_or_create(n->d, enforcer_path));
+
+	makeflow_log_file_expectation(n->d, enforcer_paths);
```
nhazekam: You will want to create a function that sets the files for exists, rather than expected. This way they should behave properly when garbage collection is enabled.

dthain: Isn't the idea here that you log 'expected', then create the file, then log 'exists'?

nhazekam: From what I understood talking with Tim, the set of files in enforcer_paths are files that are created prior to the task being run, so outside the scope of any node and in the node prep step. We have two choices to log it properly, log expected when we make the list, then log exists when we create it, both prior to node submission, or simply log it as exists. The important idea that that prior to running we know the files exist, and then afterwards the node will look at these files and move them to completed. If they are in expected, we clean them in clean mode, but in GC we only look at completed files, so the files will never be moved to existing->completed->deleted despite us creating them.

dthain: I see -- so, it sounds like the problem is that the wrapper needs to do something (in this case, log a file exists) when a job completes.  Is there a hook for a wrapper to take an action on job exit?

nhazekam: To be more clear, the enforcer_paths are inputs that are created before the node is run. Following a node's completion we move outputs from expected to existing, and inputs from existing to completed if they are no longer needed. The issue is that if the files aren't logged prior to the node's completion we treat the inputs as user provided and don't delete them. Wrapper outputs already get logged as expected->existing, but the inputs are not because a node does not explicitly create it.

dthain: So, is the wrapper-created file treated the same as an input file in every respect?
Or is it a third kind of file?

makeflow/src/makeflow.c
```diff
@@ -1417,6 +1461,10 @@ int main(int argc, char *argv[])
 		free(log_format);
 	}
 
+	if(enforcer) {
+		makeflow_wrapper_enforcer_init(enforcer, parrot_path);
```
nhazekam: I see that we have a default value for parrot_path, but do we want to force the user to set the value? If not what is the expected behavior if they don't set a value and its not in the local dir?

trshaffer:
I've made some cleanup-type changes as we discussed, and pulled in changes to master since this branch to fix a conflict from dec84002e8ec54dfdda7154829988f14fbe46a02, so this should merge cleanly

nhazekam:
In the process of testing this, I ran into an issue. When attempting to resume a partial makeflow that used --enforcement for the first and second run I get this error:

2016/02/15 11:54:39.46 makeflow[10434] fatal: could not create temp directory: File exists
Received signal 15, will try to clean up remote resources. Send signal 4 more times to force exit.

This definitely needs to be addressed as the ability to resume partial makeflows is key. I assume it has something to do with the statically named files used in enforcement. @trshaffer

trshaffer:
In a previous iteration, I had been using random names, and filesystem operations would fail if something happened to exist. Since we're going with deterministic names based on job number, seeing leftover files is to be expected. Since they're just temp files, it's usually OK to continue. The previous commit should allow resume without complaining about EEXIST

dthain:
@trshaffer @nhazekam what remains to be done in this PR?

trshaffer:
Just rebased off master, so it merges cleanly and gets the commits to quote wrapper commands from after I started this branch

nhazekam:
There is an issue with running this in Work Queue. The empty directories that are created to map /tmp /var/tmp and /dev/shm to local directories are not sent to Work Queue as they are empty.

Two fixes that are provided directly from the wrapper is to either:
Touch a file in each directory locally so a file exists that is transferred.
- or -
Create and delete the local directories as part of the wrapper script.

A third fix that may be more substantial in scope is to have WQ send the structure to the worker if there are no files in the structure. @btovar mentioned there may have been a partial effort to add this from Mike, but it wasn't evident from a cursory look where that would be.

@trshaffer @dthain Let me know your thoughts on this.

makeflow/src/makeflow_wrapper_enforcement.c
```diff
@@ -0,0 +1,162 @@
...
+			fatal("could not copy parrot: %s");
+		}
+	}
+	close(local_parrot);
+	close(host_parrot);
+
+	makeflow_wrapper_add_input_file(w, local_parrot_path);
+	makeflow_wrapper_add_input_file(w, xxstrdup(enforcer_pattern "%%"));
+	makeflow_wrapper_add_input_file(w, xxstrdup(mountlist_pattern "%%"));
+	makeflow_wrapper_add_input_file(w, xxstrdup(tmp_pattern "%%"));
```
nhazekam: These input files no longer need to be specified as they are created at the worker and deleted there as well. Removing line 68-70 fixed an issue with WQ failures.

trshaffer:
Now the script will create a temp directory and give the program a friendly suggestion to use it. `/` is readonly by default, so the program will just get blocked if it tries to write to the real `/tmp`. Rules to allow writes to input/output files will override that, so things *should* work when the work directory is somewhere under `/tmp`

makeflow/src/makeflow_wrapper_enforcement.c
```diff
@@ -0,0 +1,138 @@
...
+	while((f=list_next_item(input_list))) {
+		fprintf(enforcer, "$PWD/%s\trwx\n", f->filename);
+	}
+	list_first_item(output_list);
+	while((f=list_next_item(output_list))) {
+		fprintf(enforcer, "$PWD/%s\trwx\n", f->filename);
+	}
+	fprintf(enforcer, "EOF\n\n");
+	fprintf(enforcer, "mkdir -p \"%s\"\n", tmp_path);
+	fprintf(enforcer, "export \"TMPDIR=%s\"\n", tmp_path);
```
nhazekam: I think TMPDIR should have the absolute path to the new temporary directory and not just the local name of the directory.

nhazekam:
Just to bring this back up. I think this was ready to go. We had tested it on a larger Makeflow and it worked. The only issues we were having at the last check was with JAVA execution. @trshaffer please bring this up to date and we can take a fresh look at it. Important to note the umbrella work @hmeng-19 is doing will conflict on lines in makeflow.c so it may be worth while to update after #1458 is merged.

trshaffer:
Messy rebase, but it's updated to the current `master`, and seems to be functioning as expected. @nhazekam care to have a look?

PR #1014, opened 2015-11-09
Supplementary groups

trshaffer:
These should block `getgroups`/`setgroups` with `EPERM` when using `--fake-setuid`. If blocking `getgroups` is unnecessary, then a21e7a9701076aa2c94a7852e0c2e3eeab98cb7d just needs to be reverted.

### DIFF
--- a/parrot/src/pfs_dispatch.cc
+++ b/parrot/src/pfs_dispatch.cc
@@ -1493,9 +1493,13 @@ static void decode_syscall( struct pfs_process *p, int entering )
 
 		case SYSCALL32_getgroups32:
 		case SYSCALL32_getgroups:
+			if (entering && pfs_fake_setgid)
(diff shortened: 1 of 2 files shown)


batrick:
I think this should be done even without `--fake-setuid`? We don't pass the other setuid/setgid calls to the kernel too.

trshaffer:
I was afraid that unconditionally blocking `getgroups` would cause problems with existing programs. For `setgroups`, that definitely makes sense

batrick:
I'm sorry I think I was confusing. I'll comment in the patch itself.

trshaffer:
So this will unconditionally block `setgroups`, and allow `getgroups` as long as Parrot isn't messing around with the gid.

batrick:
That's right.

Issue #1009, opened 2015-11-09
"exit" and "release" have the same meaning at worker


btovar:
I would like to keep "release" with the current meaning: worker simply disconnect from master, and turn "exit" into terminating the worker.

From the master, shut down workers when queue exits should use exit, as "release" is implied.

(Funny story: I incorrectly set worker's timeout to 30 hours. condor_rm sent all my jobs to X state, and I had no way to kill the workers.)

@dthain, @nhazekam

dthain:
Ok here

Issue #706, opened 2015-04-13
Fixing Work Queue Foreman


nhazekam:
As per the discussion with @btovar and @dthain, this ticket should outline different areas of work to better model the interactions of the master<->foreman<->workers. The goal is to allow existing functionality to work with the foreman structure, and allow additional functionality to utilize foreman as it is added. To fix the foreman here are key areas to understand and model.
- Communication between master, foreman, and workers on task information
  - Currently, the master send task information out to foreman and workers, but the only information that is retrieved is the done status and the resulting files.
  - The end goal would allow information to move back and forth from master and worker through chains of foreman without losing important information about execution due to the encapsulating nature of foreman task management.
  - Initially, this may be addressed by passing a doc about task execution back to the master, using the same structure as exists for sending task information to workers.
- Management of policy attributes in the master
  - Currently, policies created and managed by the master are not communicated or enforced by the foreman. This causes some of the functionality to either not work function differently from expectations when foreman are used.
  - The end goal would be a standard way to communicated policies to foreman. This should work for chains of foreman and be clear how Work Queue refreshes these policies so behavior is understood.
- Revise structure to model internal structure as closely to API calls where appropriate
  - As we revise and add functionality to the master and foreman, to maintain clarity the API functions should be clearly related to the structures.
  - Though this is not an issue currently, this is included to help model the above work.

Let me know if I missed something substantial or if the direction is not quite right.

dthain:
That's a good start.  Before we proceed any further, let's add add specific examples of where the foreman breaks existing functionality, so that when we consider changes, we can first consider whether the change fixes or breaks that particular aspect.

nhazekam:
For specific examples of how foreman breaks existing functionality, there are two that come to mind.

1. Blacklist - Blacklisting workers if currently managed at the master. If there is a worker misbehaving or not executing properly, the master can blacklist the worker. However, if the worker is used by the foreman, the master does not know the name of the worker, and therefore can only blacklist the foreman losing all of its workers.

2. Fast Abort - To utilize fast abort currently both the master and the foreman must have fast abort turned on. However, in this case the statistics used to determine an aborted task are calculated separately at the master and all foreman, so the behavior is not as expected.

@btovar Please comment on other cases where foreman breaks existing functionality.

btovar:
task.host used to describe the site on which the task executed. This is no longer true for foremen, as the task.host has the ip of the foreman, not of the worker.

annawoodard:
[...]

PR #1015, opened 2015-11-10
Worker exit vs release

btovar:
Fix for #1009


nhazekam:
Looks good to me. Aside from foreman->workers, are there other instances of using broadcast that would make sense and replace current code that loops over workers?

btovar:
Not currently, but probably there will be more given #706

nhazekam:
Ok, sounds good. Thanks for that fix Ben.

Issue #1003, opened 2015-11-04
source tarball includes test makeflow log files


btovar:
For example, from a clean autobuild tarball:

2015/11/04 12:53:06.35 makeflow[76025] notice: MAKEFLOW_INPUTS is not specified
2015/11/04 12:53:06.35 makeflow[76025] notice: MAKEFLOW_OUTPUTS is not specified
parsing dirs/testcase.subdir.01.makeflow...
checking dirs/testcase.subdir.01.makeflow for consistency...
dirs/testcase.subdir.01.makeflow has 1 rules.
recovering from log file dirs/testcase.subdir.01.makeflow.makeflowlog...
will retry failed rule: mkdir -p mydir; cp input/hello mydir/1.txt; cp input/hello mydir/2.txt
cleaning filesystem...

We should not be recovering or retrying anything, and this is the first time the rule is tried.

This seems to be the case with all makeflow tests.

btovar:
Mmm, it seems this is just the clear part, when using the option -c. Perhaps we should change the "retry" messages if in cleaning mode?

PR #1020, opened 2015-11-11
Log recovery silently resets nodes if cleaning

nhazekam:
We still reset the node state, we just don't report that we are retrying the node if we are cleaning.
Address #1003. @btovar

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -201,7 +201,7 @@ static void makeflow_node_force_rerun(struct itable *rerun_table, struct dag *d,
 Decide whether to rerun a node based on batch and file system status.
 */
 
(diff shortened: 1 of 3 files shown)


makeflow/src/makeflow.c
```diff
@@ -201,7 +201,7 @@ static void makeflow_node_force_rerun(struct itable *rerun_table, struct dag *d,
 Decide whether to rerun a node based on batch and file system status.
 */
 
-void makeflow_node_decide_rerun(struct itable *rerun_table, struct dag *d, struct dag_node *n)
+void makeflow_node_decide_rerun(struct itable *rerun_table, struct dag *d, struct dag_node *n, int silent)
```
btovar: It looks good.

Please add a comment to explain why we added the silent argument (i.e., how it relates to clean_mode).

PR #1021, opened 2015-11-11
Batch Job Feature API

nhazekam:
This branch is looking at having the batch-queue express whether it supports a qiven feature as opposed to Makeflow having to switch on the queue type to see if it supports. The goal is that if new types are added or behavior changes it would be addressed in the batch_job*.c instead of adding more options to Makeflow. @btovar Still working on this and testing, but let me know your thoughts.

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -369,8 +369,6 @@ static char * makeflow_file_format( struct dag_node *n, struct dag_file *f, stru
 	switch (batch_queue_get_type(queue)) {
 		case BATCH_QUEUE_TYPE_WORK_QUEUE:
 			return string_format("%s=%s,", f->filename, remotename);
-		case BATCH_QUEUE_TYPE_CONDOR:
-			return string_format("%s,", remotename);
 		default:
 			return string_format("%s,", f->filename);
 	}
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -940,7 +940,7 @@ int main(int argc, char *argv[])
 	const char *work_queue_keepalive_timeout = NULL;
 	const char *work_queue_master_mode = "standalone";
 	const char *work_queue_port_file = NULL;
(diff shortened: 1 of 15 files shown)


makeflow/src/makeflow.c
```diff
@@ -975,7 +970,7 @@ int main(int argc, char *argv[])
 	}
 	s = getenv("WORK_QUEUE_FAST_ABORT_MULTIPLIER");
 	if(s) {
-		wq_option_fast_abort_multiplier = atof(s);
+		wq_option_fast_abort_multiplier = xxstrdup(s);
```
btovar: There is batch_queue_set_int_option. No need to change the type of wq_option_fast_abort_multiplier. I rather have ints as ints.

makeflow/src/makeflow.c
```diff
@@ -1500,12 +1444,27 @@ int main(int argc, char *argv[])
 
 	/* Remote storage modes do not (yet) support measuring storage for garbage collection. */
 
-	if(batch_queue_type==BATCH_QUEUE_TYPE_CHIRP || batch_queue_type==BATCH_QUEUE_TYPE_HADOOP) {
+	if(!batch_queue_supports_feature(remote_queue, "gc_size")) {
```
btovar: I rather have: if( makeflow_gc_method == MAKEFLOW_GC_SIZE && !batch_queue_supports_feature(remote_queue, "gc_size")) {
...

The meaning is clearer.

btovar:
Nick, it looks great so far.

nhazekam:
Ben, can you think of other locations where this would make sense to use?

btovar:
Nick, looks good. Do you want to merge it?

nhazekam:
I am planning on doing some quick testing with SLURM and TORQUE to just verify nothing is broken.

nhazekam:
@btovar Could you take a quick glance at the changes to fast-abort support to make sure this was what you were thinking.

btovar:
Yes Nick, that looks good.

PR #1023, opened 2015-11-13
Only ld linux

btovar:
Remove resource_monitorv, which did not use LD_PRELOAD.

Remove support for OSX. We did not support measuring most resources there anyway. Compile only for linux.

Check whether the executable is a script, or statically/dynamically linked. Label the summary appropriately.


batrick:
@btovar you can use [elf_get_interp](https://github.com/cooperative-computing-lab/cctools/blob/778bf8f9ff57a7959bd93b0e5272db3863219f9b/dttools/src/elfheader.h#L18) to check for static executables. Parrot uses it [here](https://github.com/cooperative-computing-lab/cctools/blob/778bf8f9ff57a7959bd93b0e5272db3863219f9b/parrot/src/pfs_dispatch.cc#L3321). `EINVAL` indicates a static exe.

btovar:
Ah, excellent!

btovar:
So, interp is not modified and errno is not EINVAL for dynamically linked.
interp is not modified, but errno is EINVAL for statically linked.
interp is set for a script interpreter.

Is that correct?

btovar:
I see dynamics set interpreter to /lib64/ld-linux-x86-64.so.2. Thanks @batrick.

batrick:
>So, interp is not modified and errno is not EINVAL for dynamically linked.
> interp is not modified, but errno is EINVAL for statically linked.
> interp is set for a script interpreter.

>Is that correct?

Uhm, I don't understand your question. `elf_get_interp` doesn't change the interpreter, it only gets the current one for you. If it is a static executable, there will be no `PT_INTERP` header so `elf_get_interp` returns -1, with errno = EINVAL.

btovar:
Yes, I was confused by the set parameter. Got it now.

batrick:
You were looking at the helper function :)

PR #1027, opened 2015-11-16
Add IPv6 support to CCTools.

### LINKED ISSUE #591: Chirp Cookie Parsing
I'm working on supporting IPv6 better in/with HTCondor, and while I was fixing a problem with our chirp server, it occurred to me to wonder: it works for us write chirp.config files that look like this:

    ::1 12345 cookie-cookie-cookie

but I can't find a spec anywhere on what that leading host field is allowed to be.  Since our chirp server is notionally compatible with your clients (and vice versa), I thought I'd ask you.  Would it be more compatible to restrict ourselves to IPv4 addresses?

batrick:
This commit does the following:

(1) Most DNS lookups are now done by link_connect. So, link_connect now takes a
    nodename/servname string as understood by getaddrinfo(3) for connections. This
    simplifies a lot of the code and makes IPv6 connections and IPv4 fallback
    possible. link_connect tries AF_UNSPEC getaddrinfo lookups and attempts
    connections via IPv6. Ditto for link_bind.

(2) Because of (1), I've changed all internal ints which are ports to strings.
    While this is not strictly necessary, the goal is to be consistent across
    the entire stack with the use of strings referring to nodename/servname.

(3) Thrown out the usage of the datagram library for sending UDP sockets.
    Setting up a sending UDP socket is a minor exercise in programming so I've
    changed WQ/Chirp to set these sockets up manually. I've also updated
    the resource monitor to use IPv6 compatible UDP sockets.

(4) The Chirp catalog management code is moved to chirp_catalog.[ch].

(5) The DNS libraries have been reworked as the new getaddrinfo_cache.[ch]
    dttools library. It was necessary to rework DNS caching to be compatible with
    IPv6 due to getaddrinfo(3) returning different results based on the query type
    (AF_UNSPEC/AF_INET/AF_INET6).

(6) Hostname lookup now in hostname.[ch], moved from domain_name_cache.[ch].
    Reworked to use caching getaddrinfo(3).

(7) Chirp may now connect to IPv6 hosts. We use the RFC 2732 format with one
    extension:

    IPv6address ":" port.

    This makes the port mandatory. This is abstracted by the new
    link_connect_nodeserv function.

(8) New link_getpeername/link_getlocalname functions replace
    link_address_local/link_address_remote.

Fixes #591.


batrick:
Lots of code to look at here which touches everything. Needs a good code review and discussion.

dthain:
Hmm, there are a lot of unrelated changes here in a single commit, it would have been better to discuss first before writing code.

`domain_name` was deliberately isolated from `link` b/c hostname lookups have some odd semantics and failure modes that I wanted the user to explicitly confront.  It's slightly more code for the caller, but leads to more correct programs.  (e.g. "no such hostname" requires a different sort of recovery than "cannot connect")

Also, the `domain_name_cache` module plays an important role b/c name lookup caching is implemented very different (and sometimes not at all) on different operating systems, and we want to retain control over the semantics of those lookups.  This also ensures that real (vs cached) lookups get logged in the debug stream, which is important for troubleshooting.

I'm not sure what is gained by removing the datagram abstraction.  The goal there (as with link) is to abstract away the complexity of dealing with the bsd socket interface, and simply represent addresses as strings.  Many new contributors struggle with this, and our interfaces make things a lot simple.

Perhaps I'm over-simplifying this, but wouldn't it be more straightforward to expand `LINK_ADDRESS_MAX` to accommodate IPV6, then have `domain_name_lookup` return an address in either form, then allow `link_connect` to recognize whether it's dealing with IPV4 or IPV6?  That way, the code changes could be isolated to just the `domain_name` and the `link` modules.

batrick:
> domain_name was deliberately isolated from link b/c hostname lookups have some odd semantics and failure modes that I wanted the user to explicitly confront. It's slightly more code for the caller, but leads to more correct programs. (e.g. "no such hostname" requires a different sort of recovery than "cannot connect")

After looking through the code before making these changes, I didn't see any instances of  hostname lookup failure handling **different** from `link_connect` failure handling. I just looked at the patch again and usually usercode would print an error/debug message and return/fatal. The same thing would be done for a failed `link_connect`. The only instance of handling a `domain_name_cache_lookup` failure differently than from a `link_connect` failure was in `chirp_client.c`:

 ````
-       if(!domain_name_cache_lookup(host, addr)) {
-               errno = ENOENT;
-               return 0;
-       }
````

In any case, I think a good way to solve this is by having an `int *gai_error` argument to `link_connect`/`link_bind` that can be used to look at `getaddrinfo` errors.

>Also, the domain_name_cache module plays an important role b/c name lookup caching is implemented very different (and sometimes not at all) on different operating systems, and we want to retain control over the semantics of those lookups. This also ensures that real (vs cached) lookups get logged in the debug stream, which is important for troubleshooting.

We are still caching lookups, through `getaddrinfo_cache`. There are some instances where we do not use the cache, like when usercode is doing a single lookup (catalog).

> I'm not sure what is gained by removing the datagram abstraction. The goal there (as with link) is to abstract away the complexity of dealing with the bsd socket interface, and simply represent addresses as strings. Many new contributors struggle with this, and our interfaces make things a lot simple.

We can't use a single source UDP socket for all UDP sends (like the chirp_server contacting the catalogs). This is because the catalogs may use different address families in compatible with the socket. I originally was using datagram sockets until I came across this issue. At that point it became clear we need a UDP socket for each catalog. This also allows doing a UDP `connect` for each socket so catalog lookups would only be done once at startup. This wasn't abstracted in the datagram library (and would be a mess to add) so I did it manually. Take a look here:

https://github.com/batrick/cctools/blob/1067cae7d5abc01a1d2b1f1029f7658ec5054ceb/chirp/src/chirp_catalog.c#L36-L74

I'm open to going back to datagram sockets for sending if you prefer. The real question is whether doing a `connect` for each UDP socket is desirable. This avoids future DNS lookups for the catalog but if the catalog address changes then the `chirp_server`/WQ would need restarted.

> Perhaps I'm over-simplifying this, but wouldn't it be more straightforward to expand LINK_ADDRESS_MAX to accommodate IPV6, then have domain_name_lookup return an address in either form, then allow link_connect to recognize whether it's dealing with IPV4 or IPV6? That way, the code changes could be isolated to just the domain_name and the link modules.

This would make fallbacks a mess. What do we do if the server is on IPv6 but the client is IPv4? What if the server has an interface for both? The address may also correspond to multiple IPs which should all be tried on failure.

The primary benefit of doing things this way is we are avoiding details about the protocol, such as the format/interpretation of addresses (and even ports).  This is considered the right way to do IPv6 compatible socket code (and new IPv4 code) and is documented in various resources:

* http://www.akkadia.org/drepper/userapi-ipv6.html
* The Linux Programming Interface (pg. 1221 [1]; this is not Linux specific code)

[1] https://books.google.com/books?id=2SAQAQAAQBAJ&lpg=PP1&dq=linux%20programming%20interface%20getaddrinfo&pg=PA1221#v=onepage&q=getaddrinfo&f=false

batrick:
> In any case, I think a good way to solve this is by having an int *gai_error argument to link_connect/link_bind that can be used to look at getaddrinfo errors.

I want to add: the old code didn't even have access to the `gai_error` result so it could not respond differently to various lookup errors.

dthain:
Ok, I see that the situation is inherently complex, so the code solution may not be simple.  But I think we can do this by using and modifying existing APIs, rather than throwing out whole modules.  I still want to some some degree of insulation from the POSIX socket APIs.

For example, if we need a distinct socket to report to each catalog, then we should create a new datagram object for each connection, and change the datagram API to bind to a single remote address.

Likewise, I don't see the benefit to passing around service strings rather than port numbers.  `/etc/services` is generally not used b/c it is peculiar to each machine, and doesn't take into account multiplicity of the same service.  (e.g. An HTTP service is not on always port 80.)  We should just pass around integer port numbers, and convert that to a string if needed for `getaddrinfo`.  Make the API accommodate the user, not the other way around.

Connecting to multiple addresses also brings up some failure issues that have to be solved.  Suppose that I do a `link_connect` with a timeout of 60 seconds.  `getaddrinfo` returns two IPV4 addresses and two IPV6 addresses.  Is there some strategy for dividing up our effort among those addresses.

There are also policy issues.  I'm sure that whatever we choose to do, the user will have some knowledge about IPV4 versus IPV6 in reality.  Is there a way for them to apply controls that require/prefer one over the other?  I am very worried that we may accidentally set up a situation where our code picks one, because the addresses appear in DNS, but the addresses are not actually routable in practice.

Last but not least, I am concerned that this is a much bigger project than it first appears to be, once you take into account testing, documentation, and so forth.  Is this really what you want to work on while trying to graduate?

dthain:
@batrick I appreciate your efforts on this, but I'm going to close out the pull request.  We need to back up and take a different approach for IPV6.

PR #1031, opened 2015-11-18
Fix struct jx anonymous union warning.

btovar:
Add --std=c99, which is the standard we are aiming for.

Add --fms-extensions, to explicitly allow the anonymous union in struct
jx. (c11 has anonymous unions, but before that, they are a gcc
extensions)


dthain:
If I understand correctly, the problem is that anonymous unions are not part of C99, right?

`-fms-extensions` is a pretty broad brush to paint with.  Are there any other gcc settings that might more narrowly accept this syntax?

(Or perhaps we should just stick to strict C99?)

btovar:
Correct. They are part of c11, though. We could use gnu99, but it may be in the same league as -fms-extensions.

Strict C99 would not be that bad. The only other place we failed was using typeof instead of __typeof__ in the monitor.

In jx, rather than saying j.int_value, we would have to say j.data.int_value, or something alike.

dthain:
Ok, let's stick to straight C99 and use j->u.int_value .

Probably better for our long term hygiene.

dthain:
See #1046

PR #1033, opened 2015-11-18
Monitor errors

btovar:
Decouple monitor and task exit status.
Monitor exit status:

RM_SUCCESS    0: no monitor errors, task exit status is 0
RM_TASK_ERROR 1: no monitor errors, task exit status is non-zero.
RM_OVERFLOW   2: no monitor errors, task terminated given of resources limits.
RM_MONITOR_ERROR 3: task could not be executed (missing executable, no
writing permissions for summaries, etc.)

This should make easier the retries at makeflow when implementing the complete resources loop.


btovar:
This means that the task exit status is only available in the summary file...

btovar:
Forgot to update man page, do not merge yet.

PR #1038, opened 2015-11-19
Disk alloc fix

nkremerh:
Fixes the way disk_alloc_delete finds and resolves the correct name of the mountpoint that is passed to losetup. This is necessary to de-allocate loop devices with a mountpoint pathname greater than 62 characters long.

### DIFF
--- a/dttools/src/disk_alloc.c
+++ b/dttools/src/disk_alloc.c
@@ -163,13 +163,11 @@ int disk_alloc_delete(char *loc) {
 	int k;
 	int max_mount_path_length = 62;
 
-	//Copy only pathname of the mountpoint without extraneous characters
+	//Copy only pathname of the mountpoint without extraneous paretheses
 	for(k = 1; k < loop_dev_path_length; k++) {
 		loop_mountpoint_array[k-1] = loop_mount[k];
 	}
-	for(k; k < 128; k++) {
-		loop_mountpoint_array[k] = '\0';
-	}
+	loop_mountpoint_array[k] = '\0';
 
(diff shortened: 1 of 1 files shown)


nkremerh:
@nhazekam Would you please take a look at this?

dttools/src/disk_alloc.c
```diff
@@ -121,46 +133,43 @@ int disk_alloc_delete(char *loc) {
...
+	char loop_dev[128], loop_info[128], loop_mount[128];
+	FILE *loop_find;
+	losetup_args = string_format("losetup -j %s", device_loc);
+	loop_find = popen(losetup_args, "r");
+	fscanf(loop_find, "%s %s %s", loop_dev, loop_info, loop_mount);
+	pclose(loop_find);
+	loop_mount[0] = '\0';
+	loop_mount[strlen(loop_mount) - 1] = '\0';
+	loop_dev[strlen(loop_dev) - 1] = '\0';
+	if(strncmp(loop_mount, device_loc, 62) + 47 == 0) {
```
nhazekam: Take a look at strncmp's return, I think you are comparing null to '\'. You will likely need to move the string over to remove the first character as opposed to just setting it to NULL. Should only go forward if the comparison succeeds.

dthain: I'll add that using magic numbers here (62,47) is poor practice.  The reader cannot tell the significant of those numbers, and you are going to forget sooner than you think.  Generate the number with something like `strlen("Some expected string")` or at least a really good comment.

nhazekam:
In addition to the comments you have already added, you should add debug statements everywhere that you either return as a failure or jump with goto error. This will clarify it for future users and keep it straight in your head what each error resulted from.

PR #1044, opened 2015-11-20
SGE script config

### LINKED ISSUE #1040: sge_submit_workers configure time option
For example:

./configure --sge-parameters='#$ -pe smp $cores'

--sge-parameters could be used several times at the command line. Each time specifies a line that should be copied at the top of sge_submit_workers.

For Notre Dame, ./configure.afs would specify the above: --sge-parameters='#$ -pe smp $cores'

btovar:
Fixes #1040 (bugs and taxes!)

Adds --sge-parameter  to configure, which can be used several times.
Prints a warning when resources are manually specified.
Uses the old options in configure.afs.

@nhazekam


work_queue/src/work_queue_submit_common
```diff
@@ -171,8 +176,14 @@ parse_arguments_common()
 	done
 
 	set_up_master_address $1 $2 $3 $4
+
+	if [ $specified_resources = 1 -a $SGE_WARNING = 1 ]
+	then
+		echo "Worker resources were manually specified. Remember to also describe your resources according to your local qsub system (e.g., -p '-pe smp 4'. See also the --sge-parameter configuration option.)" 1>&2
```
dthain: Suggest that you explicitly name the "configure" script here, otherwise it may not be clear to the user.
Also add an item to the Makeflow manual explaining in more detail.

PR #1047, opened 2015-11-25
Wq loop

btovar:
wq loop for resident memory resource.

(do not merge yet, quite experimental.)


dthain:
It looks like the right approach to me.
Any reason not to also have a histogram for disk and cpu time?

btovar:
I just want to figure it out with one resource, and then copy that to the other resources.

btovar:
Splitting this hydra into several prs...

PR #1048, opened 2015-12-01
Move BatchJob to Separate Module

dthain:
Note that the dependency tree is now:
```
makeflow wavefront: batch_job
batch_job: work_queue chirp
```
`work_queue_pool` seems like it's in the wrong place at first, but it only invokes `batch_job` not `libwork_queue` so it makes sense to stay in `batch_job`

This is a preparatory step to clean up our makefiles...


dthain:
@nhazekam @btovar how does this look?

nhazekam:
The move looks good to me. I understand why we are moving work_queue_pool, but do we also want to move *_submit_workers? They seem like it could go either.

dthain:
Fixed commits lost in the merge.

Well, the idea is that work queue can stand alone, separately of the batch_job module.  If you wanted to build and install work_queue by itself, then you would also want the submit scripts, which don't use the batch job module.

nhazekam:
OK, makes sense to me.

batch_job/src/Makefile~
```diff
@@ -0,0 +1,38 @@
+include ../../config.mk
```
btovar: This is a backup file (Makefile~)

batch_job/src/batch_job.h~
```diff
@@ -0,0 +1,190 @@
+/*
```
btovar: Another backup file (batch_job.h~)

dthain:
I am embarrased!

btovar:
I like a lot having batch_job as its own thing.

dthain:
@btovar I'll leave it to you to merge.

btovar:
Got it.

batch_job/src/#batch_job_hadoop.c#
```diff
@@ -1,3 +1,8 @@
+/*
```
btovar: Is the name between #...# on purpose?
If it is not used, why are we keeping the file?

btovar:
Ah! somebody did a commit .

nhazekam:
Sorry! Probably me.

PR #1062, opened 2015-12-06
Catch transient authentication failures

batrick:
I've used these changes for all of my experiments with Confuga for the CCPE paper. So the changes have received quite a bit of testing.

### DIFF
--- a/chirp/src/chirp_client.c
+++ b/chirp/src/chirp_client.c
@@ -556,7 +556,7 @@ static int ticket_translate(const char *name, char *ticket_subject)
 	buffer_putfstring(Benv, "CHIRP_TICKET_NAME=%s", name);
 	env[0] = buffer_tostring(Benv);
 
-	result = shellcode(command, env, Bout, Berr, &status);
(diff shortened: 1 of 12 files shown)


batrick:
@btovar @dthain any idea why 212a8d5 is necessary? I would think that the chirp python shared object would statically link to the dttool's MD5 library. Apparently it's using whatever was bundled with Python.

dthain:
No objections here.

btovar:
The problem may be the compilation. The shared object includes libchirp.a and libdttools.a just as any other .o object file (i.e., it makes the contents of the .a files 'shared'). Once this happens, all the bets are off, and the python environment may be playing LD games.

There might be a better way to compile the shared object, so that this does not happen.

batrick:
I'm not really interested in hunting down why this happened. If there are no objections to the simple fix, I'd like to get this merged.

batrick:
Sorry that was a little short. I agree Ben that's probably what's going on. We may need to take a look at this later as it's come up before with the debug library. Perhaps we will need to namespace dttools using some macro wizardry.

PR #1066, opened 2015-12-08
Worker categories

btovar:
Adds feature labels to worker. This is an experimental feature for @asherkhb. Two interface additions:

--provides option for a worker. It may specified several times. 
t.specify_requirement(feature)

worker features are treated as any other resource. If a task specifies a feature requirement, it may only run in a worker labeled with that feature. A task without an explicit feature may run on any worker. In the future, we want to generalize this to user-defined resources (e.g., number of licenses).

Example:
ta.specify_requirement("analysis_local_data")
tb.specify_requirement("merge_local_data")
// tc does not specify requirement feature.

// worker 1
work_queue_worker --provides analysis_local_data ...
// worker 2
work_queue_worker --provides analysis_local_data --provides merge_local_data  ...
// worker 3
work_queue_worker ...

ta may run on workers 1 and 2. tb may run only on worker 2. tc may work on any worker.

@nhazekam

### DIFF
--- a/doc/man/work_queue_worker.m4
+++ b/doc/man/work_queue_worker.m4
@@ -57,6 +57,7 @@ OPTION_PAIR(--gpus, n)Set the number of GPUs this worker should use. (default=0)
 OPTION_PAIR(--memory, mb)Manually set the amount of memory (in MB) reported by this worker.
 OPTION_PAIR(--disk, mb)Manually set the amount of disk space (in MB) reported by this worker.
 OPTION_PAIR(--wall-time, s)Set the maximum number of seconds the worker may be active.
+OPTION_PAIR(--provides, feature)Specifies a user-defined feature the worker provides.
 OPTION_ITEM(-v, --version')Show version string.
 OPTION_ITEM(-h, --help')Show this help message.
 OPTION_PAIR(--docker, image) Enable the worker to run each task with a container based on this image.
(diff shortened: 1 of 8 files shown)


dthain:
I'm concerned that we are overlapping two different feature sets with the same term.

If I recall correctly, an existing use of "category" is to allow for distinct collection of resource usage statistics at the master, so that we can do resource sizing and fast abort.  In that case, you want tasks of any category to run on any machine, right?

btovar:
You are correct. What about --provides and specify_requirement?

dthain:
Can we accomplish the same thing by making this a consumable resource?

btovar:
We definitely can. I think this is a first step towards that. (I've just need to add counters to the labels specified.)

nhazekam:
This would seem to allow multiple labels for a task, which is not a bad thing. How do we plan on setting a value for the counter? Is it user specified?

Also, since this is being treated as a resource, are we restricting the worker to only accept tasks that need that resource or can any task that fits work there?

btovar:
Nick, 
Yes, the counter is given by the user. Think of the resource "disk", only that it is a resource that the user defines (e.g., licenses, or connections to a database).

Since it is just a resource, any task that fits will run there. In other words, tasks have worker requirements (based on accounted resources), but workers do not have task requirements.

btovar:
Move towards custom consumable resources.

Issue #1063, opened 2015-12-07
Parrot should implement `getgroups` and `setgroups` for `--fake-setuid`


batrick:
This is necessary to resolve #981.

Just to summarize some implementation details:

* If Linux returns the primary group id with the supplemental group ids, we should too.
* Just use a statically sized array of `gid_t` in the pfs_process struct, no dynamic list.

tommasoboccali:
thanks! no big hurry, in the other way (condor outside parrot) it works ;)

tom

batrick:
Fixed by #1067.

PR #1067, opened 2015-12-08
Implement getgroups/setgroups for --fake-setuid

trshaffer:
This is for #1063

### DIFF
--- a/parrot/src/pfs_dispatch.cc
+++ b/parrot/src/pfs_dispatch.cc
@@ -1497,14 +1497,48 @@ static void decode_syscall( struct pfs_process *p, int entering )
 
 		case SYSCALL32_getgroups32:
 		case SYSCALL32_getgroups:
-			if (entering && pfs_fake_setgid)
-				divert_to_dummy(p,-EPERM);
+			if (entering && pfs_fake_setgid) {
+				/* Make sure the spare slot's open */
+				if (p->ngroups <= PFS_NGROUPS_MAX) {
+					for (int i = 0; i < p->ngroups; i++) {
+						if (p->groups[i] == p->egid)
+							return;
+					}
+					/* Since we didn't see the egid, append it to match Linux behavior */
+					p->groups[p->ngroups] = p->egid;
+					++p->ngroups;
+				}
+
+				if (args[0] == 0) {
+					divert_to_dummy(p,p->ngroups);
+				} else if (p->ngroups > args[0]) {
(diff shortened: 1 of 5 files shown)


parrot/src/pfs_dispatch.cc
```diff
@@ -1497,14 +1497,48 @@ static void decode_syscall( struct pfs_process *p, int entering )
...
 		case SYSCALL32_getgroups32:
 		case SYSCALL32_getgroups:
-			if (entering && pfs_fake_setgid)
-				divert_to_dummy(p,-EPERM);
+			if (entering && pfs_fake_setgid) {
+				/* Make sure the spare slot's open */
+				if (p->ngroups <= PFS_NGROUPS_MAX) {
+					for (int i = 0; i < p->ngroups; i++) {
+						if (p->groups[i] == p->egid)
+							return;
```
batrick: I think you meant to do `break;` instead of `return;` here but that's also probably not what you want. You want to jump out of the if on line 1502?

Something things to think about: what should happen if we do `setgid` before `getgroups`? What happens if the `setgroups` omits the primary group id?

trshaffer: Copy-paste error, my mistake. I had that part in a function before, and maybe it really should be in one.
My intent is that `setgroups` can be whatever, but on a call to `getgroups`, we add the egid if it's not present. If some program happens to frequently switch uid, call `getgroups`, and not drop extra groups, it might be possible to accumulate groups that way.

batrick:
> My intent is that `setgroups` can be whatever, but on a call to `getgroups`, we add the egid if it's not present.

Makes sense to me.

> If some program happens to frequently switch uid, call `getgroups`, and not drop extra groups, it might be possible to accumulate groups that way.

I think just do a simple test to see what Linux does in this situation so we can match behavior.

Also, I like the idea of a separate getgroups function in `pfs_process.cc`.

trshaffer:
Seems I wasn't looking closely enough, logind or PAM or someone *initially* adds the egid to the group list, but on subsequent calls to `setgroups` the kernel doesn't add it back, so I can just drop that bit. Since the Parrot's group list gets initialized with the actual supplementary groups, our behavior should match the host system without any extra work.

Putting a `getgroups` function in `pfs_process` does look better, and doesn't have dispatch digging around in the process struct.

50eddb3cf27f1980342cae73fde1f5ccb538938c should cover those.

batrick:
Looks great, thanks for being thorough!

PR #1068, opened 2015-12-10
Prevent openssl from opening files in $HOME.

batrick:
This was causing a significant slowdown as openssl would stat files in AFS that
it could not access.

@dthain: this is what caused the ~5 second turnaround time for push transfers in some of the Confuga experiments.


dthain:
Now that must have been a bear to track down.  Glad you found it.

However, I am concerned that setting `HOME` unilaterally may have unexpected consequences down the road.  How about a more targeted setting of `RANDFILE` instead?

https://www.openssl.org/docs/faq.html

PR #1071, opened 2015-12-10
Add batch job amazon for use in Makeflow and Work Queue

cshinaver:


dthain:
Charles - This is on the right track.   To be complete, you need to fix any build errors that result (http://ccl.cse.nd.edu/software/autobuild) and add some brief documentation about the feature to the man page (`doc/man/makeflow.m4`) and the user's manual (`doc/man/makeflow.html`)

makeflow/src/makeflow.c
```diff
@@ -1217,6 +1225,12 @@ int main(int argc, char *argv[])
 				if(log_format) free(log_format);
 				log_format = xxstrdup(optarg);
 				break;
+			case LONG_OPT_AMAZON_CREDENTIALS_FILEPATH:
+				amazon_credentials_filepath = optarg;
```
nhazekam: You should use xxstrdup(optarg) instead of just optarg.

makeflow/src/makeflow.c
```diff
@@ -1217,6 +1225,12 @@ int main(int argc, char *argv[])
...
+				break;
+			case LONG_OPT_AMI_IMAGE_ID:
+				ami_image_id = optarg;
```
nhazekam: You should use xxstrdup(optarg) instead of just optarg.

batch_job/src/amazon_ec2.sh
```diff
@@ -0,0 +1,156 @@
+#!/bin/sh
```
nhazekam: Since a large section of batch_job_amazon.c is devoted to creating this file, is it necessary to have it also included? Not sure, but it feels unneeded. If you do remove it I think we want to add it to the gitignore for cctools so it does not get inadvertently added later on.

cshinaver: The file that created is not this file. The file that is created is `_temp_amazon_ec2_script.sh`. This file is where that file comes from. I kept this file in case any changes needed to be made to the generated script. Editing the generated file in C (specifically dealing with escaping and newlines) is a big pain, so I wrote a perl script `convert.pl` that will pull from the `amazon_ec2.sh` file, escape it, and print it into the variable in the `batch_job_amazon.c` file.

nhazekam: Ok, that makes sense. I will defer to @dthain and @btovar on whether this file and convert.pl are included in the final pull request, but I can see why its included now. Thanks for the clarification.

dthain:
I'll make a slight change to clean up the build process, then merge this.

Issue #984, opened 2015-10-23
add a timeout to work queue pool


btovar:
Requested by @klannon.

When the pool does not find any masters in the catalog, it would be nice if it exit by itself after a given timeout.

PR #1080, opened 2016-01-06
Work Queue Pool Timeout

NeilB879:
Made it so that when there are no masters in the work queue pool after a given amount of time work_queue will timeout.

### DIFF
--- a/batch_job/src/work_queue_factory.c
+++ b/batch_job/src/work_queue_factory.c
@@ -635,7 +635,7 @@ static const struct option long_options[] = {
 	{"amazon-credentials", required_argument, 0, LONG_OPT_AMAZON_CREDENTIALS},
 	{"amazon-ami", required_argument, 0, LONG_OPT_AMAZON_AMI},
 	{"autosize", no_argument, 0, LONG_OPT_AUTOSIZE},
-	{"factory_timeout", required_argument, 0, LONG_OPT_FACTORY_TIMEOUT},
+	{"factory-timeout", required_argument, 0, LONG_OPT_FACTORY_TIMEOUT},
 	{0,0,0,0}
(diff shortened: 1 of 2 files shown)


dthain:
Looks like the right idea.  Two minor changes:

- Call it "factory timeout" throughout; the name changed since this bug was filed.
- Add a brief item to the man page for the tool.

dthain:
And what happens if the user doesn't set a factory timeout?

NeilB879:
@dthain I have changed it to factory-timeout, updated the man page, and fixed the behavior if factory-timeout isn't set.

Is there anything else that needs to be added to this?

dthain:
Neil, how did you go about testing this code?

NeilB879:
I ran the command:

./work_queue_factory -M test --factory-timeout 5 -T condor
To test the timeout working properly when set. This displays an error message and then exits, however since in the mainloop there is a 30 second sleep period it will actually timeout in a much longer period  then intended (unless timeout is about 30 seconds), I am unsure if it changing this sleep timer based on the factory-timeout would be beneficial.

I also have used the makeflow command:
makeflow -M test -T wq ../test/syntax/collect.makeflow &

To generate a master (with the name test)

And then run with some variations with the first command to test that it properly, exits/runs when it should.

dthain:
I don't think it's necessary to bring the resolution down below 30 seconds.
The main point is garbage collection, not precise timing.

dthain:
@btovar merge when you are ready
fixes #984

btovar:
On it.

btovar:
@NeilB879: Note I made factory_timeout_start a local variable. 

Also, added factory-timeout as a field to the configuration file.

PR #1081, opened 2016-01-06
umbrella communicates with OSF

hmeng-19:
umbrella communicates with OSF
    
        1) Upload the dependencies of one umbrella spec to OSF, and create
        a new umbrella spec with data dependencies from OSF; Upload this
        new umbrella spec to OSF
    
        2) Support dependencies from OSF during runtime


hmeng-19:
Umbrella in this pull request requires a separate python package `requests` to communicate with REST web services. 
Please rewrite it using packages inside python by default.

hmeng-19:
Instead of rewriting all the OSF-relative functions using the `urllib2` package from vanilla python, changing the `requests` package into an optional dependency of Umbrella.

hmeng-19:
@ben, @dthain , thanks for the suggestions.

Issue #1094, opened 2016-01-14
Makeflow: Large DAGs Load Slowly on Lustre


dthain:
(Nick, I know you are already working on this, just making a record of it.)

Per our discussion today, add a switch to makeflow that skips the initial stat of files when the user gives a `--skip-file-check` argument (or something to that effect.)

Then part: Set an alarm in the consistency checking code, so that if it takes more than a reasonable amount of time (say, 15 seconds) print an informational message and suggest the use of the option.

Also @btovar

btovar:
CJ reported that #1095 works for him.

PR #1095, opened 2016-01-14
Skip file check

nhazekam:
Addresses #1094 @btovar check over quickly.

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -923,7 +923,7 @@ static void show_help_run(const char *cmd)
 	printf(" %-30s Add node id symbol tags in the makeflow log.		(default is false)\n", "   --log-verbose");
 	printf(" %-30s Run each task with a container based on this docker image.\n", "--docker=<image>");
 	printf(" %-30s Load docker image from the tar file.\n", "--docker-tar=<tar file>");
-	printf(" %-30s Indicate user trusts inputs exist.\n", "--skip-file-check");
+	printf(" %-30s Do not check whether inputs exist. (Useful for very large workflows on shared filesystems.)\n", "--skip-file-check");
 	printf(" %-30s Indicate preferred master connection. Choose one of by_ip or by_hostname. (default is by_ip)\n", "--work-queue-preferred-connection");
 
 	printf("\n*Monitor Options:\n\n");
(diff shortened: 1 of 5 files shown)


makeflow/src/makeflow.c
```diff
@@ -922,6 +923,7 @@ static void show_help_run(const char *cmd)
 	printf(" %-30s Add node id symbol tags in the makeflow log.		(default is false)\n", "   --log-verbose");
 	printf(" %-30s Run each task with a container based on this docker image.\n", "--docker=<image>");
 	printf(" %-30s Load docker image from the tar file.\n", "--docker-tar=<tar file>");
+	printf(" %-30s Indicate user trusts inputs exist.\n", "--skip-file-check");
```
btovar: "Do not check whether inputs exit (Useful for very large workflows on shared filesystems.)"

Also, add to man page.

makeflow/src/makeflow_log.c
```diff
@@ -275,7 +275,7 @@ void makeflow_log_recover(struct dag *d, const char *filename, int verbose_mode,
 	dag_count_states(d);
 
 	// Check for log consistency
-	if(!first_run) {
+	if(!first_run && !skip_file_check) {
```
btovar: Should we simply be skipping the stat, and not the whole check?

nhazekam: The only operation this loop does it to determine if a file should exist according to the log, and if it should we need to stat the file and make sure it does and wasn't modified. There is no added benefit to entering the loop if we aren't going to stat the file. Since AND is used in the comparison, the first set of checks say this should be here and then we verify with stat. If we don't verify, there is no reason to stat, let me know if this should off.

nhazekam:
@btovar Let me know if you have concerns about the log recovery still. It should be noted that we still perform makeflow_check that logically verifies the Makeflow and the log keeps its old entries for files, it just doesn't recheck those file.

makeflow/src/makeflow_log.c
btovar: Ok, thanks. The rest looks good.

btovar:
Merged in dd087640b6e83e8cb98a5d6b1951668bda3f3036.

PR #1100, opened 2016-01-15
disk_alloc_create filesystem argument

nkremerh:
Adds filesystem argument to disk_alloc_create so the fs is no longer hard coded into the function


nkremerh:
What do you think, @dthain and/or @nhazekam?

dthain:
Looks good.

PR #1108, opened 2016-01-20
Critical Fix to Parrot-Bxgrid support

dthain:
Use of signed random number when selecting replicas led crashes 50% of the time.
(This didn't come up until recently, b/c bxgrid was using an old parrot binary.)


dthain:
@batrick please take a look

batrick:
Looks good to me. Merging.

PR #1112, opened 2016-01-20
disk_alloc help messages

### LINKED ISSUE #1110: Disk Allocator Usage
So, I attempted to use the disk allocator today in advance of explaining it to the CRC folks, but I didn't get very far:

```
 dthain@cclws15~$ disk_allocator --help
Invalid parameters defined.
 dthain@cclws15~$ disk_allocator --version
Invalid parameters defined.
 dthain@cclws15~$ disk_allocator
Segmentation fault (core dumped)
```

@nkremerh Please produce a useful help and version message, just like the rest of cctools.

nkremerh:
Fixes #1110

### DIFF
--- a/dttools/src/disk_allocator.c
+++ b/dttools/src/disk_allocator.c
@@ -10,16 +10,53 @@ See the file COPYING for details.
 #include <inttypes.h>
 #include <unistd.h>
 #include <string.h>
-
+#include "getopt.h"
 #include "disk_alloc.h"
 #include "stringtools.h"
+#include "cctools.h"
+
+static void show_help(const char *cmd) {
+
+	fprintf(stdout, "Use: %s [options]\n", cmd);
+	fprintf(stdout, "Where options are:\n");
+	fprintf(stdout, " %-30s This message\n", "-h,--help=<flag>");
+	fprintf(stdout, " %-30s Version\n", "-v,--version");
+	fprintf(stdout, " %-30s Loop device allocation\n", "create <target directory> <size (i.e. 100MB)> <filesystem>");
+	fprintf(stdout, " %-30s Loop device deallocation\n", "delete <target directory>");
+	fprintf(stdout, "\n");
+	return;
+}
 
 int main(int argc, char *argv[]) {
 
 	char *func = argv[1];
 	char *loc = argv[2];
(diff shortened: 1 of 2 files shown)


dttools/src/disk_allocator.c
```diff
@@ -10,16 +10,53 @@ See the file COPYING for details.
...
+	fprintf(stdout, "\n");
+	return;
+}
 
 int main(int argc, char *argv[]) {
 
 	char *func = argv[1];
 	char *loc = argv[2];
-	char *fs = argv[3];
+	char *fs = argv[4];
```
batrick: These array accesses are not safe without checking argc.

The right thing to do is let option parsing happen first, then pull extra arguments. Look at `man getopt_long` and the `optind` variable. Also look at Parrot:

https://github.com/cooperative-computing-lab/cctools/blob/master/parrot/src/pfs_main.cc#L1142

dttools/src/disk_allocator.c
```diff
@@ -10,16 +10,53 @@ See the file COPYING for details.
...
 	int result;
+	signed char c;
```
batrick: `getopt_long` returns an `int`.

dttools/src/disk_allocator.c
```diff
@@ -10,16 +10,53 @@ See the file COPYING for details.
...
 #include <string.h>
-
+#include "getopt.h"
 #include "disk_alloc.h"
 #include "stringtools.h"
+#include "cctools.h"
+
+static void show_help(const char *cmd) {
+
+	fprintf(stdout, "Use: %s [options]\n", cmd);
```
batrick: This needs to inform what the required arguments are. I believe you require 3: `func`, `loc`, and `fs`?

nkremerh:
@nhazekam, would you take a gander at this as well?

nhazekam:
Also add a man page to doc to have this information in both. I am not sure what protocol is on this but is there any need to explain what the arguments are for the command? @btovar @dthain

btovar:
Adding to the man page some examples on the most common usages is encouraged. Explaining the arguments is always needed, but only helps if you already know what are you looking for.

Issue #1116, opened 2016-01-22
MAKEFLOW_INPUTS are sometimes deleted


btovar:
If a file is listed in MAKEFLOW_INPUTS, but it is not declared as an input for any rule, makeflow will try to delete it with -c. It should either ignore it, or treat it like a constraint that should be met before running the dag.

btovar:
As an example, when running makeflow -c with the example in #1117, I get:

2016/01/22 07:43:39.72 makeflow[5537] notice: Makeflow: Couldn't delete /usr/bin/convert: Permission denied

PR #1118, opened 2016-01-22
Do not clean inputs

nhazekam:
Update for #1116 @btovar Take a look and make sure this resolves the issue.


btovar:
Thanks @nhazekam, works for me.

PR #1119, opened 2016-01-22
DVDT resources mini-loop with makeflow

btovar:
When a rule fails because of resources, it is then tried with the maximum allocation per category.
First allocation is computed automatically as jobs complete.

It works in all batch systems (not only wq).

(@dthain, @nhazekam, @charleszheng44)


nhazekam:
@btovar This looks good. I like that categories are now a unified entity for makeflow and work queue. This pull request touches a lot in makeflow and work queue, is there anything I should test, or have you done more comprehension testing?

PR #1122, opened 2016-01-22
add network support to the resource_monitor

btovar:
Example:
./resource_monitor -Omon --sh "curl www.google.com > /dev/null"

...
 "bytes_received":    [20259,"B"],
 "bytes_sent":    [169,"B"],
...

### DIFF
--- a/dttools/src/rmonitor_types.h
+++ b/dttools/src/rmonitor_types.h
@@ -75,6 +75,13 @@ struct rmonitor_io_info
 	uint64_t delta_bytes_faulted;
 };
 
+struct rmonitor_bw_info
+{
+	uint64_t start;
+	uint64_t end;
+	uint64_t bit_count;
+};
+
 struct rmonitor_file_info
 {
 	uint64_t n_references;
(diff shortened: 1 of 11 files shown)


dthain:
This is definitely moving in the right direction. One thing wasn't clear to me from a quick look -

Are you distinguishing between file I/O and network I/O based on the system call (read vs recv) or based on how the fd was created?  (open vs socket)

btovar:
As of now, based on the call because it was easy. So, if you read()/write() a socket, we would miss that. 

I have not figure out where to do the bookkeeping for the socket types. The easiest would be as global variables in the library helper, but that sounds ugly.

Also, I was thinking of just accumulating all the bytes read or received in some form, and then subtract what linux told me came from disk (same for sent). That way I do not have to investigate the socket type. Does that sound reasonable?

dthain:
My guess is that you will get into trouble comparing Linux stats against system call traces -- the two sources of information just have too many ways to diverge.

Looking at the call type is fine for now, I thin kthe socket type is the way to go in the long run.

Ok to commit, since it's moving the in right direction.

PR #1125, opened 2016-01-25
Show all compiled-in services under /

### LINKED ISSUE #1107: Parrot should have all services listed in /
Currently, this is done -- only for chirp -- in the local driver as a hack:

https://github.com/cooperative-computing-lab/cctools/blob/96c33d8e2fe44c2a4d33b475105badc272096f5c/parrot/src/pfs_service_local.cc#L370-L372

An upcoming PR moves that code to `pfs_dir.cc` but other services also need to be added (dynamically).

trshaffer:
I *think* this should resolve #1107. This adds entries under / for all services compiled into Parrot

### DIFF
--- a/parrot/src/pfs_dir.cc
+++ b/parrot/src/pfs_dir.cc
@@ -28,9 +28,30 @@ extern "C" {
 pfs_dir::pfs_dir( pfs_name *n ) : pfs_file(n)
 {
 	iterations = 0;
-	/* FIXME This should include all services, not just Chirp. See issue #1107. */
 	if(strcmp(n->path, "/") == 0) {
 		append("chirp");
+		append("multi");
+		append("anonftp");
+		append("ftp");
+		append("http");
+		append("grow");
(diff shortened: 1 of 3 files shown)


batrick:
Can you think of a way to make this more dynamic? Perhaps unify the list of services in 

https://github.com/cooperative-computing-lab/cctools/blob/9f1cb12e8e1e6cf49539278aae2b5ddb67aa4c38/parrot/src/pfs_service.cc#L349-L400

[This is a programming exercise: make it so we add services to a single location instead of two.]

trshaffer:
I shifted the service name mappings into a hash table, so now service lookup really is just a lookup. I also put the service definitions in `pfs_main` so it's feasible to update the available services based on command line options, environment, etc.

The one failure that shows up across the board in autobuild is the O_DIRECTORY bug. If it's a problem, I can merge master into this branch so all the tests pass.

Issue #1104, opened 2016-01-18
`open` of directory without `O_DIRECTORY` does not allow `getdents`


batrick:
If a directory is not opened with `O_DIRECTORY`, then Parrot does not allow getting its entries via `getdents`. It will return `ENOTDIR`.

`pfs_table.cc` only opens directories with this flag:

https://github.com/cooperative-computing-lab/cctools/blob/master/parrot/src/pfs_table.cc#L597-L615

`O_DIRECTORY` was not meant to be interpreted this way. The man page says:

````
       O_DIRECTORY
              If pathname is not a directory, cause the open to fail.  This flag was added in kernel version 2.1.126, to avoid  denial-of-service  problems
              if opendir(3) is called on a FIFO or tape device.
````

So it's only intended to make `open` fail if the file is not a directory.

Right now I think the fix may be to try `pname.service->getdir(&pname)` first. If that fails with `ENOTDIR`, then proceed with a regular `pname.service->open(&pname,flags,mode)`. What do you think @dthain?

This was reported by Rodney Walker. You can reproduce it using:

`mkdir foo && ./parrot_run -d all -o debug /bin/tar -czf foo.tgz foo/`

batrick:
@dthain, this fixes Rodney's problem:

https://github.com/batrick/cctools/commit/415315ef65610f2a1d70d91cb19fa34a6296e768

but I think the O_DIRECTORY logic in pfs_table.cc is still wrong. If tar were executed on a directory in Chirp, we get this:

````
$ rm foo.tgz debug; ./parrot_run -d all -o debug /bin/tar czf foo.tgz /chirp/disc01.crc.nd.edu/
/bin/tar: Removing leading `/' from member names
/bin/tar: /chirp/disc01.crc.nd.edu: Cannot open: Is a directory
/bin/tar: Exiting with failure status due to previous errors
$ grep -C5 'Is a directory' debug 
2016/01/18 10:39:45.41 parrot_run[3398] <child:3400> syscall: mprotect
2016/01/18 10:39:45.41 parrot_run[3398] <child:3399> syscall: openat
2016/01/18 10:39:45.41 parrot_run[3398] <child:3399> debug: complete_at_path: `/chirp/disc01.crc.nd.edu' -> `/chirp/disc01.crc.nd.edu'
2016/01/18 10:39:45.41 parrot_run[3398] <child:3399> libcall: open /chirp/disc01.crc.nd.edu 657664 0
[...]

PR #1126, opened 2016-01-25
Parrot POSIX Directory Semantics

trshaffer:
This is a more complete fix for #1104. These commits try to bring several main services more in line with POSIX behavior. To avoid breaking all the other services, there's a marker for services that can correctly do `O_DIRECTORY` (local, chirp, cvmfs, and irods at the moment). I'm still not confident that all the corner cases and errnos are correct, so I'll test more with these services. All the others shouldn't see a difference from current behavior.

### DIFF
--- a/parrot/src/pfs_service_cvmfs.cc
+++ b/parrot/src/pfs_service_cvmfs.cc
@@ -1153,10 +1153,6 @@ class pfs_service_cvmfs:public pfs_service {
 		return 1;
 	}
 
-	virtual int can_open_dirs() {
-		return 1;
-	}
-
 	virtual pfs_file *open(pfs_name * name, int flags, mode_t mode) {
 		struct cvmfs_dirent d;
 
--- a/parrot/src/pfs_service_cvmfs.cc
+++ b/parrot/src/pfs_service_cvmfs.cc
@@ -1165,63 +1165,13 @@ class pfs_service_cvmfs:public pfs_service {
 			return 0;
 		}
 
-		if((!S_ISDIR(d.mode)) && (flags&O_DIRECTORY)) {
-			errno = ENOTDIR;
+		/* cvmfs_open does not work with directories (it gives a 'fail to fetch' error). */
+		if(S_ISDIR(d.mode)) {
+			errno = EISDIR;
 			return 0;
 		}
 
 		debug(D_CVMFS,"open(%s)",name->rest);
-
-		if(S_ISDIR(d.mode)) {
-			struct cvmfs_dirent d;
-
-			/*
-			If the root of the CVFMS filesystem is requested,
-			we must generate it internally, containing the
-			list of the known filesystems.
-			*/
-
-			if(!name->host[0]) {
-				pfs_dir *dir = new pfs_dir(name);
-				dir->append(".");
-				dir->append("..");
-				CvmfsFilesystemList::const_iterator i    = cvmfs_filesystem_list.begin();
-				CvmfsFilesystemList::const_iterator iend = cvmfs_filesystem_list.end();
-				for(; i != iend; ++i) {
-					cvmfs_filesystem *f = *i;
-					/* If the host begins with dot, then it is a wildcard entry. */
-					/* Otherwise, it is a normal entry. */
-					const char *host = f->host.c_str();
-					if(host && host[0]!='.') {
-						dir->append(host);
-					}
-				}
-				return dir;
-			}
-
-			/*
-			Otherwise, go to CVMFS for the directory liting.
-			*/
-
-			pfs_dir *dir = new pfs_dir(name);
-
-			char **buf = NULL;
-			size_t buflen = 0;
-
-			int rc = compat_cvmfs_listdir(name->rest, &buf, &buflen);
-
-			if(rc<0) return 0;
-
-			int i;
-			for(i = 0; buf[i]; i++) {
-				dir->append(buf[i]);
-				free(buf[i]);
-			}
-			free(buf);
-
-			return dir;
-		}
-
 		int fd = compat_cvmfs_open(name->rest);
 
 		if(fd<0) return 0;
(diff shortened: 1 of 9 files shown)


parrot/src/pfs_table.cc
```diff
@@ -654,7 +650,12 @@ pfs_file * pfs_table::open_object( const char *lname, int flags, mode_t mode, in
 			} else {
 				file = pname.service->open(&pname,flags,mode);
 			}
-			free(fd);
+		} else if(!pname.service->can_open_dirs() && (flags&O_DIRECTORY)) {
+			// This is a hack to maintain compatibility with the interpretation
+			// of O_DIRECTORY used in some of the services. If getdents fails
+			// on certain remote services or similar strangeness,
+			// this is a good place to start looking.
+			file = pname.service->getdir(&pname);
```
batrick: Now that we have `can_open_dirs()`, we don't need to implement `getdir` for `local`, `cvmfs`, and `chirp`. I think it's safe to remove the those.

parrot/src/pfs_service_irods.cc
```diff
@@ -92,28 +92,30 @@ class pfs_service_irods : public pfs_service {
 
 		struct irods_file *ifile;
 
-		ifile = irods_reli_open(name->hostport,name->rest,flags,mode);
+		if(!(flags&O_DIRECTORY)) {
+			ifile = irods_reli_open(name->hostport,name->rest,flags,mode);
+		}
 		if(ifile) {
 			return new pfs_file_irods(name,ifile);
+		} if((errno == EISDIR) || (flags&O_DIRECTORY)) {
```
batrick: `else if`?

parrot/src/pfs_service_cvmfs.cc
```diff
@@ -1165,82 +1169,73 @@ class pfs_service_cvmfs:public pfs_service {
...
-
-		char **buf = NULL;
-		size_t buflen = 0;
-
-		debug(D_CVMFS, "getdir(%s)", name->rest);
-		int rc = compat_cvmfs_listdir(name->rest, &buf, &buflen);
+		int fd = compat_cvmfs_open(name->rest);
 
-		if(rc<0) return 0;
+		if(fd<0) return 0;
```
batrick: This needs a `delete dir;` prior to return. (Missing in the prior code too...)

batrick: Whoops wrong, part of the patch! Disregard.

parrot/src/pfs_service_cvmfs.cc
```diff
@@ -1165,82 +1169,73 @@ class pfs_service_cvmfs:public pfs_service {
...
-				}
-			}
-			return dir;
-		}
+			int rc = compat_cvmfs_listdir(name->rest, &buf, &buflen);
 
-		/*
-		Otherwise, go to CVMFS for the directory liting.
-		*/
+			if(rc<0) return 0;
```
batrick: Put `pfs_dir *dir = new pfs_dir(name);` below this if statement so we don't leak memory when `compar_cvmfs_listdir` fails.

dthain:
After looking at this PR in more detail, I'm more confused than when I started.

The old behavior was this:

In the Linux system call API, a user can call open() on either a file or a directory.  However, the various remote services tend to distinguish clearly between file opens and directory opens, so they have open(), getdir(), and getlongdir().  Linux open() would eventually reach pfs_table.cc::open_object, which would then examine the open flags (and possibly other things) and then decide whether to invoke service->open or service->getdir.

As I read the new code, it seems like it is combining the open() and getdir() operations in each of the underlying services, and then devolving the check for open flags down to each driver.  This seems counterproductive to me, since it is taking logic that was in one place, and then copying it out to multiple drivers, which now have inconsistent methods.

Looking back at the original bug #1104, my sense is that the problem needs to be solved by adjusting the logic in pfs_table::open_object(), not by refactoring the service modules.  But perhaps I haven't fully understood what's going on.

@trshaffer can you clarify exactly what the bug is here, and how you are attempting to fix it?

trshaffer:
The old behavior was to use the `O_DIRECTORY` flag to determine whether to use `open()` or `getdir()`, which caused problems for odd programs that open directories without `O_DIRECTORY`, hence the original issue. For correct behavior, we need to detect if the path we're opening is *not* a directory and fail if `O_DIRECTORY` was passed, and otherwise return a file/directory depending on what the underlying service does.

I was trying to push this logic down to the services so that they can do these checks under whatever semantics they define. If we assume that no remote service can open directories with `open()`, and that they all fail with appropriate errnos, We can probably keep the `O_DIRECTORY` logic in `pfs_table` and fudge the errnos there.

The other issue is that I didn't want to make too many assumptions about every service, and was worried that changing this at `pfs_table` would introduce more  irregular bugs that only show up down the road on particular services. Thus I was marking major services and trying to maintain previous (incorrect) behavior on the other services.


I'll take another shot at this via `pfs_table` and see if I run into any other issues

dthain:
Ok, that's a helpful explanation, I think we are on the right track.

So the problem is that a Unix program can issue open() and then we don't know whether to expect read() or getdents() next, even though only one of those is possible.

Of the underlying services, I believe that only the local service is the one that would actually support that behavior, right?

So, one approach is that Unix open() could first stat() the underlying file, and then decide whether to open as a file or directory, right?

batrick:
> So the problem is that a Unix program can issue open() and then we don't know whether to expect read() or getdents() next, even though only one of those is possible.
>
>Of the underlying services, I believe that only the local service is the one that would actually support that behavior, right?

This affects all services. If Rodney had run `tar` instead on something in /cvmfs or /chirp, he would have had the same failure. If `tar` opens the directory without `O_DIRECTORY`, then `pfs_table` will open a chirp or cvmfs directory as a regular file and fail.

So: we should not rely on/use `O_DIRECTORY` in `pfs_table`. Instead, the service should figure out whether or not it is opening a directory and return the appropriate `pfs_file` or its subclass (`pfs_dir`). If the flags included `O_DIRECTORY` and the file is not a directory, then the service should return a failure.

>So, one approach is that Unix open() could first stat() the underlying file, and then decide whether to open as a file or directory, right?

We already have local fixed. It does an `open` and then `fstat`:

https://github.com/cooperative-computing-lab/cctools/commit/cf6c32992446850e611855794180fa8b91e8db86

batrick:
>Of the underlying services, I believe that only the local service is the one that would actually support that behavior, right?

Sorry I misread your question. Yes, to my knowledge, `local` (unix) is the only service that returns a handle that can be used for `getdents` or `read`.

trshaffer:
Are the there any restrictions on how remote services implement `open()` and `getdir()`? If we require that remote services fail with `EISDIR` when `open()`ing directories, we might be able to have `pfs_table` look at the combination of `O_DIRECTORY`, `O_WRONLY`/`O_RDWR`, and `errno` to figure out the right thing to return. We would just need to make local a special case

dthain:
Tim you are right, that's what @batrick and I were just discussing.. . He will elaborate here, and then you can take another crack at it.

batrick:
@trshaffer, as @dthain said we think having `service::open` return `EISDIR` is the right way to go to minimize changes to services. I guess the real bug here (from Parrot's perspective) was that `local` would happily open directory as an ordinary file.

So I suggest the following changes:

(1) Modify `pfs_table::open_object` to do `pfs_service::getdir` if `O_DIRECTORY ` is set. Otherwise do `pfs_service::open`. If that fails with `EISDIR`, try `getdir`.
(2) The other loose end then is the `local` driver which needs to properly return `EISDIR`. Go ahead and revert cf6c329's changes and fix `open` to return `EISDIR` when the file is a directory.

trshaffer:
This pile of commits seems to be passing tests. This time around, most logic is in `pfs_table` and we tell local to fail if we try to `open()` a directory.

dthain:
This approach looks good to me; I'll leave it to @batrick for detailed review.

parrot/src/pfs_table.cc
```diff
@@ -660,10 +677,26 @@ pfs_file * pfs_table::open_object( const char *lname, int flags, mode_t mode, in
 				file = pfs_cache_open(&pname,flags,mode);
```
batrick: I believe we need to check for `EISDIR` here as well as the other `pfs_cache_open`. At this point, I think it's best to have an `open_directory` function that handles the `flags` logic...

trshaffer:
OK, I just put in checks for `pfs_cache_open`

batrick:
Good work @trshaffer!

PR #1128, opened 2016-01-26
Update Chirp HDFS Docs (and maybe code)

dthain:
The world has changed somewhat since we originally did the code for these environment variables.  It seems that libhdfs is now a separate package, and the library usually appears in /usr/lib64/libhdfs.so.  So, I have changed the docs to reflect the more common situation.

If the user doesn't set `LIBHDFS_PATH`, our code will search for it in `HADOOP_HOME`, which I think is no longer a common configuration.

I suggest that we simplify that setup to be `dlopen` on either `$LIBHDFS_PATH` (if set) otherwise `libhdfs.so` and skip searching through `HADOOP_HOME`.

(But we still need to search `JAVA_HOME` for `libjvm.so`)

@batrick What do you think?

### DIFF
--- a/doc/chirp.html
+++ b/doc/chirp.html
@@ -871,7 +871,8 @@ limitations, discussed below). Perhaps best of all, client jobs will no longer
 have any Hadoop or Java (version) dependencies.</p>
 
 <p>To run a Chirp server as a frontend to your HDFS filesystem,
-you will need to first set several environment variables which
+you will need to install the <tt>libhdfs-devel</tt> package
+and then set several environment variables which
 describe your Hadoop installation.  JAVA_HOME and HADOOP_HOME
 LIBHDFS_PATH should be set to indicate the location of the libhdfs library.
 Common values for the Cloudera Hadoop installation would be this:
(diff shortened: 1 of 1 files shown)


batrick:
> If the user doesn't set LIBHDFS_PATH, our code will search for it in HADOOP_HOME, which I think is no longer a common configuration.
>
> I suggest that we simplify that setup to be dlopen on either $LIBHDFS_PATH (if set) otherwise libhdfs.so and skip searching through HADOOP_HOME.

Since we always use `$LIBHDFS_PATH` if available, I don't see the harm in keeping the search through `$HADOOP_HOME` if `$LIBHDFS_PATH` is not set?

dthain:
Fine with me, does the doc update seem reasonably clear to you?

batrick:
Yup, merging.

batrick:
After playing around with running the chirp server myself, I think I now understand what you were saying.

> I suggest that we simplify that setup to be dlopen on either $LIBHDFS_PATH (if set) otherwise libhdfs.so and skip searching through HADOOP_HOME.

So we want to do `dlopen("libhdfs.so")` if `$LIBHDFS_PATH` is not set and let the loader find it in the system paths? I think that makes sense. We could even keep the nasty `$HADOOP_HOME` fallback.

In any case, we still sadly need `$HADOOP_HOME` because the `$CLASSPATH` needs updated with all of the Hadoop jar files. : /

PR #1130, opened 2016-01-26
Fix Chirp HDFS Crash on rmall

dthain:
libhdfs crashes if you call unlink(x) and x does not exist.
"rmall" needs to go through do_unlink to check for errors first.

### DIFF
--- a/chirp/src/chirp_fs_hdfs.c
+++ b/chirp/src/chirp_fs_hdfs.c
@@ -279,20 +279,20 @@ HDFS is known to return bogus errnos from unlink,
 so check for directories beforehand, and set the errno
 properly afterwards if needed.
 */
-static INT64_T do_unlink(const char *path)
+static INT64_T do_unlink(const char *path, int recursive )
 {
 	struct chirp_stat info;
 
 	if(do_stat(path, &info) < 0)
 		return -1;
 
-	if(S_ISDIR(info.cst_mode)) {
+	if(!recursive && S_ISDIR(info.cst_mode)) {
 		errno = EISDIR;
 		return -1;
 	}
 
 	debug(D_HDFS, "unlink %s", path);
-	if(hdfs_services->unlink(fs, path, 0) == -1) {
+	if(hdfs_services->unlink(fs, path, recursive) == -1) {
 		errno = EACCES;
 		return -1;
 	}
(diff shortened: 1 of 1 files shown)


dthain:
@batrick please double check for me

batrick:
Trying to reproduce this, I just get this error:

````
2016/01/26 13:29:42.42 chirp_server[22634] chirp: rmdir /foo
2016/01/26 13:29:42.42 chirp_server[22634] hdfs: stat /users/pdonnel3/foo
2016/01/26 13:29:42.42 chirp_server[22634] hdfs: stat /users/pdonnel3/.__acl
2016/01/26 13:29:42.42 chirp_server[22634] hdfs: opening file /users/pdonnel3/.__acl (flags: 0) for reading; mode: 600
2016/01/26 13:29:42.43 chirp_server[22634] hdfs: pread 0 65535 0
2016/01/26 13:29:42.43 chirp_server[22634] hdfs: pread 0 65535 20
2016/01/26 13:29:42.43 chirp_server[22634] hdfs: close /users/pdonnel3/.__acl
2016/01/26 13:29:42.43 chirp_server[22634] hdfs: rmdir /users/pdonnel3/foo
2016/01/26 13:29:42.44 chirp_server[22634] chirp: zoiks, I don't know how to transform error 5 (Input/output error)
2016/01/26 13:29:42.44 chirp_server[22634] chirp: = -127 (Input/output error)
````

Is this the crash? Or is it a segfault?

File: chirp/src/chirp_fs_hdfs.c [hunk 1]
```diff
@@ -563,7 +564,7 @@ static INT64_T chirp_fs_hdfs_rmdir(const char *path)
 {
 	RESOLVE(path)
 	debug(D_HDFS, "rmdir %s", path);
-	return hdfs_services->unlink(fs, path, 1);
+	return do_unlink(path,1);
```
batrick: So this part bothered me because I think `rmdir` here should not be recursive. Would you agree?

dthain:
I think you are seeing the right thing.  In this case the chirp server drops the connection, and then parrot just retries endlessly. With the patch, you see this behavior instead:

2016/01/26 13:46:20.81 chirp_server[27769] hdfs: rmdir /testing/hosts
2016/01/26 13:46:20.81 chirp_server[27769] hdfs: stat /testing/hosts
2016/01/26 13:46:20.81 chirp_server[27769] chirp: = -3 (No such file or directory)

chirp/src/chirp_fs_hdfs.c
dthain: I'm inclined to agree, but I thought the existing behavior might have been deliberate.
Let's me see if it works the other way...

batrick: I'll merge the PR as it is and fork an issue for this.

PR #1133, opened 2016-01-27
Makeflow Storage Footprint

nhazekam:
@btovar Give this a look when you can. Let me know if there are questions, particularly about descendant calculations.


makeflow/src/dag_file.c
```diff
@@ -12,14 +12,20 @@ See the file COPYING for details.
 
 #include <stdlib.h>
 
+#define MAKEFLOW_DEFAULT_FILE_SIZE 1024
+////*1024*1024
```
btovar: Is this in MB?  In any case, please add a comment indicating the units.
Also, remove the comment on line 16.

makeflow/src/dag_file.c
```diff
@@ -12,14 +12,20 @@ See the file COPYING for details.
...
+
 struct dag_file * dag_file_create( const char *filename )
 {
 	struct dag_file *f = malloc(sizeof(*f));
 	f->filename = xxstrdup(filename);
 	f->needed_by = list_create();
+	f->file_size = 0;
+	f->est_size = MAKEFLOW_DEFAULT_FILE_SIZE;
```
btovar: With so many variables, let's be verbose: f->estimated_size?

makeflow/src/dag_file.h
```diff
@@ -26,14 +26,23 @@ typedef enum {
 	DAG_FILE_STATE_UP
 } dag_file_state_t;
 
+typedef enum {
+	DAG_FILE_TYPE_INPUT,
+	DAG_FILE_TYPE_OUTPUT,
+	DAG_FILE_TYPE_INTERMEDIATE
```
btovar: Could you add comments to these ones? Specially because  DAG_FILE_TYPE_INTERMEDIATE could be input/output of a node, but INPUT/OUTPUT refers to the dag. Correct?

makeflow/src/dag_file.h
```diff
@@ -26,14 +26,23 @@ typedef enum {
...
-	struct list     *needed_by;              /* List of nodes that have this file as a source */
-	struct dag_node *created_by;             /* The node (if any) that created the file */
-	int    ref_count;                        /* How many nodes still to run need this file */
-	time_t creation_logged;                  /* Time that file creation is logged */
-	dag_file_state_t state;                  /* Enum: DAG_FILE_STATE_{INTIAL,EXPECT,...} */
+	struct list     *needed_by;		/* List of nodes that have this file as a source */
+	struct dag_node *created_by;	/* The node (if any) that created the file */
+	int    ref_count;				/* How many nodes still to run need this file */
+	uint64_t file_size;				/* File size used for calculating node size */
+	uint64_t est_size;				/* Estimated file size used for calculating node size */
```
btovar: estimated_size?

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
 	target->created_by = n;
 }
 
+void dag_node_prepare_node_size(struct dag_node *n)
```
btovar: I suggest to add documentation to these functions. In particular, why do we compute the size of the file differently whether it should exists or not?

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+	while((f = list_next_item(n->target_files))){
+		if(dag_file_should_exist(f))
+			n->target_size += f->file_size;
+		else
+			n->target_size += f->est_size;
+	}
+
+	set_first_element(n->descendants);
+	while((s = set_next_element(n->descendants))){
+		dag_node_prepare_node_size(s);
```
btovar: Note this function does not need to be recursive. It will be more efficient if you transverse the d->nodes list.

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+	}
+}
+
+uint64_t dag_node_determine_parent_weight(struct dag_node *n)
+{
+	struct dag_node *d;
```
btovar: We usually call the dag "d". I suggest picking another letter for this node.

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+	while((d = set_next_element(n->ancestors))){
+		weight += d->target_size;
+	}
+
+	return weight;
+}
+
+uint64_t dag_node_determine_child_weight(struct dag_node *n)
+{
+	struct dag_node *d;
```
btovar: Same as above, let's keep d for dag.

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+uint64_t dag_node_determine_descendant_weight(struct dag_node *n)
+{
+	struct dag_node *d, *e;
+	struct dag_node_size *s, *t;
+	struct set *tmp = set_create(0);
+
+	set_first_element(n->descendants);
+	while((d = set_next_element(n->descendants))){
+		set_push(tmp, d);
+		list_first_item(d->res_nodes);
```
btovar: residual_nodes?
weight_nodes?

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+
+	list_delete(n->res_nodes);
+	list_delete(n->wgt_nodes);
+	list_delete(n->run_nodes);
+
+	set_first_element(n->descendants);
+	if(set_size(n->descendants) > 1){
+		n->res_nodes = list_create();
+		n->wgt_nodes = list_create();
+		n->run_nodes = list_create();
```
btovar: running? already_ran?

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+		set_first_element(n->descendants);
+
+		int comp = 1;
+		while(comp){
+			d = set_next_element(n->descendants);
+			s = list_next_item(d->res_nodes);
+			while((d = set_next_element(n->descendants))){
+				t = list_next_item(d->res_nodes);
+				if((t && !s) || (s && !t) || (t->n->nodeid != s->n->nodeid))
```
btovar: Why are you comparing by ids? Aren't these the same object in memory?

makeflow/src/dag_node.c
```diff
@@ -18,6 +18,16 @@ See the file COPYING for details.
 #include <stdlib.h>
 #include <unistd.h>
 
+int dag_node_comp(void *item, const void *arg)
```
btovar: Why do you compare by nodeid? Simply saying d == e should be enough.

Unless we have to objects in memory with the same nodeid. If that is so, something went really wrong.

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+		}
+
+		comp = 1;
+		while(comp){
+			d = set_next_element(n->descendants);
+			s = list_next_item(d->wgt_nodes);
+			while((d = set_next_element(n->descendants))){
+				t = list_next_item(d->wgt_nodes);
+				if((t && !s) || (s && !t) || (t->n->nodeid != s->n->nodeid))
+					comp = 0;
```
btovar: It seems you want to break as soon as comp is zero.

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+	if(set_size(n->descendants) > 1){
+		n->res_nodes = list_create();
+		n->wgt_nodes = list_create();
+		n->run_nodes = list_create();
+		set_first_element(n->descendants);
+
+		int comp = 1;
+		while(comp){
+			d = set_next_element(n->descendants);
+			s = list_next_item(d->res_nodes);
```
btovar: You seem to be getting next items without setting first element first.

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+
+			set_first_element(n->descendants);
+			if(comp)
+				list_push_tail(n->wgt_nodes, s);
+		}
+
+		uint64_t node_wgt, max_wgt, tmp_wgt, res_size;
+		max_wgt = 0;
+		set_first_element(tmp);
+		while((d = set_next_element(tmp))){
```
btovar: From here then on we need comments...

makeflow/src/dag_node.c
```diff
@@ -220,6 +249,209 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
...
+	list_push_tail(n->res_nodes, dag_node_size_create(n, n->target_size));
+
+	if((n->parent_wgt >= n->child_wgt) && (n->parent_wgt >= n->descendant_wgt)){
+		list_push_tail(n->wgt_nodes, dag_node_size_create(n, n->parent_wgt));
+	} else if((n->child_wgt >= n->parent_wgt) && (n->child_wgt >= n->descendant_wgt)){
+		list_push_tail(n->wgt_nodes, dag_node_size_create(n, n->child_wgt));
+	} else {
+		list_push_tail(n->wgt_nodes, dag_node_size_create(n, n->descendant_wgt));
+	}
+	n->updated = 1;
```
btovar: updated in which respect? footprint_updated?

makeflow/src/dag_node.h
```diff
@@ -72,9 +83,19 @@ struct dag_node {
 	time_t previous_completion;
 
 	struct dag_node *next;              /* The next node in the list of nodes */
+
+	int updated;						/* Int indicating this node has updated its size */
```
btovar: size_updated, footprint_updated?

makeflow/src/dag_node.h
```diff
@@ -72,9 +83,19 @@ struct dag_node {
...
+};
+
+struct dag_node_size {
+	struct dag_node *n;
+	uint64_t size;
```
btovar: Why do we have another structure for this? Isn't it easier to add size as a field of node?

makeflow/src/makeflow.c
```diff
@@ -911,6 +926,7 @@ static void show_help_run(const char *cmd)
 	printf(" %-30s Work Queue keepalive timeout.			   (default is %ds)\n", "-t,--wq-keepalive-timeout=<#>", WORK_QUEUE_DEFAULT_KEEPALIVE_TIMEOUT);
 	printf(" %-30s Work Queue keepalive interval.			  (default is %ds)\n", "-u,--wq-keepalive-interval=<#>", WORK_QUEUE_DEFAULT_KEEPALIVE_INTERVAL);
 	printf(" %-30s Show version string\n", "-v,--version");
+	printf(" %-30s Set storage limit within which Makeflow will execute\n", "--storage-limit");
```
btovar: Copy --wq-keepalive-interval format to indicate a required argument.

nhazekam:
@btovar I think this is ready for a look. If sections need more comments let me know. Also this is quite a few commits and I think for the most part they can be quashed is need be.

dttools/src/set.c
```diff
@@ -174,6 +202,28 @@ int set_insert(struct set *s, const void *element)
...
 
+int set_insert_set(struct set *s, struct set *s2)
+{
+	set_first_element(s2);
+	const void *element;
+	while((element = set_next_element(s2))){
+		set_insert(s, element);
+	}
+
+	return 1;
```
btovar: Here and below you always return 1. Either return void, or the number of elements added.

makeflow/src/dag_file.c
```diff
@@ -12,14 +12,19 @@ See the file COPYING for details.
 
 #include <stdlib.h>
 
+#define MAKEFLOW_DEFAULT_FILE_SIZE 1024*1024*1024 // Default size of 1GB
```
btovar: Use GIGABYTE from macros.h

makeflow/src/dag.c
```diff
@@ -420,5 +421,17 @@ int dag_local_jobs_running( struct dag *d )
 	return itable_size(d->local_job_table);
 }
 
-
+uint64_t dag_absolute_filesize( struct dag *d )
```
btovar: This function counts files even if they have been garbage collected. Is that ok?

nhazekam: Good point. I'll take that into account.

makeflow/src/dag_node.c
```diff
@@ -49,22 +50,213 @@ struct dag_node *dag_node_create(struct dag *d, int linenum)
...
-	case DAG_NODE_STATE_FAILED:
-		return "failed";
-	case DAG_NODE_STATE_ABORTED:
-		return "aborted";
-	default:
-		return "unknown";
+	struct dag_node_footprint *f;
+
+	f = malloc(sizeof(struct dag_node_footprint));
+	memset(f, 0, sizeof(struct dag_node_footprint));
```
btovar: Just a little thing:

You can either do: f = calloc(1, sizeof(struct dag_node_footprint)); which automatically zeros f, or
use bzero(f, sizeof(struct dag_node_footprint)); instead of memset(..., 0, ...

makeflow/src/dag_node.c
```diff
@@ -49,22 +50,213 @@ struct dag_node *dag_node_create(struct dag *d, int linenum)
...
+
+	f->children_updated = 0;
+	f->size_updated = 0;
+	f->footprint_updated = 0;
+	f->terminal_updated = 0;
+
+	return f;
+}
+
+void dag_node_footprint_delete(struct dag_node_footprint *f);
```
btovar: Why not move the definiton here? Why do it with the forward declaration?

makeflow/src/dag_node.c
```diff
@@ -49,22 +50,213 @@ struct dag_node *dag_node_create(struct dag *d, int linenum)
...
+	set_delete(f->footprint_max_files);
+
+	set_delete(f->res_files);
+	set_delete(f->wgt_files);
+	set_delete(f->max_wgt_files);
+
+	free(f);
+}
+
+int dag_node_compare(struct dag_node *node1, struct dag_node *node2, dag_node_sort_t type)
```
btovar: This function is cryptic... Add some comments, and choose better names for size{1,2,3,4}

btovar: In fact, if you are only using this function for the comparators below, I would eliminate this function and add the code to the corresponding comparators.

makeflow/src/dag_node.c
```diff
@@ -223,6 +415,607 @@ void dag_node_add_target_file(struct dag_node *n, const char *filename, char *re
 	target->created_by = n;
 }
 
+/* Function that calculates the three different footprint values and
```
btovar: What are those three different footprints?

I would add a detailed description of the algorithm somewhere in this file, referring to functions as appropriate.

dttools/src/set.c
nhazekam: If we return a number, should we return the number added, or the number of element that now exist in the set. More clearly, what should we return in the element already exists in the set.

dthain:
Nick, please revisit this after finals.  I would like to see it merged before we forget how it works...

dthain:
What will it take to get this one merged?

nhazekam:
I will resolve the conflicts and I think this is good to go. Not sure what needs to happen in terms of documentations for use.

PR #1136, opened 2016-01-28
Fix PATH setting in cctools README

hmeng-19:
Fix PATH setting in cctools README


hmeng-19:
@btovar , is this ready to be merged?

btovar:
Go ahead.

PR #1137, opened 2016-02-01
remove exagerated makeflow D_NOTICEs

btovar:


makeflow/src/makeflow_gc.c
```diff
@@ -147,16 +147,14 @@ int makeflow_clean_file( struct dag *d, struct batch_queue *queue, struct dag_fi
 		makeflow_log_file_state_change(d, f, DAG_FILE_STATE_DELETE);
 
 		if(!silent)
-			debug(D_NOTICE, "Makeflow: Deleted path %s\n", f->filename);
+			debug(D_MAKEFLOW_RUN, "Makeflow: Deleted path %s\n", f->filename);
```
nhazekam: I understand removing the others, but this is the only location where the removal of a file is reported to the user. In older versions of Makeflow we had a comment such as "makeflow deleted file ..." that always printed when a file was cleaned. @btovar Should clean reported removed files if not in debug?

btovar: Are you ok with makig it a fprintf(stdout, ...)? A D_NOTICE does not seem to fit here.

nhazekam: That seems fine, though I am also ok with not reporting it. I just wanted to clarify that it was originally printed to the user and I used D_NOTICE to pass it through a cctools output stream instead of directly to the user.

PR #1147, opened 2016-02-09
Makeflow example simple

### LINKED ISSUE #1146: Makeflow example should not write rules depending on `/usr/bin/curl`
Since ff2e362b, `makeflow/example/example.makeflow`:

````
capitol.jpg: $CURL
    LOCAL $CURL -o capitol.jpg $UR
````

This is troublesome because curl may not be on the execution site and there is no way to check for it. This causes the simple makeflow+chirp|confuga test to fail:

````
$ makeflow --auth=unix -T chirp --working-dir=chirp://disc01.crc.nd.edu:9194/users/pdonnel3/ makeflow/example/example.makeflow
parsing makeflow/example/example.makeflow...
checking makeflow/example/example.makeflow for consistency...
makeflow: /usr/bin/curl does not exist, and is not created by any rule.
````

Something else to consider: while having an example for resource descriptions is great, it may be intimidating for new users that benefit from a dead simple example.

### PR DESCRIPTION
**Pull request by btovar**
Fixes #1146.

### DIFF
--- a/makeflow/example/example.categories.makeflow
+++ b/makeflow/example/example.categories.makeflow
@@ -1,0 +1,55 @@
+# This is a sample Makeflow script that retrieves an image from the web,
+# creates four variations of it, and then combines them into an animation.
+# It produces the same output as example.makeflow, but it introduces the syntax
+# to group rules in sets, called categories. The computational resources needed
+# for a rule in a category can also be optionally specified.
+
+# A Makeflow script is a subset of the Makefile language (see the manual
+# for more information). For example, for convenience we can define the
+# textual substitutions:
+CURL=/usr/bin/curl
+CONVERT=/usr/bin/convert
+URL=http://ccl.cse.nd.edu/images/capitol.jpg
+
+# We specify the set of inputs and outputs. This is not required, but strongly
+# recommended.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: makeflow/example/example.categories.makeflow [hunk 1]
```diff
@@ -0,0 +1,55 @@
...
+
+capitol.360.jpg: capitol.jpg
+	$CONVERT -swirl 360 capitol.jpg capitol.360.jpg
+
+# If a rule is preceded by LOCAL, it executes at the local site.
+CATEGORY=preprocess
+CORES=1
+MEMORY=10
+DISK=20
+capitol.jpg:
```
batrick: So is the plan right now to not add the `curl` and `convert` dependencies?

I think it would be great if we had another example that dealt with distributing the locally installed `curl` and `convert` along with its implicit dependencies. I believe we would use `starch` for that right?

btovar: Correct. 

I am just reverting to the example.makeflow before all the fancy changes. A more complete example would be excellent, yes.

batrick: Alright, I'll fork an issue for that and see about merging this as-is.

PR #1155, opened 2016-02-09
Prevent openssl from opening ~/.rand.

batrick:
Commit a303060 missed the chirp client code.


batrick:
@btovar this needs to be in the next release. The disc Confuga cluster is not operational without this fix.

PR #1161, opened 2016-02-12
Linearish wq codes

btovar:
As requested by @matz-e.  task->result only returns a single error code, rather than an "or" of codes.

The only codes that are ever or-ed are missing stdout and missing output. This pr makes it so that if both are present, only missing output is returned.

Also, care was taken so that the older codes (missing input, output, stdout) may still be compared with & and ==, while newer codes can be compared with ==. With this, a simple switch case can be done on task->result, rather than multiple ifs with & in the condition.


dthain:
I approve the general idea.

PR #1163, opened 2016-02-12
Clarify WQ Build Instructions

dthain:
Use `$HOME/cctools` in the Work Queue manual, since that's where beginners are going to install.  (And nobody told them to set `$CCTOOLS`.)


btovar:
Looks good. May merge to include in 5.4?

dthain:
Go for it.

Issue #1171, opened 2016-02-26
[umbrella] +x flag is not set when using "plain" mode in software section


Baguage:
I'm trying to write Umbrella spec for a simple math application, and I ran into a problem in "plain" mode.

Here is my spec:
[bertini.umbrella.txt](https://github.com/cooperative-computing-lab/cctools/files/149039/bertini.umbrella.txt)

When I'm trying to run it, I'm getting

> /bin/sh: /software/bertini: Permission denied

It happens because /software/bertini file does not have +x attribute. The only solution currently is to make file executable in Umbrella cache manually.

I'm sending a pull request with the fix for this problem. I tested it on this example, seems to work.

hmeng-19:
@Baguage , I think the real reason is that the permission of the original source - `/tmp/tmp/BertiniLinux64_v1.5/input` is not set correctly.
Change the permissions of all the mountpoints to be `0755` may introduce some security problems. 
Why not try to change the permission of `/tmp/tmp/BertiniLinux64_v1.5/input` before starting an Umbrella job? Let me know if this works?

Baguage:
Oops, sounds like I missed you response, @hmeng-19 
This happens when I use "plain" mode in software section, so there is no .tar.gz package with permissions. 

I will review my pull request again, but I believe it only sets permissions to 755 for software mountpoints and only in "plain" mode. Without doing that, "plain" mode is effectively broken - executable file will be downloaded, but won't run.

hmeng-19:
@Baguage , in "plain" mode, umbrella copies the file into its own cache and does not really mess up with its access permission. I can not really can reproduce this problem.

I would suggest create a `.tar.gz` file for each software dependency, and put binary into its `bin` subdir, and libraries (if any) into its `lib` subdir.

hmeng-19:
@Baguage , #1579 should fix the problem here. Let me know if it works for you.

hmeng-19:
@Baguage , #1579 has been merged into the master. Please test the master to see whether it works for you.

PR #1172, opened 2016-02-26
Set +x attribute for files in software section

Baguage:
Proposing fix for #1171

### DIFF
--- a/umbrella/src/umbrella.py
+++ b/umbrella/src/umbrella.py
@@ -1009,6 +1009,7 @@ def software_install(mount_dict, env_para_dict, software_spec, meta_json, sandbo
 							subprocess_error(cmd, rc, stdout, stderr)
 				else:
 					mount_dict[mountpoint] = mount_value
+					os.chmod(mount_value, 0755)
 
 
 def data_install(data_spec, meta_json, sandbox_dir, mount_dict, env_para_dict, osf_auth, name=None):


dthain:
@hmeng-19 please let us know if this PR makes sense.

hmeng-19:
@dthain , I am waiting for the feedback from Alex about my suggestion about #1171 .

hmeng-19:
BTW, I do not think it should be Umbrella's duty to set any file permission. The file permission should be preserved when a user archives a package. Umbrella tries to maintain the file permission as much as possible.

dthain:
@hmeng this pull request has been outstanding for a while now.  Please decide whether we should accept this in the current form, or solve the problem in a different way.

hmeng-19:
@Baguage , I have trouble to reproduce the problem you have. You can show me the problem when I am back. Then we can review this PR. Sounds good?

hmeng-19:
@dthain , I would leave the PR for now. I am also okay if you want to close it for now.

dthain:
Let me propose an alternate approach to this:

I think the underlying issue here is that a remote storage system will not necessary record the desired permissions bits associated with a file.  If Umbrella downloads a given file, it will be given mode bits according to the current umask, which isn't necessarily what the user wants.  And, while some items from software should be marked as executable, that won't always be the case.

So, let me suggest an alternate:

Umbrella should accept a `mode` field wherever a file can be specified.  If the permissions are important for a particular file, then the user can do something like this:

```
source: "http://....",
checksum: abc123...,
mode: 0755
```
Umbrella can easily look for that an set it, and then user only needs to give it if it is significant.

What do you think?

hmeng-19:
@dthain , a software dependency pointed by the `source` field is usually a directory, not a single file.

Baguage:
@dthain - that sounds like a good idea. As @hmeng-19 pointed out "mode" is only applicable for "plain" format.
If we are on the same page with the solution, I can update my pull request with the support for "mode" option.

hmeng-19:
@Baguage , sounds good. Go ahead.

dthain:
This pull request has been open for many months now.
Is it still in progress, or should I close it?

hmeng-19:
@dthain , you can close this one for now.
I will work on a new PR today to implement your suggestion: adding a new field, `mode` to specify the file mode.

hmeng-19:
@dthain, @Baguage  , please see #1579 .

PR #1173, opened 2016-02-29
Add in-Parrot file copy via BTRFS_IOC_CLONE

trshaffer:
This pull request adds handling of the `BTRFS_IOC_CLONE` ioctl to trigger an in-Parrot copy via `copyfile_slow`. With a recent version of coreutils, this functionality is available through `cp --reflink`. This feature masks the *real* BTRFS_IOC_CLONE, which might not be an issue. Also, I wasn't sure about the desired semantics for a clone operation. At the moment, this does not seek to the beginnings of files before copying.

### DIFF
--- a/parrot/src/syscall_parrot.tbl
+++ b/parrot/src/syscall_parrot.tbl
@@ -15,4 +15,3 @@
 1010 parrot parrot_debug sys_parrot_debug
 1011 parrot parrot_mount sys_parrot_mount
 1012 parrot parrot_unmount sys_parrot_unmount
-1013 parrot parrot_fcopyfile sys_parrot_fcopyfile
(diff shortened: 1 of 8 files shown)


dthain:
Oh darn, the user has to ask for this feature explicitly on the command line?
I was under the impression it was a silent optimization....

parrot/src/syscall_parrot.tbl
```diff
@@ -14,4 +14,5 @@
 1009 parrot parrot_search sys_parrot_search
 1010 parrot parrot_debug sys_parrot_debug
 1011 parrot parrot_mount sys_parrot_mount
-1012 parrot parrot_unmount sys_parrot_unmount
\ No newline at end of file
+1012 parrot parrot_unmount sys_parrot_unmount
+1013 parrot parrot_fcopyfile sys_parrot_fcopyfile
```
batrick: Ya @trshaffer, this 1013 syscall isn't necessary. [If you wanted to add a new syscall, you need to implement it in `pfs_dispatch*.cc`.] Go ahead and remove it.

trshaffer:
I really can't imagine a purpose for `fcopyfile` besides completeness, so yeah, I'll change the table back. Also, while reading the coreutils changelog to update Parrot's documentation with the reflink copy feature, I noticed that very recent versions of `mv` will try a reflink copy without user intervention when moving across mount points, so we get that for free

PR #1174, opened 2016-03-01
Update Makeflow docs with wrapper substitution rules

trshaffer:
I updated the section about wrappers to match the current rules on command substitution. Forgot to put this in the original PR


batrick:
Looks good. Please also add the rationale for `[]` and the default.

trshaffer:
Done

PR #1177, opened 2016-03-04
Chirp-HDFS rmdir should not be recursive

### LINKED ISSUE #1132: Chirp's hdfs driver does recursive unlink for `rmdir`
Here:

https://github.com/cooperative-computing-lab/cctools/blob/c44b4b9e7b299a23313c48e103fd2fa9263d7e69/chirp/src/chirp_fs_hdfs.c#L567

This should be a non-recursive unlink to be consistent with the other file systems.

dthain:
Fixes #1132
@batrick please double check


batrick:
Worked for me.

Issue #1073, opened 2015-12-14
Project: Add Expressions to JX Library


dthain:
Extend the JX library to permit simple expressions that can be evaluated within the context of another JSON document.  This will make it handy to filter and evaluate JSON data wherever we use it, so that things like this become easy:

```
chirp_status --where '(avail/total)>.9`
```

Here is the idea: The function `jx_evaluate` should take an arbitrary JSON expression, possibly containing symbols and operators, look them up in the JSON object `context`, and then return a new expression with everything resolved, like this:

```
struct jx * jx_evaluate( struct jx * expr, struct jx *context );
```

For example, suppose that `context` is this:

```
`{ "name":"honda", "mass": 1800, "speed": 65.7 }`
```

Then `name` would evaluate to `"honda"`, `speed>10` would evaluate to `true`, and `{ "energy": mass*speed*speed/2.0}` would evaluate to `{"energy":3884760}`

Here is a plan to go about it:

1 - Add new `JX_EXPR` node type to `jx.[ch]` and enumerate the various operators, so that you can create expressions manually like this:

```
j = jx_expr( jx_integer(10), JX_MULTIPLY, jx_symbol(speed) );
```

2 - Modify `jx_print.[ch]` to accommodate printing of expressions.
3 - Modify `jx_parse.c` to parse expressions using a simple recursive descent parser.
4 - Modify `jx_evaluate.c` to evaluate expressions down to constant values.

Once that is all working, then you can easily modify `chirp_status` to take an optional command-line argument, and display only the objects that match.

dthain:
And the follow up to this project is that the same expressions can then be used to generate arbitrary queries in DeltaDB.

dthain:
Almost done here.  They last major piece missing are we need to be able to evaluate some functions, particularly:

- `range(a,b)` returns `[a, a+1, ... b]`
- `regex(string,regex)` returns true if string matches regex
- `foreach(x,array,expr)` evaluates expr for each value of x in array.

Once we can do that, then we can represent a range of jobs with syntax like this:

```
foreach(x,range(1,10),
{
   cmd: "sim.exe -p "+(x*2)+" -o output."+x,
   output: "output."+x
   . . .
}
)
```

dthain:
@trshaffer What's the current status of JX expressions in Makeflow?

trshaffer:
[...]

PR #1180, opened 2016-03-07
JX Expressions

dthain:
First pass at adding expressions to the JX library.  Supports equality expressions == != <= < > >= and arithmetic expressions + * - / % where + is concatenation on strings.

Still not quite ready:
- Needs a test suite
- Needs to support unary +/- which were previously treated as part of the integer/double tokens.
- Needs to support boolean operations for deltadb.


dthain:
Fixes #1073

dthain:
@pivie please take a few minutes to check out this branch and try out some deltadb queries.  See if you find anything surprising.

pivie:
I got it working and used it for a few queries. 

This worked:   --where "cores_inuse>1500"
But with this, the second part seemed to be ignored:   --where "cores_inuse>1000 && cores_inuse<1500"
While this resulted in nothing being returned:   --where "(cores_inuse>500)&&(cores_inuse<1500)"

Given that we are working with NoSQL data, it is not surprising, but perhaps unsettling, that using non-existent (misspelled) fields are perfectly fine. I think I had to look at the database files to figure out which field names to use. I couldn't think of a way to show all the fields names using deltadb_query.

dthain:
@pivie can you give me the exact command lines showing this problem?

I am able to get the expected output with these lines:
```
./deltadb_query --db /afs/crc.nd.edu/group/ccl/data/catalog.history --where 'cores_inuse>1500' --output name --output cores_inuse --from "2016-01-01"
./deltadb_query --db /afs/crc.nd.edu/group/ccl/data/catalog.history --where 'cores_inuse>1500 && cores_inuse<6000' --output name --output cores_inuse --from "2016-01-01"
```

You are right about the problem of finding the appropriate properties.  With deltadb, you can see the raw json output if you don't give any --output options, so that may help.  A future project may be to track which properties have no successful object lookups, and report those to the user...

pivie:
I don't have access to that directory in afs, so I can't run your command
directly, I use a copy of the data. Try the following:

deltadb_query --db /afs/crc.nd.edu/group/ccl/data/catalog.history --from
2016-03-01 --filter name=daspos2.crc.nd.edu --filter type=wq_master --epoch
--output project --output 'tasks_complete' --output workers_busy --output
cores_inuse --output total_execute_time --output total_bytes_received
--output total_bytes_sent --every 1d --where "cores_inuse<1500" | grep
census

deltadb_query --db /afs/crc.nd.edu/group/ccl/data/catalog.history --from
2016-03-01 --filter name=daspos2.crc.nd.edu --filter type=wq_master --epoch
--output project --output 'tasks_complete' --output workers_busy --output
cores_inuse --output total_execute_time --output total_bytes_received
--output total_bytes_sent --every 1d --where "cores_inuse<1000" | grep
census

deltadb_query --db /afs/crc.nd.edu/group/ccl/data/catalog.history --from
2016-03-01 --filter name=daspos2.crc.nd.edu --filter type=wq_master --epoch
--output project --output 'tasks_complete' --output workers_busy --output
cores_inuse --output total_execute_time --output total_bytes_received
--output total_bytes_sent --every 1d --where "cores_inuse>500 &&
cores_inuse<1000" | grep census

I would expect the last one to only have the intersection of the results of
the previous ones, but I get results where cores_inuse is above 1000.
Perhaps I am getting confused about when the "where" is applied. Maybe I am
trying to use where like I would use a "having" argument.

dthain:
As an aside, you should be able to access the data in the crc afs cell, but you need to authenticate there first:
```
kinit pivie@CRC.ND.EDU
aklog -cell crc.nd.edu
```

The filter clause simply drops records from the input stream.  (A static query.)  The where clause matches and emits records as they are constructed in the internal hash table.  (A dynamic query.)

And I think I see the problem now -- you can only have one filter and one where clause.  (Just take the two requirements and join them with an AND expression.)  And a test for equality should have two equals signs.

I need to add a check to make sure that the options aren't repeated...

dthain:
Ok, try the latest one.

dthain:
Does it work for you now?

pivie:
Yes:
At first I tried --filter "name=='daspos2.crc.nd.edu' &&
type=='wq_master'", but once I saw the example in the man pages, I was able
to get the following query to work properly.

deltadb_query --db /data/pivie/catalog_history --from 2016-03-01 --filter
'name=="daspos2.crc.nd.edu" && type=="wq_master"' --epoch --output project
--output 'tasks_complete' --output workers_busy --output cores_inuse
--output total_execute_time --output total_bytes_received --output
total_bytes_sent --every 1d --where "cores_inuse>500 && cores_inuse<1500" |
grep census

As I thought about my earlier question regarding 'where' vs 'having', I
realized that the --filter option in deltadb is like 'where' in SQL and the
--where option in deltadb is like the 'having' option in SQL. With that in
mind it makes more sense.

dthain:
Ok, bombs away.
Fixes #1073

PR #1186, opened 2016-03-08
resolve #1165

### LINKED ISSUE #1165: Treat CORES as an alias to OMP_NUM_THREADS
@akahs  found the following error while running makeflow -Tlocal:
 
libgomp: Thread creation failed: Resource temporarily unavailable
Thread creation failed: Resource temporarily unavailable

The problem is that both makeflow and the task executable (/usr/bin/convert) try to use all the cores in the machine. makeflow sets CORES=1, but convert respects OMP_NUM_THREADS. For n cores, nxn processes are created and this exhausted the maximum number of concurrent processes allowed.

This is a really nice example on how different applications respect/ignore resources limits. Note that  adding OMP_NUM_THREADS would not solve the problem in general, as the executable may ignore its value, or use some other framework (e.g., MPI).

charleszheng44:
Solve issue #1165, set OMP_NUM_THREADS = 1 and export it as an environment variable.

### DIFF
--- a/makeflow/src/dag_node.c
+++ b/makeflow/src/dag_node.c
@@ -306,13 +306,33 @@ struct jx * dag_node_env_create( struct dag *d, struct dag_node *n )
 
 	struct jx *object = jx_object(0);
 
+    char *num_cores = dag_variable_lookup_string(RESOURCES_CORES, &s);
+	char *num_omp_threads = dag_variable_lookup_string("OMP_NUM_THREADS", &s);
+
+	if (num_cores && !num_omp_threads) {
+		// if number of cores is set, number of omp threads is not set,
+		// then we set number of omp threads to number of cores
+		jx_insert(object, jx_string("OMP_NUM_THREADS"), jx_string(num_cores));
+	} else if (num_omp_threads) {
+		// if number of omp threads is set, then we set number of cores
+		// to the number of omp threads
+		jx_insert(object, jx_string(RESOURCES_CORES), jx_string(num_omp_threads));
+	} else {
+		// if both number of cores and omp threads are not set, we
+		// set them to 1
+		jx_insert(object, jx_string("OMP_NUM_THREADS"), jx_string("1"));
+		jx_insert(object, jx_string(RESOURCES_CORES), jx_string("1"));
+	}
+
 	set_first_element(d->export_vars);
 	while((key = set_next_element(d->export_vars))) {
+
 		char *value = dag_variable_lookup_string(key, &s);
 		if(value) {
 			jx_insert(object,jx_string(key),jx_string(value));
 			debug(D_MAKEFLOW_RUN, "export %s=%s", key, value);
 		}
+
 	}
 
 	return object;
--- a/makeflow/src/parser.c
+++ b/makeflow/src/parser.c
@@ -149,6 +149,10 @@ void dag_close_over_environment(struct dag *d)
 			{
 				dag_variable_add_value(name, d->default_category->mf_variables, 0, value_env);
 			}
+
+			if(!value_env && !strcmp(name, RESOURCES_CORES)) {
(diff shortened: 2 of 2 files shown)


dthain:
I think the point is that OMP NUM CORES should have the same value as CORES.

charleszheng44:
Does that mean `OMP_NUM_CORES` should be set as 0 when CORES is not set?

btovar:
No, see my comment above.

charleszheng44:
I have updated the code.

makeflow/src/dag_node.c
```diff
@@ -346,8 +346,15 @@ struct jx * dag_node_env_create( struct dag *d, struct dag_node *n )
 	while((key = set_next_element(d->export_vars))) {
 		char *value = dag_variable_lookup_string(key, &s);
 		if(value) {
+			if(strcmp(key, "CORES")) {
+				jx_insert(object, jx_string("OMP_NUM_THREADS"), jx_string(value));
+			}
 			jx_insert(object,jx_string(key),jx_string(value));
 			debug(D_MAKEFLOW_RUN, "export %s=%s", key, value);
+		} else {
+			if(strcmp(key, "CORES")) {
```
btovar: How is cores defined if it does not have a value?

makeflow/src/dag_node.c
```diff
@@ -346,8 +346,15 @@ struct jx * dag_node_env_create( struct dag *d, struct dag_node *n )
 	while((key = set_next_element(d->export_vars))) {
 		char *value = dag_variable_lookup_string(key, &s);
 		if(value) {
+			if(strcmp(key, "CORES")) {
```
btovar: strcmp returns 0 if the comparison succeeds. Here you are setting OMP_NUM_THREADS to the value of each other variable, but CORES.

makeflow/src/parser.c
```diff
@@ -149,6 +149,10 @@ void dag_close_over_environment(struct dag *d)
 			{
 				dag_variable_add_value(name, d->default_category->mf_variables, 0, value_env);
 			}
+
+			if(!value_env && !strcmp(name, RESOURCES_CORES)) {
+				dag_variable_add_value(name, d->default_category->mf_variables, 0, "1");
```
dthain: I don't think you should be changing the dag unilaterally at parse time.  If cores isn't specified, then it isn't specified.

makeflow/src/dag_node.c
```diff
@@ -341,14 +341,19 @@ struct jx * dag_node_env_create( struct dag *d, struct dag_node *n )
 	char *key;
 
 	struct jx *object = jx_object(0);
+    
+    char *num_cores = dag_variable_lookup_string(RESOURCES_CORES, &s);
```
dthain: Generally speaking, check for null anytime you have a pointer returned from a function.
If it's null, then `CORES` isn't set.

dthain:
@charleszheng44 please finish up this PR, talk to @btovar if you aren't sure how to go forward.

charleszheng44:
1. If number of cores is not set
    + If `OMP_NUM_CORES` is not specified, we set number of cores to 1
    + else, we set number of cores to `OMP_NUM_CORES`

2. if number of cores is set
    + If `OMP_NUM_CORES` is not specified, we set `OMP_NUM_CORES` to the number of cores
    + If `OMP_NUM_CORES` is set, we do nothing

dthain:
Wow, what happened to this PR?

charleszheng44:
I rebase the latest master and push the change.

dthain:
@batrick Please take a look at this PR and figure out if it safe to merge.  Discuss with @charleszheng44 as needed.

batrick:
I helped him collapse his changes into a single commit based on master. If
should merge cleanly now. Charles is checking if it is still does what it's
supposed to do.

Issue #300, opened 2013-11-07
allpairs_multicore with multi-slot workers


dpandiar:
All-Pairs uses allpairs_multicore as the kernel for its tasks. allpairs_multicore has an option to specify the number of cores to use. However, this doesn't interact with the facility we have for running multi-core jobs in WQ.

dthain:
@trshaffer please take a crack at this one.

Modify allpairs_multicore to respect the CORES and MEMORY environmental variables that are set by Makeflow and WQ to tell each job what to do.

nhazekam:
I think this may have been addressed by @trshaffer 's pull request.

PR #1189, opened 2016-03-09
Check CORES and MEMORY when starting allpairs_multicore

trshaffer:
I *think* the units are correct. `allpairs_multicore` appears to use bytes for calculating blocksize, while the environment variable seems to be in megabytes.

### DIFF
--- a/dttools/src/rmonitor_poll.c
+++ b/dttools/src/rmonitor_poll.c
@@ -829,7 +829,7 @@ struct rmsummary *rmonitor_measure_host(char *path) {
 
 	if(path) {
 		path_disk_size_info_get(path, &total_disk, &file_count);
-		tr->disk = total_disk;
+		tr->disk = total_disk / MEGABYTE;
 		tr->total_files = file_count;
 	}
 	host_memory_info_get(&free_mem, &total_mem);
--- a/dttools/src/rmonitor_poll.c
+++ b/dttools/src/rmonitor_poll.c
@@ -20,6 +20,7 @@ See the file COPYING for details.
 
 #include "rmonitor_poll_internal.h"
 #include "host_memory_info.h"
+#include "path_disk_size_info.h"
 #include "load_average.h"
 
 /***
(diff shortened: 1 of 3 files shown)


trshaffer:
This is for #300

dthain:
`block_size` is not bytes, but simply the number of elements in the array.

trshaffer:
I should have been more clear, inside `block_size_estimate` it falls back to half the system's memory, measured in bytes. I'm not sure if there are any memory-intensive tests for allpairs, but I did test the CPU limit.

dthain:
This is the right approach, but let's factor out the examination of the environment and the querying of hardware properties into a module that returns everything in a resource management structure.

(Since Makelfow and other tools need to do the same thing.)

@btovar can you point @trshaffer in the right direction?

btovar:
@trshaffer:  In dttools/src/rmonitor_poll.[hc] we would add a function such as:

struct rmsummary *rmonitor_measure_host();

The returned rmsummary would have the values of cores, memory, disk, etc. The function rmonitor_measure_process should give you a good idea on how struct rmsummary is created, accessed, etc. Available fields are in dttools/src/rmsummary.h (memory and disk sizes in MB, times in usecs).

trshaffer:
Is this more what you had in mind? Right now `rmonitor_measure_host` only fills in the fields used by allpairs, but that's easy to expand.

btovar:
Nice. It looks good to me.

dthain:
Add the measurement of disk (for a specified directory) into `rmonitor_measure_host` b/c we are going to use that function in makeflow, workqueue, and everywhere else to infer the current operating conditions.

trshaffer:
I *think* that's the correct measurement for this case

dttools/src/rmonitor_poll.c
```diff
@@ -819,10 +820,18 @@ int rmonitor_measure_process_update_to_peak(struct rmsummary *tr, pid_t pid) {
...
+struct rmsummary *rmonitor_measure_host(char *path) {
+	uint64_t free_mem;
+	uint64_t total_mem;
+	int64_t total_disk;
+	int64_t file_count;
 	struct rmsummary *tr = rmsummary_create(-1);
 
+	if(path) {
+		path_disk_size_info_get(path, &total_disk, &file_count);
+		tr->disk = total_disk;
```
dthain: Do the units match here?

trshaffer:
Ah, my mistake. I was reading the wrong line

PR #1190, opened 2016-03-09
Split hostport on last : in chirp_client_connect

trshaffer:
This looks like it should address the case in the issue. Is an IP6 address without port allowed? This change would mangle that, and split the address itself

### DIFF
--- a/chirp/src/chirp_client.c
+++ b/chirp/src/chirp_client.c
@@ -70,6 +70,9 @@ See the file COPYING for details.
 /* The maximum chunk of memory the server will allocate to handle I/O */
 #define MAX_BUFFER_SIZE (16*1024*1024)
 
+/* Valid characters in an IPV6 address */
+#define IPV6_CHARS "0123456789abcdefABCDEF:"
+
 static int global_serial = 0;
 
 struct chirp_client {
@@ -349,7 +352,12 @@ struct chirp_client *chirp_client_connect(const char *hostport, int negotiate_au
 	int save_errno;
 	int port;
 
-	if(sscanf(hostport, "%[^:]:%d", host, &port) == 2) {
+	if(sscanf(hostport, "[%[" IPV6_CHARS "]]:%d", host, &port) == 2) {
+		/* IPV6 address with port */
+	} else if(sscanf(hostport, "[%[" IPV6_CHARS "]]", host) == 1) {
+		/* IPV6 address without port */
+		port = CHIRP_PORT;
+	} else if(sscanf(hostport, "%[^:]:%d", host, &port) == 2) {
 		/* use the split host and port */
 	} else {
 		strcpy(host, hostport);
--- a/chirp/src/chirp_client.c
+++ b/chirp/src/chirp_client.c
@@ -348,13 +348,9 @@ struct chirp_client *chirp_client_connect(const char *hostport, int negotiate_au
 	char host[DOMAIN_NAME_MAX];
 	int save_errno;
 	int port;
-	char *separator;
 
-	if((separator = strrchr(hostport, ':')) != NULL) {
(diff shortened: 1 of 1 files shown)


trshaffer:
This is for #591

dthain:
Arg, you are quite right, this isn't going to work if the user puts in an IPV6 address without a port.

The clearest solution would be to simply make host and port different arguments in the API.

Unfortunately, we mash together the host and port in lots of user-visible strings, for example:
```
parrot_run ls -1 /chirp
. . .
student00.cse.nd.edu:9094
student01.cse.nd.edu:9094
student02.cse.nd.edu:9094
student03.cse.nd.edu:9094
wang029.crc.nd.edu:9094
```

batrick:
I mentioned this in

https://github.com/cooperative-computing-lab/cctools/issues/591#issuecomment-61867746

The common thing to do now `[ipv6]:port` but that is not completely adopted by everyone. (It also has potential downsides as I noted in the above link.) And, IIRC, there are no new functions (like `getaddrinfo`/`getnameinfo`) which manage the parsing for you.

I would suggest adopting something completely unambiguous and rejecting anything else.

dthain:
Reading it again, your comment makes a lot of sense.
Perhaps I wasn't ready to understand it before now.  :)

So, it seems we need to handle three cases:

```
domainname
domainname:port
ipv4
ipv4:port
ipv6
ipv6:port
```

Can we distinguish those clearly by simply counting the total number of colons used?

dthain:
Six cases

batrick:
> Can we distinguish those clearly by simply counting the total number of colons used?

Unfortunately no since an ipv6 address can have 2-7 colons.

batrick:
See also this amusing rant that was released recently:

https://www.zerotier.com/blog/?p=724

dthain:
Terrifying.

How about @batrick's suggestion of [ipv6]:port ?
Easy to do that with sscanf, I think.

dthain:
That makes sense to me.  What do you think, @batrick?

batrick:
I think we got lost solving the formatting issue but forgot the original problem. In #591, Todd was seeing if we would eventually allow ipv6 addresses in Condor's chirp.config and how they would be formatted. This concerns this code:

https://github.com/trshaffer/cctools/blob/ada3cf4b51cb10336f19d98df12fdd86549d0be1/chirp/src/chirp_client.c#L322

We may have decided on the acceptable formats for the `hostport` but that doesn't address `chirp.config`. At the very least, the code above would need to format the `host` appropriately before calling `chirp_client_connect`. That is certainly more frustrating as we need to detect an ipv6 address and put square brackets around it.

I just checked what I did in my ipv6 branch to see how I resolved the problem:

https://github.com/batrick/cctools/blob/1067cae7d5abc01a1d2b1f1029f7658ec5054ceb/dttools/src/link.c#L551-L562

An ipv6 address must have a port unless it has square brackets around it. This part of the comment:

````
	 *  - "[" IPv6address "]" [":" port]
	 *  - IPv6address ":" port  // port is *required*
````

in simpler terms, it matches:

[ipv6]
[ipv6]:port
ipv6:port

That resolved the ambiguity and it works automatically for Condor's `chirp.config` without modification.

Anyway, I think deciding on a format is premature as this change does not make CCTools  compatible with ipv6 as Chirp still cannot connect to an ipv6 address. All we've done is "recognize" such an address and set a format which we may regret later.

I think we can suggest to Todd that it's okay to set an ipv6 address in the host field and without any special formatting since `chirp.config` is whitespace delimited. Our code won't work with such a cookie but that cannot be fixed until we also support ipv6.

batrick:
@trshaffer we've decided to close this due to the reasons I outlined in my last comment. We will revisit this issue in the future IPv6 compatibility work.

Issue #946, opened 2015-10-05
Project: Archival Form of Makeflow


dthain:
(Credit for this idea goes to Sandra Gesing.)

For the purposes of long-term archival, it may be useful to reduce a workflow to a single flattened shell script.  Add an option to makeflow that converts an input makeflow into a single output script, by adding a new visitor `dag_to_sh` to `makeflow/src/dag_visitors.c`.  See `dag_to_dot` and `dag_to_dax` for inspiration.

This feature could be used in combination with `makeflow_linker` to produce a completely self-contained package.

dthain:
@sandragesing would you like to take a crack at this one?

sandragesing:
Yes - I will have a look at it.

dthain:
@trshaffer take a crack at this one too, it will be an easy proof of concept.

sandragesing:
Hi Doug,

sorry that I did not have a look yet because of a lack of time but great
that somebody else takes a crack at it! I would be happy to see the outcome.

Thanks,
Sandra

------------------------------------------------------------------------------------------------------------
Sandra Gesing
Research Assistant Professor, Department of Computer Science and Engineering
Computational Scientist, Center for Research Computing
University of Notre Dame
http://www3.nd.edu/~sgesing
------------------------------------------------------------------------------------------------------------

PR #1191, opened 2016-03-09
Add dryrun mode + flat shell script output to Makeflow

trshaffer:
This is a first draft for #946. I added a new batch system, `dryrun`. If Makeflow is invoked as `makeflow -T dryrun`, it performs all validation, dependency checking, but instead of actually running jobs, it prints the command it would have used. Since issuing commands to the batch system is one of the last things Makeflow does, this will get wrappers or any other transformations that might be added.

In addition, this batch system's log is actually a flattened shell script that should generate the same results as running the makefile, albeit sequentially. Since it doesn't depend on Makeflow, the shell script from e.g. `makeflow -T dryrun -L archive.sh` can serve as an archival representation as in the issue. I'd like to try this out on a larger, more complex sample makefile, since I've just been testing simple things.

### DIFF
--- a/batch_job/src/Makefile
+++ b/batch_job/src/Makefile
@@ -10,6 +10,7 @@ EXTERNAL_LIBRARIES = ../../work_queue/src/libwork_queue.a ../../chirp/src/libchi
 SOURCES = \
 	batch_job.c \
 	batch_job_amazon.c \
+	batch_job_dryrun.c \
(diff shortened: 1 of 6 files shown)


dthain:
Add a short note to the man page, pleas.e

trshaffer:
Added

dthain:
Closes #946

PR #1192, opened 2016-03-09
Adds LD_PRELOAD library for write and open calls

nkremerh:
@nhazekam @dthain please review.


dttools/src/Makefile
```diff
@@ -120,7 +121,7 @@ HEADERS_PUBLIC = \
 	rmsummary.h \
 	timestamp.h
 
-LIBRARIES = libdttools.a
+LIBRARIES = libdttools.a lib_io_helper.so
```
btovar: The _ before lib is not customary. I suggest libio_helper.so. (In particular for linking, the name of the library would be _io_helper.)

dttools/src/Makefile
```diff
@@ -139,6 +140,9 @@ auth_globus.o: auth_globus.c
 auth_test: auth_test.o libdttools.a
 	$(CCTOOLS_LD) -o $@ $(CCTOOLS_INTERNAL_LDFLAGS) $(LOCAL_LDFLAGS) $^ $(LOCAL_LINKAGE) $(CCTOOLS_GLOBUS_LDFLAGS) $(CCTOOLS_EXTERNAL_LINKAGE)
 
+lib_io_helper.so: lib_io_helper.o
+	$(CCTOOLS_CC) -shared -ldl -fPIC $^ -o $@
```
btovar: For some reason, ubuntu does not like -ldl there. You want:

$(CCTOOLS_CC) -shared -ldl -fPIC $^ -o $@ -ldl

dthain:
Let's call it something that more directly explains what it does, like `libforce_halt_on_out_of_space`.
Perhaps you can think of something a little shorter.  :)

dttools/src/lib_io_helper.c
```diff
@@ -0,0 +1,69 @@
...
+
+	__typeof__(write) *original_write = dlsym(RTLD_NEXT, "write");
+
+	int real_count;
+	int prev_errno = errno;
+	errno = 0;
+	real_count = original_write(fd, buf, count);
+
+	if(real_count < 0 && errno == ENOSPC) {
+		original_write(2, "WRITE ERROR: device capacity reached.\n", 39);
```
btovar: Use STDERR_FILENO instead of 2.

btovar: Also, take a look at how we print messages in the debug file. Having the time of the error may prove advantageous for debugging.

dttools/src/lib_io_helper.c
```diff
@@ -0,0 +1,69 @@
...
+		return -1;
```
btovar: return real_count, rather than -1. (Here it can only be -1, but just in case somebody copy-pastes these code snippets for some other functions.)

dttools/src/lib_io_helper.c
```diff
@@ -0,0 +1,69 @@
...
+	int mode;
+	int prev_errno = errno;
+
+	va_start(ap, flags);
+	mode = va_arg(ap, int);
+	va_end(ap);
+
+	fd = original_open(path, flags, mode);
+
+	//struct statvfs free_inodes;
```
btovar: These comments should be removed.

work_queue/src/work_queue_worker.c
```diff
@@ -881,6 +882,15 @@ static int do_task( struct link *master, int taskid, time_t stoptime )
 
 	struct work_queue_process *p = work_queue_process_create(task, disk_alloc);
 
+	if(disk_alloc) {
+		char *environment_path = getenv("CCTOOLS_LIB_IO_HELPER");
+		char *preload_path = string_format("LD_PRELOAD=%s", environment_path);
```
btovar: if environment_path is NULL, doesn't this segfault?

work_queue/src/work_queue_worker.c
```diff
@@ -2166,8 +2181,13 @@ int main(int argc, char *argv[])
 			tar_fn = xxstrdup(optarg);
 			break;
 		case LONG_OPT_DISK_ALLOCATION:
+		{
+			char resolved_path[PATH_MAX];
+			char *abs_path_preloader = realpath(optarg, resolved_path);
```
btovar: On error, realpath returns NULL, which may cause a segfault below.

dttools/src/lib_io_helper.c
nkremerh: If real_count < count and real_count > 0, that is considered a successful write. It is only the successive write which is considered unsuccessful, and real_count will be set to -1 along with the relevant errno.

nkremerh:
@btovar @dthain Ready for a second pass.

dttools/src/Makefile
```diff
@@ -139,6 +140,9 @@ auth_globus.o: auth_globus.c
 auth_test: auth_test.o libdttools.a
 	$(CCTOOLS_LD) -o $@ $(CCTOOLS_INTERNAL_LDFLAGS) $(LOCAL_LDFLAGS) $^ $(LOCAL_LINKAGE) $(CCTOOLS_GLOBUS_LDFLAGS) $(CCTOOLS_EXTERNAL_LINKAGE)
 
+libforce_halt_enospc.so: libforce_halt_enospc.o
+	$(CCTOOLS_CC) -shared -ldl -fPIC $^ -o $@
```
btovar: -ldl should be at the end of the command, like:

$(CCTOOLS_CC) -shared -fPIC $^ -o $@ -ldl

Otherwise ubuntu complains. Also, a kitty cries every time I have to rewrite a comment because the first time it was ignored. Why do you want kitties to cry?

work_queue/src/work_queue_worker.c
```diff
@@ -2166,8 +2172,29 @@ int main(int argc, char *argv[])
...
+				preload_result = setenv("LD_PRELOAD", new_ld_preload, 1);
+				free(new_ld_preload);
+			}
+			else if(abs_path_preloader != NULL) {
+				preload_result = setenv("LD_PRELOAD", abs_path_preloader, 1);
+			}
+			else {
+				preload_result = 1;
+			}
+			if(preload_result != 0) {
```
btovar: We usually write: if(!preload_result) ...

dttools/src/Makefile
nkremerh: Sorry, I changed it before and undid the change due to other issues with the Makefile in between this last bit of commits. Will redo now!

btovar:
Looks good, nice job Nate.

dthain:
Looks good overall.

Would it make sense to assume that the preload library is installed in  CCTOOLS_INSTALL_DIR "/lib/libwhatever.so" ?  That would save the user some trouble in specifying it.  (I believe we do something similar with libparrot_helper.so)

btovar:
That makes sense.

Also, we need to handle the monitor and the library working together. The library most likely will kill the monitor before a clean shutdown can be done. This is probably ok, but we need to ensure exit status makes sense.

nkremerh:
@btovar @dthain Ready for third pass, hopefully ready to merge.

dthain:
Looks reasonable to me, but I'll let @btovar have the final say.

btovar:
Nate, looks good. 

Is it ready to be merged?

nkremerh:
Yes.

Issue #1195, opened 2016-03-10
Extremely long line


trshaffer:
I was wondering why my text editor was getting laggy in `dag_visitors.c`. As it happens, one of the lines is 248879 characters long. It appears to be a large XML file with newlines stripped out. I don't think a single `fprintf` statement is the correct place for this kind of data.

https://github.com/cooperative-computing-lab/cctools/blame/8196aab6913de56eb155de4315ed063f0a2ff2d8/makeflow/src/dag_visitors.c#L487

btovar:
Since this mode is not used very often, we can move it to a file somewhere in INSTALL_PATH, and read it when needed. Alternatively, perhaps there is a cytoscape style that is not 200kB.

PR #1200, opened 2016-03-14
Move 200K XML file out of source code

trshaffer:
The `style.xml` file is created with a constant name and completely static content, so I removed it from the source code and committed the style file separately. Users that want this style for Cytoscape visualization can copy it to the working directory when they run `makeflow_viz`

See #1195


dthain:
This is a necessary move, except the file will not be available at run time unless it is explicitly installed.

Suggest that you modify the install rule to put it into `$(cctools)/share/cctools/makeflow-cytoscape-style.xml` and then adjust makeflow to copy it from that location when needed.

trshaffer:
OK, it should be good. `makeflow_viz` will print a warning if it can't find the style file.

PR #1204, opened 2016-03-16
Add positional parameters to starch run.sh

hmeng-19:
Currently, `starch` assumes that the whole command is provided by its `-c` option.
	This would fail if the provided `-c` option is incomplete or the user wants to
	add more parameters to the command during runtime.

	The change here allows the user to provide more parameters during runtime.


dthain:
Don't forget the starch man page.

hmeng-19:
Please give it a new look.

PR #1206, opened 2016-03-17
Catch and abort on CLONE_UNTRACED.

batrick:
We should not let new threads/processes escape Parrot through this flag to clone.

I don't think this is related to the recently reported Parrot crash for Rodney et al.


dthain:
Seems odd to have this feature in the kernel at all.
Is there a use case that you are aware of?
Would it make sense to have parrot simply drop that flag?

batrick:
Crash I referenced is issue #1207.

batrick:
I am not aware of a use-case and I've never seen it in the wild. However, we have never checked for that flag so I have no idea if it's regularly used somewhere or not.

Issue #1178, opened 2016-03-04
Project: Makeflow JSON Representation


dthain:
Modify makeflow to emit the DAG in a straightforward JSON representation.  This will be the starting point for some interesting modifications and enhancements to Makeflow.

Start by modifying `makeflow/src/dag_visitor.c` and add a new visitor dag_to_json that takes a rule like this:

```
CPUS=5
MEMORY=10
DISK=20
output: input calib.dat
    mysim.exe input > output
```

And outputs it something like this: (details can be changed)

```
{
   cmd: "mysim.exe input >output",
   inputs: [ "input", "calib.dat" ],
   outputs: [ "output" ],
   environment: { "PATH": "/usr/bin" ],
   cores 5,
   memory: 10,
   disk: 20
}
```

Important: use the JX library in `dttools` to generate jx objects and then write them out to a file.  This will ensure that everything is escaped and constructed correctly.

I _think_ the overall structure of the file should be a JSON array.  Or perhaps it should be a dictionary, with the job id as the key?  I'm not sure.  Try one out, and then let's see how things look.

dthain:
@trshaffer take a crack at this one.

dthain:
Ok, we now have a first draft of exporting a Makelfow as JSON.
Next step is to read back that JSON file and convert it to a makeflow dag internally.
(Will require writing out a spec of exactly what we accept.)

trshaffer:
I can write out a description of the output format I've been using, which ought to lead into a discussion on parsing

trshaffer:
If I were parsing the JSON output back into a DAG, this is roughly what I'd require. A lot of the decisions (e.g. layout, which keys are required) are based on the internal data structures or current parser behavior. The variables/environment are a bit of a rough patch. I made some comments in italics in the relevant sections

# JSON Representation of Makeflows #

A Makeflow is represented as a JSON object with the following required keys.

- `"categories"`
- `"default_category"`
- `"rules"`

In addition, the following key is optional.

- `"environment"`

`"default_category"` must be one of the string keys of the categories object. The other values are described in the following sections.

## Categories ##

The categories of a Makeflow are represented as a JSON object mapping category names to category objects. Each category object contains the following keys indicating the resources required by the category.
*When the Makeflow is parsed, these are filled in with the host resources, so these are always available. Should these be optional?*

- `"cores"`
- `"disk"`
- `"memory"`
- `"gpus"`

[...]

PR #1212, opened 2016-03-17
First draft of JSON export for Makeflow

trshaffer:
For #1178 

I added a new option `makeflow_analyze -J` that reads the DAG from the given makeflow and prints a JSON representation. The output for the example included with Makeflow is [attached](https://github.com/cooperative-computing-lab/cctools/files/178743/example.json.txt). The JSON DAG object has a few main keys:

- `categories` associates variables, resource limits, etc. with category names
- `environment` is a set of environment variables to be sent with a makeflow when run
- `rules` is a list (in no particular order) of the rules that make up the makeflow

Each rule is an object that stores

- `command`
- lists of input and output filenames
- the category for that rule

I also attached [another example](https://github.com/cooperative-computing-lab/cctools/files/178744/example2.json.txt) that shows how some of the more advanced features (e.g. remote renaming, sub-makeflows, resource limits) show up.

@btovar @dthain what do you think of this format?

### DIFF
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -216,6 +216,26 @@ int jx_insert( struct jx *j, struct jx *key, struct jx *value )
 	return 1;
 }
 
+int jx_insert_unless_empty( struct jx *object, struct jx *key, struct jx *value ) {
+	switch(value->type) {
+		case JX_OBJECT:
+		case JX_ARRAY:
+			/* C99 says union members have the same start address, so
+			 * just pick one, they're both pointers. */
+			if(value->u.pairs == NULL) {
+				jx_delete(key);
+				jx_delete(value);
+				return -1;
+			} else {
+				return jx_insert(object, key, value);
+			}
+			break;
+		default:
+			return jx_insert(object, key, value);
+			break;
+	}
+}
+
 void jx_insert_integer( struct jx *j, const char *key, jx_int_t value )
 {
 	jx_insert(j,jx_string(key),jx_integer(value));
--- a/dttools/src/jx.h
+++ b/dttools/src/jx.h
@@ -176,6 +176,9 @@ struct jx * jx_remove( struct jx *object, struct jx *key );
 /** Insert a key-value pair into an object.  @param object The object.  @param key The key.  @param value The value. @return True on success, false on failure.  Failure can only occur if the object is not a @ref JX_OBJECT. */
 int jx_insert( struct jx *object, struct jx *key, struct jx *value );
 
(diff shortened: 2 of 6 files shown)


dthain:
I like it, I think it's a good start, and I'm glad to see that you have dug into some of the more challenging issues related to the makeflow-as-a-whole.

Let me suggest a few name changes to help end users:
- Make the feature accessible through `makeflow_viz -D json`, since that's the same place where we export to cytoscape, pegasus, and so forth.
- Use `environment` instead of `variables` in the categories.
- Use `inputs` and `outputs` instead of `source_files` and `target_files`
- Don't put the filename into the JSON, since the filename is external and can be changed at will.
- What is the significance of an environment variable with a null value?

I'm not sure if the rules should be an ordered array, or an object, with the rule id as the key.  Is there a good argument one way or the other?

trshaffer:
> Make the feature accessible through makeflow_viz -D json, since that's the same place where we export to cytoscape, pegasus, and so forth.

I avoided putting this into `makeflow_viz` since it's not really visualizing anything, but that's easy to move. Changing those key names and removing the filename are also no problem.

> What is the significance of an environment variable with a null value?

The DAG stores a set `export_vars` that I *think* indicates the variables that need to be taken from the makefile or host environment. Since JSON doesn't have sets, an object with null keys is one hack. I was also thinking about storing the host environment variables if not null. That way, you could try to infer host environment variables vs. category variables vs. rule variables. I'm not actually sure if this is needed, or if we should just keep the variables on the categories and rules.

> I'm not sure if the rules should be an ordered array, or an object, with the rule id as the key. Is there a good argument one way or the other?

Since the rule IDs just reflect the order Makeflow parsed them, I'm hesitant to include them in the JSON representation. An array seems cleaner, and doesn't include extraneous information, e.g.

    [
        {"command": "foo2"},
        {"command": "bar"},
        {"command": "foo1"}
    ]
vs.

    {
        "1": {"command": "foo2"},
        "2": {"command": "bar"},
        "3": {"command": "foo1"}
    }

When parsing the JSON representation, you'd end up iterating over a rules object and throwing away the keys, so it would be functionally equivalent to an array as well. Since JSON doesn't have sets, I'd prefer an array, noting that order is not important.

dthain:
I see, so this is a reflection of the fact that Make has variables (really macros) which may or may not be exported.  So, you have to distinguish between the values of the variables, and whether they are exported or not.

That actually turns out to be handy, b/c we can use the `variables` section as the context for evaluating JX expressions farther down.  So you could do this:

```
{
   variables: { "param":10,  ... },
   rules: [
        { cmd: "cmd -p "+param+" -x -y", ... }
   ]
}
```

@btovar what are the rules, exactly, for generating the environment passed along to tasks?  Do tasks only get the environment explicitly exported by makeflow, or do they inherit the environment of makeflow itself?

dthain:
Ok, I'm going to take this pull request for now.
We will have more to do in terms of parsing/printing JSON, but this is a good first step.

PR #1213, opened 2016-03-18
Pretty print JX_ARRAY

trshaffer:
The JX code prints objects with pretty indentation, but leaves arrays compact. I added some array indentation code similar to the existing code for objects


dthain:
@trshaffer please fix conflicts with recent commits.

PR #1216, opened 2016-03-19
Parrot Self Detection

### LINKED ISSUE #1203: Parrot should use mechanism other than env var to detect nested executions
Rodney Walker reported an issue with a script he wrote that may be run inside or outside of Parrot. He was trying to detect nested executions of Parrot using `ps | grep parrot_run` and only run `parrot_run` if not already inside Parrot. This was his example script:

````
ps | grep parrot_run
if [ $? -eq 1 ];then
   MYHOST=`hostname -f`
   echo No parrot on $MYHOST
   unset PARROT_ENABLED
  /home/hpc/pr58be/ri32buz2/software/cvmfs_parrot/cctools/bin/parrot_run
ls /cvmfs/atlas.cern.ch
else
  MYHOST=`hostname -f`
  echo Should be in parrot already on $MYHOST
  ps | grep parrot_run
  echo $MYHOST $?
  echo PARROT_ENABLED: $MYHOST $PARROT_ENABLED
  ls /cvmfs/atlas.cern.ch/repo
fi
````

It seems the need to for using `ps` is due to this script being part of a job submission framework which passes environment variables to jobs. So a job would begin with PARROT_ENABLED set even though it is not running Parrot.

I suggest Parrot use another mechanism like a failure from `ptrace` to detect nesting or perhaps a custom Parrot system call.

batrick:
Test if Parrot is running using parrot_version.
    
    Environment variables are not the right mechanism for testing if Parrot is
    already running. In #1203, the PARROT_ENABLED environment variable would be
    sent along with jobs to remote systems which would prevent Parrot from running.
    Instead, just try a Parrot system call to see if we're already running Parrot.
    
    Fixes #1203.

### DIFF
--- a/doc/man/parrot_run.m4
+++ b/doc/man/parrot_run.m4
@@ -57,6 +57,7 @@ OPTION_TRIPLET(-u, username, name)Use this extended username.
 OPTION_ITEM(`    --fake-setuid')Track changes from setuid and setgid.
 OPTION_ITEM(`    --valgrind')Enable valgrind support for Parrot.
 OPTION_ITEM(`-v, --version')Display version number.
+OPTION_ITEM(`    --is-running')Test is Parrot is already running.
 OPTION_TRIPLET(-w, work-dir, dir)Initial working directory.
 OPTION_ITEM(`-W, --syscall-table')Display table of system calls trapped.
 OPTION_ITEM(`-Y, --sync-write')Force synchronous disk writes.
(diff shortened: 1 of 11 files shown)


dthain:
I like this approach.

But maybe leave in `setenv("PARROT_ENABLED")` since we have encouraged that in the past as a way to update the shell prompt.  Nice to keep backwards compatibility, at least for a while.

Which reminds me, please update the manual as well, which encourages the use of `PARROT_ENABLED`.

batrick:
Done.

Issue #1207, opened 2016-03-17
Parrot crash due to lost threads


batrick:
Rodney Walker et al. got this error at an HPC center in the UK:

setparrot: pfs_table.cc:157[FINAL:0420ebc8]: Assertion 'fd == rfd ||
((0 <= rfd && rfd < pointer_count) && pointers[rfd] == __null)'
failed.

This is unrelated to #1179. The actual problem is a lost thread:

>Oh dear, this looks like a problem I thought I resolved a while ago (by updating Parrot to the newer ptrace interfaces) where a thread becomes "lost" to Parrot and eventually reappears. Near the end of the log:
>
>````
2016/03/16 12:18:34.13 parrot_run[33827] <child:34280> syscall: clone
2016/03/16 12:18:34.13 parrot_run[33827] process: waiting for blocking
event from process 34280
2016/03/16 12:18:34.13 parrot_run[33827] process: pid 34280 cloned 34338
2016/03/16 12:18:34.13 parrot_run[33827] pstree: 34280 newthread 34338
2016/03/16 12:18:34.13 parrot_run[33827] <child:34280> syscall: = 34338 [clone]
````
>
>The next event for 34338 is:
>
>````
2016/03/16 12:18:34.14 parrot_run[33827] process: pid 34338 cloned 34341
````
>
>This indicates that Parrot didn't receive any ptrace events for the thread 34338 until it cloned a new thread/process. IOW, Parrot wasn't able to intercept any of 34338's system calls. This would be why the assertion failed.
>
> I think this is probably a bug with older versions of the Linux kernel. Unfortunately, old ptrace bugs are not well documented. We'll need to discuss how to go about fixing this. It's possible a kernel update may be necessary if that's even possible for these machines...

batrick:
So for now, we think the way forward is to start by finding a way to detect this bug before the assertion failure and emit a warning for older kernels (at least the ones we suspect have this bug).

**With the assumption that we always receive a clone event for a new thread even if we eventually "lose" it**:

https://github.com/cooperative-computing-lab/cctools/blob/5ccb04599ba2ee125730981f53add80d98cf8161/parrot/src/pfs_main.cc#L359

We should add some code that makes sure we've received its initial event stop (@dthain, I misremembered, there is always an initial stop for all Linux versions). When we're using PTRACE_SEIZE (Linux >=3.4.0):

[...]

PR #1218, opened 2016-03-21
Do not use kill to wake thread from initial stop.

batrick:
Oleg Nesterov noticed a bug [1] in this code where SIGCONT could wake multiple
newly cloned threads from their initial SIGSTOP. This would cause the threads
to continue executing without stopping at system calls. That is, they would
only stop at other ptrace events like PTRACE_EVENT_CLONE or PTRACE_EVENT_EXIT.
The only way to trace system calls on a new thread is to use PTRACE_SYSCALL
after receiving its stop signal.

Possible fix for #1207.

[1] https://lkml.org/lkml/2016/3/21/644


batrick:
@btovar after this is merged I'd like to get a release out for Benjamin Wynne to try since he can somewhat reliably reproduce this bug.

batrick:
Apparently tkill isn't right either. Oleg suggest just suppressing the signal delivery by executing `ptrace(PTRACE_SYSCALL, tid, 0, 0)`.

dthain:
Nice work, @batrick!

Issue #1211, opened 2016-03-17
Specify Location of Resource Monitor Output Files Per Task in Work Queue


klannon:
Would it be possible to have the resource monitor output files go into a directory specified independently for each task, in a similar way to the way the task output file local locations can be specified?  For example, something like:

`task.specify_monitoring_dir(local)`

btovar:
Yes. We can make it so that the default is split per category. Which use case were you thinking of?

klannon:
Personally, I'd like to be able to specify a unique location for the output of each task.  Right now, all the other task related files we generate (i.e. the job outputs) go into a directory structure where each task gets its own directory.  I would like to be able to ask WQ to put the resource monitoring file in that same directory so that it's easy to associate resource usage with the rest of the task outputs (log files, etc.)  Would that be possible?

PR #1222, opened 2016-03-21
Specify monitor directory per task

btovar:
As requested by @klannon, let the user choose the name of the resource summary files.


btovar:
Fixes #1211

Issue #1220, opened 2016-03-21
Resource Monitor Delay


dthain:
Our automatic tests have gotten rather slow lately -- it takes quite a while to get through all of the makeflow tests, which used to finish in a fraction of a second.

The culprit seems to be that the category feature is enabling the resource monitor in the WQ worker, and then the RM main loop seems to make the minimum task execution time a couple of seconds.  This will have an outside impact on workflows that consist of a large number of rapid tasks.

Is there any way to improve the resource monitor so that it exits rapidly after the child completes?

btovar:
That sounds like a bug. The monitor should not take a couple of seconds to return. Let me check what's going on.

dthain:
Nice, thanks.

dthain:
Hmm, I'm still seeing huge delays in the simple test makeflows, can you look some more:
or example:
http://ccl.cse.nd.edu/a/1218bae7/1218bae7.x86_64-redhat5.log

```
--- Testing makeflow/test/TR_makeflow_001_dirs_01.sh ...          success 45s
--- Testing makeflow/test/TR_makeflow_001_dirs_02.sh ...          success 42s
--- Testing makeflow/test/TR_makeflow_001_dirs_03.sh ...          success 42s
--- Testing makeflow/test/TR_makeflow_001_dirs_04.sh ...          success 42s
--- Testing makeflow/test/TR_makeflow_001_dirs_05.sh ...          success 42s
--- Testing makeflow/test/TR_makeflow_001_dirs_06.sh ...          success 42s
--- Testing makeflow/test/TR_makeflow_001_dirs_07.sh ...          success 62s
--- Testing makeflow/test/TR_makeflow_001_dirs_08.sh ...          success 42s
--- Testing makeflow/test/TR_makeflow_001_dirs_09.sh ...          success 43s
--- Testing makeflow/test/TR_makeflow_001_dirs_10.sh ...          success 42s
--- Testing makeflow/test/TR_makeflow_001_dirs_11.sh ...          success 82s
```

btovar:
RH5 still is slow... with #1225. Checking....

dthain:
I think there is a second issue here -

I don't think we want the RM enabled unless the user has specifically asked for it, or a feature requested that requires it.   Is there a reason that it seems to be enabled for the base makeflow test cases?

btovar:
[...]

PR #1226, opened 2016-03-22
Set dummy handler for SIGCONT.

btovar:
Third time... SIGCONT was ignored by sigtimedwait in RH5 unless there is an explicit handler. Now all tests run in acceptable time.


dthain:
Go ahead and merge when ready.

btovar:
Fix for #1220

PR #1227, opened 2016-03-22
ensure only one job untar a starch SFX

hmeng-19:
Problem: When multiple commands based on the same SFX executable, let us say `convert`, are running concurrently, they all try to uncompress the tarfile inside the SFX at the same time. 

Solution: using locking to guarantee only one process, `P1`, uncompress the SFX, other processes will wait until `P1` finishes.


hmeng-19:
@dthain , please give this a look.

dthain:
Looks good.

PR #1229, opened 2016-03-23
build resource_monitor_histograms not only on Linux

btovar:


btovar:
More trouble than it's worth to get right.

PR #1234, opened 2016-03-30
Select for 1 second to reduce CPU load.

batrick:


btovar:
I assume this should be deployed as soon as possible? @dthain, @batrick

dthain:
Looks good, please merge and deploy.

btovar:
Deployed. (5.4.8)

PR #1238, opened 2016-04-04
Adds Makeflow & Work Queue recognizing loop device being full

nkremerh:
@btovar Take a look?

### DIFF
--- a/dttools/src/libforce_halt_enospc.c
+++ b/dttools/src/libforce_halt_enospc.c
@@ -59,7 +59,6 @@ ssize_t write(int fd, const void *buf, size_t count) {
 	if(real_count < 0 && errno == ENOSPC) {
 		int fd;
 		fd = open("../loop_dev_report.txt", O_RDWR | O_CREAT);
-		char *desc = sprintf("fd: %d\n", fd);
 		if(fd < 0) { original_write(STDERR_FILENO, "WRITE ERROR: could not alert Work Queue of full loop device.\n", 61); }
 		original_write(STDERR_FILENO, "WRITE ERROR: device capacity reached.\n", 39);
 		return real_count;
--- a/dttools/src/libforce_halt_enospc.c
+++ b/dttools/src/libforce_halt_enospc.c
@@ -58,7 +58,7 @@ ssize_t write(int fd, const void *buf, size_t count) {
 
 	if(real_count < 0 && errno == ENOSPC) {
 		int fd;
-		fd = open("../loop_dev_report.txt", O_RDWR | O_CREAT);
+		fd = open("../disk_alloc_report.txt", O_RDWR | O_CREAT);
 		if(fd < 0) { original_write(STDERR_FILENO, "WRITE ERROR: could not alert Work Queue of full loop device.\n", 61); }
 		original_write(STDERR_FILENO, "WRITE ERROR: device capacity reached.\n", 39);
 		return real_count;
--- a/dttools/src/libforce_halt_enospc.c
+++ b/dttools/src/libforce_halt_enospc.c
@@ -10,33 +10,14 @@
 #include <stdlib.h>
 #include <dlfcn.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <string.h>
 #include <signal.h>
 #include <inttypes.h>
 #include <stdarg.h>
 #include <unistd.h>
 #include <sys/types.h>
-
-ssize_t write(int fd, const void *buf, size_t count) {
-
-	__typeof__(write) *original_write = dlsym(RTLD_NEXT, "write");
-
-	int real_count;
-	int prev_errno = errno;
-	errno = 0;
-	real_count = original_write(fd, buf, count);
-
-	if(real_count < 0 && errno == ENOSPC) {
-		original_write(STDERR_FILENO, "WRITE ERROR: device capacity reached.\n", 39);
-		return real_count;
-	}
-
-	if(!errno) {
-	   errno = prev_errno;
-	}
-
-	return real_count;
-}
+#include <sys/stat.h>
 
 int open(const char *path, int flags, ...)
 {
@@ -64,4 +45,29 @@ int open(const char *path, int flags, ...)
 	}
 
 	return fd;
+}
+
+ssize_t write(int fd, const void *buf, size_t count) {
+
+	__typeof__(write) *original_write = dlsym(RTLD_NEXT, "write");
+
+	int real_count;
+	int prev_errno = errno;
+	errno = 0;
+	real_count = original_write(fd, buf, count);
+
+	if(real_count < 0 && errno == ENOSPC) {
+		int fd;
+		fd = open("../loop_dev_report.txt", O_RDWR | O_CREAT);
+		char *desc = sprintf("fd: %d\n", fd);
+		if(fd < 0) { original_write(STDERR_FILENO, "WRITE ERROR: could not alert Work Queue of full loop device.\n", 61); }
+		original_write(STDERR_FILENO, "WRITE ERROR: device capacity reached.\n", 39);
+		return real_count;
+	}
+
+	if(!errno) {
+	   errno = prev_errno;
+	}
+
+	return real_count;
 }
--- a/dttools/src/libforce_halt_enospc.c
+++ b/dttools/src/libforce_halt_enospc.c
@@ -20,17 +20,6 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 
-char *strconcat(char *s1, char *s2)
-{
-    char *result = malloc(strlen(s1)+strlen(s2)+1);
-	if(!result) {
-		return "";
-	}
-    strcpy(result, s1);
-    strcat(result, s2);
-    return result;
-}
-
 int open(const char *path, int flags, ...)
 {
 
@@ -49,20 +38,14 @@ int open(const char *path, int flags, ...)
 
 	if(fd == -1 && errno == ENOSPC) {
 		int err_fd;
-		char *filename;
-		char *parent = "../";
-		char *env_name = getenv("CCTOOLS_DISK_ALLOC");
-		unsetenv("CCTOOLS_DISK_ALLOC");
-		filename = strconcat(parent, env_name);
-		if(strlen(filename) == 0) {
-			fprintf(stderr, "OPEN ERROR: could not generate filename for alerting Work Queue of loop device inode exhaustion.\n");
+		char *filename = getenv("CCTOOLS_DISK_ALLOC");
+		if(!filename) {
+			fprintf(stderr, "OPEN ERROR: could not generate filename for alerting resource management system of loop device inode exhaustion.\n");
 			fprintf(stderr, "OPEN ERROR: inode capacity reached.\n");
-			free(filename);
 			return fd;
 		}
 		err_fd = open(filename, O_RDWR | O_CREAT);
-		free(filename);
-		if(err_fd < 0) { fprintf(stderr, "OPEN ERROR: could not alert Work Queue of loop device inode exhaustion.\n"); }
+		if(err_fd < 0) { fprintf(stderr, "OPEN ERROR: could not alert resource management system of loop device inode exhaustion.\n"); }
 		fprintf(stderr, "OPEN ERROR: inode capacity reached.\n");
 		return fd;
 	}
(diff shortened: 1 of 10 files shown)


makeflow/src/makeflow.c
```diff
@@ -651,7 +651,11 @@ static void makeflow_node_complete(struct dag *d, struct dag_node *n, struct bat
 
 	struct list *outputs = makeflow_generate_output_files(n, wrapper, monitor);
 
-	if(info->exited_normally && info->exit_code == 0) {
+
+	if(getenv("CCTOOLS_LOOP_DEV_FULL")) {
+		job_failed = 1;
```
dthain: I can see why you did it this way, but let's not get into the habit of passing information back via an environment variable.
Add an element to `batch_job_info` instead, which is where we send back detailed info about jobs.

work_queue/src/work_queue.c
```diff
@@ -1583,11 +1583,11 @@ static work_queue_result_code_t get_result(struct work_queue *q, struct work_que
 	timestamp_t effective_stoptime = 0;
 	time_t stoptime;
 
-	//Format: task completion status, exit status (exit code or signal), output length, execution time, taskid
-	char items[4][WORK_QUEUE_PROTOCOL_FIELD_MAX];
-	int n = sscanf(line, "result %s %s %s %s %" SCNd64, items[0], items[1], items[2], items[3], &taskid);
+	//Format: task completion status, exit status (exit code or signal), output length, execution time, loop device, taskid
+	char items[5][WORK_QUEUE_PROTOCOL_FIELD_MAX];
+	int n = sscanf(line, "result %s %s %s %s %s %" SCNd64, items[0], items[1], items[2], items[3], items[4], &taskid);
```
dthain: When changing a WQ message, do it in  backwards-compatible way, since users always manage to mix and match masters and workers from different versions.   Add the new element to the end of the string, and have the master accept both n=5 and n=6.

btovar:
Nate, 
You want to increment  the protocol version in work_queue_protocol.h (add a comment to why, too)

btovar:
Also, I can't release as 5.4, as you are changing the protocol. I can make a release candidate for 5.5, though.

btovar:
Why do you need the loop device in the result message? Could you send it as an info message? (See process_info in work_queue.c.)

dthain:
Need to bring this one up to date?

dthain:
The general approach looks good, but use the name `disk_allocation` or similar instead of `loop_dev` throughout, since allocation is the objective, and we will want to use different mechanisms (resource monitor, containers, etc) at the worker in the future, without changing makeflow or the work queue master.

nkremerh:
@btovar Would you mind taking a look at this please?

nkremerh:
@dthain I believe this is ready to merge. Can you and @btovar take a look please?

dttools/src/libforce_halt_enospc.c
```diff
@@ -65,3 +46,27 @@ int open(const char *path, int flags, ...)
...
+	__typeof__(write) *original_write = dlsym(RTLD_NEXT, "write");
+
+	int real_count;
+	int prev_errno = errno;
+	errno = 0;
+	real_count = original_write(fd, buf, count);
+
+	if(real_count < 0 && errno == ENOSPC) {
+		int fd;
+		fd = open("../disk_alloc_report.txt", O_RDWR | O_CREAT);
```
btovar: I would choose a less generic name, such as:   'cctools_disk_alloc_report.log'.

Also, why the ../? Are you assuming this is running in a task directory in work queue?

makeflow/src/makeflow.c
```diff
@@ -678,6 +682,24 @@ static void makeflow_node_complete(struct dag *d, struct dag_node *n, struct bat
 			}
 		}
 
+		if(info->disk_alloc_full) {
```
btovar: This does not play nicely with the code below if both monitor and disk allocations are activated.

dttools/src/libforce_halt_enospc.c
nkremerh: Yes, currently the only use case for this library is for a WQ worker. If we would like to make this more generic right away, please let me know.

work_queue/src/work_queue.c
```diff
@@ -1670,9 +1670,9 @@ static work_queue_result_code_t get_result(struct work_queue *q, struct work_que
 	timestamp_t effective_stoptime = 0;
 	time_t stoptime;
 
-	//Format: task completion status, exit status (exit code or signal), output length, execution time, taskid
-	char items[4][WORK_QUEUE_PROTOCOL_FIELD_MAX];
-	int n = sscanf(line, "result %s %s %s %s %" SCNd64, items[0], items[1], items[2], items[3], &taskid);
+	//Format: task completion status, exit status (exit code or signal), output length, execution time, taskid, loop device
+	char items[5][WORK_QUEUE_PROTOCOL_FIELD_MAX];
+	int n = sscanf(line, "result %s %s %s %s %" SCNd64" %s", items[0], items[1], items[2], items[3], &taskid, items[4]);
```
btovar: Rather than modify the protocol, add another entry to enum rmonitor_errors to indicate loop device exhaustion.

btovar: I'm not sure about modifying the result message here. Perhaps we can do it through enum rmonitor_errors?

work_queue/src/work_queue_worker.c
```diff
@@ -618,6 +618,13 @@ static int handle_tasks(struct link *master)
 				debug(D_WQ, "task %d (pid %d) exited abnormally with signal %d",p->task->taskid,p->pid,p->exit_status);
 			} else {
 				p->exit_status = WEXITSTATUS(status);
+				FILE *loop_full_check;
+				if(p->loop_mount == 1 && (loop_full_check = fopen("./disk_alloc_report.txt", "r"))) {
+					p->task_status = WORK_QUEUE_RESULT_RESOURCE_EXHAUSTION;
+					p->task->disk_alloc_full = 1;
+					fclose(loop_full_check);
+					unlink("./disk_alloc_report.txt");
```
btovar: I don't like that this name is fixed, as it may collide with a user file.  A better way would be to create a tmp file and pass the name to your library, perhaps with an environment variable.

dttools/src/libforce_halt_enospc.c
btovar: Aha, see my comment below. Perhaps we can make the name not to be fixed, which will help with relative paths.
One option is to always create the files as temps in /tmp, and communicate the name to your library somehow.

dthain:
What remains to be done here?

nkremerh:
@btovar Let me know what you think of this.

dttools/src/libforce_halt_enospc.c
```diff
@@ -10,32 +10,25 @@
...
-		return real_count;
-	}
-
-	if(!errno) {
-	   errno = prev_errno;
+char *strconcat(char *s1, char *s2)
+{
+    char *result = malloc(strlen(s1)+strlen(s2)+1);
+	if(!result) {
+		return "";
```
btovar: Is this correct? It is better to fail than to return an incorrect result.

dttools/src/libforce_halt_enospc.c
```diff
@@ -10,32 +10,25 @@
...
-	real_count = original_write(fd, buf, count);
-
-	if(real_count < 0 && errno == ENOSPC) {
-		original_write(STDERR_FILENO, "WRITE ERROR: device capacity reached.\n", 39);
-		return real_count;
-	}
-
-	if(!errno) {
-	   errno = prev_errno;
+char *strconcat(char *s1, char *s2)
```
btovar: I'd suggest:
char *result = xxstrdup(s1);
result = string_combine(result, s2);
return result;

dttools/src/libforce_halt_enospc.c
```diff
@@ -55,6 +48,21 @@ int open(const char *path, int flags, ...)
...
 
 	if(fd == -1 && errno == ENOSPC) {
+		int err_fd;
+		char *filename;
+		char *parent = "../";
+		char *env_name = getenv("CCTOOLS_DISK_ALLOC");
+		unsetenv("CCTOOLS_DISK_ALLOC");
+		filename = strconcat(parent, env_name);
+		if(strlen(filename) == 0) {
+			fprintf(stderr, "OPEN ERROR: could not generate filename for alerting Work Queue of loop device inode exhaustion.\n");
```
btovar: remove 'Work Queue', as the mechanism is more general.

dttools/src/libforce_halt_enospc.c
```diff
@@ -55,6 +48,21 @@ int open(const char *path, int flags, ...)
 	fd = original_open(path, flags, mode);
 
 	if(fd == -1 && errno == ENOSPC) {
+		int err_fd;
+		char *filename;
+		char *parent = "../";
```
btovar: Why ../?

dttools/src/libforce_halt_enospc.c
```diff
@@ -55,6 +48,21 @@ int open(const char *path, int flags, ...)
...
+		char *env_name = getenv("CCTOOLS_DISK_ALLOC");
+		unsetenv("CCTOOLS_DISK_ALLOC");
+		filename = strconcat(parent, env_name);
```
btovar: What is the default if CCTOOLS_DISK_ALLOC is unset?

dttools/src/libforce_halt_enospc.c
```diff
@@ -55,6 +48,21 @@ int open(const char *path, int flags, ...)
 	fd = original_open(path, flags, mode);
 
 	if(fd == -1 && errno == ENOSPC) {
+		int err_fd;
+		char *filename;
+		char *parent = "../";
+		char *env_name = getenv("CCTOOLS_DISK_ALLOC");
+		unsetenv("CCTOOLS_DISK_ALLOC");
```
btovar: Why do you unset the variable?

dttools/src/libforce_halt_enospc.c
btovar: I would recommend to have the default name as:

cctools_disk_allocation_exhausted.PID (where PID comes from getpid()).

If CCTOOLS_DISK_ALLOC is defined, then that name should not be modified.

dttools/src/libforce_halt_enospc.c
```diff
@@ -65,3 +73,39 @@ int open(const char *path, int flags, ...)
...
+	int real_count;
+	int prev_errno = errno;
+	errno = 0;
+	real_count = original_write(fd, buf, count);
+
+	if(real_count < 0 && errno == ENOSPC) {
+		int fd;
+		char *filename;
+		char *parent = "../";
+		char *env_name = getenv("CCTOOLS_DISK_ALLOC");
```
btovar: Code repetition... write a function to encompass this.

work_queue/src/work_queue.h
```diff
@@ -67,6 +67,7 @@ typedef enum {
 	WORK_QUEUE_RESULT_FORSAKEN            = 5 << 3, /**< The task failed, but it was neither a task or worker error **/
 	WORK_QUEUE_RESULT_MAX_RETRIES         = 6 << 3, /**< The task could not be completed successfully in the given number of retries. **/
 	WORK_QUEUE_RESULT_TASK_MAX_RUN_TIME   = 7 << 3, /**< The task ran for more than the specified time (relative since running in a worker). **/
+	WORK_QUEUE_RESULT_DISK_ALLOC_FULL     = 1 << 4  /**< The task filled its loop device allocation but needed more space. **/
```
btovar: Better 8 << 3, so next person does not have to think about it.

dttools/src/libforce_halt_enospc.c
nkremerh: In this case, CCTOOLS_DISK_ALLOC is unique for each task if I am remembering environment variables correctly. To address using PID instead of taskid, we get PID once we fork. At that point, I believe it is out of our hands which is why I am using taskid since we know that before we ship out the task.

nkremerh:
@btovar Hopefully this alleviates your concerns.

nkremerh:
@btovar Bump

dttools/src/libforce_halt_enospc.c
```diff
@@ -49,20 +38,14 @@ int open(const char *path, int flags, ...)
...
-		char *parent = "../";
-		char *env_name = getenv("CCTOOLS_DISK_ALLOC");
-		unsetenv("CCTOOLS_DISK_ALLOC");
-		filename = strconcat(parent, env_name);
-		if(strlen(filename) == 0) {
-			fprintf(stderr, "OPEN ERROR: could not generate filename for alerting Work Queue of loop device inode exhaustion.\n");
+		char *filename = getenv("CCTOOLS_DISK_ALLOC");
+		if(!filename) {
+			fprintf(stderr, "OPEN ERROR: could not generate filename for alerting resource management system of loop device inode exhaustion.\n");
 			fprintf(stderr, "OPEN ERROR: inode capacity reached.\n");
```
btovar: If the variable is unset that does not mean the inode cpacity is reached.

work_queue/src/work_queue_process.c
```diff
@@ -211,9 +211,11 @@ pid_t work_queue_process_execute(struct work_queue_process *p, int container_mod
 	}
 
 	if(p->loop_mount) {
-		char *pwd = get_current_dir_name();
+		char *buf = malloc(PATH_MAX);
```
btovar: If get_current_dir_name does not work on OSX, please add a fix there.

dttools/src/libforce_halt_enospc.c
```diff
@@ -55,6 +37,15 @@ int open(const char *path, int flags, ...)
 	fd = original_open(path, flags, mode);
 
 	if(fd == -1 && errno == ENOSPC) {
+		int err_fd;
+		char *filename = getenv("CCTOOLS_DISK_ALLOC");
+		if(!filename) {
+			fprintf(stderr, "OPEN ERROR: could not generate filename for alerting resource management system of loop device inode exhaustion.\n");
```
btovar: ENOSPC and inode exhaustion are different things. The message is confusing.

PR #1239, opened 2016-04-04
add makeflow support for specifying input deps

hmeng-19:
Problem:
Currently, Makeflow allows the user to use absolute paths to specify the input dependencies, which affects the portability of the makeflow.

Solution:
Limit the behavior of a makeflow within the current working directory, and allow the user to specify the source and target of each input dependency for a makeflow through `--mounts <mountfile >` option. Every line of mountfile should be in the format of `<target> <source>`.

Here is an example of a mountfile:

    curl /usr/bin/curl
    convert ../bin/convert
    data/file1 /home/bob/input1

To execute a makeflow with a mountfile, named mountfile, specifying its input dependencies:
`makeflow --mounts mountfile example.makeflow`
Before starting parsing the DAG, Makeflow first parses each line of the mountfile, and copies the specified dependency from the location specified by source field to the location specified by the target field.

To cleanup a makeflow together with the input dependencies installed through the --mounts option, run the following command:
`makeflow --mounts mountfile -c example.makeflow`
The input dependencies set up by the --mounts option can only be cleaned up when the makeflow clean mode is set to all. After Makeflow cleans up the outputs and intermediate files, each target specified in the mountfile will be removed.


hmeng-19:
@dthain , please give this a look.

dthain:
Haiyan, this is a good start on a tricky feature.  However, after seeing the code changes to makeflow, I am concerned that this is going to interact in complex ways with storage management, garbage collection, and other things going forward.

Wouldn't this be simpler to implement as a name-substitution instead?  e.g. for each entry of the form `-Mx=y`, enter `x->y` into a hash table.  Then, replace references to `dag_file->filename` with a wrapper that looks up the entry in the hashtable.  That way, you can support arbitrary filenames, instead of placing all sorts of restrictions on relative paths.

hmeng-19:
Your suggestion may work for local dependencies.
However, I try to picture this pull request as a first step of decouple workflow specification from data dependencies. In the future, we can also allow the mountfile  to include dependencies from the Internet.

hmeng-19:
@dthain , Before you decide the final fate of this pull request, we should discuss this problem more thoroughly.

nhazekam:
If possible, I would love to sit in on this as it directly relates to recent work in Makeflow.

hmeng-19:
@dthain , @btovar and @nhazekam , I think the first version of this feature is done. 
Please give it a look.
Two new options (`--mounts` and `--cache`) are added. You can find how to use and test it in the `doc/makeflow.html` in this branch.

makeflow/src/dag_file.h
```diff
@@ -26,6 +26,12 @@ typedef enum {
 	DAG_FILE_STATE_UP
 } dag_file_state_t;
 
+/* the type of a dependency specified in the mountfile */
+typedef enum {
+	SOURCE_LOCAL,
+	SOURCE_HTTP,
+	SOURCE_UNSUPPORTED
+} source_type;
```
nhazekam: To follow the naming scheme this should probably be dag_file_source_t.

makeflow/src/makeflow_log.c
```diff
@@ -200,6 +224,42 @@ void makeflow_log_recover(struct dag *d, const char *filename, int verbose_mode,
...
+			}
+			if(sscanf(line, "# MOUNT %" SCNu64 " %s %s %s %d", &previous_completion_time, file, source, cache_name, &type) == 5) {
+				char *p = NULL;
+				f = dag_file_lookup_or_create(d, file);
+
+				/* if the user specifies the same target in the mountfile, ignore the info from the log file */
+				if(f->source) continue;
+
+				p = xxstrdup(source);
+				if(!p) {
```
nhazekam: I am not sure that this checking is necessary as xxstrdup is used. If strdup fails then xxstrdup emits a fatal warning. @btovar am I misinterpreting xxstrdup?

makeflow/src/makeflow_log.c
```diff
@@ -200,6 +224,42 @@ void makeflow_log_recover(struct dag *d, const char *filename, int verbose_mode,
...
+				}
+				f->source = p;
+
+				p = NULL;
+				p = xxstrdup(cache_name);
+				if(!p) {
+					LDEBUG("xxstrdup(%s) failed!\n", cache_name);
+					return;
+				}
+				f->cache_name = p;
```
nhazekam: If a different mount file is used in future runs, which would be log, this would be a memory leak. Even if we don't support it currently, we should check if it is set and free if needed.

makeflow/src/mountfile.h
```diff
@@ -0,0 +1,34 @@
...
+#define MOUNTFILE_H
+
+#include "dag.h"
+
+/* mountfile_parse parses the mountfile and loads the info of each dependency into the dag structure d.
+ * @param mountfile: the path of a mountfile
+ * @param d: a dag structure
+ * @return 0 on success, -1 on failure.
+ */
+int mountfile_parse(const char *mountfile, struct dag *d);
```
nhazekam: These functions should be uniformly named as makeflow_mount_* or makeflow_mountfile_* to follow the naming conventions in makeflow.

makeflow/src/mountfile.c
```diff
@@ -0,0 +1,660 @@
...
+		}
+
+		/* log the dependency */
+		makeflow_log_mount_event(d, df->filename, df->source, df->cache_name, type);
+	}
+	list_delete(list);
+	return 0;
+}
+
+int mount_uninstall_all(struct dag *d) {
```
nhazekam: This logic belongs more in makeflow_gc as it is part of file cleaning. This would be changes to makeflow_clean_file and makeflow_clean.

makeflow/src/mountfile.c
```diff
@@ -0,0 +1,660 @@
...
+int mount_install_all(struct dag *d) {
+	struct list *list;
+	struct dag_file *df;
+	char *cache_dir;
+
+	if(!d) return 0;
+
+	if(d->cache_dir) {
+		/* check the validity of the cache_dir and create it if neccessary and feasible. */
+		if(check_cache_dir(d->cache_dir)) {
```
nhazekam: If this function creates the cache_dir, why is it only used when the user specifies the cache and not also when we create a cache for it as in the else clause of this if statement?

makeflow/src/mountfile.c
```diff
@@ -0,0 +1,660 @@
...
+			LDEBUG("mkdtemp(%s) failed: %s\n", cache_dir, strerror(errno));
+			free(cache_dir);
+			return -1;
+		}
+
+		d->cache_dir = cache_dir;
+	}
+
+	/* log the cache dir info */
+	makeflow_log_cache_event(d, d->cache_dir);
```
nhazekam: Do we want a new log event every time we run Makeflow with mount or only when the mount changes? @btovar @dthain 

Do repeated runs all contain a new log entry for cache_dir?

makeflow/src/dag_file.h
hmeng-19: Accept. I will fix these naming style problem when you are done with commenting.

makeflow/src/makeflow_log.c
hmeng-19: Here should not have any memory leakage problem.
When there are multiple items in the log file sharing the same target but having different sources, we just take the first one and ignore the remaining ones, as shown in line 243.

hmeng-19: During parsing the mountfile, if there is already one dag_file with the same target set up due to the logging system, `mountfile_parse` would free the `source` set by the logging system, and use the `source` field specified in the mountfile.

makeflow/src/mountfile.h
hmeng-19: Accept.

makeflow/src/dag_file.c
```diff
@@ -84,4 +87,32 @@ int dag_file_in_trans( const struct dag_file *f )
 		return 0;
 }
 
+int dag_file_mount_clean(struct dag *d) {
```
nhazekam: The structure of this function/location aren't very clear. This function should operate on a single dag_file and clean the cache information. If you want to do an operation on the list of dag_files that should be in dag.c.

makeflow/src/dag_file.c
```diff
@@ -84,4 +87,32 @@ int dag_file_in_trans( const struct dag_file *f )
...
+	struct list *list;
+	struct dag_file *df;
+	if(!d) return 0;
+
+	list = dag_input_files(d);
+	if(!list) return 0;
+
+	list_first_item(list);
+	while((df = (struct dag_file *)list_next_item(list))) {
+		if(df->source) {
```
nhazekam: I think line 100-107 are what should remain in dag_file.c. Then a function "dag_mount_clean" could iterate over the the files calling dag_file_mount_clean as well as cleaning the dag level information such as cache_dir.

makeflow/src/mountfile.c
hmeng-19: I personally think that leaving `mount_uninstall_all` and `mount_install_all` in the same source file is okay.

makeflow/src/mountfile.c
hmeng-19: when the user specifies a cache_dir, we need to check its validity such as the cache_dir path does not include `..`, can not be absolute path, can not be an existing file, and so on.
These checking is not applicable when the user does not specify a cache_dir.

makeflow/src/makeflow_log.c
nhazekam: However, if between different runs the mountfile is changed and a source is changed for a target, both have been logged and we are overwriting the early retrieved log event with the latest. The behavior is what we want, as we want the most recent, but it would overwrite a pointer.

makeflow/src/dag_file.c
hmeng-19: I think the structure should be clear, it gets all the input files of a dag, and try to clean up each dag_file's source and cache_name.

hmeng-19: I agree with you that dag.c should be a better location for this.

makeflow/src/dag_file.c
hmeng-19: Accept.

makeflow/src/mountfile.c
nhazekam: I understand, but if the function works for both cases, even if it is overkill, then I think we can use it for both. This would allow us to log the creation in check_cache_dir and just set the cache_dir name if it were null.

makeflow/src/makeflow_log.c
hmeng-19: Then the case gets complex, because the `dag_file->source` may be set up by the `--mounts` option, or by the older items from the log. We need to have a way to differentiate these two cases.

hmeng-19: No matter how we rank the different sources from the log file, the source field from the mountfile must win. Right?

makeflow/src/mountfile.c
hmeng-19: I am okay with both options. These two cases are added into this branch at two very different coding time points, so I did not try to merge them.

makeflow/src/makeflow_log.c
nhazekam: Yes, mountfile wins. I think I am wrong about the memory leak in this case. I missed that you shortcut if the source is already set. Though this slightly departs from log_recovery's attempt to replay the DAG state I think its fine in this case.

makeflow/src/dag_file.c
nhazekam: After looking at this again I don't think it is necessary. I think it is appropriate to free this information, but only in the context of freeing a whole dag_file and not just mount information. Is there any practical point, other than code/memory cleanliness, to free this information?

makeflow/src/mountfile.c
```diff
@@ -0,0 +1,660 @@
...
+		LDEBUG("the target (%s) should not include any symbolic link!\n", target);
+		fprintf(stderr, "the target (%s) should not include any symbolic link!\n", target);
+		return -1;
+	}
+
+	/* Check whether target already exists. */
+	if(!access(target, F_OK)) {
+		LDEBUG("the target (%s) already exists!\n", target);
+		fprintf(stderr, "the target (%s) already exists!\n", target);
+		return -1;
```
nhazekam: This if statement prevents the user from using the same commandline, even if the mountfile is the same. I thought we only prevent them from moving forward if the mountfiles were different and changed the information?

nhazekam: To clarify a user can't run:
makeflow test.makeflow --mounts mountfile
encounter and error and then again run
makeflow test.makeflow --mounts mountfile
without makeflow no longer allowing it as the targets exist as we created them in the first.

You can remove --mounts from the second for it to run, but I think that is counter-intuitive and not what we want in this case.

makeflow/src/mountfile.c
nhazekam: By having it separate, clean does not know what to do with mounted files. Though they don't have to be in the same class, I think they are part of the same process.

makeflow/src/makeflow.c
```diff
@@ -754,6 +761,11 @@ static int makeflow_check(struct dag *d)
 			if(skip_file_check || batch_fs_stat(remote_queue, f->filename, &buf) >= 0) {
 				continue;
 			}
+
+			if(f->source) {
```
nhazekam: The makeflow is incorrectly reported as inconsistent when you clean twice without intermediate runs, this is because we don't load the mountfile when we are cleaning and rely on the makeflow log to provide source.

makeflow/src/dag_file.c
hmeng-19: I like the idea that all the malloced space should be freed somewhere.
If this does not break any other part, I would suggest leaving them here.

nhazekam: If you would like to free all of the malloced space, then a dag_file_delete function should be written that is not specific to mount. In this function we only partially free malloced space for the dag_file. Also this information is cleaned and removed from the dag_file structures prior to running the workflow, which is not incorrect, but doesn't allow that information to be used be any other parts of makeflow in the future.

makeflow/src/mountfile.c
hmeng-19: If the user really wants to keep using `--mount` option, then the correct way of doing this is to clean up the cache between different runs.

nhazekam: If the cache and the mountfile have not changed at all, then I think that is still correct. The user is simply verifying that they are using the same input file by specifying the mount.

hmeng-19: Another option which allows multiple runs to share the same cache is to use `--cache` and `--mount` option together, and add a new behavoir to makeflow cleanup  - `makeflow -ctargets 1.mf`, which only cleans the targets, and leaves the cache there.

makeflow/src/makeflow.c
hmeng-19: Currently, I do not suppose the `--mounts` and `-c` options should be used concurrently.
However, I may be too strict on how the user uses these options.
Let us give it more discussion.

makeflow/src/dag_file.c
hmeng-19: Makeflow has lots of places doing memory allocation and free, and I do not think I know all of them. 
When I implement my branch, I try to guarantee all the codes I wrote should have no memory leakage.

hmeng-19: If you can guarantee there is no other places freeing the other fields of `dag_file`, I can enlarge my function to also free other fields.

dttools/src/copy_tree.h
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (C) 2013- The University of Notre Dame
```
btovar: 2016

dttools/src/copy_tree.c
```diff
@@ -0,0 +1,287 @@
+/*
+ * Copyright (C) 2013- The University of Notre Dame
```
btovar: 2016

dttools/src/copy_tree.c
```diff
@@ -0,0 +1,287 @@
...
+
+int copy_symlink(const char *source, const char *target) {
+	char linkname[PATH_MAX];
+	ssize_t r;
+
+	assert(source);
+	assert(target);
+
+	if(!access(target, F_OK)) {
+		LDEBUG("%s already exists!\n", target);
```
btovar: Why LDEBUG? You want notice() of DEBUG here, don't you?
(unless these are temporary while you test the branch)

dttools/src/copy_tree.c
```diff
@@ -0,0 +1,287 @@
...
+
+	r = readlink(source, linkname, sizeof(linkname));
+	if(r == -1) {
+		LDEBUG("readlink(`%s`) failed: %s!\n", source, strerror(errno));
+		return -1;
+	}
+	linkname[r] = '\0';
+
+	if(symlink(linkname, target)) {
+		LDEBUG("symlink(`%s`, `%s`) failed: %s\n", linkname, target, strerror(errno));
```
btovar: notice(...

dttools/src/copy_tree.c
```diff
@@ -0,0 +1,287 @@
...
+
+	if(S_ISDIR(s_stat.st_mode)) {
+		if(!create_dir(t, (int)default_dirmode)) return -1;
+		return copy_dir_real(s, t);
+	} else if(S_ISREG(s_stat.st_mode)) {
+		return copy_file_to_file(s, t);
+	} else if(S_ISLNK(s_stat.st_mode)) {
+		return copy_symlink(s, t);
+	} else {
+		LDEBUG("Ignore Copying %s: only dir, regular files, and symlink are supported!\n", s);
```
btovar: Not copying...

dttools/src/copy_tree.h
hmeng-19: Accept. 
Oops. I directly copied these copyright info from other source files and got tricked.

dttools/src/copy_tree.c
hmeng-19: Accept.

dttools/src/copy_tree.c
hmeng-19: I added a LDEBUG macro into `debug.h` which also reports the function name, line number when an error happens.

dttools/src/copy_tree.c
```diff
@@ -0,0 +1,287 @@
...
+
+		t = path_concat(target, entry->d_name);
+		if(!t) {
+			free(s);
+			rc = -1;
+			goto finish;
+		}
+
+		/* try-the-best copy: if failed to copy one direntry, move on to the next direntry instead of exiting. */
+		copy_direntry(s, t);
```
btovar: Is there a reason not to stop at the first failure? If it won't work, I rather know right away.

dttools/src/copy_tree.c
hmeng-19: I think these information are helpful for my debugging.

dttools/src/copy_tree.c
hmeng-19: Same reason with above. 
`LDEBUG` is a debug-level info, and used in all the codes I added.

dttools/src/copy_tree.c
btovar: Nice, just change them to notices, or debugs in the final version, so it looks like the rest of the debug statements that we generate.

dttools/src/copy_tree.c
hmeng-19: It should be `copying`. Right?

btovar: I meant: "Not copying...".  "Ignore copying" sounds as if you are copying but ignoring the results.

dttools/src/copy_tree.c
hmeng-19: This decision here is always difficult to me. Either way is okay and acceptable to me.
I just chose to copy as much as possible.

hmeng-19: Oops. It seems that I forgot to check the exit code of `copy_direntry`. I will fix that.

dttools/src/copy_tree.c
hmeng-19: Okay. Sounds good. Accept.

dttools/src/copy_tree.c
```diff
@@ -0,0 +1,287 @@
...
+	if(!q) {
+		LDEBUG("xxstrdup(%s) failed: %s!\n", s, strerror(errno));
+		return NULL;
+	}
+
+	p = q;
+
+	while(*p) {
+		char old_p;
+		size_t n = strspn(p, "/") + strcspn(p, "/");
```
btovar: Clever.

dttools/src/copy_tree.c
hmeng-19: Okay. May I leave the `LDEBUG` macro in the `debug.h` in the final version?

dttools/src/copy_tree.c
hmeng-19: Thanks.

dttools/src/path.c
```diff
@@ -462,4 +463,118 @@ char *path_which(const char *exec) {
...
+	s1 = strlen(p1);
+	s2 = strlen(p2);
+
+	p = malloc(sizeof(char) * (s1 + s2 + 2));
+	if(!p) {
+		fprintf(stderr, "path_concat malloc failed: %s!\n", strerror(errno));
+		return NULL;
+	}
+
+	snprintf(p, s1+s2+2, "%s/%s", p1, p2);
```
btovar: Does it matter if s1 ends in '/' ?

dttools/src/copy_tree.c
btovar: Of course! I could have used something like it before...

dttools/src/path.c
hmeng-19: It does not affect the correctness of the code. However, it indeed looks a little ugly.

btovar: I'm thinking of your code that looks for for a directory predecessor.  Wouldn't the predecessor of 'b' in  a//b be confused with root '/' ?

hmeng-19: I would leave it like that if you allow. 
Let me know if you demand me to fix it.

hmeng-19: A easy fix is to scanning from the beginning to end and delete the extra `/`.

btovar: Perhaps it would be better. As I see your code below you do seem to compute the position of '/' often, and tracking a bug of a path that looks like 'a///b//c' might be ugly.

makeflow/src/dag_file.c
```diff
@@ -84,4 +87,32 @@ int dag_file_in_trans( const struct dag_file *f )
...
+			df->source = NULL;
+		}
+		if(df->cache_name) {
+			free(df->cache_name);
+			df->cache_name = NULL;
+		}
+	}
+	list_delete(list);
+
+	if(d->cache_dir) {
```
btovar: From @batrick  I've recently learned that you can do a free(ptr) on an NULL ptr. So these ifs can be removed.

dttools/src/path.c
hmeng-19: `strspn` should cover all the consecutive `/` until the first non-backslash character.
For cleanliness, I'd better fix this just in case.

hmeng-19: Agree. I will fix it. Accept.

makeflow/src/dag_file.c
hmeng-19: Accept.

makeflow/src/makeflow_log.c
```diff
@@ -66,6 +68,16 @@ collected - the number of files were collected in this garbage collection cycle.
 time_spent - the length of time this cycle took.
 total_collected - the total number of files has been collected so far since the start this makeflow execution.
 
+Line format: # MOUNT timestamp target source cache_name type
+
+timestamp - the unix time (in microseconds) when this line is written to the log file.
```
btovar: :)

makeflow/src/makeflow_log.c
btovar: Correct.
(However, if source is NULL, xxstrdup also does a fatal out of memory, which is misleading.)

makeflow/src/mountfile.c
```diff
@@ -0,0 +1,660 @@
+/*
+Copyright (C) 2008- The University of Notre Dame
```
btovar: In the future we may want to move this to dttools, as we can add similar functionality to other tools (e.g., a wq master).

makeflow/src/mountfile.c
```diff
@@ -0,0 +1,660 @@
...
+	}
+	return 0;
+}
+
+/* md5_cal calculates the md5 checksum of string s.
+ * @param s: a string pointer
+ * return the md5 checksum of s on success, return NULL on failure.
+ * The caller should free the returned string.
+ */
+char *md5_cal(const char *s) {
```
btovar: We seem to do this often. Could you move this one to md5.* in dttools (and name it md5_of_string?)

makeflow/src/mountfile.c
btovar: Yes, I thought this is what we concluded last time.

makeflow/src/mountfile.c
btovar: Logging every time seems more foolproof to me.

makeflow/src/makeflow_log.c
hmeng-19: I just realized I forgot to add `CACHE` into the area. Will fix it.

makeflow/src/mountfile.c
hmeng-19: Okay. I need to do some file renaming to match these new files to the naming style of `makeflow. 
We can talk about moving it to dttools after I finish renaming them.

makeflow/src/mountfile.c
hmeng-19: Of course. No problem.

makeflow/src/mountfile.c
hmeng-19: @btovar , what are you referring here? Too many comments there.

makeflow/src/makeflow_log.c
hmeng-19: @btovar , your suggestion is to check where `source` is NULL before calling `xxstrdup(source)`?

makeflow/src/mountfile.c
btovar: To your "Another option ..."

hmeng-19:
@nhazekam , @btovar , thanks for your efforts of reviewing the pull request and giving so many good suggestions.

Here is a modification plan incorporating most of your big suggestions:
1) support makeflow edge cases (thanks Nick for listing all these edge cases out):
      (a) using `--mounts` and `-c` options concurrently with/without an existing log file;
      (b) using `--mounts`, `--cache` and `-c` options concurrently with/without an existing log file;
      (c) running `makeflow --mounts 1.mounts` multiple times without cleanup between different runs;
2) rename source/header files and function/struct/enum names to follow the naming style of makeflow;
3) remove duplicate bachslashes in a path for safety;
4) move the `mounts` cleanup function into makeflow_gc;
5) move `md5_cal` into `dttools/src/md5.*`;
6) replace `LDEBUG` with `debug`;
7) avoid checking the exit code of `xxstrdup`;
8) modify the copyright starting year in the source files;

Excpet for these changes, all the other tiny changes we agreed on in this thread will also be fixed.

hmeng-19:
@btovar , @nhazekam , I am done with the fixes on my side. Please give it a new look. 

I rebased this PR. The new changes are the following commits:
 
    Fix copyright year          33b91ce
    move md5_cal into dttools/src/md5.h         628f766
    avoid checking the exit code of xxstrdup            e6ddc70
    add dag_mount_clean         5b3a4fd
    move the mounts cleanup function into makeflow_gc           80f3280
    rename mountfile.* to makeflow_mounts.*         1b7d784
    rename source_type to dag_file_source_t         cf379f0
    support makeflow --mounts edge cases …          1588277
    remove duplicate bachslashes in path_concat         031b79d
    replace LDEBUG with debug           9a50390
    stop check whether the target path has symlink          327bd5f

hmeng-19:
@nhazekam , @btovar , to make it easier to resolve the conflicts between this PR and the cctools master, I squashed all the commits in this PR into a single one, https://github.com/cooperative-computing-lab/cctools/pull/1239/commits/0ce138a13510650978b2650ac3d81a3560e09b2d .

hmeng-19:
@nhazekam , @btovar , I rebased this PR and merged the conflicts. PTAL.

btovar:
I'll do that ASAP. QAPLA'!

hmeng-19:
@btovar , what is QAPLA'?

btovar:
Good luck, success, in klingon...

doc/makeflow.html
```diff
@@ -545,6 +545,75 @@ <h2 id="resources">Resources and Categories<a class="sectionlink" href="#resourc
...
+<p>Here is an example of a mountfile:</p>
+<code>curl /usr/bin/curl
+convert ../bin/convert
+data/file1 /home/bob/input1
+1.root http://myresearch.org/1.root
+</code>
+
+<p>To limit the behavoir of a makeflow inside the current working directory, the <tt>target</tt> field should satisfy the following requirements:</p>
+<ul>
+  <li>the target path should be non-existing;</li>
```
btovar: Do you mean:
the target path should not exist; ?

makeflow/src/makeflow.c
```diff
@@ -1605,6 +1631,19 @@ int main(int argc, char *argv[])
 	if (change_dir)
 		chdir(change_dir);
 
+	/* Prepare the input files specified in the mountfile. */
+	if(mountfile && !clean_mode) {
+		/* check the validity of the mountfile and load the info from the mountfile into the dag */
+		printf("checking the consistency of the mountfile ...\n");
+		if(makeflow_mounts_parse_mountfile(mountfile, d)) {
+			fprintf(stderr, "Fails to parse the mountfile (%s)!\n", mountfile);
```
btovar: Failed to parse the mountfile: '%s'.

btovar:
Tried a mountfile like (not including +++++):

++++++
mycurl     /usr/bin/curl
myconvert /usr/bin/convert
++++++

Got:

../src/makeflow  --mount example.mount example.mount.makeflow
parsing example.mount.makeflow...
checking the consistency of the mountfile ...
Fails to parse the mountfile (example.mount)!

doc/makeflow.html
```diff
@@ -545,6 +545,75 @@ <h2 id="resources">Resources and Categories<a class="sectionlink" href="#resourc
...
+  <li>the target path should be relative;</li>
+  <li>the target path should not include any symbolic link; (For example,
```
btovar: any symbolic links;

doc/makeflow.html
```diff
@@ -545,6 +545,75 @@ <h2 id="resources">Resources and Categories<a class="sectionlink" href="#resourc
...
+  if the target path is <tt>a/b/c/d</tt>, then <tt>a</tt>, <tt>b</tt>,
+  <tt>c</tt> should not be symbolic link.) This limitation aims to avoid the makeflow
```
btovar: should not be symbolic links.

doc/makeflow.html
```diff
@@ -545,6 +545,75 @@ <h2 id="resources">Resources and Categories<a class="sectionlink" href="#resourc
...
+  leaving the current working directory through symbolic links.</li>
+  <li>the target path should not include doubledots (<tt>..</tt>) to limit the
+  behavoir of the makeflow within the current working directory.  For example, <tt>a/../b</tt>
```
btovar: to limit makeflow to the current working directory.

doc/makeflow.html
```diff
@@ -545,6 +545,75 @@ <h2 id="resources">Resources and Categories<a class="sectionlink" href="#resourc
...
+  <li>the source path can include doubledots (<tt>..</tt>);</li>
+  <li>the source path should only be regular files, directories, and symbolic links;</li>
+  <li>the source path should not be any ancestor directory of the target to avoid infinite loop;</li>
+</ul>
+
+<p>To execute a makeflow with a mountfile, named <tt>mountfile</tt>, specifying its input dependencies:</p>
+
+<code>makeflow --mounts mountfile example.makeflow</code>
+
+<p>Before starting execute the workflow, Makeflow first parses each line of the mountfile when the <tt>--mounts</tt> option is set, and copies the specified dependency from the location specified by <tt>source</tt> field into a local cache, and then links the <tt>target</tt> to the item under the cache. Makeflow also records the location of the local cache and the info (source, target, filename under the cache dir, and so on) of each dependencies specified in the mountfile into its log.</p>
```
btovar: Before execution,

doc/makeflow.html
```diff
@@ -545,6 +545,75 @@ <h2 id="resources">Resources and Categories<a class="sectionlink" href="#resourc
...
+
+<p>To only cleanup the local cache and all the links created due to the mountfile:</p>
+<code>makeflow -ccache example.makeflow</code>
+
+<p>By default, makeflow would create a unique directory under the current working directory to hold all the dependencies introduced by the mountfile.
+However, Makeflow also allows the user to provide a cache dir through the <tt>--cache cache_dir</tt> option, which can be empty or already has some of the depedencies, or all the dependencies.</p>
+
+<p>To limit the behavoir of a makeflow inside the current working directory, the <tt>cache_dir</tt> should satisfy the following requirements:</p>
+<ul>
+  <li>the cache_dir path should be relative;</li>
```
btovar: relative to what?

doc/man/makeflow.m4
```diff
@@ -37,7 +37,7 @@ execution engine.
 
 SUBSECTION(Commands)
 OPTIONS_BEGIN
-OPTION_TRIPLET(-c, --clean, option)Clean up: remove logfile and all targets. If option [intermediates, outputs] is specified, only indicated files are removed.
+OPTION_TRIPLET(-c, --clean, option)Clean up: remove logfile and all targets. If option [intermediates, outputs, cache] is specified, only indicated files are removed.
```
btovar: If option is one of  intermidiates, otuputs, or cache, then only ....

dttools/src/copy_tree.c
```diff
@@ -0,0 +1,279 @@
...
+#include <unistd.h>
+
+#include "copy_tree.h"
+#include "copy_stream.h"
+#include "create_dir.h"
+#include "debug.h"
+#include "path.h"
+#include "xxmalloc.h"
+
+mode_t default_dirmode = S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH;
```
btovar: In CCTools, we usually define such global variables with static, as:

static mode_t default_dirmode ...

dttools/src/copy_tree.c
btovar: Let's stop at the first failure. The faster we return failure the better.

makeflow/src/makeflow.c
btovar: They should be able to be used concurrently. Please fix this.

doc/makeflow.html
hmeng-19: Yes.

makeflow/src/makeflow.c
hmeng-19: @btovar , this should be fixed with the current PR.

hmeng-19: @btovar , did you test this feature and failed?

btovar: I keep getting:

Fails to parse the mountfile (example.mount)!

But as far as I can tell, my mount file follows your documentation... I'll send them to you by email.

hmeng-19:
@btovar , https://github.com/cooperative-computing-lab/cctools/pull/1239/commits/009cd5d5bf049c65ff925692f32838e7f2f2f19f should fix all your new suggestions. PTAL.

hmeng-19:
@nhazekam , @btovar , I rebased this PR on top of the ccl master, and resolved all the conflicts. PTAL.

hmeng-19:
@nhazekam , I added a new file, `makeflow/example/example.mount.README`, which includes how to test the mount feature.

hmeng-19:
@nhazekam , https://github.com/cooperative-computing-lab/cctools/pull/1239/commits/74cc73397a5572f16a41ab2e1f27dd5a7df5aa42 pointed out the consistency between the target and the usage in the prerequisites of a rule. PTAL.

nhazekam:
I think you are good to squash the commits.

hmeng-19:
@nhazekam , I squashed the commits and rebased it. PTAL.

Issue #1187, opened 2016-03-09
Project: Failover Catalog Server


dthain:
Establish a fail-over scheme for the catalog server, so that the outage of one machine doesn't bring everything to a halt.

Step 1 - Allow the environment variable CATALOG_HOST to contain a list of hostnames.  Let the default value be "catalog.cse.nd.edu;backup-catalog.cse.nd.edu"  Factor out the various places where we update a single catalog via datagram_send into a common module that will send the same packet to a list of servers.

Step 2 - Update catalog_query_create to accept a list of servers, and then query them in order.  If the first one doesn't respond in a reasonable time, move on to the second one.

Step 3 - Optimize the failover case -- if the first one is not responding quickly, then continue to use the backup instead of always retrying the primary.  (Until the backup fails.)

dthain:
@trshaffer do you want to take a try at this one?

trshaffer:
Will all catalog servers in the list need to use the same port?

dthain:
Seems to me that the might each have different ports.
We may need to clean up the API...

trshaffer:
It seems like I've been running into trouble with the networking APIs lately. For this one, it would be also be nice to fall back to other addresses returned by DNS, which I think would tie into some of the IPv6 discussion with @batrick.

dthain:
What sort of trouble?

trshaffer:
Oh, just the host:port a bit ago, and the internal API here will be a little awkward. Should I be looking at minimal changes to make this work, or thinking about e.g. API changes for IPv6 integration?

dthain:
Short answer: only change the API where really necessary.
Let's talk about it today.

dthain:
I think this one is done, right?

trshaffer:
As far as I know

PR #1243, opened 2016-04-05
Backup catalog server

trshaffer:
Here's a first draft of backup catalog server support for #1187. For sending updates or creating queries, everything now takes a string of hosts to try. The default is `catalog.cse.nd.edu;backup-catalog.cse.nd.edu`. Multiple servers are specified in a semicolon delimited list, and each host can optionally have a port specified. For updates, the various places that directly sent datagrams now use a common function that sends to all hosts. Queries are attempted in order, unless a query to a host previously failed, in which that host is tried after other options. I tried to split this PR up into individual commits because changing the catalog interface touches a lot of areas (Makeflow, Work Queue, Chirp, Confuga, etc. ). Also, some catalog stuff is exposed in Work Queue's library interface, so I tried not to break compatibility.

### DIFF
--- a/chirp/src/chirp_server.c
+++ b/chirp/src/chirp_server.c
@@ -88,7 +88,6 @@ static char        address[LINK_ADDRESS_MAX];
 static time_t      advertise_alarm = 0;
 static int         advertise_timeout = 300; /* five minutes */
 static int         config_pipe[2] = {-1, -1};
-static struct      datagram *catalog_port;
 static char        hostname[DOMAIN_NAME_MAX];
 static int         idle_timeout = 60; /* one minute */
 static UINT64_T    minimum_space_free = 0;
@@ -135,16 +134,6 @@ static int space_available(INT64_T amount)
 		errno = ENOSPC;
 		return 0;
 	}
-}
-
-int update_one_catalog(void *catalog_host, const void *text)
-{
-	char addr[DATAGRAM_ADDRESS_MAX];
-	if(domain_name_cache_lookup(catalog_host, addr)) {
-		debug(D_DEBUG, "sending update to %s:%d", (char*) catalog_host, CATALOG_PORT);
-		datagram_send(catalog_port, text, strlen(text), addr, CATALOG_PORT);
-	}
-	return 1;
 }
 
 static void downgrade (void)
@@ -253,7 +242,7 @@ static int update_all_catalogs(const char *url)
 
 	char *message = jx_print_string(j);
 
-	list_iterate(catalog_host_list, update_one_catalog, message);
+	list_iterate(catalog_host_list, (int (*) (void *, const void *)) catalog_query_send_update, message);
 
 	free(message);
 	jx_delete(j);
@@ -2219,7 +2208,6 @@ int main(int argc, char *argv[])
 		opts_write_port_file(port_file, chirp_port);
 
 	starttime = time(0);
-	catalog_port = datagram_create(0);
 	if(manual_hostname) {
 		strcpy(hostname, manual_hostname);
 	} else {
--- a/chirp/src/chirp_server.c
+++ b/chirp/src/chirp_server.c
@@ -242,7 +242,7 @@ static int update_all_catalogs(const char *url)
 
 	char *message = jx_print_string(j);
 
-	list_iterate(catalog_host_list, (int (*) (void *, const void *)) catalog_query_send_update, message);
+	list_iterate(catalog_host_list, (list_op_t) catalog_query_send_update, message);
 
 	free(message);
 	jx_delete(j);
--- a/doc/catalog.html
+++ b/doc/catalog.html
@@ -1,0 +1,82 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
+<html lang="en">
+
+<head>
+<meta http-equiv="Content-Type" content="text/html;charset=utf-8">
+<link rel="stylesheet" type="text/css" href="manual.css">
+<title>Chirp Protocol Version 2</title>
+<style type="text/css">
+.command {
+    margin: 10px;
+    border: 1px solid black;
+    font-weight: bold;
+}
+</style>
+</head>
+
+<body>
+
+<div id="manual">
+<h1>Catalog Servers</h1>
+
+<p style="text-align: right;"><b>Last edited: 12 April 2016</b></p>
+
+<p>To check the status of all of the Chirp servers that are available for use,
+consult the Chirp <a href="http://catalog.cse.nd.edu:9097">storage catalog</a>.
+This web page is a list of all known Chirp servers and their locations.</p>
+
+<h2 id="specifying">Specifying Catalog Servers<a class="sectionlink" href="#specifying" title="Link to this section.">&#x21d7;</a></h2>
+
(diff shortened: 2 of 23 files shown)


dthain:
FYI, `backup-catalog.cse.nd.edu` doesn't exist yet.  I will request that it be a DNS alias for cclweb01 for the time being, and then look for a better place to put it...

chirp/src/chirp_server.c
```diff
@@ -253,7 +242,7 @@ static int update_all_catalogs(const char *url)
 
 	char *message = jx_print_string(j);
 
-	list_iterate(catalog_host_list, update_one_catalog, message);
+	list_iterate(catalog_host_list, (int (*) (void *, const void *)) catalog_query_send_update, message);
```
dthain: Would be clearer to cast to `list_op_t` here.

(I wish C was smart enough to detect type compatibility in function-pointer arguments, but oh well.)

dttools/src/catalog_query.c
```diff
@@ -13,45 +14,147 @@ See the file COPYING for details.
...
+		return NULL;
+		break;
+	}
+	return next ? next + 1 : NULL;
+}
 
-	char *url = string_format("http://%s:%d/query.json", host, port);
-	struct link *link = http_query(url, "GET", stoptime);
-	free(url);
+struct catalog_query *catalog_query_create(const char *hosts, struct jx *filter_expr, time_t stoptime)
```
dthain: This function is getting a bit hairy, can you factor it into one function that queries one host, and a parent function that works on the whole list?

dttools/src/catalog_server.c
```diff
@@ -203,7 +191,7 @@ static void update_all_catalogs(struct datagram *outgoing_dgram)
 	char *text = jx_print_string(j);
 	jx_delete(j);
 
-	list_iterate(outgoing_host_list, update_one_catalog, text);
+	list_iterate(outgoing_host_list, (int (*) (void *, const void *)) catalog_query_send_update, text);
```
dthain: list_op_t

dttools/src/catalog_update.c
```diff
@@ -127,15 +118,11 @@ int main(int argc, char *argv[]) {
 
 	char *text = jx_print_string(j);
 
-	char address[DATAGRAM_ADDRESS_MAX];
-	if (domain_name_cache_lookup(host, address)) {
-		datagram_send(d, text, strlen(text), address, port);
-	} else {
-		fatal("unable to lookup address of host: %s", host);
+	if(catalog_query_send_update(host, text) < 1) {
+		fatal("Unable to send update");
```
dthain: Avoid `fatal()` unless we are really deep into something with no recovery -- it gives a rather disruptive error message.  `fprintf(stderr)` is usually the right thing to do in `main`

dthain:
Overall, it looks pretty good and is a general simplification of the code.  I noted a few cleanups inline.

In the meantime, please update the documentation regarding the catalog server -- it's discussed in several places in makeflow, chirp, and workqueue.  Suggest that you factor out the discussion of how to configure and run a catalog server out of chirp.html and into a new catalog.html, that explains the new list and backup server behavior.  Then, the other manuals can briefly describe the most important details and point the user to the catalog manual for the complete story.

I'm still waiting on the new DNS alias, so we cannot merge yet...

dthain:
Ok, `backup-catalog.cse.nd.edu` is now running, so you can test against that.

dttools/src/catalog_query.c
```diff
@@ -98,53 +120,43 @@ struct catalog_query *catalog_query_create(const char *hosts, struct jx *filter_
...
-
-		if(h->down) {
-			debug(D_DEBUG,"catalog server at %s is back up", h->host);
-			set_first_element(down_hosts);
-			while((n = set_next_element(down_hosts))) {
-				if(!strcmp(n, h->host)) {
-					free(n);
-					set_remove(down_hosts, n);
-					break;
+		struct jx *j = catalog_query_send_query(h->url, stoptime);
```
dthain: Ok, this is much clearer about what is happening.
Now, let's look at the use of stoptime.
Suppose that the client calls `catalog_query_create(hosts,time(0)+60)`
How long will a single attempt of `catalog_query_send_query` work before giving up and allowing another catalog to be tried?

trshaffer: Huh, I had assumed `stoptime` worked as a timeout. Looking at `link.h`, I see that is not the case.

When a stop time is specified, should we divide up the total time among the servers, or find the relative time and let each request take the same? If we divide the time up, a large number of servers could push the time for each request too low for any of them to succeed. Since the link API uses absolute times, if we give each request the relative time specified, callers might be surprised that the query runs past the absolute stoptime.

dthain: Well, the outermost question is easy:

When the user calls `catalog_query_create` with a particular stoptime, that's an absolute value: when `time(0)>stoptime`, the function must return failure.

(BTW, that's why I chose this odd convention -- it's easy to pass stoptime down a chain of functions and apply the same test, even if the functions take some time to execute.  Passing around timeouts gets tricky when functions take non-zero time.)

So, now the question is how to divide that fixed amount of time among the servers.

My first inclination is to say that we should pick some fixed values (say 5 seconds) and try each server for that amount of time, round-robin until the stoptime expires.

But I'm open to other suggestions.

trshaffer: That seems like a pretty reasonable solution. I can add a check to respect `stoptime` and just keep going around the list until then

trshaffer:
Now `catalog_query` gives each server 5 seconds, and round robins around the sorted list. If none of the catalog servers answer, the query will spin until `stoptime`.

dthain:
The code looks pretty good, but the documentation needs some reworking.
Start from the beginning, assuming that the user doesn't know what the catalog is.
Explain what it is used for, how to query it, how to browse it, then describe
any special details relative to Chirp and Work Queue.

doc/catalog.html
```diff
@@ -19,12 +19,16 @@
 <div id="manual">
 <h1>Catalog Servers</h1>
 
-<p style="text-align: right;"><b>Last edited: 12 April 2016</b></p>
+<p style="text-align: right;"><b>Last edited: 29 April 2016</b></p>
+
+<p>Catalog servers function as connection points for tools that need to interact remotely. Tools can send free-form updates to a catalog server, and can query catalog servers to find other hosts matching some criteria. Catalog updates are sent via UDP, and the catalog server exposes an HTTP interface to view status and make queries.</p>
 
 <p>To check the status of all of the Chirp servers that are available for use,
```
dthain: Everything, not just Chirp

doc/catalog.html
```diff
@@ -19,12 +19,16 @@
 <div id="manual">
 <h1>Catalog Servers</h1>
 
-<p style="text-align: right;"><b>Last edited: 12 April 2016</b></p>
+<p style="text-align: right;"><b>Last edited: 29 April 2016</b></p>
+
+<p>Catalog servers function as connection points for tools that need to interact remotely. Tools can send free-form updates to a catalog server, and can query catalog servers to find other hosts matching some criteria. Catalog updates are sent via UDP, and the catalog server exposes an HTTP interface to view status and make queries.</p>
```
dthain: JSON

dthain:
Looks pretty good, please bring the PR up to date, and let's merge it.

trshaffer:
Just rebased off the current `master`

Issue #1231, opened 2016-03-30
#1180 (jx expressions) breaks resource monitor tools


btovar:
Before #1180, we could have many json objects per file. This was not proper json, but it did allow the monitor tools to process the resource summaries quite fast. I could force the many objects in a file to be in an array, but then the whole array would need to be read before the first object can be processed. A single file may contain thousands of summaries, so this is undesirable.

Any thoughts?

dthain:
That sounds like a reasonable use case, and it's pretty similar to the test case `TR_jx.sh` which expects to parse and return one item at a time.  Can you put in a test that breaks?

(The only parsing wrinkle here is that `[a][x]` could indicate two array literals, each containing one item, or an array literal `[a]` followed by an array lookup of item in position x.  This is solved by the hack of adding a semicolon between the two.)

dthain:
Is this one solved?

trshaffer:
The above test is passing. As far as I know, this is working in @btovar's use case

PR #1246, opened 2016-04-07
Add jx_parser_yield to support streaming JSON parsing

trshaffer:
This is in reference to #1231

Since JX expressions allows lookups, the parser has to scan beyond the end of the object/array, and then unscan (if there's no lookup). Since the library JX supports parsing TCP streams directly, seeking back isn't really an option, so to unscan, the parser stores the extra character and uses it on the next scan. Unfortunately for @btovar's use case, `jx_parse_stream` automatically creates and destroys the parser, throwing out the unscanned character and getting the stream out of sync on subsequent calls, hence the garbled test output.

I added a new function, `jx_parser_yield`, that parses a single value and leaves the parser intact for subsequent calls. If someone needs to do streaming JX parsing, I think it's not unreasonable that they dig around in the lower-level functions a bit, so I didn't touch the convenience wrappers (e.g. `jx_parse_stream`). I've updated `jx_count_obj_test.c` to use the new streaming interface, and now the corresponding test passes.

### DIFF
--- a/dttools/src/jx_count_obj_test.c
+++ b/dttools/src/jx_count_obj_test.c
@@ -32,7 +32,9 @@ int main(int argc, char **argv) {
 	int count = 0;
 
 	struct jx *j;
-	while((j = jx_parse_stream(stream))) {
+	struct jx_parser *p = jx_parser_create(0);
+	jx_parser_read_stream(p, stream);
+	while((j = jx_parser_yield(p))) {
 		char *str = jx_print_string(j);
 		fprintf(stdout, "%s\n", str);
 
@@ -41,6 +43,7 @@ int main(int argc, char **argv) {
 
 		count++;
 	}
+	jx_parser_delete(p);
 
 	if(count == n) {
 		exit(0);
--- a/dttools/src/jx_parse.c
+++ b/dttools/src/jx_parse.c
@@ -658,6 +658,17 @@ static struct jx * jx_parse_finish( struct jx_parser *p )
 	return j;
 }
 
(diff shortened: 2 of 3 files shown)


dthain:
Aha, good diagnosis of the problem.  But perhaps a simpler solution is to indicate that repeated use of `jx_parse_stream()` on a single stream is incorrect.  I think the correct approach is `jx_parser_create(); jx_parse(); jx_parser_delete();` right?

@btovar

trshaffer:
The helper functions go though jx_parse_finish, which does some extra error checking. We could just require that error checking manually, but I think having a yield function makes for cleaner code, since e.g. you can use it as the condition for a while loop, as in the test
On Apr 7, 2016 16:42, Douglas Thain <notifications@github.com> wrote:Aha, good diagnosis of the problem.  But perhaps a simpler solution is to indicate that repeated use of jx_parse_stream() on a single stream is incorrect.  I think the correct approach is jx_parser_create(); jx_parse(); jx_parser_delete(); right?

@btovar

btovar:
I like that yield does the error checking. 

I find the name of parse_stream confusing, though. It seems to indicate to me a stream of json objects, which is what yield does.

dthain:
If you both like the solution, then go ahead.  :)

`parse_stream` in that context refers to the stdio stream it is reading from, like `parse_string` reads from a string and `parse_file` reads from a file name.  Maybe there is a clearer naming scheme.

btovar:
Thanks @trshaffer!

PR #1257, opened 2016-04-11
Report unmatched tasks

btovar:
Adds -A to work_queue_status, so that the number of workers that can run tasks, per category, are reported. E.g.:

work_queue_status -A localhost 9001
````
CATEGORY          WAITING  FIT-WORKERS    MAX-CORES   MAX-MEMORY     MAX-DISK

my_category             1            0            4          ???          ???

default                 0            0          ???          ???          ???

manual-label            1            0            5          ???          ???
````

@klannon, is this what you were thinking about?

Also available with the API, the value of FIT-WORKERS is stats.workers_able, as whole and per category.


btovar:
Cleaning... Do not merge yet.

Issue #1250, opened 2016-04-11
JX library failure


trshaffer:
While trying to run `chirp_status`, on the current master I get an error.

    [tshaffe1@cclws19 src]$ ./chirp_status  -d all
    2016/04/11 12:35:48.31 chirp_status[28267] debug: ./chirp_status version 6.0.0 [master:8b2b9d1c] (released 2016-04-11 08:00:51 -0400)
    2016/04/11 12:35:48.31 chirp_status[28267] debug: Built by tshaffe1@cclws19 on 2016-04-11 12:26:11 -0400
    2016/04/11 12:35:48.31 chirp_status[28267] debug: System: Linux cclws19 2.6.32-431.17.1.el6.x86_64 #1 SMP Fri Apr 11 17:27:00 EDT 2014 x86_64 GNU/Linux
    2016/04/11 12:35:48.31 chirp_status[28267] debug: Configuration: --strict
    2016/04/11 12:35:48.31 chirp_status[28267] http: connect catalog.cse.nd.edu port 9097
    2016/04/11 12:35:48.31 chirp_status[28267] dns: looking up name catalog.cse.nd.edu
    2016/04/11 12:35:48.31 chirp_status[28267] dns: catalog.cse.nd.edu is 129.74.152.166
    2016/04/11 12:35:48.31 chirp_status[28267] tcp: connecting to 129.74.152.166:9097
    2016/04/11 12:35:48.31 chirp_status[28267] tcp: made connection to 129.74.152.166:9097
    2016/04/11 12:35:48.31 chirp_status[28267] http: GET http://catalog.cse.nd.edu:9097/query.json HTTP/1.1
    Connection: close
    Host: catalog.cse.nd.edu
    User-Agent: Mozilla/5.0 (compatible; CCTools 6.0.0 [master:8b2b9d1c] Parrot; http://ccl.cse.nd.edu/)
    2016/04/11 12:35:48.32 chirp_status[28267] http: HTTP/1.1 200 OK
    2016/04/11 12:35:48.32 chirp_status[28267] http: Date: Mon Apr 11 12:35:48 2016
    2016/04/11 12:35:48.32 chirp_status[28267] http: Server: catalog_server
    2016/04/11 12:35:48.32 chirp_status[28267] http: Connection: close
    2016/04/11 12:35:48.32 chirp_status[28267] http: Access-Control-Allow-Origin: *
[...]

PR #1266, opened 2016-04-14
read doubles with e,E notation

btovar:
Fix for #1250.


trshaffer:
Works for me
On this PR, floats and that Catalog output are parsing correctly.

dthain:
I like your approach, it's subtly tail-recursive.

PR #1272, opened 2016-04-18
Handle symbolic links within Chirp root.

### LINKED ISSUE #380: [Proposal] Chirp Symlink Handling
# Chirp Symbolic Link handling

Chirps handling of symbolic links is currently broken. This issue discusses how
symlink can be used and how we would like symlink to function.  It also
includes possible solutions to achieve desired behavior.

`int symlink(const char *target, const char *linkname);`

`target` is some data that usually must be less than PATH_MAX (4096).
`linkname` is the path to the new link.

`target` can be many things, some interpretations do not allow others:
* An absolute link to some other path in `--root`.
  e.g. `/a -> /b`
* A relative link to some other path in `--root`.
  e.g. `a -> b` or `a -> ../b`
* A relative/absolute link to some other path outside of `--root`:
  e.g. `/bxgrid -> /data/bxgrid` or `/bxgrid -> ../bxgrid`
  (where `/data/chirp` is `--root`)
* A path/datum interpreted by the client, such as Parrot.
  e.g. `null -> /dev/null`
  The application might make a symlink to `/dev/null` and then want to use
  that link later on in the application.

## Desired Behavior

Loosely speaking, ideal behavior would be:
* During any pathname resolution, all symlinks under `--root` are interpreted as if Chirp were chroot to `--root`.
* `target` is returned unmodified by `readlink`.

This behavior is not easily obtainable though. So we have these goals instead:

### Requirements:
* Whatever the user sets for `target` MUST be returned by `readlink`. Any other
  behavior would be unacceptable where the client is using Chirp in conjunction
  with other file systems.

### Really nice to have:
* The server does not need to resolve pathname components to check for links.
  For example, `open /a/b` does not require checking if `a` is a symlink and
  verifying it points within `--root`.
* Path resolution of symlinks by the kernel (during e.g. chirp_fs_local_open)
  does not allow leaving `--root`. That is, the user can't construct a symlink
  that allows them to escape `--root`.

### Nice to have:
* Path resolution of absolute symlinks by the kernel use `--root`, not the
  server file system `/`.

## Solution 1a

[Chirp prior to commit 25d6adb0288ecb13c2c826fe2f1af355768fb59b used this
solution.] All `symlink` calls store `target` literally without modification.

### Problems:
* Normal (i.e. kernel) path resolution won't work with absolute symlink
  targets.
* The user can easily escape `--root`. Protections completely rely ACLs to protect
  paths outside of `--root`.

### Benefits:
* `target` is unmodified.

## Solution 1b

All `symlink` calls store `target` literally without modification. Any operation
which requires pathname resolution will require checking each component for
a symlink which must be resolved.

### Problems:
* Normal (i.e. kernel) path resolution won't work with absolute symlink
  targets. [Therefore, the Chirp server must handle all pathname resolution.]
* Slows down any operation that requires pathname resolution. Requires a
  call to `readlink` for each pathname component.

### Benefits:
* ALL relative/absolute symlinks are resolved by the Chirp server, so we can
  easily ensure that the user can't escape `--root`. Links created outside of
  Chirp CANNOT point outside of `--root`.
* `target` is unmodified --> any "magic" in symlink handling does not leak into the file system.

## Solution 2

All `symlink` calls with an absolute `target` prepend `--root`. Calls to
`readlink` remove `--root`. [This solution is partially implemented in master.
`readlink` does not remove `--root`.]

### Problems:
* Links created outside of Chirp CAN point outside of `--root`.
* All links become invalid if any pathname component of `--root` is changed.
* Relative `target` can point outside of `--root`.

### Benefits:
* Less runtime overhead.
* Kernel path resolution can be used.

## Solution 3

All `symlink` calls with an absolute `target` prepend `.__root`. This
is another per-directory link managed by the Chirp server which points
to the actual root. This can be expressed with a relative path. For
example:
````
.__root  ->  ../../
a        ->  .__root/b
````

### Problems:
* Links created outside of Chirp CAN point outside of `--root`.
* Some (?) links become invalid if `--root` is changed to a
  sub/super-directory.  e.g. `--root=super/root` or `--root=root/sub`
* Relative links can point outside of `--root`.

### Benefits:
* Kernel path resolution can be used.

## Solution 4

Have two symlinks created for each `chirp_symlink`. One is used for pathname
resolution, is always relative, and never points outside of `--root` (at least,
at the time of creation). The other symlink is the raw `target` and is returned
on `readlink`. Possible example:

````
.__link:a    ->  /b
a            ->  ../b
````

### Problems:
* Links created outside of Chirp CAN point outside of `--root`.
* Chirp must ensure paths do not include a component with a `.__` prefix.
  (Otherwise, you can use `.__link:a` to escape `--root`.)
* Each symlink requires two files. Unlink must also remove `.__link:...`.

### Benefits:
* Some (?) links become invalid if `--root` is changed to a
  sub/super-directory.  e.g. `--root=super/root` or `--root=root/sub`
* All Chirp-created symlinks cannot escape `--root` (unless `-root` changes, as noted above).

### PR DESCRIPTION
**Pull request by batrick**
This is solution 1b of issue #380.

Each operation involving a path is decomposed into components which are traversed. All symlinks are handled by Chirp without the kernel's help.

Fixes #380.

### DIFF
--- a/dttools/src/create_dir.h
+++ b/dttools/src/create_dir.h
@@ -10,13 +10,15 @@ See the file COPYING for details.
 
 /** @file create_dir.h Create a new directory recursively. */
 
+#include <sys/stat.h>
+
 /** Create a new directory recursively.
 @param path The full path of a directory.  It is not necessary for all components of the path to exist.
 @param mode The desired unix mode bits of the directory and parents.
 @return One on success, zero on failure.
 */
 
-int create_dir(const char *path, int mode);
+int create_dir (const char *path, mode_t mode);
 
 /** Create needed parent directories of a file or directory.
 @param path The full path of a file or directory.
@@ -24,6 +26,8 @@ int create_dir(const char *path, int mode);
 @return One on success, zero on failure.
 */
 
-int create_dir_parents(const char *path, int mode);
+int create_dir_parents (const char *path, mode_t mode);
 
 #endif
+
+/* vim: set noexpandtab tabstop=4: */
--- a/dttools/src/mkdir_recursive.h
+++ b/dttools/src/mkdir_recursive.h
@@ -7,6 +7,8 @@ See the file COPYING for details.
 #ifndef MKDIR_RECURSIVE_H
 #define MKDIR_RECURSIVE_H
 
+#include "compat-at.h"
+
 #include <fcntl.h>
 #include <sys/stat.h>
 
--- a/dttools/src/mkdir_recursive.h
+++ b/dttools/src/mkdir_recursive.h
@@ -1,0 +1,50 @@
+/*
+Copyright (C) 2015- The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file COPYING for details.
+*/
+
+#ifndef MKDIR_RECURSIVE_H
+#define MKDIR_RECURSIVE_H
+
+#include <fcntl.h>
+#include <sys/stat.h>
+
+/** @file mkdir_recursive.h Create a new directory recursively. */
+
+/** Create a new directory recursively.
+@param fd The directory path is relative to.
+@param path The full path of a directory.  It is not necessary for all components of the path to exist.
+@param mode The desired unix mode bits of the directory and parents.
+@return -1 on failure, 0 on success.
+*/
+
+int mkdirat_recursive(int fd, const char *path, mode_t mode);
+
+/** Create a new directory recursively.
+@param path The full path of a directory.  It is not necessary for all components of the path to exist.
+@param mode The desired unix mode bits of the directory and parents.
+@return -1 on failure, 0 on success.
+*/
+
+int mkdir_recursive(const char *path, mode_t mode);
+
+
+/** Create needed parent directories of a file or directory.
+@param fd The directory path is relative to.
+@param path The full path of a file or directory.
+@param mode The desired unix mode bits of the parent directories.
+@return -1 on failure, 0 on success.
+*/
+
+int mkdirat_recursive_parents(int fd, const char *path, mode_t mode);
+
+/** Create needed parent directories of a file or directory.
+@param path The full path of a file or directory.
+@param mode The desired unix mode bits of the parent directories.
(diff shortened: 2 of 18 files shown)

### CODE REVIEW

**Comment by batrick** [comment]
Nearly done WIP. Comments welcome.

**Comment by dthain** [comment]
How portable is `openat()` and friends?

**Comment by batrick** [comment]
It's POSIX.1-2008. The system call first appeared in Linux 2.6.16 (so pre RHEL5). It's also supported on Darwin and the BSDs.

**Comment by batrick** [comment]
@dthain if you're cool with this, I think it's ready to merge now.

File: dttools/src/create_dir.h [hunk 1]
```diff
@@ -10,20 +10,32 @@ See the file COPYING for details.
...
+#include <sys/stat.h>
+
 /** Create a new directory recursively.
 @param path The full path of a directory.  It is not necessary for all components of the path to exist.
 @param mode The desired unix mode bits of the directory and parents.
 @return One on success, zero on failure.
 */
 
-int create_dir(const char *path, int mode);
+static inline int create_dir (const char *path, mode_t mode)
```
dthain: Inlines and macros tend to cause confusion.

Unless there is some really good performance reason, let's just make these
ordinary functions that invoke the new functions.

(e.g. old binaries will still link correctly.)

dttools/src/mkdir_recursive.h
```diff
@@ -0,0 +1,34 @@
...
+
+/** Create a new directory recursively.
+@param fd The directory path is relative to.
+@param path The full path of a directory.  It is not necessary for all components of the path to exist.
+@param mode The desired unix mode bits of the directory and parents.
+@return -1 on failure, 0 on success.
+*/
+
+int mkdirat_recursive(int fd, const char *path, mode_t mode);
+#define mkdir_recursive(path,mode) mkdirat_recursive(AT_FDCWD,path,mode)
```
dthain: Avoid macros unless there is a really solid reason.
Make these real functions that invoke the underlying ones.

dthain:
General approach is good, but pls clean up some of the macro use.

Issue #1287, opened 2016-04-26
http request format error


hmeng-19:
Problem:
Currently, the http request constructed by the `http_query_size_via_proxy` function inside `dttools/src/http_query.c` always set the `Request-URI` of the http request to be `absoluteURI`, such as `http://ccl.cse.nd.edu/research/data/4_cubes.pov`.
This would fails with a 403 (Forbidden) http response when no http proxy is used. 

Solution:
According to the specification of the http protocol 5.1.2 (https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html), the `Request-URI` field should be `absoluteURI` when a http proxy is used.
When there is no http proxy to be used, the `Request-URI` field should be `abs_path` (such as `/research/data/4_cubes.pov`).

hmeng-19:
Please see the PR #1288 .

hmeng-19:
Fixed by #1288 .

PR #1288, opened 2016-04-26
Http request format fix

hmeng-19:
change the `Request-URI` field of http requests:
 When there is no proxy to be used, the `Request-URI` field should be `abs_path`.

### DIFF
--- a/dttools/src/Makefile
+++ b/dttools/src/Makefile
@@ -115,6 +115,7 @@ HEADERS_PUBLIC = \
 	category.h \
 	debug.h \
 	hash_table.h \
+	http_query.h \
 	int_sizes.h \
 	jx.h \
 	link.h \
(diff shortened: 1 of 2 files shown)


hmeng-19:
This PR aims to fix #1287 .

hmeng-19:

hmeng-19:
@btovar , how do you think about this?

hmeng-19:
It looks that autobuild can not send us email notification.
The record here http://ccl.cse.nd.edu/software/autobuild/ shows it should not have any bit damage to the code.

btovar:
I asked suggest@cse about the emails. They have not responded yet. The error is:

   ----- Transcript of session follows -----
<btovar@nd.edu>... Deferred: alt4.aspmx.l.google.com.: No route to host
Warning: message still undelivered after 4 hours
Will keep trying until message is 5 days old

btovar:
Rather than goto, you can use two scanf's in succession, or better yet, use the functions that Patrick added in pattern.h.

hmeng-19:
@btovar , thanks for pointing out.
I double checked the code and noticed that `goto` is not really necessary here. So I just removed it.

Let me know whether this is ready to merge.

btovar:
It looks good to me.

Issue #1286, opened 2016-04-25
Look into using GCC's memory leak detection during build tests


batrick:
Added in [GCC v4.8](https://gcc.gnu.org/gcc-4.8/changes.html). See also this [blog post](http://lemire.me/blog/2016/04/20/no-more-leaks-with-sanitize-flags-in-gcc-and-clang/).

trshaffer:
Once I fixed a small leak in `resource_monitor/src/piggybacker.c`, I could compile cctools with those checks. For `makeflow example.makeflow`

    =================================================================
    ==10039==ERROR: LeakSanitizer: detected memory leaks

    Direct leak of 192 byte(s) in 6 object(s) allocated from:
        #0 0x7f72f836449a in __interceptor_malloc /build/gcc/src/gcc-5-20160209/libsanitizer/asan/asan_malloc_linux.cc:38
        #1 0x4faeb4 in cctools_list_create cctools/dttools/src/list.c:16
        #2 0x4fdf8c in list_duplicate cctools/dttools/src/list.c:375
        #3 0x404f9b in makeflow_generate_output_files cctools/makeflow/src/makeflow.c:152
        #4 0x40885a in makeflow_node_complete cctools/makeflow/src/makeflow.c:647
        #5 0x40a5c2 in makeflow_run cctools/makeflow/src/makeflow.c:830
        #6 0x40e012 in main cctools/makeflow/src/makeflow.c:1590
        #7 0x7f72f701e70f in __libc_start_main (/usr/lib/libc.so.6+0x2070f)

    Direct leak of 192 byte(s) in 6 object(s) allocated from:
        #0 0x7f72f836449a in __interceptor_malloc /build/gcc/src/gcc-5-20160209/libsanitizer/asan/asan_malloc_linux.cc:38
        #1 0x4fbd02 in new_node cctools/dttools/src/list.c:120
        #2 0x4fc79c in cctools_list_push_tail cctools/dttools/src/list.c:190
        #3 0x4fe042 in list_duplicate cctools/dttools/src/list.c:378
        #4 0x404f9b in makeflow_generate_output_files cctools/makeflow/src/makeflow.c:152
        #5 0x4074b1 in makeflow_node_submit cctools/makeflow/src/makeflow.c:474
        #6 0x407e8d in makeflow_dispatch_ready_jobs cctools/makeflow/src/makeflow.c:574
        #7 0x40a43c in makeflow_run cctools/makeflow/src/makeflow.c:817
[...]

PR #1289, opened 2016-04-26
Add option to compile with -fsanitize=address and -fsanitize=undefined

trshaffer:
This is for #1286.

I added a new option to the `configure` script to compile with these options. N.B. that GCC>=4.8 is required. If configured as

    ./configure --sanitize

all programs will exit with non-zero status and print details in the case of leaks. The numerous changed files are due to test scripts that compile their own code. Extra linker flags are required for those tests, so I passed them through an environment variable. I'm not sure if this is the most correct way to do it, though

### DIFF
--- a/configure
+++ b/configure
@@ -245,6 +245,7 @@ do
 		--debug)
 			optdebug=1
 			optstrict=1
+			optsanitize=1
 			;;
 		--prefix)
 			install_path=$(abspath "${arg}")
--- a/configure
+++ b/configure
@@ -231,7 +231,7 @@ do
 			opt=${opt%%=*}
 			;;
 			# empty arg only for these options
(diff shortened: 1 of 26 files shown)


trshaffer:
@nhazekam just pointed out that `-fsanitize=undefined` seems to be unsupported on GCC 4.8.2

batrick:
Go ahead and turn on `--sanitize` as well for `--debug`, in `configure`.

nhazekam:
@btovar Was this merge ready? I think we had concerns about compiler versions compatibility.

nhazekam:
@trshaffer @btovar can we revert and have Tim rollback to c199604 instead of 5098db9. This would not add sanitize to the debug options in config, the user has to explicitly add -fsanitize.

PR #1294, opened 2016-04-27
if task cannot run in the worker anymore, forsake it

btovar:
This is needed for when the worker is launched without the --memory option, but a task has explicit memory requirements. If the free available memory of the system is consumed by some other process the task never got to run in the worker, and it was never returned to the master.


btovar:
We would like something similar for foremen, but I am still thinking about a clean way to do it.

PR #1269, opened 2016-04-15
Wq category no info

btovar:
Use largest resources seen in workers for categories that do not define a maximum allocation.

Experimental, do not merge yet.

PR #1305, opened 2016-05-02
Wq alloc by max seen

btovar:
Part 2 of #1269.

update first allocation using max worker/max seen per category.

First commit is: 
 	Compute allocation every n tasks. 			d3b2be5


dthain:
Looks reasonable, make sure to test for memory leaks...

PR #1306, opened 2016-05-02
Wq transactions log

btovar:
Part 3 of #1269.

First commit is:
 	Add tasks transactions log. 			1523bfa


dthain:
The approach looks good to me.

PR #1307, opened 2016-05-03
Remove unused is_seekable logic.

batrick:
All services report files as seekable. The local service no longer opens
unseekable files since the introduction of native file descriptors.

This resolves an issue privately reported by Benjamin Wynne. In his case, an
attached fd during the bootstrap phase was wrongly marked as unseekable. The
fix for that is obviated by this commit which simply removes all the unused
seekable logic.

### DIFF
--- a/parrot/src/pfs_service_local.cc
+++ b/parrot/src/pfs_service_local.cc
@@ -91,17 +91,13 @@ class pfs_file_local : public pfs_file
 {
 private:
 	int fd;
-	int is_a_pipe;
-	int is_a_named_pipe;
 	pfs_off_t last_offset;
 
 public:
-	pfs_file_local( pfs_name *name, int f, int i ) : pfs_file(name) {
+	pfs_file_local( pfs_name *name, int f ) : pfs_file(name) {
 		assert(f >= 0);
 		fd = f;
 		last_offset = 0;
-		is_a_pipe = i;
-		is_a_named_pipe = -1;
 	}
 
 	virtual int canbenative (char *path, size_t len) {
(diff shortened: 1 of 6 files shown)


dthain:
Hold on, I think you are missing something important here.

`pfs_service_{http|ftp}::is_seekable()` return zero, because they are streaming protocols and the files must be read sequentially.

`pfs_table::open_object` calls `is_seekable` to determine if the cache should be used.  That is, an open on a non-seekable service requires that the file be downloaded and placed into the cache, so that it can become seekable.

Without that capability, then `lseek` on a file obtained by HTTP/FTP  fails in strange and silent ways.

batrick:
Oh, I missed that some services were relying on default behavior provided by the `pfs_service` class. Okay, I'll go back and fix that...

batrick:
How about this one @dthain.

parrot/src/pfs_service_local.cc
```diff
@@ -153,25 +149,6 @@ process to sleep and wait for actual input to become ready.
 		result = ::read(fd,data,length);
 		if(result>0) last_offset = offset+result;
 
-		if(result==0 && offset==0) {
```
dthain: The reason for this roundabout code is explained by the comment above.  Does that reason no longer apply?

parrot/src/pfs_service_local.cc
```diff
@@ -326,7 +303,7 @@ process to sleep and wait for actual input to become ready.
 
 	virtual int is_seekable()
 	{
-		return !is_a_pipe;
```
dthain: This flag existed so that one gets a proper errno when attempting to seek on a pipe.  Does that issue no longer come up?

parrot/src/pfs_service_local.cc
batrick: Right, named pipes are no longer opened as Parrot files. Instead, they are opened as native file descriptors. That logic is handled here:

https://github.com/cooperative-computing-lab/cctools/blob/89b41ac35584347a24baa4c3a42be29e184309f0/parrot/src/pfs_service_local.cc#L103-L110

parrot/src/pfs_service_local.cc
batrick: That issue no longer exists. All pipes/sockets/fifos are accessed via native file descriptors.

PR #1315, opened 2016-05-04
Wq factory autosize fix

btovar:
Compute disk size in kB correctly for condor.
Use same default values as condor_submit_workers
worker has the same size as TARGET.


btovar:
In a further pr I'll add the default resources to ./configure, with corresponding documentation.

PR #1319, opened 2016-05-05
Reclaim space used by replicas of deleted files.

batrick:


batrick:
@dthain This branch is done and builds/tests okay in autobuild. Shall I merge?

Issue #1303, opened 2016-05-02
Make Catalog More Responsive


dthain:
Under certain conditions, the catalog server gets blocked up and doesn't respond to queries.  here is what's going on:

The catalog server works by accepting updates via udp, which is fast and non blocking.  For queries, it forks to service the request.  This makes it easy to give the client a snapshot of the catalog state as it existed when the query arrived, since we don't have to worry about updates.   However, we don't want an infinite number of catalog processes running at once, so the server has an upper limit of 50 query processes running.  When that limit is reached, it stops accepting connections.  As a result, queries are delayed until another one finishes.

Now, what's happening is that we are getting a bunch of queries from off campus.  For whatever reason, the queries are issued, but then are slow reading all of the data off of the TCP stream.  So, the catalog server process sits there, trying to fulfill the request, and the next thing you know, we have 50 blocked processes.

Currently, the catalog server tries forever to service the request.  What we need to do in the query handler is to set an alarm for, say, 15 seconds.  If the client cannot pull the data off the stream and give up.

Also, for debugging purposes, it would be nice if the catalog server would change its process title to "catalog_server [128.x.y.z]" so we can see which process is handling which client.  (See change_process_title.h)

dthain:
I think this one is done, right?

trshaffer:
Yep, this was already implemented

PR #1322, opened 2016-05-05
Change command name on catalog_server query processes

trshaffer:
For #1303

I added a bit of code so that query prosesses' command lines to show the remote address. As to the query timeout, there's already a `--timeout` option implemented. I tested it with a local server instance and `nc` to simulate a slow client. It does indeed end query processes after a fixed amount of time (60 s by default). Is this option set too high for the production catalog server?


dthain:
Aha, you are right about the timeout; I didn't look at the code before requesting the change.
And 60s timeout is too high, I have lowered it to 5s on the production server, and let's see if that helps.
I'll merge as soon as this passed autobuild,.

PR #1323, opened 2016-05-05
Default resources in ./configure  

btovar:
Sets default resources for work_queue_worker, work_queue_factory, and the submission scripts.

default_cores=1
default_memory=1024  # MB, e.g., 1 GB
default_disk=1024    # MB, e.g., 1 GB

Modify with something like: 

./configure --default-cores 4 --default-memory 2048 --default-disk 10240


dthain:
I must have misunderstood our previous discussion.  I thought you were setting a default resource consumption for *tasks*, not for *workers*.

Wouldn't it make more sense to give tasks a default size, and for the worker to, by default, measure its surroundings?

btovar:
Setting default resources per worker we make the surroundings less brittle. The problem with measuring surroundings is that such surroundings change during the lifetime of the worker. For cores and memory this may work ok, but for disk not so much. Two workers running on the same node step on each other. 

Even with Nate's allocations, we would not want to reserve a big chunk of the disk because it was free. By default the workers write to /tmp, which we do not want to fill up.

Both tasks and workers could go hand in hand. Tasks by default use 1 core, 1GB of memory and 1GB disk. By default, workers are created to run one such task.

PR #1236, opened 2016-04-01
Prune

pivie:
This branch has at least some of the whitespace fixed.

PR #1338, opened 2016-05-18
Prune with Umbrella HEP example (not WQ). Execute in client mode.

pivie:
HEP is example works again using Umbrella (but not with Umbrella and Work Queue yet).
Python client can execute workflows so that the final export can exist in the workflow script.


pivie:
This deprecates #1236

dthain:
Merge this one?

PR #1340, opened 2016-05-19
Tmp dir per worker

btovar:
Adds cache/tmp,  pointed to with environment variable WORKER_TMPDIR.

Keeping tmpdirs per task broke parrot cache sharing in lobster. The idea is for tasks to read WORKER_TMPDIR and set parrot paths accordingly.


dthain:
No good deed goes unpunished.

PR #1342, opened 2016-05-19
Makeflow JSON import

trshaffer:
Here's a first pass at a parser for the Makeflow JSON format (#1178). I added the `--json` flag to Makeflow to switch to the JSON format. I made some adjustments to the output format as well. If I dump JSON from `example.makeflow`, I get

    {
      "default_category":"default",
      "categories":
        {
          "default":
            {
              "environment":
                {
                  "CONVERT":"/usr/bin/convert",
                  "CURL":"/usr/bin/curl",
                  "URL":"http://ccl.cse.nd.edu/images/capitol.jpg",
                  "MAKEFLOW_INPUTS":"",
                  "MAKEFLOW_OUTPUTS":"capitol.montage.gif"
                }
            }
        },
      "rules":
        [
          
          {
            "command":"/usr/bin/convert -delay 10 -loop 0 capitol.jpg capitol.90.jpg capitol.180.jpg capitol.270.jpg capitol.360.jpg capitol.270.jpg capitol.180.jpg capitol.90.jpg capitol.montage.gif",
            "allocation":"first",
            "outputs":
              [
                "capitol.montage.gif"
              ],
            "inputs":
              [
                "capitol.jpg",
                "capitol.90.jpg",
                "capitol.180.jpg",
                "capitol.270.jpg",
                "capitol.360.jpg"
              ],
            "category":"default",
            "local_job":false
          },
          
          {
            "command":"/usr/bin/convert -swirl 90 capitol.jpg capitol.90.jpg",
            "allocation":"first",
            "outputs":
              [
                "capitol.90.jpg"
              ],
            "inputs":
              [
                "capitol.jpg"
              ],
            "category":"default",
            "local_job":false
          },
          
          {
            "command":"/usr/bin/convert -swirl 180 capitol.jpg capitol.180.jpg",
            "allocation":"first",
            "outputs":
              [
                "capitol.180.jpg"
              ],
            "inputs":
              [
                "capitol.jpg"
              ],
            "category":"default",
            "local_job":false
          },
          
          {
            "command":"/usr/bin/convert -swirl 270 capitol.jpg capitol.270.jpg",
            "allocation":"first",
            "outputs":
              [
                "capitol.270.jpg"
              ],
            "inputs":
              [
                "capitol.jpg"
              ],
            "category":"default",
            "local_job":false
          },
          
          {
            "command":"/usr/bin/convert -swirl 360 capitol.jpg capitol.360.jpg",
            "allocation":"first",
            "outputs":
              [
                "capitol.360.jpg"
              ],
            "inputs":
              [
                "capitol.jpg"
              ],
            "category":"default",
            "local_job":false
          },
          
          {
            "command":"/usr/bin/curl -o capitol.jpg http://ccl.cse.nd.edu/images/capitol.jpg",
            "allocation":"first",
            "outputs":
              [
                "capitol.jpg"
              ],
            "inputs":
              [
              ],
            "category":"default",
            "local_job":true
          }
        ]
    }

Although this runs, it's picked up a some cruft on its way through Makeflow. I put a [cleaner example JSON file](https://github.com/trshaffer/cctools/blob/5adf4c6fea99d727dbbd72b9879637ac62925517/makeflow/example/example.json) in the PR. In particular, per-category and per-rule environment variables seem to be working. Some outstanding things:

- We might want to special case the `"environment"` key on the DAG object. As in the example, global variables have to be somewhat awkwardly specified on a category. If we add top-level `"environment"` to the default category, it's possible to specify rules and environment variables without (visibly) using categories at all.
- I didn't include `MAKEFLOW_INPUTS` and `MAKEFLOW_OUTPUTS`. They don't really fit as variables, but Make syntax doesn't have a lot of options. In JSON, should we put `"inputs"` and `"outputs"` keys on the top-level object? This is consistent with the convention used within the rules, reflects their use as inputs/outputs of the DAG as a whole, and lends itself to better structure than string variables.
- JX expressions are not yet implemented. I've been keeping the `"variables"` key to use as context for that.

### DIFF
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -442,6 +442,20 @@ struct jx  *jx_copy( struct jx *j )
 	return 0;
 }
 
+struct jx *jx_merge(struct jx *j, ...) {
+	va_list ap;
+	va_start (ap, j);
+	struct jx *result = jx_object(NULL);
+	for (struct jx *next = j; jx_istype(next, JX_OBJECT); next = va_arg(ap, struct jx *)) {
+		for (struct jx_pair *p = next->u.pairs; p; p = p->next) {
+			jx_delete(jx_remove(result, p->key));
+			jx_insert(result, jx_copy(p->key), jx_copy(p->value));
+		}
+	}
+	va_end(ap);
+	return result;
+}
+
 int jx_pair_is_constant( struct jx_pair *p )
 {
 	return jx_is_constant(p->key)
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -183,6 +183,16 @@ jx_int_t jx_lookup_integer( struct jx *object, const char *key )
 	}
 }
 
+int jx_lookup_boolean( struct jx *object, const char *key )
+{
+	struct jx *j = jx_lookup(object,key);
+	if(j && jx_istype(j,JX_BOOLEAN)) {
+		return !!j->u.boolean_value;
+	} else {
+		return 0;
+	}
+}
+
 double jx_lookup_double( struct jx *object, const char *key )
 {
 	struct jx *j = jx_lookup(object,key);
--- a/dttools/src/jx.h
+++ b/dttools/src/jx.h
@@ -279,7 +279,7 @@ struct jx * jx_iterate_values(struct jx *j, void **i);
  * @param i A variable to store the iteration state.
  * @return A pointer to each key in the object, and NULL when iteration is finished.
  */
-struct jx * jx_iterate_key(struct jx *j, void **i);
+struct jx * jx_iterate_keys(struct jx *j, void **i);
 
 /** Merge an arbitrary number of JX_OBJECTs into a single new one. The constituent objects are not consumed. Objects are merged in the order given, i.e. a key can replace an identical key in a preceding object. The last argument must be NULL to mark the end of the list. @return A merged JX_OBJECT that must be deleted with jx_delete. */
 struct jx *jx_merge(struct jx *j, ...);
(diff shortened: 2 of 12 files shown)


dthain:
That's interesting, it didn't occur to me that we would still be able to do environment substitution in this way.  And it's simpler in the sense that Makelfow doesn't do the substitution -- it happens in the shell execution.

Is `"#"` a customary way of doing comments in json?

trshaffer:
I've never seen `"#"` before, I just like the look of it. We could come up with a standard way to handle comments, since it's nice to be able to include them

Are JX expressions mature enough to try doing variable substitution?

dthain:
I would suggest the more boring `comment ="xyz"` but only because that's how umbrella does it.

Yes, you should try `jx_eval`-- I won't promise it's bug free, but you should give it a go.

btovar:
Actually json does not have comments, so comment = "..." would be the recommended practice.

trshaffer:
I did that environment variable shorthand and `jx_eval`ed the command strings, so here's the new [example.jx](https://github.com/trshaffer/cctools/blob/25915d942ee311551453fcbea114876c61d7e712/makeflow/example/example.jx). I'll evaluate expressions on the input/output filenames next. I was also thinking of adding a few magic variables on the rules:

- `RULENO` to give the rule's number in the input file
- `INPUTS` and `OUTPUTS` as the rule's input/output lists. Especially if we can do some forEach  in JX, these would be useful in building the command line

dthain:
@btovar do you see any show-stoppers in here?  We can continue to refine the definition of the JSON format, but if this is a good first pass, we should take it.

btovar:
Looks good.

Issue #1326, opened 2016-05-05
Confuga should provide a run-time tool to add/remove SN


batrick:
Currently Confuga limits SN membership changes to HN start-up. An administration tool should be developed which allows the administrator to modify the storage node membership (and possibly other future state) during run-time.

Solution should be aware of #1324 and #1325.

This issue supersedes #828.

Issue #1325, opened 2016-05-05
Confuga SN should be uniquely identified by a persistent name


batrick:
Currently, Confuga will utilize a SN based on the `host:port` provided at start-up. The HN will also differentiate and identify storage nodes based on this `host:port`. This is problematic for two reasons:

* A rebooted storage node that has lost its state (underlying data) due to some kind of reboot, eviction, etc. is no longer "the same storage node". The HN should treat it as a new storage node.
* If the catalog server sees the SN by a different `host:port`, then the HN is unable to associate the SN with the status information from the catalog.

Confuga can solve this by creating its own naming system for storage nodes. In the same way as Ceph's OSDs, a storage node should be given a UUID which is persistent and allows the HN to uniquely identify a SN.

I suggest adding a persistent UUID, stored on the backend file system, to the Chirp server code which is included with the catalog status update. This UUID code can then be used by the HN to learn how to contact storage nodes (i.e. learning the `host:port`).

Issue #1324, opened 2016-05-05
Confuga should verify SN identity in the cluster


batrick:
Currently, Confuga will utilize a SN based on the `host:port` provided at start-up. The HN does no verification that the Chirp server operating on that `host:port` is the same the user wants to use (i.e. another user may have started a Chirp server on that `host:port`).

The proposed fix for this is to have a password file stored on the Chirp server in a location only the HN (or owner) can read. Setting up this password file should be automated by a tool which adds storage nodes to the cluster. More on that problem in an upcoming issue.

PR #1344, opened 2016-05-21
Create UUID for Chirp server to send to Catalog.

batrick:
This is a prereq for #1324, #1325, and #1326.

### DIFF
--- a/chirp/src/chirp_filesystem.h
+++ b/chirp/src/chirp_filesystem.h
@@ -24,7 +24,7 @@ enum {
 typedef struct CHIRP_FILE CHIRP_FILE;
 
 struct chirp_filesystem {
-	int (*init) ( const char url[CHIRP_PATH_MAX], uuid *uuid );
+	int (*init) ( const char url[CHIRP_PATH_MAX], cctools_uuid_t *uuid );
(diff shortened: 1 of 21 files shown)


batrick:
Apparently there is a standard API via uuid.h in darwin/linux. I will see if that's usable instead of what I wrote in dttools.

batrick:
I elected to use the simple uuid implementation I wrote. I think this is ready for merging.

dthain:
It's fine (even preferable) to have our own uuid implementation, but I would like to avoid relying on unusual gcc path options.  If necessary, please call it "cctools_uuid.h" or something like that.

dthain:
While I understand that, technically, `_t` is a reserved suffix, that's the customary usage in most C code, including ours.  `uuid_t` or `cctools_uuid_t` would be preferable to `uuid`.

(This will be a very handy feature for tracking and reporting.)

batrick:
`uuid_t` is off the table because Darwin includes its own `uuid.h` in some of its system headers so the conflict cannot be avoided.

I will make the other changes you've suggested.

batrick:
@dthain please merge if you think this is ready.

PR #1345, opened 2016-05-23
Improve Confuga SN management/membership

### LINKED ISSUE #1324: Confuga should verify SN identity in the cluster

### LINKED ISSUE #1325: Confuga SN should be uniquely identified by a persistent name

### LINKED ISSUE #1326: Confuga should provide a run-time tool to add/remove SN

batrick:
This commit adds a new CLI tool (with accompanying C API) for adding and removing SN in Confuga.
    
        $ confuga_adm confuga://... sn-add [-p <password file>] [-r <root>] <"uuid"|"address"> <uuid|address>
    
        and
    
        $ confuga_adm confuga://... sn-rm <"uuid"|"address"> <uuid|address>
    
    The root directory of the storage node set via the "-r" switch is the location
    the HN will use to store SN state. The default root directory is "/.confuga".
    
    The password file passed via "-p" is expected to hold the same contents as a
    file on the storage node located at '/password' under the storage node's root
    directory set via the '-r' switch. Usually the user doesn't want to explicitly
    set this up so the HN will automatically create this password file if the '-p'
    option is not used.
    
    Examples:
    
        $ confuga_adm confuga:///path/to/confuga/ sn-add uuid 46BE3610-42F5-4B7C-B9D1-2B3421C4F172
        OR
        $ confuga_adm confuga:///path/to/confuga/ sn-add address localhost:1024
    
        $ confuga_adm confuga:///path/to/confuga/ sn-rm uuid 46BE3610-42F5-4B7C-B9D1-2B3421C4F172
        OR
        $ confuga_adm confuga:///path/to/confuga/ sn-rm address localhost:1024
    
    The HN will authenticate a SN by using the SHA1 hash RPC to check the password
    file at "/<root>/password" is correct. The HN does not reauthenticate every new
    session as sessions are handled at the Chirp reli layer. Instead, the HN
    reauthenticates a storage node every 15 minutes.
    
    Storage nodes are now remembered by the HN using UUIDs broadcast as part of
    normal Catalog updates. This allows the HN to distinguish between storage nodes
    operating with the same address and port (localhost/LAN).
    
    The HN is now also capable of removing storage nodes. This is done by marking a
    storage node to be removed which prevents it from being used for future jobs
    and new replicas. The HN will remove its replicas so long as sufficient
    replicas exist in the cluster. The HN will replicate as necessary so that the
    SN can eventually be removed. Once all replicas have been removed and jobs have
    completed, the SN will be removed.

    Fixes #1324.
    Fixes #1325.
    Fixes #1326.

### DIFF
--- a/chirp/src/.gitignore
+++ b/chirp/src/.gitignore
@@ -14,4 +14,5 @@ chirp_migrate
 chirp_matrix_roc
 chirp_status
 chirp_partition_stream
+confuga_adm
 libchirp_client.so
--- a/chirp/src/Makefile
+++ b/chirp/src/Makefile
@@ -15,7 +15,9 @@ OBJECTS = chirp_tool.o chirp_fuse.o $(OBJECTS_CONFUGA) $(OBJECTS_LIBRARY) $(OBJE
 OBJECTS_CONFUGA = $(SOURCES_CONFUGA:%.c=%.o)
 OBJECTS_LIBRARY = $(SOURCES_LIBRARY:%.c=%.o)
 OBJECTS_SERVER = $(SOURCES_SERVER:%.c=%.o)
-PROGRAMS = chirp chirp_get chirp_put chirp_server chirp_status chirp_benchmark chirp_stream_files chirp_fuse chirp_distribute
+PROGRAMS = $(PROGRAMS_CHIRP) $(PROGRAMS_CONFUGA)
+PROGRAMS_CHIRP = chirp chirp_get chirp_put chirp_server chirp_status chirp_benchmark chirp_stream_files chirp_fuse chirp_distribute
+PROGRAMS_CONFUGA = confuga_adm
 PUBLIC_HEADERS = chirp_global.h chirp_multi.h chirp_reli.h chirp_client.h chirp_stream.h chirp_protocol.h chirp_matrix.h chirp_types.h chirp_recursive.h confuga.h
 SCRIPTS = chirp_audit_cluster chirp_server_hdfs
 SOURCES_CONFUGA = confuga.c confuga_namespace.c confuga_replica.c confuga_node.c confuga_job.c confuga_file.c confuga_gc.c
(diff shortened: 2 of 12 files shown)


batrick:
This is based on #1344. #1344 should be merged before this.

batrick:
@btovar if you have time, maybe give this a shot and see how it works for you.

btovar:
Great, thanks Patrick.

batrick:
I rebased this on master to get the latest changes from #1344. This should be good to go but someone should give it a whirl before merging.

PR #1346, opened 2016-05-24
JX Functions

trshaffer:
I just finished a first draft of function evaluation in JX. As a proof of concept, I implemented Python's `range` and `str` functions, and the foreach construction discussed the other day. This is certainly *not* finished, and I'd like to think about the syntax and semantics more, but I thought I'd share this draft to get some comments. I added some of the new function syntax to the JX tests.

[Input](https://github.com/trshaffer/cctools/blob/b007cfbfb481964e7723358d7b531630801ba25a/dttools/test/jx.input#L128)
[Expected](https://github.com/trshaffer/cctools/blob/b007cfbfb481964e7723358d7b531630801ba25a/dttools/test/jx.expected#L270)

I'm thinking about supporting keyword arguments in addition to positional ones. I've also been talking to @nhazekam about the minimal set of functions needed to make this useful to users. I think we'll want a few basic operations at least, such as concatenate arrays, merge objects, string -> number, etc. Any thoughts/suggestions @dthain?


trshaffer:
I went through the source code and wrote up a description of the JX specific things as they are now

https://gist.github.com/trshaffer/43f84d0745fa20ccb6599e75e4b5ec8b

dthain:
Tim, I missed your writeup before.  Lots of good detail in there.  Maybe just add a bit to the introduction to explain purpose and use cases for JX, and explain the Venn diagram.  (e.g. jx >= json, evaluated jx is always json, is jx a subset of js?)

Then, I would like you to take a look at JS expressions and see to what extent JX semantics differ from JS.  There is the business of null on invalid types.  Are there any others?  My initial idea was to make JX a subset of JS, but I'm not sure if that is entirely achievable and/or desirable.

You pointed out that JX distinguishes between DOUBLE and INTEGER, whereas JSON only has NUMBER.  That was more an oversight than a deliberate choice on my part.  Do we gain anything by distinguishing the two?

trshaffer:
I added some some more to the introduction of that Gist.

The only places that distinguish between integers and doubles are array indexing and `range`. The single number type seems to be a big regret with JSON, as languages generally split that into one or more real numeric types, and JSON-derivative formats often do the same. It shouldn't be a big deal to have two types as long as the docs mention it.

As long as you're not trying to do mixed-type operations, JX should be pretty much close to JS. As soon as you mix types, [all bets are off](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators#Equality). For JS compatibility, we really out to be using `===`. We actually diverged a little with array concatenation:

    [1, 2] + [3, 4]
    JX => [1, 2, 3, 4]
    JS => "1,23,4"

The JavaScript-y way to do that would be `Array.concat([1, 2], [3, 4])`. If we reintroduce `concat` to JX, we could probably write some JS shims for the functions, except `foreach`. I don't think it's possible to write a JS function to do that without rewriting the JX with lambda functions. Overall, I think JavaScript's handling of types, extensive use of anonymous functions, and standard library don't really line up with the goals of JX. I've actually had Python in mind more than JS, since it's sensible, easy, and well known.

As to returning null on invalid operations, it's not really like Python or JS. For operators, JS does *something* and returns a value (generally not what you were expecting), and throws an exception on bad function calls. Python throws exceptions for both these cases. We could consider adding a `JX_ERROR` type to be returned on bad operations that unwinds the eval. That would be helpful to diagnose errors, since we could report a nice error string mentioning the operations and data that failed.

trshaffer:
I just rebased off the current master. The main addition is error reporting during evaluation. I added a new type, `JX_ERROR` to indicate an evaluation failure. This JX type is only returned from a call to `jx_eval` where a function or operator was misapplied, indicating that the result of evaluation is undefined. In the same way that it's necessary to check that `jx_parse` succeeded, callers to `jx_eval` must check if the returned `struct jx` is of type `JX_ERROR`. If not, `jx_eval` succeeded and the evaluated result contains only plain JSON types. If `jx_eval` returns an error, the caller should handle that appropriately. The error type has a field that holds an arbitrary JX structure for info on the failure. I've been doing JS style errors, so errors are objects that have at least a `"name"` and `"message"` field.

I intentionally didn't make syntax to read in errors or serialize them faithfully, since errors need to be  handled where they occur. If a caller doesn't check for errors after `jx_eval` and simply prints the result, any `JX_ERROR` is printed as `&ERROR&`. This isn't valid JX, JSON, JS, XML, or any other language I can think of, so it ought to make it easier to track down where errors are occurring. The intent is to make things fail loudly rather than letting some garbage value or serialized error escape into other places. I updated the `jx_test` program to print errors to `stderr`

    expression: join([1,2])
    value:      &ERROR&
    error:      {"name": "SyntaxError", "message": "joined items must be strings", "func": join([1, 2]), "line": 240, "file": "jx_function.c"}

trshaffer:
I've expanded the errors to have an external representation that we can read in and out. The current syntax is

    Error{"source":"jx_eval","name":"unsupported operator","message":"unsupported operator on boolean","operator":true%false,"file":"jx_eval.c","line":59,"code":1}

Since we might be using these to pass error status between different programs in addition to just `jx_eval`, I made the `"source"` key mandatory. Errors in general are free form, but the `"source"` gives an indication of who generated the error, and what format the rest of it is. I defined some standard fields and error types for the source `"jx_eval"`, and updated the JX spec I linked to above. I'm imagining a `"work queue"` source, for example, that Makeflow would know how to parse, and other sources can design errors any way they want without interfering with JX internal uses.

trshaffer:
Per a discussion with @nhazekam, I also made `"name"` and `"message"` required keys. This way, there needs to be a conscious decision to make a nondescript error

dthain:
This is looking good overall.  If you would, please take your writeup about JX in gist, and add it to the documentation in the repo under `doc/jx.html` or similar.

trshaffer:
Getting ready for use in Makeflow, I added a partial evaluation mode for JX expressions. This makes me a little uneasy, but I can't think of a way to do per-rule variable scopes without this or some sort of quoting mechanism that end users would need to deal with. The problem arises if the Makeflow rules array is built from multiple pieces, e.g. by concatenating several smaller arrays of rules. It's pretty common to have some setup rules, one or more types of production rules, and some cleanup/aggregation rules. If Makeflow will do per-rule variable scope and expose the input and output arrays to the JX within each rule, Makeflow needs to dig into the rule bodies and handle the various fields. If the rules array is dynamically generated, the JX might instead contain an operator or function that will generate the rules. If Makeflow just evaluates the entire rules array, *all* variables and functions get evaluated, so there's no opportunity for Makeflow to do per-rule evaluation.

If per-rule variable scopes aren't necessary for Makeflow, I'd remove this partial eval code. Otherwise, we need to be able to expand the structure of a JX document without completely evaluating it. This approach has a smaller end-user impact than some sort of quoting mechanism, but seems like a bad idea...

dthain:
I think partial eval approach is powerful, and we will need to do something like this for makeflow.
However, the selection of resolution methods that you have is probably going to grow and change
along with various use cases.  (e.g. makeflow might have several levels of lookup that eventually fall back on `getenv`)

To keep the library stable, let me suggest a different approach.  Make one underlying function that invokes a callback function to evaluate symbols:

```
typedef struct jx * (*lookup_t) ( const char *name );
struct jx * jx_eval_1( struct jx *expr, lookup_t lookup );
```

And then modify `jx_eval` to invoke `jx_eval_1` with `jx_lookup` as the callback function.

Then, you can implement whatever lookup function you like in makeflow.

How does that sound?

dthain:
@trshaffer per our discussion, please back out the partial eval feature, and let's merge the implementation of functions.

trshaffer:
I've reverted the partial eval and rebased off the current master, so I think this should be ready to merge

PR #1347, opened 2016-05-25
Fixes mount issue in disk_alloc

nkremerh:
Bug fix: adds option to mkfs which makes the process of instantiating a loop device more explicit


nkremerh:
@dthain @btovar This is a small bug fix. Would you mind looking it over?

btovar:
@nkremerh: for posterity, could you add a comment in the code describing the mkdev arguments?

PR #1350, opened 2016-05-25
Modifies disk_alloc_create to redirect lengthy output to /dev/null

nkremerh:
This prevents the terminal from becoming swamped with lengthy, obscure output.
@dthain @btovar, could you take a look please?


btovar:
You may want to redirect stderr to /dev/null too, so that all messages come from your error checking.

nkremerh:
@btovar How about this?

btovar:
Thanks Nate.

PR #1352, opened 2016-05-26
bug fix: export correctly CORES, MEMORY, and DISK when task execs

btovar:
Fixes a bug as reported by @nhazekam


nhazekam:
@btovar Works and looks good to me. Thanks for the quick turn around.

PR #1354, opened 2016-05-26
Changes to wq wait loop to autoregulate capacity

btovar:
@dthain, as discussed. Let me know if this looks ok before I add the time stats measurements.


dthain:
I think it's pretty close; the only bit I don't like is the semi-busy-waiting in link_poll.  That sort of busy waiting always seems to come back and bite us.  (And it also imposes a 250ms penalty on each task dispatch each time around the loop.  That will add up for large applications.)

Can you figure out a way to factor it so that the main loop polls with a timeout of zero at first, and then does a long sleep only if all other work has been dispatched?

btovar:
Now we only wait if nothing interesting happened in the previous cycle. The wait, when it occurs, is never more than 1/4 of a second.

btovar:
@dthain, it should be ready now.

PR #1357, opened 2016-05-31
Makeflow condor logs

btovar:
Generate output and error files in the submit file:

RULENUMBER.condor.error and RULENUMBER.condor.output.

### DIFF
--- a/batch_job/src/batch_job.c
+++ b/batch_job/src/batch_job.c
@@ -201,9 +201,9 @@ const char *batch_queue_type_string()
 }
 
 
-batch_job_id_t batch_job_submit(struct batch_queue * q, const char *cmd, const char *extra_input_files, const char *extra_output_files, struct jx *envlist, const struct rmsummary *resources)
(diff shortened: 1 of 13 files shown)


dthain:
Hmm, I'm not thrilled with this approach, since it seems to be adding yet another special case to behavior that is already inconsistent within the batch job interface.

Can we step back and consider if there is a more consistent way that stderr could be handled through the batch job interface?

btovar:
What about this one instead?

makeflow --error-dir=somedir

Error files are written to somedir, for batch systems that support it (right now only condor implemented). The targer error file is passed as an argument to submit job.

dthain:
Let's talk about it.

nhazekam:
@btovar Has this been resolved? Is it still a priority for 5.5?

btovar:
Not resolved yet.  A clean solution won't be ready for 5.5.

nhazekam:
As this is not going to be resolved currently I am removing myself as an assignee.

PR #1358, opened 2016-05-31
JX match functions

trshaffer:
I wrote up some access functions for unwrapping JX structs. The common case I had in mind (and encountered a lot with JX functions and Makeflow import) is reading a value of a specific type into a local variable. There are specialized functions for lookups in objects, so these do something similar for `struct jx`s in general. In addition to the extra opportunities for the compiler to check things, these functions are more readable in code than digging around in struct fields and enum variants.

The `jx_match_*` functions follow a common interface. Each type has a specialized match function. If the struct is the requested type, a pointer to the raw field inside the `struct jx` is returned, or `NULL` otherwise. This can be used to directly modify the `struct jx`, or more commonly, can be used as a truth value to make sure that the type was correct.  Each function also accepts an optional pointer to a local variable. The unwrapped value is written by value (or copied for a heap allocated type).

The [header documentation](https://github.com/trshaffer/cctools/blob/6981308b799434d97f22bc2e93aadfd8f330bfdb/dttools/src/jx_match.h) is pretty comprehensive, so that's a good place to look for details and examples of use.

### DIFF
--- a/dttools/src/jx.h
+++ b/dttools/src/jx.h
@@ -217,7 +217,7 @@ int jx_is_constant( struct jx *j );
 void jx_export( struct jx *j );
 
 /** Iterate over the values in an array.
- * The iteation state is stored by the caller in an opaque pointer variable. When starting iteration,
+ * The iteration state is stored by the caller in an opaque pointer variable. When starting iteration,
  * the caller MUST pass the address of a pointer initialized to NULL as i. It is undefined behavior
  * to pass a non-NULL iterator variable not set by a previous call. Subsequent calls should
  * use the same variable to continue iteration. After the initial call, the value of j is ignored.
@@ -236,7 +236,7 @@ void jx_export( struct jx *j );
 struct jx * jx_iterate_array(struct jx *j, void **i);
 
 /** Iterate over the values in an object.
- * The iteation state is stored by the caller in an opaque pointer variable. When starting iteration,
+ * The iteration state is stored by the caller in an opaque pointer variable. When starting iteration,
  * the caller MUST pass the address of a pointer initialized to NULL as i. It is undefined behavior
  * to pass a non-NULL iterator variable not set by a previous call. Subsequent calls should
  * use the same variable to continue iteration. After the initial call, the value of j is ignored.
@@ -256,7 +256,7 @@ struct jx * jx_iterate_array(struct jx *j, void **i);
 struct jx * jx_iterate_values(struct jx *j, void **i);
 
 /** Iterate over the keys in an object.
- * The iteation state is stored by the caller in an opaque pointer variable. When starting iteration,
+ * The iteration state is stored by the caller in an opaque pointer variable. When starting iteration,
  * the caller MUST pass the address of a pointer initialized to NULL as i. It is undefined behavior
  * to pass a non-NULL iterator variable not set by a previous call. Subsequent calls should
  * use the same variable to continue iteration. After the initial call, the value of j is ignored.
--- a/dttools/src/jx.h
+++ b/dttools/src/jx.h
@@ -216,4 +216,63 @@ int jx_is_constant( struct jx *j );
 /** Export a jx object as a set of environment variables.  @param j A JX_OBJECT. */
 void jx_export( struct jx *j );
 
+/** Iterate over the values in an array.
+ * The iteation state is stored by the caller in an opaque pointer variable. When starting iteration,
+ * the caller MUST pass the address of a pointer initialized to NULL as i. It is undefined behavior
+ * to pass a non-NULL iterator variable not set by a previous call. Subsequent calls should
+ * use the same variable to continue iteration. After the initial call, the value of j is ignored.
+ *
+ *     struct jx *item;
+ *     for (void *i = NULL; (item = jx_iterate_array(j, &i));) {
+ *         printf("array item: ");
+ *         jx_print_stream(item, stdout);
+ *         printf("\n");
+ *     }
+ *
+ * @param j The JX_ARRAY to iterate over.
+ * @param i A variable to store the iteration state.
+ * @return A pointer to each item in the array, and NULL when iteration is finished.
+ */
+struct jx * jx_iterate_array(struct jx *j, void **i);
+
+/** Iterate over the values in an object.
+ * The iteation state is stored by the caller in an opaque pointer variable. When starting iteration,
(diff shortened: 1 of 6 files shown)


dthain:
This looks like the right approach to me, @nhazekam please take a look at the details.

dttools/src/jx.h
```diff
@@ -216,4 +216,63 @@ int jx_is_constant( struct jx *j );
 /** Export a jx object as a set of environment variables.  @param j A JX_OBJECT. */
 void jx_export( struct jx *j );
 
+/** Iterate over the values in an array.
+ * The iteation state is stored by the caller in an opaque pointer variable. When starting iteration,
```
nhazekam: Small spelling mistake, second word. Happens in the other two descriptions as well.

dttools/src/jx_match.c
```diff
@@ -0,0 +1,198 @@
...
+		return &j->u.double_value;
+	} else {
+		return NULL;
+	}
+}
+
+char **jx_match_string(struct jx *j, char **v) {
+	if (jx_istype(j, JX_STRING)) {
+		if (v) {
+			*v = strdup(j->u.string_value);
```
nhazekam: Here, as per our discussion, I think if the strdup fails we should return NULL. That way if the match is successful we know that v is also usable if requested. This can be bypassed by setting v to NULL as a parameter in the case were you just want a point.

dttools/src/jx_match.h
```diff
@@ -0,0 +1,265 @@
...
+ * and a sequence of item specifications. Each item
+ * spec includes a JX type, the location of a pointer that will reference
+ * the extracted value, and for objects, a key name. The match functions
+ * process each specification in turn, stopping on an invalid spec,
+ * or NULL, and returning the number of items succesfully matched.
+ *
+ * @code
+ *     jx_int_t a;
+ *     double b;
+ *     switch (jx_match_positional(j, &a, JX_INTEGER, &b, JX_DOUBLE)) {
```
nhazekam: Misnamed: jx_match_array?

dttools/src/jx_match.h
```diff
@@ -0,0 +1,265 @@
...
+ * type of the matched value. The pseudo-type JX_ANY is provided
+ * for this purpose. Since the type of the value matched by JX_ANY is
+ * unknown, the caller might want to use the other match functions to
+ * handle whatever cases they're interested in. The following example
+ * uses most of the matching functionality.
+ *
+ * @code
+ *     struct jx *a;
+ *     struct jx *b;
+ *     if (jx_match_keyword(j, &a, JX_ARRAY, "array key", &b, JX_ANY, "???") == 2) {
```
nhazekam: Misnamed: jx_match_object?

nhazekam:
Overall the code looks good and makes sense to me. I appreciate the amount of documentation and examples to show the intended use case/behavior. 

Two things that I would appreciate @dthain @btovar to give input on.
1. Included are some compiler warnings about expected usage, warning the compiler/user if the program does not check the result of the match. I think it is a useful warning, but is that something we should include in CCTools? I am all for it, but wanted to get your opinions.
2. The match functions for Arrays and Objects utilize a null pointer as ending the matching process. @trshaffer and I have discussed the possible merits of allowing null args to allow a positional match where you are looking for the kth item. This would require some code changes and I wanted to talk about it here so the changes could be made before its merged and where we are still developing the functionality.

dttools/src/jx_match.c
```diff
@@ -0,0 +1,198 @@
...
+jx_int_t *jx_match_integer(struct jx *j, jx_int_t *v) {
+	if (jx_istype(j, JX_INTEGER)) {
+		if (v) {
+			*v = j->u.integer_value;
+		}
+		return &j->u.integer_value;
+	} else {
+		return NULL;
+	}
+}
```
btovar: For functions like this one, I rather have:

```C
jx_int_t  jx_match_integer(struct jx *j, int *v) {
    if(v) {
        *v = 0;
    }
	if (jx_istype(j, JX_INTEGER)) {
		if (v) {
                      *v = 1;
                }
           return j->u.integer_value;
		}
	return 0;
}
```

btovar: If you do not like that, you have to change your function to:

```C
const jx_int_t *jx_match_integer(struct jx *j, jx_int_t *v) { ```

as we do not want the pointer to be able to modify its value.

File: dttools/src/jx_match.c (hunk 2 above)
Comment (btovar) [inline dttools/src/jx_match.c]: With my recommendation above, we would avoid the ** pointer:

```C
char *jx_match_string(struct jx *j, int *v) {
```

which is easier to think about.

dttools/src/jx_match.c
trshaffer: Returning a non-`const` pointer was actually intentional. Usually either

- the caller just wants to read the value, so put a copy in the `v` argument and check the return as a truth value.
- the caller wants to modify the value, so pass `NULL` in `v` and grab the returned pointer. The caller can do arbitrary reads/modifications directly on value in the structure without `malloc()`ing anything.

Are we making an assumption that JX structures are immutable? I don't think we have enough library built up to do that efficiently, and allowing direct modifications would be useful for incrementally building things.

btovar: Not immutable, but they should be modifiable only through the API.

trshaffer: Which API are you referring to? There are currently no functions for in-place modifications to values.

trshaffer:
I just cut out some of the functions that were included for completeness but unlikely to be useful. Also, the match functions all return `int` now indicating the result of the match, so no worries about pointer trickiness and bad casts. I still like the extra compiler checks.

    double x;
    jx_match_double(j, &x);                         // compiler warning
    if (!jx_match_double(j, &x)) return NULL;       // OK

dthain:
Extra static checking by the compiler is a good thing, as long as we are reasonably sure that they don't introduce compatibility/portability bugs.  c.f. gcc attribute flags for checking varargs formats in debug.h and other places.

dthain:
And I agree that we should not encourage direct modification of JX structures.  When that need becomes apparent, we should add a suitable API.

trshaffer:
I'm checking the compiler and version before defining the attribute symbols. On a different compiler or GCC before the features were introduced, the symbols are `#define`d empty. I grabbed that `__GNUC_PREREQ__` macro from Bionic libc, so it ought to be plenty compatible.

nhazekam:
The test looks good. Merging so this can be used in other jx library pull requests.

Issue #1348, opened 2016-05-25
wrapping commands does not work with -Tcondor


btovar:
E.g.:

Found illegal unescaped double-quote: "/usr/bin/convert -swirl 180 capitol.jpg capitol.180.jpg"
The full arguments you specified were: /afs/nd.edu/user34/btovar/usr/x86_64/redhat6/cctools/bin/resource_monitor --no-pprint --with-output-files=mon/resource-rule-2 -L 'cores: 1' -L 'memory: 3' -L 'disk: 1' -V 'category:default' --sh "/usr/bin/convert -swirl 180 capitol.jpg capitol.180.jpg"
couldn't submit batch job, still trying...
submitting job: /afs/nd.edu/user34/btovar/usr/x86_64/redhat6/cctools/bin/resource_monitor --no-pprint --with-output-files=mon/resource-rule-1 -L 'cores: 1' -L 'memory: 3' -L 'disk: 1' -V 'category:default' --sh "/usr/bin/convert -swirl 90 capitol.jpg capitol.90.jpg"

This one should be fixed before release.

btovar:
There are the condor rules:

The entire string representing the command line arguments is surrounded by double quote marks. This permits the white space characters of spaces and tabs to potentially be embedded within a single argument. Putting the double quote mark within the arguments is accomplished by escaping it with another double quote mark.

    The white space characters of spaces or tabs delimit arguments.

    To embed white space characters of spaces or tabs within a single argument, surround the entire argument with single quote marks.

    To insert a literal single quote mark, escape it within an argument already delimited by single quote marks by adding another single quote mark.

Example:

arguments = "3 simple arguments"

Produces:

argument 1: 3
argument 2: simple
argument 3: arguments

Another example:

arguments = "one 'two with spaces'	3"

Produces:

argument 1: one
argument 2: two with spaces
argument 3: 3

And yet another example:

arguments = "one ""two"" 'spacey ''quoted'' argument'"

Produces:

argument 1: one
argument 2: "two"
argument 3: spacey 'quoted' argument

Notice that in the new syntax, the backslash has no special meaning. This is for the convenience of Windows users.

PR #1362, opened 2016-06-01
Added function to escape Condor, called in batch_job_condor

nhazekam:
Added function that escapes quotes which is required in a condor submit file. Needed as quotes are used in wrapping of commands with wrappers and resource monitor.

### DIFF
--- a/batch_job/src/batch_job_condor.c
+++ b/batch_job/src/batch_job_condor.c
@@ -115,7 +115,9 @@ static batch_job_id_t batch_job_condor_submit (struct batch_queue *q, const char
 
 	fprintf(file, "universe = vanilla\n");
 	fprintf(file, "executable = condor.sh\n");
-	fprintf(file, "arguments = %s\n",cmd);
+	char *escaped = string_escape_condor(cmd);
+	fprintf(file, "arguments = %s\n", escaped);
+	free(escaped);
 	if(extra_input_files)
 		fprintf(file, "transfer_input_files = %s\n", extra_input_files);
 	// Note that we do not use transfer_output_files, because that causes the job
--- a/dttools/src/stringtools.c
+++ b/dttools/src/stringtools.c
@@ -53,6 +53,46 @@ char *string_escape_shell( const char *str )
 	return result;
 }
 
+/*
+ * Based on HTCondor documentation:
+ * -The white space characters of spaces or tabs delimit arguments.
+ * -To embed white space characters of spaces or tabs within a single argument, 
+ * surround the entire argument with single quote marks.
+ * -To insert a literal single quote mark, escape it within an argument already 
+ * delimited by single quote marks by adding another single quote mark.
+ *
+ * We surround the whole string with double quotes to enable quote escaping in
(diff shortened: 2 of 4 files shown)


nhazekam:
Resolves #1348 @btovar test when able.

dthain:
Please add a test case.

nhazekam:
On it.

btovar:
It does not quite work for me.

Add the following to example.makeflow, just before the first rule:

export MAGICK_THREAD_LIMIT=4

MEMORY=512
DISK=100
CORES=4

and run with:  makeflow -Tcondor example.makeflow  --monitor=mon -dall

(you need to that from /tmp)

You'll see only one rule completes.

btovar:
This is what condor ends up seeing:

Arguments = ".//resource_monitor --no-pprint --with-output-files=mon/resource-rule-1 -L ''''cores:' '4'''' -L ''''memory:' '512'''' -L ''''disk:' '100'''' -V ''''category:default'''' --sh \"/usr/bin/convert -swirl 90 capitol.jpg capitol.90.jpg\""

It does not look quite right.

nhazekam:
I am taking a look right now. I'll see what change I made that caused this.

nhazekam:
I was able to get it to run:

".//resource_monitor --no-pprint --with-output-files=./resource-rule-0 -L '''cores: 2''' -L '''memory: 10''' -L '''disk: 20''' -V '''category:merge''' --sh ""/usr/bin/convert -delay 10 -loop 0 capitol.jpg capitol.90.jpg capitol.180.jpg capitol.270.jpg capitol.360.jpg capitol.270.jpg capitol.180.jpg capitol.90.jpg capitol.montage.gif"" "

Unfortunately, when you have a directory for monitor, such as with the above example, it fails to find the file. looking into that now.

btovar:
Nick, see: #1364

dthain:
If inducing the desired behavior from Condor is too complicated, perhaps we might consider changing the behavior of MF/RM to live within the constraints.

(For example, we might consider having Makeflow consolidate all the resource records into a single output file.  We don't have to do that, just pointing it out as an option.)

btovar:
Nick, should we merge this one?

nhazekam:
Were you able to successfully run resource monitor through condor? If so I
think it's good to go and doesn't effect any other systems.
On Jun 2, 2016 12:14 PM, "Benjamin Tovar" <notifications@github.com> wrote:

Nick, should we merge this one?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/cooperative-computing-lab/cctools/pull/1362#issuecomment-223342033>,
or mute the thread
<https://github.com/notifications/unsubscribe/AE_SNyLOCWRq5kFl922C1anJi9I-iAyUks5qHwFXgaJpZM4IrrA_>
.

btovar:
It does  if I do: makeflow --monitor=.

nhazekam:
To reiterate: the remaining issue at this point for resource monitor in Condor is directories. Pointing to that issue #1364

Issue #1361, opened 2016-06-01
Makeflow Example Categories issue


nhazekam:
Categories in Makeflow are not behaving as expected, from my understanding. Currently the resources exported to a node, regardless of the category, are the last resource set. In the example this comes up as every rule getting the resources for the "preprocessing" category.

nhazekam:
I have addressed it in a branch, but I will touch base with @btovar to make sure this conforms with the expected behavior.

PR #1363, opened 2016-06-01
Makeflow Variable are set to most recent Category, not default

nhazekam:
Setting to default overwrote the different variable which resulted in every task using the default category values. No values were set in the actual categories. Now every variable is tied to the most recent defined category and not the default, unless that was the most recent.


nhazekam:
Resolves #1361 @btovar take a look.

btovar:
Excellent, thanks Nick.

Issue #1366, opened 2016-06-02
Batch Options garbled using SLURM


nhazekam:
When testing on slurm, I was experiencing issues with jobs failing to submit. It only occurred after the first job submitted and sbatch complained about a missing file, where the file name was garbage characters.

nhazekam:
The issue occurs in how we set the batch options. I will create a pull request with changes.

nhazekam:
Addressed by creating a copy of the batch options in makeflow.c.

PR #1367, opened 2016-06-03
Batch job option fix

nhazekam:
This waits to free until after we have made a duplicate of the new value. This prevents an issue if the old value passed back in. e.g. as in Makeflow

previous = batch_queue_get_option (q, "batch-options");
if(new-options)
    batch_queue_set_option(q, "batch-options", new-options);

run task

if(previous)
    batch_queue_set_option(q, "batch-options", previous);

Set option duplicates the passed value, and removes the previous value. When removing the previous value, it also frees that value as batch_job created it in the first place. The issue occurs when you pass back in the original value that you retrieved and it has already been freed. This happens regardless of whether you set new options or not. 

This pull request still frees the internal copy when a new value is set. It also returns a copy on get so we don't have issues if a new value is set while a pointer is held.

Instead of the solution here we could respect const char and not strdup the value, relieving the need to free at all in batch options. 

Before this is merged I need to go through and make sure memory leaks weren't added by strdup on get_option and feature.

Definitely open to a better solution, but this works in SLURM. @btovar @dthain

### DIFF
--- a/batch_job/src/batch_job.c
+++ b/batch_job/src/batch_job.c
@@ -148,13 +148,13 @@ int batch_queue_port(struct batch_queue *q)
 void batch_queue_set_option (struct batch_queue *q, const char *what, const char *value)
 {
 	char *current = hash_table_remove(q->options, what);
-	free(current);
 	if(value) {
 		hash_table_insert(q->options, what, xxstrdup(value));
 		debug(D_BATCH, "set option `%s' to `%s'", what, value);
 	} else {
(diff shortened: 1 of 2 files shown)


nhazekam:
@btovar How does that look.

nhazekam:
Segfaults on no batch options. I am making a change.

btovar:
Nick, the general approach seems fine to me. Merge when ready.

nhazekam:
Addresses #1366

Issue #1364, opened 2016-06-02
rmonitor does not work with Condor and output in directories


btovar:
We rely on Condor to bring back files in the task scratch directory. This does not work when the resource_monitor writes its output to a directory.

btovar:
Some options:

- Specify transfer_output_files in Condor, and add a periodic remove so that jobs are never held when output is missing.

- The monitor wrapper tells the monitor to always write in the task top-most directory. When the task finishes, the wrapper (locally) moves the output to the desired destination. This is what we do in Work Queue, but with remote renaming.

btovar:
@nhazekam: The monitor works correctly on -Tsge. Please check with slurm on xsede.

dthain:
IIRC, Condor can be induced to retrieve a directory of output, if it is specified in the submit file.

btovar:
Yes, but we do not specify any of the outputs in condor. This is so that if the task fails, the condor job is not held. We could change that, though, telling condor never to hold a job on missing output.

PR #1373, opened 2016-06-07
Resource Monitor file directories in Condor

nhazekam:
Fix that uses the basename of Resource Monitor files remotely on condor. 

If resource monitor is enabled with Condor, we write the monitor command to use a file in the cwd, not in a directory. This way the monitor output if sent back by Condor. As Condor does the actual transport of the files, there isn't code in batch_job_condor to intercept and rename the files. To put them in the correct location a move function is used to rename the files in the expected directory prior to checking if the output exist.

@btovar Give this a look, though it is not an ideal solution. 

Continuation fix for Issue #1364

### DIFF
--- a/batch_job/src/batch_job.c
+++ b/batch_job/src/batch_job.c
@@ -74,6 +74,7 @@ struct batch_queue *batch_queue_create(batch_queue_type_t type)
 	q->data = NULL;
 
 	batch_queue_set_feature(q, "local_job_queue", "yes");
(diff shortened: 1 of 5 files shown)


btovar:
Nick, could you resubmit this one?

btovar:
There are a couple of problems:

When specifying --monitor=.
Failed to move resource-rule-1.summary -> ./resource-rule-1.summary

When specifying --monitor=mon
mon directory is not created.

If mon directory is created by hand, the workflow still fails.

Issue #1142, opened 2016-02-04
Makeflow ignores return value of work_queue_task_specify_file


btovar:
Example:

output: /my/absolute/path
    ....

Work queue work_queue_task_specify_file prints an error on encountering the absolute path, and does not add the file to the task. Makeflow does not check for this and continues. The task only succeeds if the absolute path is globally accessible (e.g. afs).

Proposal: Any errors from work_queue_task_specify_file should be fatal to makeflow.

@nhazekam, @dthain: thoughts?

nhazekam:
Would this check be better handled in makeflow_check_batch_consistency? This would force us to note when a file is specified as an absolute path but then we could decide it it is supported as a whole for each batch system as opposed to just Work Queue. makeflow_check_batch_consistency is where we currently check for remote renaming support.

btovar:
I think now you know the code better... I have a present for you!

btovar:
Checking in makeflow_check_batch_consistency should be ok, but note that this is in addition to that. work_queue_specify_file may fail in other ways.

btovar:
After talking with @nhazekam:

1) makeflow will produce a notice for the absolute paths in work queue, as we do for remote renaming for other batch systems.
2) makeflow will produce a fatal exit on work_queue_specify_file.

nhazekam:
Resolved in the referenced pull request. 

Solution: 
Batch_job* now has a feature to express whether absolute paths are allowed in a rule or not. 
If not allowed, Makeflow prevents rules from being executed in that queue if they have a remote absolute path.
Work Queue now produces a fatal error if an absolute path is incorrectly used, instead of printing to stderr, but continuing with computation.

Does not change all WQ errors in specify file to fatal.

PR #1374, opened 2016-06-07
Makeflow prevent absolute paths in WQ

nhazekam:
Checks that the batch job system supports absolute path filenames/remote-names. Exits with message and rule number of first encountered problem.

Also exits if specify file fails. I can make this a fatal error w/message, but did not as WQ already prints error message.

@btovar take a look.

### DIFF
--- a/batch_job/src/batch_job_work_queue.c
+++ b/batch_job/src/batch_job_work_queue.c
@@ -25,12 +25,12 @@ static void specify_files(struct work_queue_task *t, const char *input_files, co
 			if(p) {
 				*p = 0;
 				if(!work_queue_task_specify_file(t, f, p + 1, WORK_QUEUE_INPUT, caching_flag)){
-					fatal("Failed to specify file %s->%s in Work Queue", f, p+1);
+					fatal("Error while specifying file: %s->%s.", f, p+1);
 				}
 				*p = '=';
 			} else {
 				if(!work_queue_task_specify_file(t, f, f, WORK_QUEUE_INPUT, caching_flag)){
-					fatal("Failed to specify file %s in Work Queue", f);
+					fatal("Error while specifying file: %s.", f);
 				}
 			}
 			f = strtok(0, " \t,");
@@ -46,12 +46,12 @@ static void specify_files(struct work_queue_task *t, const char *input_files, co
 			if(p) {
 				*p = 0;
 				if(!work_queue_task_specify_file(t, f, p + 1, WORK_QUEUE_OUTPUT, caching_flag)){
-					fatal("Failed to specify file %s->%s in Work Queue", f, p+1);
+					fatal("Error while specifying file: %s->%s.", f, p+1);
 				}
 				*p = '=';
 			} else {
 				if(!work_queue_task_specify_file(t, f, f, WORK_QUEUE_OUTPUT, caching_flag)){
-					fatal("Failed to specify file %s in Work Queue", f);
+					fatal("Error while specifying file: %s.", f);
 				}
 			}
 			f = strtok(0, " \t,");
--- a/batch_job/src/batch_job_work_queue.c
+++ b/batch_job/src/batch_job_work_queue.c
@@ -25,12 +25,12 @@ static int specify_files(struct work_queue_task *t, const char *input_files, con
 			if(p) {
 				*p = 0;
 				if(!work_queue_task_specify_file(t, f, p + 1, WORK_QUEUE_INPUT, caching_flag)){
-					exit(1);
+					fatal("Failed to specify file %s->%s in Work Queue", f, p+1);
 				}
 				*p = '=';
 			} else {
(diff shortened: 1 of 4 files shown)


nhazekam:
Addresses #1142

batch_job/src/batch_job_work_queue.c
```diff
@@ -24,10 +24,14 @@ static void specify_files(struct work_queue_task *t, const char *input_files, co
 			p = strchr(f, '=');
 			if(p) {
 				*p = 0;
-				work_queue_task_specify_file(t, f, p + 1, WORK_QUEUE_INPUT, caching_flag);
+				if(!work_queue_task_specify_file(t, f, p + 1, WORK_QUEUE_INPUT, caching_flag)){
+					exit(1);
```
btovar: We want a fatal message here.

nhazekam:
@btovar Confirm that this does what we would like. Makeflow disallows absolute paths when they are not supported in the batch system, currently only WQ, so we should not see a specify file failure, but if we do it is fatal.

batch_job/src/batch_job_work_queue.c
```diff
@@ -24,10 +24,14 @@ static void specify_files(struct work_queue_task *t, const char *input_files, co
 			p = strchr(f, '=');
 			if(p) {
 				*p = 0;
-				work_queue_task_specify_file(t, f, p + 1, WORK_QUEUE_INPUT, caching_flag);
+				if(!work_queue_task_specify_file(t, f, p + 1, WORK_QUEUE_INPUT, caching_flag)){
+					fatal("Failed to specify file %s->%s in Work Queue", f, p+1);
```
btovar: Message is confusing, as the file was specified. You want something like: Error while specifying file.

batch_job/src/batch_job_work_queue.c
```diff
@@ -13,7 +13,7 @@
 #include <string.h>
 #include <errno.h>
 
-static void specify_files(struct work_queue_task *t, const char *input_files, const char *output_files, int caching_flag )
+static int specify_files(struct work_queue_task *t, const char *input_files, const char *output_files, int caching_flag )
```
btovar: You always return 0. No need to change the void.

batch_job/src/batch_job_work_queue.c
```diff
@@ -76,7 +85,8 @@ static batch_job_id_t batch_job_wq_submit (struct batch_queue * q, const char *c
 
 	t = work_queue_task_create(cmd);
 
-	specify_files(t, extra_input_files, extra_output_files, caching_flag);
+	if(specify_files(t, extra_input_files, extra_output_files, caching_flag))
```
btovar: On error, fatal exits, thus we do not need to check here.

makeflow/src/makeflow.c
```diff
@@ -784,7 +785,7 @@ static int makeflow_check_batch_consistency(struct dag *d)
 
 		if(itable_size(n->remote_names) > 0 || (wrapper && wrapper->uses_remote_rename)){
 			if(n->local_job) {
-				debug(D_ERROR, "remote renaming is not supported on locally. Rule %d.\n", n->nodeid);
+				debug(D_ERROR, "remote renaming is not supported locally. Rule %d.\n", n->nodeid);
```
btovar: "not supported with -Tlocal or LOCAL execution."

btovar:
Did you test it?

nhazekam:
@btovar It has now been tested and should be good to go. The updates to work_queue.c only changed the absolute path errors to fatal, while several other errors are still only fprintf(stderr, ...).

work_queue/src/work_queue.c
```diff
@@ -3934,7 +3934,7 @@ int work_queue_task_specify_file(struct work_queue_task *t, const char *local_na
 	// the worker(the worker on which the task will be executed) is unlikely to
 	// be known. Thus @param remote_name should not be an absolute path.
 	if(remote_name[0] == '/') {
-		fprintf(stderr, "Error: Remote name %s contains absolute path.\n", remote_name);
+		fatal("Error: Remote name %s contains absolute path.\n", remote_name);
```
btovar: Nitpick... Remote name '%s' is an absolute path.\n

nhazekam: But totally correct. Thanks for that keen eyes!

btovar:
Looks good Nick. Merge when you are ready.

work_queue/src/work_queue.c
```diff
@@ -3865,7 +3865,7 @@ int work_queue_task_specify_url(struct work_queue_task *t, const char *file_url,
 		return 0;
 	}
 	if(remote_name[0] == '/') {
-		fprintf(stderr, "Error: Remote name %s contains absolute path.\n", remote_name);
+		fatal("Error: Remote name %s is an absolute path.\n", remote_name);
 		return 0;
```
nhazekam: Should I remove the return value as it fatal should exit?

PR #1376, opened 2016-06-10
Do not submit new workers until old workers connect.

btovar:
This helps with two issues:

- A master may have reach its capacity, and new workers do not have a
  chance to connect to it anyway.

- A worker job may be idle (e.g., condor) or in the queue (e.g., sge),
  and submitting new workers is counterproductive. In particular,
  consider a master that does not have many tasks to run anymore. As old
  workers time-out, the condor idle jobs start to run and connect to the
  master. They are never given tasks, and in turn time out, which is a
  waste of resources.


dthain:
Per our discussion the other day, this one makes sense to me.
Ok to merge?

PR #1378, opened 2016-06-16
Flatten resource monitor output in Condor

nhazekam:
@btovar This is attempt 2 take a look and let me know what you think. #1364

### DIFF
--- a/batch_job/src/batch_job.c
+++ b/batch_job/src/batch_job.c
@@ -75,6 +75,7 @@ struct batch_queue *batch_queue_create(batch_queue_type_t type)
 
 	batch_queue_set_feature(q, "local_job_queue", "yes");
 	batch_queue_set_feature(q, "absolute_path", "yes");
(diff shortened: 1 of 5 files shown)


btovar:
Nick, works for me, but it seems there are some conflicts with the current master.

nhazekam:
Thanks for pointing that out, had not noticed.

nhazekam:
Before we merge this, because the directory, if one is used, is created by the monitor wrapper should I delete this in clean? It is not currently logged or tracked.

btovar:
If we create the directory (was not there before) and if after -c the directory is empty, then yes, we should remove it.

nhazekam:
I'll work on adding that handling.

btovar:
Nick is this ready to be merged? Did you test it?

nhazekam:
Yes, just for documentation, if the directory doesn't exist Makeflow will make it, if Makeflow created it Makeflow will delete it.

btovar:
Great, thanks Nick

PR #1379, opened 2016-06-17
Add ability to add lists and sets to sets, also set duplicate

nhazekam:
Enhancement to sets to allow for duplicating and inserting sets and lists.

### DIFF
--- a/dttools/src/set.h
+++ b/dttools/src/set.h
@@ -100,7 +100,7 @@ This call will return 1 if all elements of s2 exist or are added to the set.
 Also note that you cannot insert a null set into the set.
 @param s A pointer to a set.
 @param s2 A pointer to a set to be inserted.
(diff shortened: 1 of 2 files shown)


dttools/src/set.h
```diff
@@ -47,16 +48,31 @@ while(element = set_next_element(s)) {
 
 struct set *set_create(int buckets);
 
+/** Duplicate a set from an existing set.
```
btovar: For all these new functions, please add a warning in the documentation saying that pointers are not duplicated, and that one has to watch for double frees.

nhazekam: You are referring to the pointers of the elements in the set, correct?

nhazekam: Is the NOTE I added sufficient, only should it be more informative?

PR #1381, opened 2016-06-17
Fix dag file sizes

nhazekam:
Update variable names and add file size information to dag_file.

Mechanism for setting estimated size: 
FILE=something
FILESIZE=estimatedsize

also required a small fix to prevent segfault when setting the same variable several times in the same rule/node context.

Also, cleans up makeflow logging to be slightly more consistent.

### DIFF
--- a/makeflow/src/dag_file.c
+++ b/makeflow/src/dag_file.c
@@ -9,6 +9,7 @@ See the file COPYING for details.
 
 #include "xxmalloc.h"
 #include "list.h"
+#include "macros.h"
 
 #include <stdlib.h>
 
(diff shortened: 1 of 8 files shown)


dthain:
Per our discussion today, let's not overload the meaning of variables. Instead, please introduce a new syntax like this:

```
SIZE filename size
```

Or perhaps:

```
#SIZE filename size
```

Whatever we do here will be imperfect but better handled in the json format.

@btovar any suggestions here?

btovar:
I like

SIZE filename size

as I do not like comments that aren't really comments. We can do the same for resources.

However, to make parsing easier, I would say:

.SIZE filename size
.CORES 4

etc.

nhazekam:
I am removing the SIZE parsing from this pull request and opening a new one. #1383

nhazekam:
@btovar Could you give this a look now that the FILESIZE section is removed. This will be added back in a later pull request.

nhazekam:
@btovar Once this is merged I will create the directives branch for specifying resources and file sizes.
Let me know if anything from this needs to be changed.

btovar:
Nick, it looks good.

PR #1390, opened 2016-06-23
disk loop allocation cleanup

btovar:
@nkremerh please review (and test).


nkremerh:
I believe we chose to stick with the "disk allocation" terminology over "loop device" due to semantics (see #1238) because we may use methods of _disk allocation_ other than loop devices in the future. What do you think?

btovar:
Uhm, forgot about that.

btovar:
what about know?

nkremerh:
Looks great! Tested for functionality, looks good to me.

PR #1391, opened 2016-06-27
Fixes help message to match disable-cache implementation

nkremerh:

PR #1397, opened 2016-06-30
Work Queue Disable Caching

nkremerh:


nkremerh:
@btovar @nhazekam What do you think?

work_queue/src/work_queue.c
```diff
@@ -950,11 +952,15 @@ high throughput and low latency.
 static work_queue_result_code_t get_file_or_directory( struct work_queue *q, struct work_queue_worker *w, struct work_queue_task *t, const char *remote_name, const char *local_name, int64_t * total_bytes)
 {
 	// Remember the length of the specified remote path so it can be chopped from the result.
-	int remote_name_len = strlen(remote_name);
 
 	// Send the name of the file/dir name to fetch
 	debug(D_WQ, "%s (%s) sending back %s to %s", w->hostname, w->addrport, remote_name, local_name);
-	send_worker_msg(q,w, "get %s 1\n",remote_name);
+	if(disable_cache) {
```
dthain: Yikes -- this is the wrong way to go about it.
Even when files are not cached between jobs, they still need to into the cache directory,
so that the worker can name them into the right place, and then account for the storage.
The master will still delete the files when the task is done.

dthain:
Ok, the problem here is that you are trying to disable caching by modifying the guts of WQ.
WQ already has the ability to control whether individual files are cached or not, by passing `WORK_QUEUE_CACHE` or `WORK_QUEUE_NOCACHE` in the flags to `work_queue_task_specify_file`.

If your goal is to simply disable caching in Makeflow+WQ , all you need to do is edit `batch_job_work_queue.c` and make sure that `WORK_QUEUE_CACHE` is not used.  In fact, it looks like there is already some code to that effect in there, and just needs to be turned on.

If your goal is to disable caching in WQ generally, then you simply need to suppress the `WORK_QUEUE_CACHE` flag somewhere in `work_queue.c`.

nkremerh:
@btovar @nhazekam Would either of you like to take a look at this?

makeflow/src/makeflow.c
```diff
@@ -1370,7 +1370,7 @@ int main(int argc, char *argv[])
 				}
 				break;
 			case LONG_OPT_DISABLE_BATCH_CACHE:
-				cache_mode = 0;
+				cache_mode = 1;
```
btovar: cache_mode = 1  should be with cache on, otherwise it gets confusing with the double negatives.

makeflow/src/makeflow.c
```diff
@@ -1571,7 +1571,7 @@ int main(int argc, char *argv[])
 	batch_queue_set_option(remote_queue, "priority", priority);
 	batch_queue_set_option(remote_queue, "keepalive-interval", work_queue_keepalive_interval);
 	batch_queue_set_option(remote_queue, "keepalive-timeout", work_queue_keepalive_timeout);
-	batch_queue_set_option(remote_queue, "caching", cache_mode ? "yes" : "no");
+	batch_queue_set_option(remote_queue, "disable-cache", cache_mode ? "yes" : "no");
```
btovar: Why change to disagle-cache? I rather have enable-cache ? yes, no. It is easier to think without the double negation.

makeflow/src/makeflow.c
btovar: Also, note that for WQ, WORK_QUEUE_NOCACHE is 0, while CACHE is 1. Not that it matters, as it is an implementation detail of the constant, but it is better to keep this consistent.

nkremerh:
This was resolved in  #1391. This was simply a documentation issue.

Issue #1383, opened 2016-06-20
Makeflow Directives


nhazekam:
Directives is the name of special variables that direct the execution of nodes. This is a manner of expressing information that is different from normal variables. The need for this comes from wanting to inform a dag/node about information relevant to execution such as filesizes, resources, and mounts.

Prior to this nodes were able to specify resources using variables as each resource had a single value, CORES=4, MEMORY=1024... 

However, in the case of filesizes and mounts there is additional information that needs to pulled from the line.

The proposal is to add .NAME directives so that this information can be parsed in the manner appropriate with the type of directive.

examples:
.SIZE filename size
.MOUNT source target
.CORES number  --or-- .RESOURCES CORES number

@dthain @btovar Thoughts?

btovar:
I like .RESOURCES CORES X

nhazekam:
This is now implemented in #1399

nhazekam:
This is implemented and merged in #1399.

PR #1399, opened 2016-07-05
Makeflow directives

nhazekam:
This introduces directives, name courtesy of @btovar, which provides a flexible way to specify Makeflow specific information and parse out more tokens for a given type of directive. In this pull request .RESOURCE and .SIZE were added, with the possibility of .MOUNT or others being added later.

@btovar @dthain give it a look and let me know what you think.


dthain:
A quick sanity check -- Where we find it necessary to extend the Makeflow syntax, it should not unduly conflict with existing Make syntax.

Is this consistent with the use of directives like `.FORCE` in GNU Make?

nhazekam:
They don't appear to be the same. .FORCE and .PHONY are ways of forcing rules to run or always made (simplification), whereas the directives in this case are used to inform Makeflow of information about nodes, files, or the DAG in general. They do not affect if/when rules are executed.

btovar:
But they are consistent, in the sense that they inform the make program, rather than the dag. This is why we chose the . (dot) notation.

dthain:
Ok, Ben addressed my concern.

btovar:
.RESOURCE notaresource 123 should generate an error.

.NOTADIRECTIVE
does generate an error, but the error message has garbage, probably a memory error.

.RESOURCE CORES aword
works, but it shouldn't, as we are expecting a number.

nhazekam:
I will address the second and see if I can make the error more clear. As for the first and the third, .RESOURCE uses the current logic for variables, which allows both. I can add handling that only permits certain resource specifications {CORES, DISK, MEMORY} and types{ints}, but it currently doesn't as variables are generic.

btovar:
Since .RESOURCE SOMETHING has more of a syntax flavor, let's generate an error for those we do not recognize.

nhazekam:
@btovar I add more syntax information and also added a couple of tests. let me know if this is sufficient or not quite the expected cases.

btovar:
Nick, ready to merge if you are.

nhazekam:
Resolve #1383

PR #1400, opened 2016-07-05
Added standalone catalog query tool.

dthain:


dthain:
Botched pull request.

PR #1402, opened 2016-07-07
fix logic for cache flag in bindings

btovar:
Under some conditions, the logic in the bindings would always cache files.

Discovered by  @WenzhaoLi and @annawoodard.


dthain:
Aha, was that the cause of the timed-out X509 proxy?

btovar:
Indeed!

PR #1409, opened 2016-07-19
small fixes for resource monitor histograms

btovar:


btovar:
.RESOURCE notaresource 123   should generate an error.

.NOTADIRECTIVE 
does generate an error, but the error message has garbage, probably a memory error.

.RESOURCE CORES aword
works, but it shouldn't, as we are expecting a number.

nhazekam:
@btovar Is this intended for #1399 ?

btovar:
no... why?

nhazekam:
.RESOURCE is not integrated in master or this pull request. These may be issues with the referenced pull request, but I don't think this applies to RM Histograms.

btovar:
My mistake, I commented in the wrong pr.

Issue #1396, opened 2016-06-30
Feature: Modify Makeflow to send catalog updates.


dthain:
Modify Makeflow to send its own periodic updates to the catalog server, indicating the number of jobs that are blocked, ready, running, failed, and complete, with a report type of "makeflow". 

Note that this report is distinct from the WQ level report, which only sees things relevant to WQ.  The Makeflow level report should be sent whenever the -N option is given, regardless of whether WQ is enabled.  So, the options "-N projectname -T condor" would indicate that jobs should be sent to condor, while makeflow level reports are sent to the catalog.

Nekel-Seyew:
Implemented, and created a pull request to merge my working copy with repo.

PR #1412, opened 2016-07-21
Addition of Feature Request: Modify Makeflow to send catalog updates (issue #1396)

Nekel-Seyew:
I created two files to store the additional code, which really consists of just one function. I then added code to the makeflow_run method so that at the beginning of the method it sends an update, then once every minute it sends an update, then at the end, it sends an update all to the catalog server. 

I also added to .gitignore the autogenerated files from my IDE, so as to not clutter up the repo.

I then added my files to the makefile requirements.

### DIFF
--- a/makeflow/src/.gitignore
+++ b/makeflow/src/.gitignore
@@ -2,3 +2,4 @@ makeflow
 makeflow_analyze
 makeflow_linker
 makeflow_viz
+nbproject/
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -40,6 +40,8 @@ See the file COPYING for details.
 #include "parser.h"
 #include "parser_jx.h"
 
+#include "makeflow_catalog_reporter.h"
+
 #include <fcntl.h>
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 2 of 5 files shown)


makeflow/src/makeflow_catalog_reporter.c
```diff
@@ -0,0 +1,75 @@
...
+This software is distributed under the GNU General Public License.
+See the file COPYING for details.
+*/
+
+#include "dag.h"
+#include "timestamp.h"
+#include "stringtools.h"
+#include "list.h"
+#include "makeflow_summary.h"
+#include "../../dttools/src/catalog_query.h"
```
dthain: I believe you can write this simply as `catalog_query.h` and the build system will set the paths correctly.

makeflow/src/.gitignore
```diff
@@ -2,3 +2,4 @@ makeflow
 makeflow_analyze
 makeflow_linker
 makeflow_viz
+nbproject/
```
dthain: Noise?

makeflow/src/makeflow.c
```diff
@@ -874,7 +880,13 @@ static void makeflow_run( struct dag *d )
 	struct dag_node *n;
 	batch_job_id_t jobid;
 	struct batch_job_info info;
+        timestamp_t last_time = timestamp_get();
+        int first_report = 1;
 
+        //reporting to catalog
+        if(in_n_mode){
```
dthain: I think you can simplify to `if(project)`

dthain:
Looks pretty good.  A couple more things to add to the report:
- Time at which the makeflow started.
- Batch system type (local, condor, etc)

Then next question: how will the end user view these reports?  Maybe `makeflow_status`?

makeflow/src/makeflow.c
Nekel-Seyew: I had uncertainty, because project is set some place other than just in the -N flag, so I wanted to have a separate, known variable that will only be true if the -N flag is passed

Nekel-Seyew:
Closed due to errors from timestamp formatting.

Issue #1410, opened 2016-07-20
[Parrot] Parrot_run fails with full IceCube simulation


bgharper:
@dthain @hmeng-19 

It seems to occur fairly randomly, as sometimes it will successfully run and trace (as it did when I created the package). 

```

sh-4.1$ ./fullSimulation.py 
2016/07/20 15:05:15.71 parrot_run[506691] <child:506871> fatal: cannot run application which does not allow tracing (CLONE_UNTRACED)
2016/07/20 15:05:15.71 parrot_run[506691] <child:506871> notice: received signal 15 (Terminated), killing all my children...
2016/07/20 15:05:15.71 parrot_run[506691] <child:506871> notice: sending myself 15 (Terminated), good
bye!
 Terminated
```

Attached is a zip of the fullSimulation.py script, although it requires an icecube setup on CVMFS. [Here](https://drive.google.com/open?id=0B9jyC8sqxSFpLVZVOFdERlFuemM) is a link to the package I created in one of the successful runs (~250mb/650mb uncompressed). I recommend running as root to avoid any gid/uid issues. 

[fullSimulation.zip](https://github.com/cooperative-computing-lab/cctools/files/374598/fullSimulation.zip)

dthain:
Ok, the first thing we need to do here is understand what application is setting `CLONE_UNTRACED`, which is an unusual request to make in the first place...

bgharper:
Dumping an strace of a regular run shows the following calls to clone:
`clone(child_stack=0, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7f207cde49d0) = 508235
`
(repeated twelve times because num cpus = 12, with same flags, just different addresses and PIDs)

Where it doesn't look like it's being set by any of them by default. Dumping an strace on a (failed) parrot_run shows:

```
[billobob@uc3-c007 icecube]$ cat strace.parrot | grep -C 3 CLONE
open("/tmp/parrot.53362/parrot-instance.3c4sPk/parrot-fd.cLib53", O_RDONLY|O_DIRECTORY) = 5
dup2(5, 4094)                           = 4094
close(5)                                = 0
clone(child_stack=0, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7ff64374dab0) = 509874
[...]

PR #1413, opened 2016-07-21
Fix Parrot Clone Bug

dthain:
Only check cone flags on a real clone, not on fork.


dthain:
May fix #1410

batrick:
Don't forget 32 bit! ;)

PR #1414, opened 2016-07-22
Addition of Feature Request: Modify Makeflow to send catalog updates and makeflow_status (issue #1396)

Nekel-Seyew:
Implemented the request in issue #1396 as well as implementing the makeflow_status program which retrieves the sent information from the catalog server.


dthain:
Ok, please ask @nhazekam to do a code review.
And, don't forget to update `doc/makeflow.html` and add a manpage for `makeflow_status`.

nhazekam:
I'll take a look at it later today.

doc/makeflow.html
```diff
@@ -212,6 +212,12 @@ <h3 id="workqueue.projectnames">Project Names<a class="sectionlink" href="#workq
 
 <code>work_queue_worker -N proj1 -N proj2 -N proj3</code>
 
+<p>
+In addition to creating a project name using the -N option, this will also trigger automatic reporting to the 
+designated catalog server. The Port and Server address are taken from the environment variables <b>CATALOG_HOST</b> 
+and <b>CATALOG_PORT</b>. If either variable is not set, then the addresses "catalog.cse.nd.edu,backup-catalog.cse.nd.edu" and port 9097 will be used.
+</p>
+
```
dthain: Suggest that you clearly separate two things in the docs:
- Indicate earlier on that -N can be used with any batch system type.
- Indicate here that -N is the easiest way to match workers to masters.

dttools/src/rmonitor_poll.c
```diff
@@ -723,6 +723,7 @@ char *rmonitor_get_command_line(pid_t pid)
 		if(cmdline[i] == '\0')
 			cmdline[i] = ' ';
 	}
+        fclose(fline);
```
Nekel-Seyew: working on it.

makeflow/src/makeflow_status.c
```diff
@@ -0,0 +1,246 @@
...
+#include "getopt_aux.h"
+#include "cctools.h"
+#include "xxmalloc.h"
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+/*static struct jx_table headers[] = {
```
dthain: The table headers looks like the right idea; any reason not to use it?

dthain:
@Nekel-Seyew this is almost done, so let's finish it.
Please answer my question above.

Nekel-Seyew:
I didn't use the js_table headers because at first they were not outputting the right information, and I decided that I could handle the information output myself with printf.

dthain:
That's fine, then remove the dead code that is commented out...

makeflow/src/makeflow_catalog_reporter.c
```diff
@@ -0,0 +1,86 @@
...
+    
+    int tasks_completed = 0;
+    int tasks_aborted   = 0;
+    int tasks_waiting   = 0;
+    int tasks_running   = 0;
+
+    for (n = d->nodes; n; n = n->next) {
+        state = n->state;
+        if (state == DAG_NODE_STATE_FAILED && !list_find(failed_tasks, (int (*)(void *, const void *)) string_equal, (void *) fn))
+            list_push_tail(failed_tasks, (void *) n->command);
```
dthain: If I read this correctly, you are keeping a list of failed tasks, so as to de-duplicate them.
I think you can just do `tasks_failed++` here, since each makeflow node is already unique.
(It is possible for two different tasks to have the same command string, since the file namespace can vary per task.)

makeflow/src/makeflow_catalog_reporter.c
```diff
@@ -0,0 +1,86 @@
...
+    
+    //transmit report here
+    //creates memory
+    char* host = string_format("%s:%i",CATALOG_HOST,CATALOG_PORT);
+    
+    char username[USERNAME_MAX];
+    username_get(username);
+    
+    timestamp_t now= timestamp_get();
+    //char timestamp[25];//uint64 largest number is only 21 digits long, give it 25 for good measure
```
dthain: Dead code

makeflow/src/makeflow_catalog_reporter.c
Nekel-Seyew: This was the method used in the other makeflow_reporter tool, which is why I preserved this form. But I can simplify and eliminate needing to use the list.

makeflow/src/makeflow_catalog_reporter.c
Nekel-Seyew: This code doesn't actually reflect when the tasks were started. Going to fix.

dthain:
@trshaffer will this work with the new multi-catalog update approach?

doc/makeflow.html
Nekel-Seyew: the second point is indicated earlier on lines 188-189.

makeflow/src/makeflow.c
```diff
@@ -874,7 +880,14 @@ static void makeflow_run( struct dag *d )
 	struct dag_node *n;
 	batch_job_id_t jobid;
 	struct batch_job_info info;
+        timestamp_t last_time = timestamp_get();
+        timestamp_t start = timestamp_get();
+        int first_report = 1;
 
+        //reporting to catalog
+        if(in_n_mode){
```
trshaffer: For someone reading the code, it's not immediately obvious what "N mode" is. Rather than commenting, it might be helpful to give this variable a more descriptive name like `send_reports` or something similar

makeflow/src/makeflow_catalog_reporter.c
```diff
@@ -0,0 +1,72 @@
...
+    //creates memory
+    char* host = string_format("%s:%i",CATALOG_HOST,CATALOG_PORT);
+    
+    char username[USERNAME_MAX];
+    username_get(username);
+    
+    const char* batch_type = batch_queue_type_to_string(type);
+    
+    //creates memory
+    char* text = string_format("{\"type\":\"makeflow\",\"total\":%i,\"running\":%i,\"waiting\":%i,\"aborted\":%i,\"completed\":%i,\"failed\":%i,\"project\":\"%s\",\"owner\":\"%s\",\"time_started\":%" PRIu64 ",\"batch_type\":\"%s\"}",
```
trshaffer: Is there a reason this needs to format the JSON itself? I'd prefer to build a JX structure and dump the string representation. That way, we can avoid duplicating serialization code

makeflow/src/makeflow_catalog_reporter.c
```diff
@@ -0,0 +1,72 @@
...
+            tasks_completed++;
+        else if(state == DAG_NODE_STATE_RUNNING)
+            tasks_running++;
+        else if(state == DAG_NODE_STATE_WAITING)
+            tasks_waiting++;
+    }
+    
+    //transmit report here
+    //creates memory
+    char* host = string_format("%s:%i",CATALOG_HOST,CATALOG_PORT);
```
trshaffer: There's no need to add the port yourself. Since we have two servers in the default `CATALOG_HOST`, this actually just specifies a port for the second one. `CATALOG_PORT` is somewhat deprecated; it's just the default port to use if none is specified in one of the entries in `CATALOG_HOST`. You can just pass `CATALOG_HOST` directly to `catalog_port_send_update`, it'll figure out the ports.

makeflow/src/makeflow_status.c
```diff
@@ -0,0 +1,268 @@
...
+int first_comma(char* in){
+    int i=0;
+    int max = strlen(in);
+    for(i=0; i<max; ++i){
+        if(in[i] == ',')
+            return i;
+    }
+    return -1;
+}
+char* add_port(char* in, int port){
```
trshaffer: I see you've noticed the issues with a single port option applied to multiple hosts. It's best to avoid parsing the host list yourself, there's logic elsewhere that handles it, so you can just treat it as an opaque string.

makeflow/src/makeflow_status.c
```diff
@@ -0,0 +1,268 @@
...
+            case 't':
+                timeout = string_time_parse(optarg);
+                break;
+            case 'u':
+                username = xxstrdup(optarg);
+                break;
+            case 's':
+                server = xxstrdup(optarg);
+                break;
+            case 'p':
```
trshaffer: I don't think there should be a `-p` option as it doesn't play nicely with multiple catalog servers. The `-s` option could hold a host list, then you don't have to worry about ports. The user can choose any host/port combos, e.g.
- `"server1"`
- `"server1:2000"`
- `"server1,server2:2232,server3"`

makeflow/src/makeflow_catalog_reporter.c
Nekel-Seyew: It was written as JSON because at the time, 1) I wasn't too familiar with JX, 2) Knew that JX was a superset of JSON, so JSON was a fine way to go, 3) Wasn't sure what all to send at first, so writting it out by hand was the fastest/easiest way to go.

I can convert it to JX if that fits better.

makeflow/src/makeflow_catalog_reporter.c
Nekel-Seyew: Oh, good. Will change that then. But shouldn't CATALOG_PORT be refactored out of the code then, since it's not used anymore?

trshaffer: It is used, but mostly as a way for users to set the default port. When sending updates or making catalog queries, we just pass around a host list without parsing it. The code in `catalog_query.c` grabs a default from `CATALOG_PORT` if the host list didn't specify one

makeflow/src/makeflow_catalog_reporter.c
trshaffer: As long as you only use JSON types, the JX will be plain JSON. If you didn't see it already, have a look at `jx_print.h`

makeflow/src/makeflow.c
Nekel-Seyew: I can change it to `n_switch_on` because the `-n` option activates two things.

trshaffer: It's not the wording, just that "N switch" or "N mode" doesn't relate to anything in the code beyond the option parsing. If the `-N` flag controls two things, it might be clearer to have two variables, even if they're simply set together

nhazekam:
Also the binary should not be included in the pull request. Also add to .gitignore to prevent in the future.

makeflow/src/.gitignore
```diff
@@ -2,3 +2,4 @@ makeflow
 makeflow_analyze
 makeflow_linker
 makeflow_viz
+nbproject/
```
nhazekam: Update for the executable

makeflow/src/.gitignore
```diff
@@ -2,3 +2,5 @@ makeflow
 makeflow_analyze
 makeflow_linker
 makeflow_viz
+makeflow_status
+nbproject/
```
nhazekam: Please remove nbproject

PR #1416, opened 2016-07-22
Histogram data structure

btovar:
Factor out the histogram code from category.c into its own data structure.

I still need to change category.c and resource_monitor_histograms.c


nhazekam:
Looks like a good abstraction. Aside from the test you added, does this work as you intended? @btovar

PR #1422, opened 2016-07-27
Fixes remote naming conflict bug in All-Pairs

### LINKED ISSUE #1421: Bug: Remote naming conflicts in All-Pairs
Currently, All-Pairs shortens all input files such that only the base name of the file is left (./my/dir/a.txt simply becomes a.txt). This can introduce remote name conflicts in Work Queue if two files have the same name but are in different directories.

nkremerh:
Fixes #1421


dttools/src/text_list.c
```diff
@@ -75,4 +76,27 @@ void text_list_delete(struct text_list *t)
 	free(t);
 }
 
+struct text_list *text_list_allpairs_remote_create(const char *path, const char *set)
```
dthain: I'm going to nitpick here.  `text_list` is a general purpose data structure, so it should include any application specific code.  This function is very specific to the allpairs tools, so it should be a module in the allpairs directory and invoke `text_list_create/append` externally.

PR #1423, opened 2016-08-04
Close chirp binding

btovar:
As reported by @matz-e. 

Chirp clients keep the connection open once the python object was destroyed.

### DIFF
--- a/chirp/src/chirp_reli.c
+++ b/chirp/src/chirp_reli.c
@@ -135,7 +135,7 @@ static INT64_T connect_to_file( struct chirp_client *client, struct chirp_file *
 	return 1;
 }
 
-static void invalidate_host( const char *host )
+void chirp_reli_disconnect( const char *host )
 {
 	struct chirp_client *c;
 	c = hash_table_remove(table,host);
@@ -173,7 +173,7 @@ struct chirp_file * chirp_reli_open( const char *host, const char *path, INT64_T
 			} else {
 				if(errno!=ECONNRESET) return 0;
 			}
-			invalidate_host(host);
+			chirp_reli_disconnect(host);
 		} else {
 			if(errno==ENOENT) return 0;
 		}
@@ -224,7 +224,7 @@ INT64_T chirp_reli_close( struct chirp_file *file, time_t stoptime )
 				if(result>=0 || errno!=ECONNRESET) return result; \
 			} \
 			if(errno==ESTALE) return -1; \
-			invalidate_host(file->host); \
+			chirp_reli_disconnect(file->host); \
 		} else { \
 			if(errno==ENOENT) return -1; \
 			if(errno==EPERM) return -1; \
@@ -454,7 +454,7 @@ INT64_T chirp_reli_fsync( struct chirp_file *file, time_t stoptime )
 			if(result>=0) { \
 				return result; \
 			} else if (errno == ECONNRESET) { \
-				invalidate_host(host); \
+				chirp_reli_disconnect(host); \
 			} else if (errno == EAGAIN) { \
 				if (_NOEAGAIN) \
 					return result; \
@@ -983,7 +983,7 @@ static INT64_T chirp_reli_bulkio_once( struct chirp_bulkio *v, int count, time_t
 	failure:
 	for(i=0;i<count;i++) {
 		struct chirp_bulkio *b = &v[i];
-		invalidate_host(b->file->host);
+		chirp_reli_disconnect(b->file->host);
 	}
 	errno = ECONNRESET;
 	return -1;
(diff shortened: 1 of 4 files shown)


dthain:
But is this really a problem?

If the process attempts another operation on the same host, the client library will re-use the existing connection, if possible.  No more than one connection will exist to a given remote host.

matz-e:
For us, it was daemonizing…

I would, from a more formal standpoint, argue that this would also be closer to expected behavior (I was surprised to see that connections were cached):

* when I open a file in e.g. python, I expect it to be closed again when the file object drops out of scope (using the `with` statement).
* when I use an interface to create a connection I would also expect that the connection is closed when the object drops out of scope.
* if that is not the case, I would at least expect that there is a way to manually close the connection opened.

btovar:
I agree that Chirp is improved by the current pr. However, I still don't understand why would chirp write to your database. Are you by any chance using fixed file descriptors somewhere?

matz-e:
We don't use fixed file descriptors, as far as I know (may have to dig a bit into the CMS provided software we use.

I've used another python package to track open files and connections, and the integer file descriptor that is used by Chirp gets re-used by the database after Lobster turns into a daemon. As in, if I'm listing open files before becoming a daemon and afterwards, I'm seeing the same integer assigned to a connection to `deepthought` and to `lobster.db`, respectively.

I'm currently wondering how you would notice in C that a file descriptor has been closed?

dthain:
So, if the python package is tracking the fds opened and closed by *python* itself, then it won't see the fds used by the underlying C libraries.  And keep in mind that there isn't necessarily a one-to-one mapping between Chirp python objects and fds.  The same connection to a server will be shared by multiple open files since they are relatively expensive to create.  And, a single Chirp object might consume multiple fds, if it needs to contact a catalog server, open a multi-volume, etc...

matz-e:
* With the python package that I'm using to look at open fds, I can see Chirp's C file descriptors.  This is purely observatory, I do not do anything with the file-descriptors but complain to the user that there are open ones (and crash).
* (Looked at the source for this) The `python-daemon` package actually uses a "low-level" (pydoc) `os.close` on *all* file descriptors up to the limit set on the OS or 2048 if the limit is infinite. This does not rely on tracking at all.
* As a user of an object-oriented language, I did not expect Chirp to cache and share connections… I'm aware that no such guarantee was explicitly made by Chirp, though.

dthain:
Wait a minute -- are any Chirp operations getting called before `daemonize` ?

batrick:
I think the essence of the issue here is that the Python binding is overlaying a "server connection object" on top of Chirp reli. This is probably not the right abstraction to put on top of chirp_reli...

@matz-e: with this change, if you have two Client objects and close one of them, the other Client's connection gets closed too (because they "share" the same connection in chirp_reli.cc). [Now, chirp_reli.cc is smart enough to reconnect to the Chirp server and reopen file descriptors but that's beside the point.]

btovar:
That's not really a big deal. I was thinking of adding a counter to the python class, and only close the  link when the counter goes to zero.

matz-e:
@batrick: Also, @btovar mentioned that the connections get reset whenever the authentication changes.  So having the connection temporarily closed is somewhat consistent.

I'd prefer every python object having it's own connection, but since I'm not going to submit a PR for that, I can live with how it is currently.

btovar:
What if I rename Client to Connection, and closing a connection needs to be called explicitly by the programmer?  Creating a new connection again simply resets the connection, as it is now. This would make the abstraction more clear: only one connection.

Another option is to tag the host with a label, such as "<client:SOMENUMBER>host:port". When connecting, we discard the <client:...> part, but we use the whole string inside the hash table of chirp reli. If a host does not have such label, it is treated as before. As far as I can see, only the connection function should be modified in this way.

dthain:
Guys, I think we are over-solving the problem here.

The `chirp_reli` interface was designed to hide connection state, precisely because connection management is an ugly topic.  Connections are slow and expensive, they can fail, must be reconnected, difficult to share correctly, etc.  All of that works and is safely hidden from the user, who simply creates a client object and then accesses files without regard to connections.  And, you get efficient behavior, even if the user creates multiple objects, creates and deletes them rapidly, etc.

Now, there is a small cost to doing this, which Matthias has observed: there is no longer a direct relationship between the client object and the kernel level fds consumed.  If an extra fd hanging around is a problem, then the new function introduced by Ben cleans it up.

Trust me on this -- adding more state management on top of this is just going to make things more complex, with no benefit to the user.  We have the right solution now.

btovar:
Works for me! If the current pr works for @matz-e, then I can merge it and close the issue.

matz-e:
It solves the most recent "somewhat" immediate failure. We still have a similar issue that I need to track down.

Issue #1415, opened 2016-07-22
Update Parrot to New libcvmfs


dthain:
Update `pfs_service_cvmfs.cc` to use the new cvmfs API:
https://github.com/cvmfs/cvmfs/pull/1634

Key bit: Set up a way to pass in arbitrary cvmfs options from the command line

dthain:
FYI @trshaffer @btovar @jblomer

trshaffer:
My preference would be to use JSON to pass CVMFS options into Parrot as key/value pairs. This would fit well with the v2 API, and avoid parsing and manipulating large option strings. Could we deprecate most of Parrot's CVMFS-related options and use the v2 API if it's available and the user gives a JSON config, else fall back to the existing config syntax? This should maintain backwards compatibility and encourage users to switch to JSON config

dthain:
That works for me.

dthain:
Finished some time ago.

PR #1425, opened 2016-08-09
Use new libcvmfs interface if available

trshaffer:
@jblomer added a new interface for option handling and initialization for libcvmfs rev. 23, so this PR updates our Parrot connector. I *think* this works for any libcvmfs version. I also made some changes on our end. I'd like to pass configuration to libcvmfs as directly as possible, so I marked some of the Parrot specific or legacy parts as deprecated. New users can use any desired options from the [libcvmfs docs](http://cvmfs.readthedocs.io/en/latest/cpt-configure.html). Aside from Parrot-internal options like repo switching, `--cvmfs-option` and `--cvmfs-option-file` should be preferred for configuring CVMFS in Parrot, since they use the same syntax as the FUSE module and libcvmfs.

Also, is there a reason Parrot explicitly disallows direct connections and requires a proxy?

### DIFF
--- a/parrot/src/pfs_main.cc
+++ b/parrot/src/pfs_main.cc
@@ -46,6 +46,7 @@ extern "C" {
 #include "tracer.h"
 #include "xxmalloc.h"
 #include "hash_table.h"
+#include "jx.h"
 }
 
 #include <fcntl.h>
@@ -124,6 +125,8 @@ bool pfs_cvmfs_repo_switching = false;
 char pfs_cvmfs_alien_cache_dir[PATH_MAX];
 char pfs_cvmfs_locks_dir[PATH_MAX];
 bool pfs_cvmfs_enable_alien  = true;
+char pfs_cvmfs_option_file[PATH_MAX];
+struct jx *pfs_cvmfs_options = NULL;
 
 int pfs_irods_debug_level = 0;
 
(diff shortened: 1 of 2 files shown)


trshaffer:
See #1415

dthain:
Parrot explicitly requires a proxy b/c the main repository operators are not set up to handle direct access traffic.  This protects against the situation where a relatively small number of direct clients could ruin performance for everyone by slamming the central repo.

trshaffer:
OK, just checking if it's still needed. The proxy check happens on both old and new libcvmfs now.

Should there be a log warning if a user tries the new option args with old libcvmfs? Also, it seems like Parrot is running out of short argument letters

dthain:
Well, as a general rule, combinations that don't work at all should result in an error.  Things that work but are not preferred should result in a warning.  So, if the new option format would not work at all with the old libcvmfs, that should be an error.

It's ok if the options are only in long format without a short letter.

btovar:
Tim, is this ready to be merged?

trshaffer:
I don't know of any outstanding issues, and it's been working well for my benchmarks here, but I'd still like to verify that nothing changed for the old library interface. If there's no hurry, I'd like to go over it again and test it next week when I'm back

trshaffer:
All right @btovar, this seems to be behaving as expected under various libcvmfs/option combinations

PR #1432, opened 2016-08-17
Adds option to change Work Queue Factory cycle period

nkremerh:
@btovar What do you think?


btovar:
I think we still want to check everything every 30s, as the factory does other stuff in addition to just launching workers. You may want to add the option just for new workers.

dthain:
@nkremerh what's the problem you are trying to fix here?

Generally speaking, it's good to have a knob like this, in case our tools do something uncivilized at run time.  But, we don't want to encourage the user to query more frequently than about 30s -- batch systems just don't respond that quickly, and rapid queries tend to cause unnecessary load.

nkremerh:
Implementing this feature requires more changes to the underlying factory code, so consider this pull request closed and this idea tabled for now.

Issue #1198, opened 2016-03-14
Documentation for Makeflow linker needs to be updated


hmeng-19:
The documentation for Makeflow linker requires to be updated:
http://ccl.cse.nd.edu/software/manuals/makeflow.html#linking
It seems `-b` option has been removed from `makeflow`.

dthain:
Go right ahead!

hmeng-19:
@dthain , I do not think any current options of Makeflow aims to collect the dependencies of a given makeflow into a dir, as described http://ccl.cse.nd.edu/software/manuals/makeflow.html#linking .

hmeng-19:
@dthain , we can either close the issue for now, or remove http://ccl.cse.nd.edu/software/manuals/makeflow.html#linking from the Makeflow user manual page.

hmeng-19:
Fixed by 1435.

PR #1435, opened 2016-08-19
Fix doc for makeflow linking deps

hmeng-19:
The dependency tracking of makeflow has been removed from `makeflow` into `makeflow_analyze`.
This PR updates the Makeflow user manual to reflect this change.


hmeng-19:
@dthain , PTAL.

hmeng-19:
This should fix #1198 .

PR #1440, opened 2016-08-19
Mesos Support for Makeflow

charleszheng44:
This is the initial version of makeflow mesos mode. By enabling this mesos mode, makeflow can run on top of mesos clusters. 
1. Procedures is as follows
   - makeflow process the dags and write task detail to text file "task_info"
   - makeflow start the makeflow mesos scheduler (source can be found in `$CCTOOLS/bin/mf_mesos_scheduler`)
   - mesos scheduler read the task information and submit each task to mesos master
   - mesos master receive resource offer from mesos slaves
   - mesos master bundle resource offer with makeflow tasks
   - mesos master create makeflow mesos executor(source can be found in `$CCTOOLS/bin/mf_mesos_executor`) 
   - mesos master launch each task with a executor on mesos slaves
   - when tasks are complete, mesos executor inform mesos master and master write the status of finished tasks to text file "task_state" 
   - At the mean time, makeflow is keep pooling the "task_state" and wait for all tasks complete
2. Some problems of current version
   - Can not specified resource requirement for each task, it is easy to implement but we need to figure out a proper way to do it (specify the resource requirement based on makeflow categories?)
   - Each executor can only launch one task, may be we can reserve the executor on the slaves for launching multiple tasks to avoid the overhead?
   - To fit the mesos setting on disc cluster, I hard code some system variables in the `$CCTOOLS/bin/mf-mesos-executor`, I am trying to find a more general way to solve this problem


dthain:
Charles, this is coming along nicely, and you have the right approach.
I'm note a couple of small things along the way...

batch_job/src/batch_job_mesos.c
```diff
@@ -0,0 +1,379 @@
...
+#define NUM_OF_TASKS 4096
+
+static int counter = 0;
+static int finished_tasks[NUM_OF_TASKS];
+static int num_finished_tasks = 0;
+static time_t old_time;
+static int is_first_time = 1;
+
+// mesos task struct
+typedef struct mesos_task{
```
dthain: To be consistent with the rest of the cctools, please name the structure and function like so:

```
struct mesos_task {};
struct mesos_task * mesos_task_create( . . .);
void mesos_task_delete( struct mesos_task * );
```

You might even move the data type into `mesos_task.[ch]` to keep it separate.

batch_job/src/batch_job_mesos.c
```diff
@@ -0,0 +1,379 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <signal.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <time.h>
+
+#define FILE_TASK_INFO "task_info"
```
dthain: How about `.mesos_task_info` and `.mesos_task_state` so that we don't pollute the user's directory view?

batch_job/src/batch_job_mesos.c
```diff
@@ -0,0 +1,379 @@
...
+#define FILE_TASK_STATE "task_state"
+
+#define NUM_OF_TASKS 4096
+
+static int counter = 0;
+static int finished_tasks[NUM_OF_TASKS];
```
dthain: Use an `itable` here, so that the task numbers can grow arbitrarily.

batch_job/src/batch_job_mesos.c
```diff
@@ -0,0 +1,379 @@
...
+	int task_id;
+	char *task_cmd;
+	int num_input_files;
+	char **task_input_files;
+	int num_output_files;
+	char **task_output_files;
+} mesos_task;
+
+// get list of input files 
+char **build_str_lst_from_str(int *num_str, const char *str)
```
dthain: This looks a lot like `dttools/src/text_list.h` which loads an array of strings from a file.  Use that instead, just add a method to create a `text_list` from a string.

batch_job/src/batch_job_mesos.c
```diff
@@ -0,0 +1,379 @@
...
+	return 0;
+}
+
+// check is file is exist
+int is_file_exist(const char *path) 
+{
+	return access(path, F_OK) != -1;
+}
+
+// check the is file is modified since @old_time
```
dthain: Yikes, pass in old_time as a parameter, to minimize use of globals.

batch_job/src/batch_job_mesos.c
```diff
@@ -0,0 +1,379 @@
...
+int is_file_modified(const char *path) {
+	struct stat file_stat;
+	int err = stat(path, &file_stat);
+	if (err != 0) {
```
dthain: exit() seems like an extreme action here

batch_job/src/batch_job_mesos.c
```diff
@@ -0,0 +1,379 @@
...
+	int task_id = ++counter;
+
+	debug(D_BATCH, "task %d is ready", task_id);
+	struct batch_job_info *info = malloc(sizeof(*info));
+	memset(info, 0, sizeof(*info));
+	info->submitted = time(0);
+	info->started = time(0);
+	itable_insert(q->job_table, task_id, info);
+
+	FILE *fp_1;
```
dthain: Style: give this file pointer a name that reflects the purpose of the file.

batch_job/src/batch_job_mesos.c
```diff
@@ -0,0 +1,379 @@
...
+
+	destroy_mesos_task(mt);
+	fclose(fp_1);
+
+	return task_id;
+}
+
+static batch_job_id_t batch_job_mesos_wait (struct batch_queue * q, struct batch_job_info * info_out, time_t stoptime)
+{
+
```
dthain: Expand this into a more detailed comment (at the top of the file) that explains the relationship between batch_job_mesos and the script.

batch_job/src/batch_job_mesos.c
```diff
@@ -0,0 +1,379 @@
...
+					free(info);
+					fclose(fp);
+
+					finished_tasks[num_finished_tasks] = task_id;	
+					num_finished_tasks++;
+
+					return task_id;
+				}
+			}
+		//} else {
```
dthain: Remove dead code.

makeflow/src/makeflow.c
```diff
@@ -78,6 +79,7 @@ an example.
 */
 
 #define MAX_REMOTE_JOBS_DEFAULT 100
+#define MF_DONE_FILE "makeflow_done"
```
dthain: What's the purpose of this file?

makeflow/src/makeflow.c
```diff
@@ -1669,6 +1680,43 @@ int main(int argc, char *argv[])
...
 	}
 
+	if (batch_queue_type == BATCH_QUEUE_TYPE_MESOS) {
+		pid_t mesos_PID;
+		mesos_PID = fork();			
+		char *mesos_cwd;
+		mesos_cwd = path_getcwd();
+
+		char *mesos_py = "/afs/nd.edu/user37/ccl/software/external/mesos-0.26.0/amd64_linux26/lib/python2.6/site-packages";
+		char *cctools_path = getenv("CCTOOLS");
```
dthain: We don't generally encourage users to set the environment variable `CCTOOLS`.
Instead, let's use the compile-time install path, which is (I think) `CCTOOLS_INSTALL_DIR`

makeflow/src/makeflow.c
```diff
@@ -1669,6 +1680,43 @@ int main(int argc, char *argv[])
...
+		char *exe_py_path = string_format("%s/bin/mf_mesos_scheduler", cctools_path);
+		char *envs[] = {"LD_PRELOAD=/afs/nd.edu/user37/ccl/software/external/gcc-4.9.3/amd64_linux26/lib64/libstdc++.so.6:/afs/nd.edu/user37/ccl/software/external/svn-1.9.4/amd64_linux26/lib/libsvn_delta-1.so", "CCTOOLS=/afs/crc.nd.edu/user/c/czheng2/cctools", NULL};
```
dthain: Ok, this is what you needed to do in order to run mesos in our local environment.
But, I don't think it's the general case, esp when people install in /usr/bin.
I think you should consider this to be a setting peculiar to Notre Dame and just set it manually before you run Makeflow.

batch_job/src/batch_job_mesos.c
charleszheng44: Create `mesos_task.h` and `mesos_task.c` and remove the `typedef` keyword

batch_job/src/batch_job_mesos.c
charleszheng44: rename `task_[info\state]` to `mesos_task_[info\state]`

batch_job/src/batch_job_mesos.c
charleszheng44: remove the global variable `old_time`

batch_job/src/batch_job_mesos.c
charleszheng44: rename `fp_1` to `task_info_fp`

batch_job/src/batch_job_mesos.c
charleszheng44: Replace int array with `itable`

batch_job/src/batch_job_mesos.c
charleszheng44: Replace `char **build_str_lst_from_str(int *num_str, const char *str) ` with a new function `struct text_list *text_list_load_str(const char *inp_str)`

batch_job/src/batch_job_mesos.c
charleszheng44: remove useless `if...else...` clause

makeflow/src/makeflow.c
charleszheng44: add `mesos-path` option, which can be used by specify the path of mesos

makeflow/src/makeflow.c
charleszheng44: it is created by makeflow to inform the mesos scheduler that the workflow is done.

batch_job/src/batch_job_mesos.c
charleszheng44: remove the `is_file_modified(const char *path)` function

makeflow/src/makeflow.c
charleszheng44: Use `readlink("/proc/self/exe", buf, buf_size)` to get the path of `makeflow` executable, instead of using `CCTOOLS`

batch_job/src/batch_job_mesos.c
charleszheng44: Add comments at the beginning of the file

dthain:
Note: waiting for patch to substitute AFS cp with http put/get

dthain:
@charleszheng44 please talk to @btovar and figure out how to clean up this pull request and rebase it on master with only your commits.

dthain:
Superseded by #1587

Issue #1431, opened 2016-08-16
add a new field into an umbrella spec to mark the necessity of archiving it


hmeng-19:
Currently, `umbrella upload` uploads all the dependencies specified in an  Umbrella specification into OSF or S3. 

However, sometimes the user may want to only archive some deps with `umbrella upload`. 

Proposed plan:
Add a new optional field, named `upload`, into each dependency object to mark whether that dependency is needed to be archived or not. 
The `upload` field can be set to `true` (the default value) or `false`.

hmeng-19:
@dthain , PTAL.

dthain:
Ok with me

hmeng-19:
Fixed by #1441 .

PR #1441, opened 2016-08-19
Add `upload` field into umbrella spec

hmeng-19:
this boolean field can be used to mark whether a dependency should be uploaded into s3/osf or not.


hmeng-19:
This PR aims to fix #1431 .

hmeng-19:
@dthain , PTAL.

PR #1444, opened 2016-08-23
Add umbrella spec for an easy makeflow

hmeng-19:
This PR creates an Umbrella spec for an easy makeflow, the image conversion one.


hmeng-19:
@dthain , PTAL.

Issue #1320, opened 2016-05-05
Umbrella output freedom


pivie:
Would it be possible to allow Umbrella output files to be specified via the command line, even if they are not specified in the specification file? This would allow multiple tasks to use the same generic environment specification when the tasks are very similar. It would also allow a third party workflow manager (such as Prune) to add extra outputs to a task without having to modify the specification, again encouraging the reuse of generic environment specifications.

hmeng-19:
@pivie , as we discussed on Thursday, one way to avoid this problem is to add the directory, `D`, which includes all the potential outputs, into the `output` section. Then you can use the `--output` option of umbrella to obtain `D` and use the stuff under it.

hmeng-19:
The reason why we added the `output` section, together with its `files` and `dirs` subsections,  into the umbrella spec is to differentiate which output is a regular file, which output is a dir.
This information is necessary when the `docker` execution engine is used. Each item in your `--output` option will be treated as a docker volume and pulled back from your docker container. However, by default, docker treats the target of a docker volume to be a directory on your host machine. The only way to tell docker that a docker volume should be a file is to create the target before starting the docker container.

hmeng-19:
When a user set the `--output` option of umbrella to be `<container_path1>=<local_path1>, <container_path2>=<local_path2>`, umbrella does not know the file type of these two paths. Therefore, it checks the `output` section of the umbrella spec to figure out whether a path is under the `files` subsection, or under the `dirs` subsection.

hmeng-19:
To give the umbrella user more freedom of setting outputs, I suggest we can change the format of the `--output` option of umbrella to accept two styles of settings: 
the 1st one is `<container_path>=<local_path>`;
the 2nd one is `<container_path>=<local_path>=<file_type>` (file_type can be `f` for files or `d` for dirs).
(Of course, the users have freedom to mix these two styles if they like.)

When the first style is used, we treat it as we currently do;
When the second style is used, we will not check the `output` section of the umbrella spec, because the user already tells us the file type. We can directly create the target for docker volumes based on the file type info the user provides.

How do you like this?

pivie:
That would work. Do you think it would be better to ignore the output from
the umbrella spec when there is an output on the command line, or should
the two be combined?

hmeng-19:
[...]

PR #1445, opened 2016-08-24
Allow the umbrella users to specify output types

hmeng-19:
As #1320 pointed out, it is expected that an umbrella user may want to obtain extra outputs which are not specified in the `output` section of an Umbrella spec.

This PR extends the format of the `--output` option of umbrella to accept two styles of settings: 
the 1st one is `<container_path>=<local_path>`;
the 2nd one is `<container_path>=<local_path>:<file_type>` (file_type can be `f` for files or `d` for dirs).
(Of course, the users have freedom to mix these two styles if they like.)

When the first style is used, we treat it as we currently do;
When the second style is used, we will not check the output section of the umbrella spec, because the user already tells us the file type.


hmeng-19:
@pivie , please take a look.

hmeng-19:
@pivie , I will merge this PR for now since I tested using some Umbrella examples on my side, and my following work needs this feature.

PR #1447, opened 2016-08-25
Automatically determine terminal width.

matz-e:
Helps enormously when master names are long and would be severly truncated
by default.  Setting the COLUMNS environment variable helps, too, but has
to be reset if the terminal changes in size.

### DIFF
--- a/work_queue/src/work_queue_status.c
+++ b/work_queue/src/work_queue_status.c
@@ -22,6 +22,7 @@ See the file COPYING for details.
 #include <errno.h>
 #include <string.h>
 #include <stdlib.h>
+#include <sys/ioctl.h>
 
 typedef enum {
 	FORMAT_TABLE,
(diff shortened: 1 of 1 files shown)


btovar:
Thanks @matz-e !

btovar:
It does not quite work for me on tmux/rxvt, TERM=screen-256color.  Is this expected?

matz-e:
I'll check. Does it give you the default width of 80 columns?

btovar:
​Yes.​

matz-e:
xterm and iterm2 work for me if I start the terminal, run tmux, ssh to earth, and then execute the status. Even if I resize the window afterwards.

Is that the sequence of commands you're running? (at least you get the default, still)

matz-e:
urxvt on Mac, tmux, ssh to earth still works for me. I guess I have to switch the order of tmux and ssh? Tmux is not available on earth, though.

matz-e:
When I use screen within the ssh-session in urxvt, this still works for me… I can play around more once I'm home and with a Linux.

annawoodard:
> Tmux is not available on earth, though.

You can grab it here for testing: `~awoodard/local/bin/tmux`

btovar:
I have my own tmux :)
It worked on my laptop, and I did modify the term settings on my desktop, so probably is on my end.

PR #1452, opened 2016-08-26
Enhance string_combine

hmeng-19:
Currently, `string_combine` does not allow its arguments to be `NULL`.

The PR fixes this.


hmeng-19:
@btovar , @dthain , PTAL.

btovar:
Looks good.

PR #1456, opened 2016-08-31
Makeflow mode directive

btovar:


nhazekam:
Looks good to me. For clarification, does the change to batch_job_work_queue.c relate to adding category modes? Just seems unrelated to me, though I may just not know its use. @btovar

makeflow/src/parser.c
```diff
@@ -473,21 +473,41 @@ static int dag_parse_directive(struct lexer *bk, struct dag_node *n)
...
 
-		lexer_preppend_token(bk, t2);
-		lexer_preppend_token(bk, t);
+		if(set_var) {
+			lexer_preppend_token(bk, t2);
+			lexer_preppend_token(bk, t);
 
-		dag_parse_variable_wmode(bk, n, '=');
+			dag_parse_variable_wmode(bk, n, '=');
+		} else {
```
nhazekam: Doe we need this else case? Also, good catch on not pushing those back on.

btovar:
The transactions log is how I debugged that the modes were correctly activated. You are correct that it could be an independent pull request.

makeflow/src/parser.c
btovar: We do not. I'll remove it.

nhazekam:
I think it is fine in this pull request, just wasn't sure how it was related.

batch_job/src/batch_job_work_queue.c
```diff
@@ -108,6 +108,10 @@ static batch_job_id_t batch_job_wq_wait (struct batch_queue * q, struct batch_jo
 		{
 			return -1;
 		}
+
+		char *transactions = string_format("%s.transactions", q->logfile);
+		work_queue_specify_transactions_log(q->data, transactions);
+		free(transactions);
```
nhazekam: More of a meta question, but is there a reason we wait to specify the log and transaction log until batch_job_*_wait, as opposed to batch_job_*_create?

btovar: It might be historical. For other batch systems we ensure the file is open, so when adapting for wq, this would be the way to go. But don't know, I am making stuff up.

PR #1458, opened 2016-09-01
Run each makeflow rule as an Umbrella job

hmeng-19:
Add `--umbrella-spec` and `--umbrella-binary` option into makeflow
Add `dag_node->umbrella_spec` to track the umbrella spec needed for each rule
Add `makeflow_wrapper_umbrella` model

### DIFF
--- a/dttools/src/path.h
+++ b/dttools/src/path.h
@@ -44,4 +44,9 @@ Returns NULL if none is found.
 */
 char *path_which(const char *exec);
 
+/* path_join_two_strings joins two strings, and adds sep between them.
+ * The caller should free the space pointed by the returned pointer.
+ */
+char *path_join_two_strings(const char *s, const char *t, const char * sep);
+
 #endif
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -37,6 +37,7 @@ See the file COPYING for details.
 #include "makeflow_wrapper.h"
 #include "makeflow_wrapper_docker.h"
 #include "makeflow_wrapper_monitor.h"
+#include "makeflow_wrapper_umbrella.h"
 #include "parser.h"
 #include "parser_jx.h"
 
@@ -51,6 +52,7 @@ See the file COPYING for details.
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdbool.h>
 
 /*
 Code organization notes:
@@ -134,6 +136,7 @@ static int log_verbose_mode = 0;
  *  * files. */
 static struct makeflow_wrapper *wrapper = 0;
 static struct makeflow_monitor *monitor = 0;
+static struct makeflow_wrapper_umbrella *wrapper_umbrella = 0;
 
 /* is true only if the -N command is used. Useful for catalog reporting.
  */
@@ -142,7 +145,7 @@ static int catalog_reporting_on = 0;
 /* Generates file list for node based on node files, wrapper
  *  * input files, and monitor input files. Relies on %% nodeid
  *   * replacement for monitor file names. */
-static struct list *makeflow_generate_input_files( struct dag_node *n, struct makeflow_wrapper *w, struct makeflow_monitor *m )
+static struct list *makeflow_generate_input_files( struct dag_node *n, struct makeflow_wrapper *w, struct makeflow_monitor *m, struct makeflow_wrapper_umbrella *u )
 {
 	struct list *result = list_duplicate(n->source_files);
 
@@ -154,10 +157,14 @@ static struct list *makeflow_generate_input_files( struct dag_node *n, struct ma
 		result = makeflow_wrapper_generate_files(result, m->wrapper->input_files, n, m->wrapper);
 	}
 
+	if(u){
+		result = makeflow_wrapper_generate_files(result, u->wrapper->input_files, n, u->wrapper);
+	}
+
 	return result;
 }
 
-static struct list *makeflow_generate_output_files( struct dag_node *n, struct makeflow_wrapper *w, struct makeflow_monitor *m )
+static struct list *makeflow_generate_output_files( struct dag_node *n, struct makeflow_wrapper *w, struct makeflow_monitor *m, struct makeflow_wrapper_umbrella *u )
 {
 	struct list *result = list_duplicate(n->target_files);
 
@@ -167,6 +174,10 @@ static struct list *makeflow_generate_output_files( struct dag_node *n, struct m
 
 	if(m){
 		result = makeflow_wrapper_generate_files(result, m->wrapper->output_files, n, m->wrapper);
+	}
+
+	if(u){
+		result = makeflow_wrapper_generate_files(result, u->wrapper->output_files, n, u->wrapper);
 	}
 
 	return result;
@@ -189,7 +200,7 @@ static void makeflow_abort_all(struct dag *d)
 		printf("aborting local job %" PRIu64 "\n", jobid);
 		batch_job_remove(local_queue, jobid);
 		makeflow_log_state_change(d, n, DAG_NODE_STATE_ABORTED);
-		struct list *outputs = makeflow_generate_output_files(n, wrapper, monitor);
+		struct list *outputs = makeflow_generate_output_files(n, wrapper, monitor, wrapper_umbrella);
 		list_first_item(outputs);
 		while((f = list_next_item(outputs)))
 			makeflow_clean_file(d, local_queue, f, 0);
@@ -201,7 +212,7 @@ static void makeflow_abort_all(struct dag *d)
 		printf("aborting remote job %" PRIu64 "\n", jobid);
 		batch_job_remove(remote_queue, jobid);
 		makeflow_log_state_change(d, n, DAG_NODE_STATE_ABORTED);
-		struct list *outputs = makeflow_generate_output_files(n, wrapper, monitor);
+		struct list *outputs = makeflow_generate_output_files(n, wrapper, monitor, wrapper_umbrella);
 		list_first_item(outputs);
 		while((f = list_next_item(outputs)))
 			makeflow_clean_file(d, remote_queue, f, 0);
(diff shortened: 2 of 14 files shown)


hmeng-19:
@btovar , @nhazekam , @dthain , PTAL.

dttools/src/path.h
```diff
@@ -44,4 +44,9 @@ Returns NULL if none is found.
 */
 char *path_which(const char *exec);
 
+/* path_join_two_strings joins two strings, and adds sep between them.
+ * The caller should free the space pointed by the returned pointer.
+ */
+char *path_join_two_strings(const char *s, const char *t, const char * sep);
```
nhazekam: Could also be useful for writing remote rename.

makeflow/src/makeflow.c
```diff
@@ -1599,6 +1626,19 @@ int main(int argc, char *argv[])
 		makeflow_gc_method = MAKEFLOW_GC_ALL;
 	}
 
+	if(wrapper_umbrella && wrapper_umbrella->spec) {
```
nhazekam: For now umbrella should also require that the binary is specified. I understand that there may be cases when umbrella exists locally, but in Makeflow we assume that you are specifying the executable files in most cases. This also ensures consistent version across nodes.

makeflow/src/makeflow.c
```diff
@@ -1599,6 +1626,19 @@ int main(int argc, char *argv[])
...
+		debug(D_MAKEFLOW_RUN, "setting dag_node->umbrella_spec...\n");
+		struct dag_node *cur;
+		cur = d->nodes;
+		while(cur) {
+			dag_node_set_umbrella_spec(cur, wrapper_umbrella->spec);
```
nhazekam: This seems redundant. I understand having an umbrella spec stored in the node for the eventuality that each node could have a different spec, but when they all share the same umbrella spec it seems to make more sense to only specify it in the umbrella wrapper. Then in the future when node can have different umbrella specs, the dag_node spec is only set when it is different from the standard.

makeflow/src/makeflow_wrapper_umbrella.c
```diff
@@ -0,0 +1,169 @@
...
+
+	w->wrapper = makeflow_wrapper_create();
+	w->spec = NULL;
+	w->binary = NULL;
+	return w;
+}
+
+void makeflow_wrapper_umbrella_set_spec(struct makeflow_wrapper_umbrella *w, const char *spec) {
+	struct stat st;
+	if(lstat(spec, &st) == -1) {
```
nhazekam: Currently in Makeflow we use batch_fs_stat to check consistently with the batch system. Check makeflow.c for two examples.

makeflow/src/makeflow_wrapper_umbrella.c
```diff
@@ -0,0 +1,169 @@
...
+		fatal("lstat(`%s`) failed: %s\n", spec, strerror(errno));
+	}
+	if((st.st_mode & S_IFMT) != S_IFREG) {
```
nhazekam: @dthain @btovar Should we start checking if the file is regular or is this overkill for Makeflow?

makeflow/src/makeflow_wrapper_umbrella.c
```diff
@@ -0,0 +1,169 @@
...
+		fatal("the --umbrella-spec option of makeflow should specify a regular file\n");
+	}
+	w->spec = spec;
+	debug(D_MAKEFLOW_RUN, "setting wrapper_umbrella->spec to %s\n", spec);
+}
+
+void makeflow_wrapper_umbrella_set_binary(struct makeflow_wrapper_umbrella *w, const char *binary) {
+	struct stat st;
+	if(lstat(binary, &st) == -1) {
```
nhazekam: Same as above.

makeflow/src/makeflow_wrapper_umbrella.c
```diff
@@ -0,0 +1,169 @@
...
+	umbrella_input_opt = create_umbrella_opt(queue, input_files, false);
+	debug(D_MAKEFLOW_RUN, "umbrella input opt: %s\n", umbrella_input_opt);
+
+	debug(D_MAKEFLOW_RUN, "output_files: %s\n", output_files);
+	umbrella_output_opt = create_umbrella_opt(queue, output_files, true);
+	debug(D_MAKEFLOW_RUN, "umbrella output opt: %s\n", umbrella_output_opt);
+
+	// construct umbrella_command
+	if(!batch_queue_supports_feature(queue, "remote_rename")) {
+		if(!w->binary) {
```
nhazekam: Same as above, for consistency we should require that the binary is specified. This can be discussed, but in general with Makeflow we don't assume information about the execution environment. This is different when using programs like Docker that need special permissions to be installed, but when its software we are writing I think we can require it to be specified. @btovar @dthain

nhazekam:
As a general comment, how much has this been tested? Also has it been tested with other wrappers to see how they interact and that wrappers still work in the absence of umbrella. The last one should be fine, but we should still test it.

makeflow/src/makeflow.c
hmeng-19: @nhazekam , the logic here is that if the user does not set up the `--umbrella-spec`, it is not necessary to set up the umbrella wrapper.
That means, if the user only sets `--umbrella-binary` without setting `--umbrella-spec`, we do nothing.

nhazekam: You also never require the binary to be set. I understand that we only want to do something if spec is set, but for Makeflow we should also have the binary. There is a comment on this in the umbrella wrapper. Assuming we need both, then the logic should check for both.

makeflow/src/makeflow_wrapper_umbrella.c
hmeng-19: My point for this is:
If the user provides the umbrella binary via `--umbrella-binary`, then we use it.
If the user does not provide the umbrella binary via `--umbrella-binary`, we expect to find the umbrella binary through `$PATH` on the execution node. This is useful when we later run work_queue_worker as an umbrella job, which delivers both work_queue_worker and umbrella .

makeflow/src/makeflow.c
hmeng-19: @nhazekam , I got your point.
The only reason I keep it is it does not really break anything. And I indeed have plan to adding support for specifying umbrella specs for each rule.
I would keep it for now if redundancy is your only concern here.

makeflow/src/makeflow_wrapper_umbrella.c
nhazekam: I understand this is a we previously discussed. However, I did comment then that in Makeflow we typically specify the executable. When assuming there is an executable for umbrella at the worker we are also assuming the version is correct and consistent with all other nodes.

makeflow/src/makeflow.c
hmeng-19: This is deliberate. So that the code is not broken no matter the umbrella binary is set or not.
If the umbrella binary is set, we will send the binary to the execution node and the execution node will run `./umbrella run ...`;
If the umbrella binary is not set, we will not send the binary to the execution node and the execution node will run `umbrella run ...`.

makeflow/src/makeflow.c
nhazekam: If we are ever setting for each specific node then there will be no need to loop over them all and set the same one (we could still use the generic one in the wrapper). This is duplicate information and it isn't used in the current implementation.

hmeng-19: Okay. If you really want, I can remove it for now. 
But the next PR will definitely have these code again.

makeflow/src/makeflow.c
nhazekam: After further discussion, not specifying the executable is fine, but if it is not specified we should emit a warning that umbrella will be used at the execution site.

makeflow/src/makeflow_wrapper_umbrella.c
nhazekam: As with the above comment, this is allowable with the knowledge that the user should be warned that we are assuming the executable exists at the work site. Disregard my previous comments on this.

makeflow/src/makeflow_wrapper_umbrella.c
hmeng-19: @nhazekam , the prototype of `batch_fs_stat` requires a `struct batch_queue *`:
```
int batch_fs_stat (struct batch_queue *q, const char *path, struct stat *buf)
```

`makeflow_wrapper_umbrella_set_spec` is called during the step of parsing options, when the batch_queue has not been initialized yet.

hmeng-19:
@nhazekam , here are the testings for this PR, which is under `cctools/makeflow/example/makeflow_umbrella`:
```
To test makeflow with umbrella using local execution engine:
$ makeflow --umbrella-binary $(which umbrella) --umbrella-spec convert_S.umbrella example.makeflow

To test makeflow with umbrella using wq execution engine:
$ makeflow -T wq --umbrella-binary $(which umbrella) --umbrella-spec convert_S.umbrella example.makeflow

To test the case when `--wrapper` and `--umbrella-spec|binary` are used at the same time:
$ makeflow --wrapper 'time -p /bin/sh -c []'  --umbrella-binary $(which umbrella) --umbrella-spec convert_S.umbrella example.makeflow
$ makeflow -T wq --wrapper 'time -p /bin/sh -c []'  --umbrella-binary $(which umbrella) --umbrella-spec convert_S.umbrella example.makeflow
```
For each case, I also tested the case when the `--umbrella-binary` option is not set.

hmeng-19:
@nhazekam , https://github.com/cooperative-computing-lab/cctools/pull/1458/commits/6798f4e9e0a7bdae50b11f3ef9081145a6ea896e  and https://github.com/cooperative-computing-lab/cctools/pull/1458/commits/8fdeb0a36f26bc57dd65f9528ebdddfd77cea9b7 should fix your concerns, PTAL.

hmeng-19:
@nhazekam , fixed the conflicts and squashed the commits into a single one, as show in https://github.com/cooperative-computing-lab/cctools/pull/1458/commits/1866d35bea34f0281a5a1bceabacd5d409a7f7cd. PTAL.

nhazekam:
When testing it everything worked. The only issue is that when I run makeflow -c example.makeflow, the umbrella.log is left behind and not cleaned. This file should be added to the log so it is properly removed.

nhazekam:
We can merge. I'll open an issue about the umbrella.log.

PR #1465, opened 2016-09-07
Wq graph fixes

btovar:
Adds stack graph to show time master utilization.

Cleans up capacity computation. (Nate, please check it is still correct. I do not think we wanted to weight the difference, but the total.)


btovar:
new graph looks like this (for shrimp):

makeflow wqlog times-stacked

dthain:
I like it!

dthain:
What do you make of the increased internal time at the end?

PR #1469, opened 2016-09-14
Wq maker update

nhazekam:
This is an update to WQ_MAKER tested for use in Jetstream. 

The main goal of this change to wq_maker is to no longer rely so heavily on the existing MAKER code base. This code now makes calls to MAKER as opposed to using the MAKER libraries for calls. We split by contig/scaffold and make a separate invocation to MAKER for each. We use the basename and maker locally to organize as the same output after the distributed portion completes.

The old version, which was named wq_maker is now wq_maker_legacy. The new version is named, wq_maker.2.31.8 and wq_maker. I can remove and reissue the pull request if we want only one of those names. This has been tested in Jetstream, but there are small changes coming in the future, likely for clarity and better stat output. @btovar take a quick glance

Things to add in future: Handling of multiple input genome files from maker_opts.ctl, and allowing them to be split differently. Add stat output for runs to show performance and estimate total output. Clarify resource specification if needed.


btovar:
Nice job Nick.

Can we run all workflows of wq_maker.legacy with the new wq_maker without making any changes? If so, I would not keep the legacy one.

PR #1481, opened 2016-09-16
Give error code when no more tickets are available.

### LINKED ISSUE #1477: Chirp Server Crashes on Ticket Auth
This succeeds: `chirp -a hostname disc02.crc.nd.edu`

This fails: `chirp -a ticket -a hostname disc02.crc.nd.edu` b/c the chirp server crashes after accepting the ticket authentication method, but having no tickets to deal with.

This is a high priority to fix, since the default configuration (no -a options) tries the ticket method and causes the crash, even when the user does nothing special.

btovar:
Fixes #1477.

The problem was that chirp_auth_ticket always returned success. Now it returns failure by default, unless a success is explicitly stated.

### DIFF
--- a/dttools/src/auth_ticket.c
+++ b/dttools/src/auth_ticket.c
@@ -278,7 +278,7 @@ static int auth_ticket_accept(struct link *link, char **subject, time_t stoptime
 		}
 	}
 
-	rc = 0;
+	rc = 1;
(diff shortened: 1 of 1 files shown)


dthain:
Makes sense to me, but I am trusting you to test it....

PR #1484, opened 2016-09-21
Makeflow Automatic Archiving

piercecunneen:
In this pull request, we create a new option for makeflow to preserve each node within a caching directory. 

We create a cache id for each node by getting the hash of each input file and the command used to generate the outputs and then concatenating each hash into a buffer and hashing that buffer. This cache id is used as the sub directory name within the caching directory. 

Within each sub directory, we store the output files, the cache id of each ancestor node, and the original makeflow source file.

If the --preserve option is passed to makeflow, it will skip submitting a node to a batch system if the node already has been preserved


dthain:
Good! Looks like you are on the right path.

One code structure issue:  Move all of the new functions that you added to `makeflow.c` into a new module `makeflow_cache.c` and call them `makeflow_cache_populate()`  etc...  Set things up so that the only additions to `makeflow.c` are invocations of the `makeflow_cache_*` functions.

Then, for the location of the cache, make it `/tmp/makeflow.cache.$(UID)` by default, overridable by and option.

makeflow/src/makeflow_cache.c
```diff
@@ -0,0 +1,146 @@
+/*
+Copyright (C) 2014- The University of Notre Dame
```
dthain: 2016

makeflow/src/makeflow_cache.c
```diff
@@ -0,0 +1,146 @@
...
+  char *caching_file_path, *output_file_path, *source_makeflow_file_path, *ancestor_file_path;
+  char *ancestor_cache_id_string = NULL;
+  struct dag_node *ancestor;
+  struct dag_file *f;
+  int sucess;
+  FILE *fp;
+
+  caching_file_path = xxstrdup(d->caching_directory);
+  caching_file_path = string_combine_multi(caching_file_path, n->cache_id, "/outputs", 0);
+  sucess = batch_fs_mkdir(queue,caching_file_path, 0777, 1);
```
dthain: Actually, for the purposes of caching data, please use fopen, mkdir, on the cache itself, rather than batch_fs_fopen, batch_fs_mkdir, etc.  The latter are only necessary for the files that are in the dag itself.

makeflow/src/makeflow_cache.h
```diff
@@ -0,0 +1,29 @@
+/*
+Copyright (C) 2014- The University of Notre Dame
```
dthain: 2016

dthain:
@piercecunneen I liked the demo that you gave yesterday.

One mechanical thing -- please name the cache directory `/tmp/makeflow.cache.$(UID)` so that each user on the system has their own top level directory.  Then inside that directory, you can have `jobs/XY` and `files/XY` where XY is the first two chars of the checksum.  (That will avoid polluting `/tmp` unnecessarily.)

makeflow/src/makeflow.c
```diff
@@ -828,6 +849,31 @@ static void makeflow_node_complete(struct dag *d, struct dag_node *n, struct bat
 				makeflow_log_file_state_change(d, f, DAG_FILE_STATE_COMPLETE);
 		}
 
+		/* store node into archiving directory  */
```
dthain: I'm concerned that this code is repeating the sequence of events necessary to fully expand the command string and file lists, which normally happens elsewhere.  That will be fragile if/when those sequences changes.  Can you find a way instead to record and save the command and files as they existed at the point of submission?

makeflow/src/makeflow.c
```diff
@@ -950,9 +996,10 @@ static void makeflow_run( struct dag *d )
         }
 
 	while(!makeflow_abort_flag) {
+		did_find_archived_job = 0;
 		makeflow_dispatch_ready_jobs(d);
 
-		if(dag_local_jobs_running(d)==0 && dag_remote_jobs_running(d)==0 )
+		if(dag_local_jobs_running(d)==0 && dag_remote_jobs_running(d)==0 && did_find_archived_job == 0 )
```
dthain: A bit clunky, but I think it's the right solution.
Please add a comment here to explain why it is necessary.

makeflow/src/makeflow_archive.h
```diff
@@ -0,0 +1,44 @@
...
+void makeflow_archive_populate(struct dag *d, struct dag_node *n, char *command, struct list *inputs, struct list *outputs, struct batch_job_info *info);
+
+/* Returns true if a node has been preserved within the archiving directory*/
+int makeflow_archive_is_preserved(struct dag *d, struct dag_node *n, char *command, struct list *inputs, struct list *outputs);
+
+/* copies files from archiving directory to working directory */
+int makeflow_archive_copy_preserved_files(struct dag *d, struct dag_node *n, struct list *outputs);
+
+/* writes the run_info files that is stored within each archived node */
+void makeflow_write_run_info(struct dag *d, struct dag_node *n, char *archive_path, struct batch_job_info *info, char *command);
```
dthain: Every public function in this model needs to be called `makeflow_archive_something()`.
(It's the only way to keep sanity in a large project.)

makeflow/src/makeflow_archive_query.py
```diff
@@ -0,0 +1,247 @@
+import hashlib
```
dthain: Call the program `makeflow_archive_query` and put a pound-bang /usr/bin/python at the top;
from a UX perspective, the user shouldn't have to care what language it's written in.

makeflow/src/dag.h
```diff
@@ -19,6 +19,8 @@ See the file COPYING for details.
 
 #include <stdio.h>
 
+#define ARCHIVING_DEFAULT_DIRECTORY "/tmp/makeflow.archive."
```
dthain: It seems odd to put this here, since it's more of a property of the execution than of the dag state itself.
Please call it `MAKEFLOW_ARCHIVE_DEFAULT_DIRECTORY` and put it in `makeflow_archive.h`

makeflow/src/makeflow_archive.h
piercecunneen: Fixed this by making my non-public members static and moving them out of the header file

makeflow/src/makeflow.c
dthain: That's one or.  Or, you could factor out the job expansion code, so that it exists in only once place..

annawoodard:
:+1::+1::+1::+1:

Very excited about this new feature, which will simplify my workflow. I create a new makeflow script for major iterations as my analysis changes, but there are some rules that change rarely. Currently I'm using a script to copy output from previous iterations over to new ones rather than re-run the rules. This will be much nicer. Thanks @piercecunneen!

makeflow/src/makeflow.c
```diff
@@ -1568,6 +1634,55 @@ int main(int argc, char *argv[])
 				if(!wrapper_umbrella) wrapper_umbrella = makeflow_wrapper_umbrella_create();
 				makeflow_wrapper_umbrella_set_spec(wrapper_umbrella, (const char *)xxstrdup(optarg));
 				break;
+			case LONG_OPT_ARCHIVE:
```
piercecunneen: has been refratored in	e679f3b

piercecunneen:
makeflow_archive_populate has become to complicated for my liking, going to refractor out parts into separate static functions in order to make code more readable/maintainable

dthain:
And here I was about ready to merge!

I'll have a look at it later this week.

piercecunneen:
@dthain no new changes, just rebased in order to get rid of merge conflicts

dthain:
Ok, I'll take a look now.

dthain:
Merged, congrads!

Issue #1489, opened 2016-09-23
missing implementations of several dag funcs


hmeng-19:
I noticed that `dag.h` includes the prototypes of `dag_input_lookup_or_create` and `dag_output_lookup_or_create`, however, I can not find their imeplementations in `dag.c`.

hmeng-19:
@nhazekam , @btovar , @dthain , PTAL.

dthain:
Dead code should be removed!  Haiyan, please fix it and make a pull request.

hmeng-19:
@dthain , just want to make it clear. 
Your point is that makeflow no longer provides these interfaces. Right?

dthain:
If the prototype serves no purpose, remove it.

hmeng-19:
#1490  should fix this.
@dthain , PTAL.

hmeng-19:
Fixed by #1490 .

PR #1490, opened 2016-09-23
Remove unimplemented prototypes from dag.h

hmeng-19:


hmeng-19:
This PR aims to fix #1489 .
@dthain , PTAL.

Issue #1468, opened 2016-09-09
Makeflow- Umbrella : When umbrella wrapper is used locally, the umbrella.log is left behind


nhazekam:
We should log this, allow the user to specify a pattern to be used for each tasks logs, and clean this when it is done. By default it should either always be collected with a name such as umbrella_%%.log, or it should be disable. @hmeng-19 I will leave this up to you for the next pull request on Umbrella in Makeflow.

hmeng-19:
@nhazekam , PTAL.
#1491 should fix this.

hmeng-19:
@nhazekam , for testing this feature, `cctools/makeflow/example/makeflow_umbrella` will be a good example.

nhazekam:
Fixed in #1491 .

PR #1491, opened 2016-09-23
Handle umbrella logfiles elegantly in makeflow

hmeng-19:
Add umbrella logfiles into makeflow log file
Add umbrella-log-prefix opt
Fix umbrella wrapper

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -540,7 +540,7 @@ static void makeflow_node_submit(struct dag *d, struct dag_node *n)
 	char *command = strdup(n->command);
 	command = makeflow_wrap_wrapper(command, n, wrapper);
 	command = makeflow_wrap_enforcer(command, n, enforcer, input_list, output_list);
-	command = makeflow_wrap_umbrella(command, wrapper_umbrella, queue, input_files, output_files);
+	command = makeflow_wrap_umbrella(command, n, wrapper_umbrella, queue, input_files, output_files);
 	command = makeflow_wrap_monitor(command, n, queue, monitor);
 
 	/* Before setting the batch job options (stored in the "BATCH_OPTIONS"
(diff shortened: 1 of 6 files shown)


hmeng-19:
This PR aims to fix #1468 .
@nhazekam , PTAL.

hmeng-19:
@nhazekam , for testing this feature, cctools/makeflow/example/makeflow_umbrella will be a good example.

makeflow/src/dag_node.c
```diff
@@ -47,6 +47,7 @@ struct dag_node *dag_node_create(struct dag *d, int linenum)
 	n->resource_request = CATEGORY_ALLOCATION_FIRST;
 
 	n->umbrella_spec = NULL;
+	n->umbrella_logfile = NULL;
```
nhazekam: With umbrella_spec we were planning (at some point) on the specification being unique to the node, are we also expecting umbrella_logfile to be unique to the node. If not this should only exist in the wrapper so the file names are consistent across all of the nodes.

makeflow/src/makeflow.c
```diff
@@ -149,6 +149,8 @@ static char *mountfile = NULL;
 static char *mount_cache = NULL;
 static int use_mountfile = 0;
 
+static char *umbrella_log_prefix = NULL;
```
nhazekam: This prefix belongs in the umbrella wrapper, not in makeflow.c. Refer to how we handle logs for the resource monitor.

makeflow/src/dag_node.c
hmeng-19: @nhazekam , good point. 
I do not think it is very meaningful to allow the user to specify an unique umbrella log file for each rule.
I will move this into the umbrella_wrapper.

makeflow/src/makeflow.c
hmeng-19: @nhazekam , sounds good. Will fix it.

hmeng-19:
@nhazekam , PTAL.
https://github.com/cooperative-computing-lab/cctools/pull/1491/commits/7bbf0c81eb9f5ae8568d20eb9dda0c0f1e2c72f0 should fix your concerns.

hmeng-19:
@nhazekam , PTAL.
As we discussed earlier, I added a new commit, https://github.com/cooperative-computing-lab/cctools/pull/1491/commits/e983a9aa842e7238d748e4ecb0817f6ac0ed7f83, to implement the `--umbrella-mode` option, which allows the user to specify the execution mode of umbrella at makeflow level (not rule level).

hmeng-19:
@nhazekam , PTAL. I rebased it and fixed the conflicts.

PR #1492, opened 2016-09-29
Allow Parrot-in-Parrot

trshaffer:
This PR adds a new syscall, `parrot_dissociate`, that tells Parrot that the process wants to make private changes to the mount entries. This behaves similarly to `MS_PRIVATE`, but each child process also gets a new namespace. Rather than trying to implement full Linux namespaces, this adds the minimum required to run Parrot-in-Parrot with working mount options. For @hmeng-19's Umbrella workflow, the inner process will notice that it's in another Parrot and simply `exec` after some option parsing. If any mounts are given, they are forwarded via syscall to the outer Parrot. The outer Parrot adds the mount entries _based on the inner Parrot's view of the filesystem_ so that inner Parrots don't need to know what their mounts are. For example

```
# --dynamic-mounts is required
./parrot_run --dynamic-mounts -- sh
# my workstation doesn't have /run
cat /run/sshd.pid 
    cat: /run/sshd.pid: No such file or directory
./parrot_mount /run /var/run rwx
# since /run doesn't really exist, the inner Parrot's path needs to
# be resolved given existing mounts
./parrot_run -M /tmp/sshd.pid=/run/sshd.pid -- sh
cat /tmp/sshd.pid
    2934
exit
cat /tmp/sshd.pid
    cat: /tmp/sshd.pid: No such file or directory
cat /run/sshd.pid
    2934
exit
```

This functionality is intentionally much more limited than a full namespace implementation. This should make it possible to run Parrot under Parrot, but that's still probably not a good idea for most applications.


dthain:
Tim, you have the right idea here, but I'm a bit concerned about making such a cross-cutting change to the pfs_table and pfs_sys APIs.  Let me suggest a simplification:  Can  you make the pointer to the namespace a member of pfs_table?  Then you don't have to modify all the entry points in pfs_table.h and pfs_sys.h and pfs_dispatch{64}.cc.  Most of the changes can then just be isolated to pfs_resolve and friends.

dthain:
A minor point, but how about `parrot_fork_namespace()` instead of `parrot_dissociate()`

trshaffer:
These new commits make this a much smaller pull request. Unfortunately, since `struct pfs_process` is C++ and resolve needs to know about who's asking for the path, I had to change to `pfs_resolve.cc`

dthain:
:+1:  for a much more parsimonious approach, but :-1: for reverting one file to C++.   :)

One UX thing -- the message indicating "most features are unavailable" is confusing to me as an end user.  Everything in Parrot still works as before, right?  It's only that -M is the only option that can be used on a nested call.

Is there a way to alert the user if they attempt to do anything other than -M in the nested invocation?

trshaffer:
I put in a blanket warning for options that will be ignored. For environment variables, I'm assuming we're likely to inherit those, and they'll be in effect unless someone intentionally changed them before the inner Parrot invocation. I can't think of a straightforward way to detect that case, though

trshaffer:
Per discussions with @dthain, I've expanded this into a richer namespace implementation. I was also able to relax some of the restrictions on `parrot_mount` by changing what `--dynamic-mounts` does under the hood.

Mount entries in Parrot are now organized into hierarchical, reference counted namespaces. A fresh copy of Parrot puts all process into the global namespace. After that, processes are free to fork their namespaces with the `parrot_fork_namespace` syscall. Whenever a process forks, its child will inherit a reference to its parent's mount namespace. Namespace changes are visible to processes subject to the following rules. Given a process `P` in namespace `N`,

- `P` may add/remove mount entries in `N`.
- any other process with a reference to `N`, will see changes by `P`.
- suppose another process `P'` forks `N`, so that `P'` is in namespace `N'`, then
  + any changes to `N'` **are not** visible in `N`.
  + any changes to `N` **are** visible in `N'`.
- when adding a mount entry to `N`, the redirect is resolved in the parent namespace of `N`.

When Parrot starts, the global mount namespace has no parent, so mounts passed in are resolved in the host filesystem. Unless the `--dynamic-mounts` flag is passed, Parrot **seals** (not sure if this is the right name for this) the global mount namespace before running the tracee by making a new child namespace. Thus all processes are, by default, locked into the namespace set up by the command line/environment variables. Processes no longer get an `EPERM` for using `parrot_mount`, and `parrot_mount --disable` simply seals the caller's namespace again. If `--dynamic-mounts` is passed, the global namespace is left unsealed, so mounts/unmounts are resolved in the host filesystem. This allows a setup script to mess with the command line mounts, then seal the global namespace so that it can no longer be modified. Likewise, a process can fork its namespace, add some mount entries, then seal it so that all children will be locked in.

The `parrot_mount` utility can add/remove mounts and seal the namespace with `parrot_mount --disable`. The new utility, `parrot_namespace`, forks the namespace and does command line mounts if running under Parrot, or just execs Parrot otherwise.

I'd like to add some debugging statements for when namespaces get created/destroyed and whatnot, but it should be OK to test out this functionality.

@hmeng-19, do you use `-M /foo=/bar`, or do you need the full mount files?

hmeng-19:
@trshaffer , I am using the environment variable `PARROT_MOUNT_FILE`,  which corresponds to `-m`.
```
 -m,--ftab-file=<file>          Use this file as a mountlist.             (PARROT_MOUNT_FILE)
```

dthain:
@jblomer You might find this interesting for the deployment cases where parrot is already deployed in uCVMFS before the user arrives to set up their own namespace.

dthain:
@hmeng I would generally encourage you to rely on command line options instead of environment variables, since they make it easier to avoid implicit bugs.

hmeng-19:
@dthain , sounds good. I opened an issue #1503 , and will fix it later.

trshaffer:
The logic to parse and apply mount files for `-m` currently lives in `pfs_resolve.c`, but `parrot_namespace` can't directly link against it since it pulls in most of Parrot. I'd rather not duplicate the parsing and helper functions between the two places, but I don't see a good way to refactor that's reachable from the client library. Any thoughts on organization @dthain?

dthain:
It's been a while since I looked at this code, but I would suggest refactoring it something like this:

Take the function which actually parses the mount file, call it `pfs_mountfile_parse` and move it into a new module `pfs_mountfile`.  That function should invoke `pfs_mount` as needed to update the mount table.  Then, link this module into `parrot_client.a` and `parrot_run`.

Inside of `parrot_client.a`, `parrot_mount` resolves to the library function and system call of that time.

Inside of `parrot_run`, define a function `parrot_mount` that updates the mount table directly.

Does that approach work?

trshaffer:
The linking's a little ugly, getting everyone to see the right things, but this should add `-m` for @hmeng-19's use case

dthain:
@trshaffer I think Haiyan is eager to use this feature, do you think it is ready to go?

trshaffer:
Yep, @hmeng-19  can try it any time. I'd like to make sure things work as expected in a real workflow

hmeng-19:
Nice, I will try it and give you my feedback.

hmeng-19:
@trshaffer , I tried the command instructions you posted at the top of this PR:
```
hmeng@daspos1 ~/cctools/bin$ ./parrot_run --dynamic-mounts -- sh
sh-4.1$ cat /run/sshd.pid 
cat: /run/sshd.pid: No such file or directory
sh-4.1$ ./parrot_mount /run /var/run rwx
sh-4.1$ ./parrot_run -M /tmp/sshd.pid=/run/sshd.pid -- sh
sorry, parrot_run cannot be run inside of itself.
version already running is 6.0.0 [1492:11a0d48b-DIRTY].
sh-4.1$ 
```
However, it failed. Did I miss sth?

hmeng-19:
@trshaffer , here is the parrot version info:
```
hmeng@daspos1 ~/cctools/bin$ ./parrot_run -v
./parrot_run version 6.0.0 [1492:11a0d48b-DIRTY] (released 2016-11-07 10:45:05 -0500)
        Built by hmeng@daspos1.crc.nd.edu on 2016-11-07 10:45:05 -0500
        System: Linux daspos1.crc.nd.edu 2.6.32-642.6.2.el6.x86_64 #1 SMP Mon Oct 24 10:22:33 EDT 2016 x86_64 x86_64 x86_64 GNU/Linux
        Configuration: --prefix /afs/crc.nd.edu/user/h/hmeng/cctools --strict
```

trshaffer:
@hmeng-19, I added end-user docs and a manpage for `parrot_namespace`, so you don't have to read this entire pull request. The intent is to replace `parrot_run` with `parrot_namespace` if you plan to do nested Parrot

hmeng-19:
@trshaffer , thanks for the fixes. I will give it a try.

hmeng-19:
@trshaffer , @dthain , I tested the latest version of this PR. The nesting feature works pretty nice for me. Thanks Tim.
My next step will be to integrate this feature into Umbrella to see how it works.
I pasted my test case here, in case you guys are interested in it.

```
hmeng@daspos1 /data/hmeng/work/umbrella_workflow/parrot_nest$ mkdir dir

hmeng@daspos1 /data/hmeng/work/umbrella_workflow/parrot_nest$ echo hello >>dir/file

hmeng@daspos1 /data/hmeng/work/umbrella_workflow/parrot_nest$ cat dir/file
hello

hmeng@daspos1 /data/hmeng/work/umbrella_workflow/parrot_nest$ parrot_namespace ls /dir
ls: cannot access /dir: No such file or directory

hmeng@daspos1 /data/hmeng/work/umbrella_workflow/parrot_nest$ parrot_namespace -M /dir=/data/hmeng/work/umbrella_workflow/parrot_nest/dir -- sh
sh-4.1$ ls /dir
file
sh-4.1$ cat /dir/file
hello
sh-4.1$ parrot_namespace cat /dir1/file
cat: /dir1/file: No such file or directory
sh-4.1$ parrot_namespace -M /dir1=/dir cat /dir1/file
hello
sh-4.1$ parrot_namespace -M /dir1=/dir -- sh
sh-4.1$ ls dir1
file
sh-4.1$ ls dir
file
sh-4.1$ parrot_namespace cat /dir2/file
cat: /dir2/file: No such file or directory
sh-4.1$ parrot_namespace -M /dir2=/dir1 cat /dir2/file
hello
sh-4.1$ parrot_namespace -M /dir2=/dir cat /dir2/file
hello
sh-4.1$ parrot_namespace -M /dir2=/data/hmeng/work/umbrella_workflow/parrot_nest/dir cat /dir2/file
hello
```

hmeng-19:
@trshaffer , from my testing, it seems that the outermost `parrot_namespace` can be replaced with `parrot_run`, once all the inner executable are `parrot_namespace`. Right?

trshaffer:
@hmeng-19 you're right, as long as you're **sure** you have the outermost one. If you want to use non-mount features across all instances, you'll have to take that route, since `parrot_namespace` doesn't allow most options

hmeng-19:
@trshaffer , is it difficult to support the `--ld-path` and `--no-set-foreground` options of `parrot_run` inside `parrot_namespace`? (These two options are also being used in the umbrella parrot mode.)
```
 -l,--ld-path=<path>            Path to ld.so to use.                      (PARROT_LDSO_PATH)
    --no-set-foreground         Disable changing the foreground process group of the session.
```

hmeng-19:
@trshaffer , I tried the `-d` debug mode of `parrot_namesapce`, and saw the `resolve` log info appears before the version info of `parrot_namespace`. Is this normal?

```
2016/11/08 11:48:07.60 parrot_namespace[21160] resolve: resolving /data/hmeng/work/umbrella_workflow/parrot_nest/dir in parent ns
2016/11/08 11:48:07.60 parrot_namespace[21160] resolve: /data/hmeng/work/umbrella_workflow/parrot_nest/dir = /data/hmeng/work/umbrella_workflow/parrot_nest/dir,7
2016/11/08 11:48:07.60 parrot_namespace[21160] debug: parrot_namespace version 6.0.0 [1492:36acb823-DIRTY] (released 2016-11-08 10:12:00 -0500)
2016/11/08 11:48:07.60 parrot_namespace[21160] debug: Built by hmeng@daspos1.crc.nd.edu on 2016-11-08 10:12:00 -0500
2016/11/08 11:48:07.60 parrot_namespace[21160] debug: System: Linux daspos1.crc.nd.edu 2.6.32-642.6.2.el6.x86_64 #1 SMP Mon Oct 24 10:22:33 EDT 2016 x86_64 x86_64 x86_64 GNU/Linux
```

hmeng-19:
@trshaffer , is it possible to add a `-o` option into `parrot_namespace` to make it easier to debug command lines like `parrot_namespace -d all -o log -- sh`?

trshaffer:
@hmeng-19 I added the `-o` flag. Note that debug options are not namespaced, so everything under a single Parrot instance shares the same config. Nested invocations can't really change that, so when running Parrot in Parrot, `-d` and `-o` don't do much.

Is this the same behavior you would want for the tty and ld.so flags? Or do you really need to be able to set different values for different nested Parrot invocations?

parrot/src/parrot_namespace.c
```diff
@@ -0,0 +1,142 @@
+/*
+Copyright (C) 2008- The University of Notre Dame
```
hmeng-19: @trshaffer , the copyright should be `2016`.

trshaffer:
@hmeng-19, each mount namespace now has its own `ld.so`, so you should be able to use the `-l` option. From what I can tell it seems to work, but I don't have a large selection of linkers on my workstation, so it's better if you can test with Umbrella

hmeng-19:
@trshaffer , thanks for the quick fixing.
I will give it a try.

hmeng-19:
@trshaffer , @dthain , here is some feedback about my current experience with integrating  `parrot_namespace` into `umbrella`. First of all, it works well.

On `daspos1.crc.nd.edu`, I started a makeflow with umbrella wrapper enabled:
```
makeflow -N hmeng-project -T wq --umbrella-binary $(which umbrella) \
--umbrella-mode parrot --umbrella-spec convert_S.umbrella example.makeflow
```

Then on `earth.crc.nd.edu`, I started `work_queue_worker` as an Umbrella job:
```
umbrella --spec work_queue_worker_S.umbrella --localdir /tmp/umbrella_test/ \
--sandbox_mode parrot --log umbrella.log run 'work_queue_worker -N hmeng-project'
```

Here is a snapshot of processes created by `umbrella` and `work_queue_worker` on `earth.crc.nd.edu`:
```
hmeng    1924766 1905916  0 14:57 pts/2    00:00:00 /usr/bin/python2 ./umbrella --spec work_queue_worker_S.umbrella --localdir /tmp/umbrella_test/ --sandbox_mode parrot --log umbrella.log run work_queue_worker -N hmeng-project
hmeng    1924778 1924766  1 14:57 pts/2    00:00:15 /tmp/umbrella_test/cache/parrot-6.0.2-x86_64-redhat6/bin/parrot_namespace -- /bin/sh -c cd /tmp/umbrella_test/tmpwGgLYZ; work_queue_worker -N hmeng-project
hmeng    1924779 1924778  0 14:57 pts/2    00:00:00 /bin/sh -c cd /tmp/umbrella_test/tmpwGgLYZ; work_queue_worker -N hmeng-project
hmeng    1924781 1924779  0 14:57 pts/2    00:00:01 work_queue_worker -N hmeng-project
hmeng    1931801 1924781  0 15:10 pts/2    00:00:00 /usr/bin/python2 ./umbrella --spec convert_S.umbrella --localdir /tmp/umbrella_test --inputs capitol.montage.gif=capitol.montage.gif,convert_S.umbrella=convert_S.umbrella,umbrella=umbrella, --output (null) --sandbox_mode parrot --log example.makeflow.umbrella.log.0 run ls /tmp/umbrella_test && sleep 60
```

You can see there are two umbrella jobs (`1924766` and `1931801`) executing concurrently, each has its own mountlists.

Here is the mountlist of `1924766`:
```
/ /tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64
/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64 /tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64
/tmp/umbrella_test /tmp/umbrella_test
/software /tmp/umbrella_test/tmpwGgLYZ/fake_mount/software
/software/umbrella-5.4.19-centos6-x86_64 /tmp/umbrella_test/cache/d4c5c5363946e1100c8a3bd216a97891/umbrella-5.4.19-centos6-x86_64
/tmp/umbrella_test/cache/d4c5c5363946e1100c8a3bd216a97891/umbrella-5.4.19-centos6-x86_64 /tmp/umbrella_test/cache/d4c5c5363946e1100c8a3bd216a97891/umbrella-5.4.19-centos6-x86_64
/tmp/umbrella_test/cache/parrot-6.0.2-x86_64-redhat6 /tmp/umbrella_test/cache/parrot-6.0.2-x86_64-redhat6
/tmp/umbrella_test/cache/parrot-6.0.2-x86_64-redhat6 /tmp/umbrella_test/cache/parrot-6.0.2-x86_64-redhat6
/software/work_queue_worker-5.4.19-centos6-x86_64 /tmp/umbrella_test/cache/2a3dc19a698a76203101b85ebe5322dd/work_queue_worker-5.4.19-centos6-x86_64
/tmp/umbrella_test/cache/2a3dc19a698a76203101b85ebe5322dd/work_queue_worker-5.4.19-centos6-x86_64 /tmp/umbrella_test/cache/2a3dc19a698a76203101b85ebe5322dd/work_queue_worker-5.4.19-centos6-x86_64
/tmp/umbrella_test/tmpwGgLYZ /tmp/umbrella_test/tmpwGgLYZ
/proc /proc
/sys /sys
/dev /dev
/run /run
/etc/hosts /etc/hosts
/etc/resolv.conf /etc/resolv.conf
/etc/passwd /etc/passwd
/etc/group /etc/group
/var/run/nscd /tmp/umbrella_test/tmpwGgLYZ/fake_mount/var/run/nscd
/var/run/nscd/socket ENOENT
```

Here is the mountlist of `1931801`:
```
/ /tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64
/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64 /tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64
/tmp/umbrella_test /tmp/umbrella_test
/tmp/umbrella_test/cache/parrot-6.0.2-x86_64-centos6 /tmp/umbrella_test/cache/parrot-6.0.2-x86_64-centos6
/tmp/umbrella_test/cache/parrot-6.0.2-x86_64-centos6 /tmp/umbrella_test/cache/parrot-6.0.2-x86_64-centos6
/tmp/umbrella_test/tmpXFa3M4 /tmp/umbrella_test/tmpXFa3M4
/proc /proc
/sys /sys
/dev /dev
/run /run
/etc/hosts /etc/hosts
/etc/resolv.conf /etc/resolv.conf
/etc/passwd /etc/passwd
/etc/group /etc/group
/var/run/nscd /tmp/umbrella_test/tmpXFa3M4/fake_mount/var/run/nscd
/var/run/nscd/socket ENOENT
/tmp/umbrella_test/tmpXFa3M4/capitol.montage.gif /tmp/worker-198023-1924781/t.7/capitol.montage.gif
/tmp/umbrella_test/tmpXFa3M4/umbrella /tmp/worker-198023-1924781/t.7/umbrella
/tmp/umbrella_test/tmpXFa3M4/convert_S.umbrella /tmp/worker-198023-1924781/t.7/convert_S.umbrella
```

doc/man/parrot_namespace.m4
```diff
@@ -0,0 +1,60 @@
...
+for CODE(parrot_run) that automatically handles nested invocations.
+
+SECTION(OPTIONS)
+
+OPTIONS_BEGIN
+OPTION_ITEM(`-M, --mount /foo=/bar')Mount (redirect) CODE(/foo)
+to CODE(/bar) (CODE(PARROT_MOUNT_STRING))
+OPTION_ITEM(`-m, --ftab-file <path>')Use CODE(<file>) as a
+mountlist (CODE(PARROT_MOUNT_FILE))
+OPTION_TRIPLET(-l, ld-path, path)Path to ld.so to use.
```
hmeng-19: @trshaffer , does the environment variable `PARROT_LDSO_PATH` work for this purpose?

parrot/src/parrot_namespace.c
```diff
@@ -31,8 +31,6 @@ static void show_help()
 	printf(optfmt, "-M", "--mount /foo=/bar", "Mount (redirect) /foo to /bar", " (PARROT_MOUNT_STRING)");
 	printf(optfmt, "-m", "--ftab-file <file>", "Use <file> as a mountlist", " (PARROT_MOUNT_FILE)");
 	printf(optfmt, "", "--parrot-path <path>", "Path to parrot_run", " (PARROT_PATH)");
-	printf(optfmt, "-d", "--debug <flags>", "Enable debugging for this subsystem", " (PARROT_DEBUG_FLAGS)");
-	printf(optfmt, "-o", "--debug-file <file>", "Send debugging to this file", " (PARROT_DEBUG_FILE)");
 	printf(optfmt, "-v", "--version", "Show version number", "");
```
hmeng-19: @trshaffer , is there a reason why we stop support `-d` and `-o`?

hmeng-19:
@trshaffer , is there a reason why we stop supporting `-d` and `-o` in `parrot_namespace`?

doc/man/parrot_namespace.m4
trshaffer: Yep, the various environment variables should work

hmeng-19:
@trshaffer , I ran into a case where I did not understand the behavior of the `-l` option of `parrot_namespace`.
Here is the mountlist of the outer-layer `parrot_namespace`:
```
/ /data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64
/data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64 /data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64
```
 
Here is the mountlist of the inner-layer `parrot_namespace`:
```
/ /tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64
/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64 /tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64
```
 
Here is the failure happened in the func `pfs_dispatch_prepexe`. I do not understand how the 7th line happened, it was the `resolve` in this line broke the process.
```
line 1: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> process: execve: /bin/bash (/data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/bin/bash) is an ordinary executable
line 2: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> process: pfs_dispatch_prepexe: forcing use of loader /tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/lib64/ld-linux-x86-64.so.2
line 3: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> libcall: open /tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/lib64/ld-linux-x86-64.so.2 0 0
line 4: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> resolve: /tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/lib64/ld-linux-x86-64.so.2 = /data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/lib64/ld-linux-x86-64.so.2,5
line 5: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> local: readlink /data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/lib64/ld-linux-x86-64.so.2 0x7fff663d0a90 1023
line 6: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> local: = 10 [readlink]
line 7: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> resolve: /data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/lib64/ld-2.12.so = /data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/lib64/ld-2.12.so,5
line 8: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> local: readlink /data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/lib64/ld-2.12.so 0x7fff663cdda0 1023
line 9: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> local: = -1 No such file or directory [readlink]
line 10: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> local: open /data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/data/hmeng/work/umbrella_workflow/work_queue_workers/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/tmp/umbrella_test/cache/93284f9e50040fe5704c0c795d72df61/centos-6.8-x86_64/lib64/ld-2.12.so 2048 0
line 11: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> local: = 2 No such file or directory [open]
line 12: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> libcall: = -1 No such file or directory [pfs_open_cached]
line 13: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> debug: pfs_dispatch_prepexe: pfs_dispatch.cc:3670[pull/1492/head:a69610b3] unix error: -1 (errno = 2) `No such file or directory'
line 14: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> process: execve: failed: No such file or directory
line 15: 2016/11/11 15:18:35.44 parrot_run[126255] <child:126270> syscall: = -2 No such file or directory [execve]
```

doc/man/parrot_namespace.m4
hmeng-19: Okay. Then you may want to add `(CODE(PARROT_LDSO_PATH))` into the doc.

hmeng-19:
@trshaffer , it seems that the `ld.so` got resolved twice at both `line 4` and `line 7`. Is there a way we can avoid the duplicate resolution happened at `line 7`?

hmeng-19:
@trshaffer , it seems that we need somehow to mark a path which has been resolved by all the mountlists at different levels of `parrot_namespace`, and avoid feeding a full-resolved path into the resolution process again. Right?

trshaffer:
So by design, mounts in a child are first resolved in the parent namespace. This allows a program to make mounts based on its current view of the filesystem without knowing how deep it is. Do you have a path or symlink written relative to the outside view of the system?

hmeng-19:
> mounts in a child are first resolved in the parent namespace.

@trshaffer , if a path in level 5 needs to be resolved, parrot first checks the mountlists in level 1, and then level 2 and then level 3, then level 4, and eventually level 5. Is this what you mean?

trshaffer:
Not exactly. When a mount entry is added in level 5, the target is first resolved in level 4. Supposing that a path doesn't match any mount entries, Parrot would be searching 5->4->3->2->1->host

hmeng-19:
> Not exactly. When a mount entry is added in level 5, the target is first resolved in level 4. Supposing that a path doesn't match any mount entries, Parrot would be searching 5->4->3->2->1->host

@trshaffer , in the case where a mount entry can be used at level 4, it will ignore the mount entry at level 5, and directly go back to 3->2->1->host, right?
This may explain lots of strange file access pattern I saw, because at every level, there is a mount entry which maps `/` to the rootfs image, and will be definitely used earlier than the mount entry at its child level.

dthain:
@hmeng-19 and I talked about this issue today, and I think we were able to reduce it down to a more succinct statement of the problem and an example.  Haiyan, can you re-state it here?

hmeng-19:
@trshaffer , @dthain , the problem here is a path which is a symbolic link gets resolved twice by the parrot path resolution module currently, and the resolution result does not make sense when parrot runs inside parrot:
```
pfs_table::open -> pfs_table::open_object -> pfs_table::resolve_name -> pfs_table::follow_symlink -> pfs_table::resolve_name -> ...
```

Here is an example with two levels of parrot:
the mountlist of the outer-level parrot:
```
/ /rootfs-1
/rootfs-1 /rootfs-1
```

the mountlist of the inner-level parrot:
```
/ /rootfs-2
/rootfs-2 /rootfs-2
```

Inside the inner-level parrot, here is what happens when we try to open a symlink file, `/lib64/libinfo.so.3`, whose target is `libinfo.so.3.7`:
```
line 1: libcall: open /lib64/libinfo.so.3
line 2: resolve: /lib64/libinfo.so.3 = /rootfs-1/rootfs-2/lib64/libinfo.so.3
line 3: local: readlink /rootfs-1/rootfs-2/lib64/libinfo.so.3
line 4: local: = 14 [readlink]
line 5: resolve: /rootfs-1/rootfs-2/lib64/libinfo.so.3.7 = /rootfs-1/rootfs-2/rootfs-1/rootfs-2/lib64/libinfo.so.3.7
line 6: local: readlink /rootfs-1/rootfs-2/rootfs-1/rootfs-2/lib64/libinfo.so.3.7
line 7: local: = -1 No such file or directory [readlink]
line 8: local: open /rootfs-1/rootfs-2/rootfs-1/rootfs-2/lib64/libinfo.so.3.7
line 9: local: = 2 No such file or directory [open]
line 10: libcall: = -1 No such file or directory [pfs_open_cached]
```
As you can see, after `readlink` returns successfully (line 4), `/rootfs-1/rootfs-2/lib64/libinfo.so.3.7` is fed into the `resolve_name` function and get resolved to `/rootfs-1/rootfs-2/rootfs-1/rootfs-2/lib64/libinfo.so.3.7` (line 5).

trshaffer:
So this doesn't seem to be an issue with my PR, as the behavior occurs on `master`, too...

To test, I set up some nested directories with `busybox`en inside so I could mount over `/`. With this mount list, things work as expected.

    /	/tmp/ns
    /tmp/ns	/tmp/ns

My PR uses the existing name resolution infrastructure, but exposes a namespaced interface. Thus, I can write out a flat mount file to achieve the same effect as @hmeng-19's example.

    /	/tmp/ns
    /tmp/ns	/tmp/ns
    /	/tmp/ns/ns2
    /ns2	/tmp/ns/ns2

Using this mount list, relative symlinks get double-resolved. Based on the debug logs, it appears the trouble's coming from `pfs_table::follow_symlink`, and indeed when I run with `--no-follow-symlinks` to disable cross-service symlinks, the issue disappears. The identity mount in the first list hides the issue, but I think the resolution behavior is still incorrect. @hmeng-19 is this the purpose of the identity mount?

hmeng-19:
> So this doesn't seem to be an issue with my PR, as the behavior occurs on master, too...

@trshaffer , you are right. The problem here is not introduced by your PR. It is the deficiency of the parrot path resolution module.

hmeng-19:
@trshaffer , the idea of creating a flat mount file at every mount level is nice, but error-prone.
Take the flat mount file you constructed as an example:
```
/   /tmp/ns (the 1st mount entry)
/tmp/ns /tmp/ns (the 2nd mount entry)
/   /tmp/ns/ns2 (the 3rd mount entry)
/ns2    /tmp/ns/ns2 (the 4th mount entry)
```

The first two entries do not have any effect in the inner-level parrot, because the 3rd mount entry has precedence over the first two entries. (However, this only causes the mount list a little bigger, and is not a real problem for us)

To make the program behave, we really need to add a mount entry like this in the flat mount file for the inner-level parrot:
```
/tmp/ns/ns2 /tmp/ns/ns2
```

After the discussion with Dr. Thain, we did not think constructing a flat mount file is a good direction to go.

@dthain, If I understood you correctly, you prefer to fix the parrot path resolution module to tackle the problem here. Right?

dthain:
Tim and I just talked about the problem  He is oging to think about it a little more and write up a proposed solution.

trshaffer:
At the moment, Parrot uses the mount list to get physical paths, but on reaching a relative symlink, Parrot appends it to the already resolved path, and tries to resolve again. When using mount lists to significantly rearrange the filesystem, this behavior lets the underlying filesystem layout leak out and causes the incorrect behavior observed toward the end of this thread.

I don't really like the approach of using explicit mount rules as @hmeng-19 suggested to short-circuit symlink resolution. This commit reworks the path resolution to use the *logical* path rather than the *physical* path when following symlinks. This is similar to Bash's behavior when `cd`ing across a symlink. I don't necessarily like that, but the previous behavior was incorrect when used with mount lists. I was initially concerned about using the logical layout, but Parrot's mount lists and existing path resolution already make textual transformations on paths rather than following the underlying filesystem layout. With this change, my previous example works correctly as

    /	/tmp/ns
    /tmp/ns	/tmp/ns
    /	/tmp/ns/ns2

And to address the concern about the unreachable rules, there's most likely a parent process that doesn't see the last rule, so for that process, the unreachable rules are at the head of the mount list.

dthain:
This looks like the right approach to me.
@hmeng does it fix your use case?

hmeng-19:
@trshaffer , @dthain , is this ready for me to test again?

trshaffer:
Yep, @hmeng-19 

You shouldn't need those `/rootfs-1 /rootfs-1` type lines anymore (unless you need them for something else)

hmeng-19:
@trshaffer , is it easy for you to rebase your PR on top of ccl master? I need some new features in the master to test this PR?

hmeng-19:
@trshaffer , do not worry. Ignore the comment. I can rebase on my side. You can delay rebasing until your PR is ready to be merged.

hmeng-19:
> You shouldn't need those /rootfs-1 /rootfs-1 type lines anymore (unless you need them for something else)
@trshaffer , quick question. Is it going to break if we keep these lines in the mountfile?

hmeng-19:
@dthain , I need the debug option for `parrot_namespace`. 
I tried to retest this PR, it did not work and I got different errors. I need a debug log to see what is going on before telling Tim more details.

trshaffer:
I don't think those lines will break anything. You can't use debug on the outermost level?

hmeng-19:
> You can't use debug on the outermost level?

Both levels are called with `parrot_namespace` inside umbrella. 
Umbrella does not have a way to differentiate these two levels.

hmeng-19:
> I don't think those lines will break anything. 

Okay. My current test experience seems that even a single level parrot_namespace does not work as expected. I will double test it before coming back to you.

hmeng-19:
@trshaffer , @dthain , the PR works. 
Nice work, @trshaffer . 
Unfortunately, I can  not reproduce the failure cases I met at the early stage of my test. But the PR works pretty well for the following use case:
```
# First terminal (each job will be wrapped into an umbrella job):
makeflow -N hmeng-project -T wq \
--umbrella-binary $(which umbrella) \
--umbrella-mode parrot \
--umbrella-spec convert_S.umbrella \
example.makeflow
```

```
# Second terminal:
umbrella --spec work_queue_worker_S.umbrella \
--localdir /tmp/umbrella_test/ \
--sandbox_mode parrot \
--log umbrella.log \
run 'work_queue_worker -N hmeng-project'
```

Currently, to test the PR, I rebased this PR on top of the ccl master, modified umbrella
to use parrot_namespace instead of parrot_run.
I can retest it after the PR gets merged.

dthain:
@btovar please review this PR.  We need to be extra careful with the inner-most parts of Parrot.

doc/parrot.html
```diff
@@ -306,6 +306,56 @@ <h2 id="nameresolution">Name Resolution<a class="sectionlink" href="#nameresolut
...
+<ul>
+<li>any changes to <em>N’</em> <strong>are not</strong> visible in <em>N</em>.</li>
+<li>any changes to <em>N</em> <strong>are</strong> visible in <em>N’</em>.</li>
+</ul>
+</li>
+<li>when adding a mount entry to <em>N</em>, the redirect is resolved in the parent namespace of <em>N</em>.</li>
+</ul>
+<p>When Parrot starts, the global mount namespace has no parent, so mounts passed in are
+resolved in the host filesystem. Unless the <tt>--dynamic-mounts</tt> flag is passed,
+Parrot <strong>seals</strong> the global mount namespace before running the tracee by making a new
```
btovar: Tim,
These mount names have lexical scope, as every time a new parrot_namespace is called, a new scope is created, with the name bindings lasting until the process ends. So they seem to me proper _closures_, and I would call them as such.

doc/parrot.html
```diff
@@ -306,6 +306,56 @@ <h2 id="nameresolution">Name Resolution<a class="sectionlink" href="#nameresolut
...
+changes. If <tt>--dynamic-mounts</tt> is passed, the global namespace is left unsealed,
+so mounts/unmounts are resolved in the host filesystem. This allows a setup script
+to modify the command line mounts, then seal the global namespace so that it can no
+longer be modified. Likewise, a process can fork its namespace, add mount entries,
+then seal it so that all children will be locked into the current view of the system.</p>
+<p>The <tt>parrot_namespace</tt> utility gives a more convenient way to create new mount namespaces
+rather than using the <tt>parrot_fork_namespace</tt> syscall directly. This utility forks the
+current mount namespace and performs any mounts specified on the command line.
+<tt>parrot_namespace</tt> detects whether or not it is already running under Parrot and does
+the right thing. When running under Parrot, <tt>parrot_namespace</tt> uses Parrot-specific
```
btovar: Reword... "does the right thing." is imprecise.

parrot/src/parrot_namespace.c
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	s = getenv("PARROT_MOUNT_FILE");
+	if (s) pfs_mountfile_parse_file(s);
+
+	s = getenv("PARROT_MOUNT_STRING");
+	if (s) pfs_mountfile_parse_string(s);
+
+	optind = 1;
+	while((c=getopt_long(argc, argv, optstring, long_options, NULL)) > -1) {
```
btovar: Rather than a reparse, I suggest saving the args to 'm' and 'M' to some other variables (initially set to NULL), and do a xxstrdup. Then do a if(var) ....

parrot/src/pfs_mountfile.c
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+	return mode;
+}
+
+void pfs_mountfile_parse_string(const char *str) {
+	assert(str);
+	char *e;
+	str = xxstrdup(str);
+	e = strchr((char *) str,'=');
```
btovar: why is this cast needed?

parrot/src/pfs_mountfile.c
```diff
@@ -0,0 +1,103 @@
...
+	if(!e) fatal("badly formed mount string: %s",str);
+	*e = 0;
+	e++;
+	if (pfs_resolve_mount(str, e, "rwx") < 0) fatal("call to parrot_mount failed: %s", strerror(errno));
```
btovar: free(str)

btovar:
Tim, other than the small comments, it looks good.

trshaffer:
OK, I pushed some changes to address those comments. As far as I know, it's ready to merge

PR #1494, opened 2016-10-04
Factory Fixes

dthain:
Fixes two problems that were preventing correct operation at Arizona:
* -N batch name cannot be more than 15 characters
* scratch directory of factory needs to be on a shared filesystem.


btovar:
This changes look good. Are they ready to be merged?

dthain:
Hold on

dthain:
Need to fix a build issue and test quickly at the crc.

dthain:
Ok, we are good now.

btovar:
Will wait for blue water fix to make the next bug release.

Issue #1495, opened 2016-10-07
WQ Master Stuck on Task Forsaken


dthain:
@btovar please check this out -

To reproduce, set up an all-pairs job like this:

```
find /usr/share/man/man1 > setA.txt
cp setA.txt setB.txt
allpairs_master setA.txt setB.txt /usr/bin/diff -N testrun -d all
```

Then run a worker like this:
```
work_queue_worker -d all -N testrun
```

Everything goes swimmingly for a while, and then the worker suddenly reports the task as forsaken:
```
orker-102744-25655/t.46/B_152_play.1.gz
2016/10/07 08:47:42.95 work_queue_worker[25655] wq: linking cache/file-0-4974a6e0148e888e1c99cb145d7e92f8-papi_component_avail.1.gz to /tmp/worker-102744-25655/t.46/B_153_papi_component_avail.1.gz
2016/10/07 08:47:42.95 work_queue_worker[25655] wq: linking cache/file-0-dc2b821e066edb43f5ee75476cdd0709-docbook2texi.1.gz to /tmp/worker-102744-25655/t.46/B_154_docbook2texi.1.gz
2016/10/07 08:47:42.95 work_queue_worker[25655] wq: linking cache/file-0-5f70386cd2376fce8abdab7c164885d8-git-write-tree.1.gz to /tmp/worker-102744-25655/t.46/B_155_git-write-tree.1.gz
2016/10/07 08:47:42.95 work_queue_worker[25655] wq: Waiting task 46 has been forsaken.
2016/10/07 08:47:42.95 work_queue_worker[25655] wq: tx to master: alive
2016/10/07 08:47:42.95 work_queue_worker[25655] wq: Sending resource description to master:
2016/10/07 08:47:42.95 work_queue_worker[25655] wq:  workers      0 inuse      1 total      1 smallest      1 largest
2016/10/07 08:47:42.95 work_queue_worker[25655] wq:     disk     26 inuse  32194 total  32194 smallest  32194 largest
2016/10/07 08:47:42.95 work_queue_worker[25655] wq:   memory      0 inuse   5737 total   5737 smallest   5737 largest
2016/10/07 08:47:42.95 work_queue_worker[25655] wq:     gpus      0 inuse      0 total      0 smallest      0 largest
2016/10/07 08:47:42.95 work_queue_worker[25655] wq:    cores      0 inuse      1 total      1 smallest      1 largest
[...]

PR #1496, opened 2016-10-07
Fix forsaken stop

btovar:
Two fixes for #1495:

1) When a task was forsaken, we did not update the accounting for that worker at the master.
2) Memory available at the worker is reported as the total memory, rather than the free memory at the time of measurement.


btovar:
The bug was a combination of two things:

1) For unlabeled tasks, we label them with the full worker resources (minus a fudge factor). Should the resources at the worker be reduced below the fudge factor before the task starts running, the task is marked as forsaken.

2) On forsaken tasks, the master was simply marking the task as ready to run, rather than also updating the accounting for the corresponding worker.

In this case, the resource reduced was memory, since we were sending memory free, rather than memory total. Since memory free does not really mean much, and it may change wildly, tasks were unnecessarily being marked as forsaken.

PR #1498, opened 2016-10-07
Work Queue Display Resources

dthain:
This is a small commit, but a potentially big change.

I propose that the time has come to make WQ multi-core by default.

Currently, a WQ worker only advertises one core, unless someone explicitly overrides this.
As machines are getting bigger and fatter, it is more naturally the case that we are either
(a) running multiple single core tasks on a machine or (b) matching multicore tasks to multicore machines.

This patch causes WQ to advertise all the cores available, unless overriden by the factory,
the batch system or the user.

Also, we explicitly printf the resources at startup, since this is a significant operational concern
and debugging challenge for the naive user.

Thoughts?
Concerns?

### DIFF
--- a/work_queue/src/work_queue_worker.c
+++ b/work_queue/src/work_queue_worker.c
@@ -152,7 +152,7 @@ static struct work_queue_resources * total_resources = 0;
 static struct work_queue_resources * total_resources_last = 0;
 
 static int64_t last_task_received  = 0;
-static int64_t manual_cores_option = 0;
+static int64_t manual_cores_option = 1;
 static int64_t manual_disk_option = 0;
 static int64_t manual_memory_option = 0;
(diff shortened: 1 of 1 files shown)


btovar:
I rather have it as it is right now. If you want automatic measurement, then use --cores=all. Taking by default the whole machine makes the worker a bad citizen. 

Both of these changes also seem geared to make easier to quickly test in the command line. If anything, I'd have new users write --cores=all to  make the point that resources are something to think about.

We also would need to check that scripts (such as sge and slurm_submit_workers) do the right thing. I know condor_submit_workers does.

dthain:
Ok, I'm going back to one core by default, but leaving the explicit output, so the user sees what is going on....

PR #1501, opened 2016-10-08
Support specifying umbrella specs in makefile

hmeng-19:
This PR adds the support to allow the user to specify an Umbrella specification for a group of rule(s) in the Makefile by putting the following directives before the rule(s) you want to apply the Umbrella spec to:

```
.UMBRELLA CATEGORY 1
.UMBRELLA SPEC convert_S.umbrella
```

In this case, the specified Umbrella spec will be applied to all the following rules until a new `.UMBRELLA CATEGORY...` directive is declared. All the rules before the first `.RESOURCE CATEGORY ...` directive will use the Umbrella spec specified by the `--umbrella-spec` option. If the `--umbrella-spec` option is not specified, these rules will run without being wrapped by Umbrella.

An example makefile with umbrella specs can be found at `cctools/makeflow/example/makeflow_umbrella/example_in_makefile.makeflow`:

```
# if the `--umbrella-spec` option is set, then these two rules will use it as their umbrella specs;
# otherwise, they will run as it is without being wrapped by the umbrella wrapper.
capitol.montage.gif: capitol.jpg capitol.90.jpg capitol.180.jpg capitol.270.jpg capitol.360.jpg
        convert -delay 10 -loop 0 capitol.jpg capitol.90.jpg capitol.180.jpg capitol.270.jpg capitol.360.jpg capitol.270.jpg capitol.180.jpg capitol.90.jpg capitol.montage.gif

capitol.90.jpg: capitol.jpg
        convert -swirl 90 capitol.jpg capitol.90.jpg

# the following three rules will use umbrella_specs/convert_S_category1.umbrella as their umbrella specs
.UMBRELLA CATEGORY 1
.UMBRELLA SPEC umbrella_specs/convert_S_category1.umbrella
capitol.180.jpg: capitol.jpg
        convert -swirl 180 capitol.jpg capitol.180.jpg

capitol.270.jpg: capitol.jpg
        convert -swirl 270 capitol.jpg capitol.270.jpg

capitol.360.jpg: capitol.jpg
        convert -swirl 360 capitol.jpg capitol.360.jpg

# the following rule will use umbrella_specs/convert_S_category2.umbrella as its umbrella spec
.UMBRELLA CATEGORY 2
.UMBRELLA SPEC umbrella_specs/convert_S_category2.umbrella
capitol.jpg:
        curl -o capitol.jpg http://ccl.cse.nd.edu/images/capitol.jpg
```

To test makefile with umbrella (without a default umbrella spec):

```
$ makeflow example_in_makefile.makeflow
```

To test makefile with umbrella (with a default umbrella spec):

```
$ makeflow --umbrella-binary $(which umbrella) --umbrella-spec convert_S.umbrella example_in_makefile.makeflow
```

### DIFF
--- a/doc/makeflow.html
+++ b/doc/makeflow.html
@@ -297,6 +297,23 @@ following three modes: local, parrot, and docker. The default value of the
 docker mode, and tries to utilize the parrot mode if the docker mode is not
 available. </p>
 
+<h2 id="umbrella_in_makefile">Specify Umbrella in Makefile<a class="sectionlink" href="#umbrella_in_makefile" title="Link to this section.">&#x21d7;</a></h2>
+
+<p> You can also specify an Umbrella specification for a group of rule(s) in the
+Makefile by putting the following directives before the rule(s) you want to apply
+the Umbrella spec to: </p>
+<code>
+.MAKEFLOW CATEGORY 1
+.UMBRELLA SPEC convert_S.umbrella
+</code>
+
+<p>In this case, the specified Umbrella spec will be applied to all the
+following rules until a new ".MAKEFLOW CATEGORY ..." directive is declared. All
+the rules before the first ".MAKEFLOW CATEGORY ..." directive will use the
+Umbrella spec specified by the <tt>--umbrella-spec</tt> option. If the
+<tt>--umbrella-spec</tt> option is not specified, these rules will run without being
+wrapped by Umbrella. </p>
+
 <h2 id="details">More Language Details<a class="sectionlink" href="#details" title="Link to this section.">&#x21d7;</a></h2>
 
 <p>The Makeflow language is very similar to Make, but it does have a few important differences that you should be aware of.</p>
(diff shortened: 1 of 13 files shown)


hmeng-19:
@dthain , @btovar , @nhazekam , PTAL.

btovar:
How does this interact with .RESOURCE CATEGORY?
I would rather have a single category specification, say .RULE CATEGORY
Both the resource monitor and umbrella then can read the category from the same directive.

hmeng-19:
@btovar , that is the part we may need to get together to sort out.
Currently, using `.RESOURCE CATEGORY ...` and `.UMBRELLA CATEGORY ...` achieve the same effect.

> I would rather have a single category specification, say .RULE CATEGORY
I like this idea.

btovar:
Perhaps .MAKEFLOW CATEGORY?

.RULE seems to indicate only a single rule, which may be confusing.

hmeng-19:
@btovar , @nhazekam , PTAL.
As we discussed, the PR replaces the makefile `.RESOURCE|.UMBRELLA CATEGORY` with `.MAKEFLOW CATEGORY`.

hmeng-19:
@nhazekam , let me know how you think about this.

PR #1513, opened 2016-10-27
Parrot Packaging: Detect File Size

hmeng-19:
Currently, `parrot_package_create` does not keep the file size info when the copy degree of a file is metadata-level copy, because using `truncate` system call on an empty file on AFS results in the `st_blocks` field becomes non-zero, which makes it difficult to determine whether a file is really empty or not. 

This PR is an attempt to keep the file size information. This may cause some applications involving AFS file accesses to break.

### DIFF
--- a/parrot/src/parrot_package_create.c
+++ b/parrot/src/parrot_package_create.c
@@ -17,6 +17,7 @@
 #include <time.h>
 #include <limits.h>
 
+
 #include "copy_stream.h"
 #include "debug.h"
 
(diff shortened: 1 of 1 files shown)


hmeng-19:
@trshaffer , try to use the parrot packaging tool in this PR to reproduce your python3 applications. It should fix the problem.
Let me know if it works for you.

hmeng-19:
@dthain , PTAL.

dthain:
If this works for our common use cases today, then that's fine.

As a long-term measure, I'm not sure that I would rely upon the semantics of `st_blocks`.  Different filesystems seem to interpret that field in different ways.

dthain:
@trshaffer Does this PR work for the python3 use case?

dthain:
@trshaffer need your thumbs up or down on this one.

trshaffer:
:+1:
I can start Python 3 under this branch

PR #1519, opened 2016-10-28
Total task info to catalog

btovar:
@nkremerh: please take a look. I removed the alpha computations from the factory, as they will be done from the master (and we did not get the correct info anyway).


nkremerh:
I approve. Your changes are pretty straightforward.

Issue #1506, opened 2016-10-20
string_cookie has crummy randomness


bbockelm:
I noticed `string_cookie` (whose random string is used `link_auth_password`) has really poor entropy.  It's actually based on `rand` (yuck!).  It weakens the handshake for the authentication.

It seems that `random_hex` could be a drop-in replacement for this usage.

Note that `random_hex` and friends still don't meet cryptographic standards for generating key material, but might be OK in this case.

dthain:
@trshaffer please take a look at this one.

PR #1522, opened 2016-10-31
Use Mersenne Twister in string_cookie

trshaffer:
This is for #1506 

A CSPRNG might be nice, but I agree it's not really necessary for our use. We wouldn't want to use a `string_cookie` for any sort of key material or anything beyond the simple handshake. It looks like Mersenne Twister was added fairly recently and hasn't made it all the way through the codebase, so thanks for the issue @bbockelm

### DIFF
--- a/dttools/src/random.c
+++ b/dttools/src/random.c
@@ -18,10 +18,9 @@
 #include <string.h>
 #include <time.h>
 
-int random_initialized = 0;
-
 void random_init (void)
 {
+	static int random_initialized = 0;
 	if (random_initialized) return;
 	int fd = open("/dev/urandom", O_RDONLY);
 	if (fd == -1) {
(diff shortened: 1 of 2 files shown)


dttools/src/random.c
```diff
@@ -18,8 +18,11 @@
 #include <string.h>
 #include <time.h>
 
+int random_initialized = 0;
+
 void random_init (void)
 {
+	if (random_initialized) return;
```
batrick: you can move this int into this funciton as a static variable

trshaffer: Done

dttools/src/random.c
```diff
@@ -33,12 +35,14 @@ void random_init (void)
 	} else {
 		uint64_t seed;
 nasty_fallback:
+		fprintf(stderr, "warning: falling back to low-quality entropy");
```
dthain: Pls use debug(D_NOTICE) here and I will go ahead and merge...

trshaffer:
Fixed, should be ready to merge

Issue #1504, opened 2016-10-19
wq_factory only works with project names


btovar:
Explicit host names are needed when makeflow is run with --work-queue-preferred-connection=by-ip

As reported by @bbockelm.

bbockelm:
Actually, I don't think it's a `makeflow` issue.  I want to be able to manage the hostname `work_queue_factory` utilizes.  For example, from my cluster headnode, I get:

```
$ work_queue_factory -T slurm -N project-bbockelm
2016/10/19 09:40:11: |submitted: 0 |needed: 8 |waiting connection: 0 |requested: 5 
PROJECT            HOST                   PORT WAITING RUNNING COMPLETE WORKERS 
masters:
project-bbockelm   external.example.com         9001       4       0        0       0 
```

Where `external.example.com` is the external hostname and firewalled off.  However, if I run the `hostname` command itself, it returns `internal.example.com`.

I should be able to override the hostname used explicitly - and it might be better to default the hostname used by the catalog server to be the same as the output of `gethostname()`.  It seems the catalog server uses the default hostname from the address returned by `getpeername()` on the TCP connection.

btovar:
You are correct. I mentioned that makeflow option to eliminate 'just use a project name' solution.

Thanks for the extra info...

PR #1523, opened 2016-10-31
Allow specification of master host and master port in Work Queue Factory

nkremerh:
Addresses #1504 

Adds -H and -p options for master host and master port respectively
Removes project/foreman name regex from function arguments

### DIFF
--- a/batch_job/src/work_queue_factory.c
+++ b/batch_job/src/work_queue_factory.c
@@ -11,6 +11,7 @@ See the file COPYING for details.
 #include "hash_table.h"
 #include "copy_stream.h"
 #include "debug.h"
+#include "domain_name_cache.h"
 #include "envtools.h"
 #include "stringtools.h"
 #include "xxmalloc.h"
@@ -19,6 +20,7 @@ See the file COPYING for details.
 #include "delete_dir.h"
 #include "macros.h"
 #include "catalog_query.h"
+#include "link.h"
 #include "list.h"
 #include "get_line.h"
 #include "getopt.h"
@@ -28,6 +30,7 @@ See the file COPYING for details.
 
 #include "jx.h"
 #include "jx_parse.h"
+#include "jx_print.h"
 #include "jx_table.h"
 
 #include <errno.h>
@@ -42,6 +45,24 @@ See the file COPYING for details.
 #include <sys/time.h>
 #include <unistd.h>
 #include <signal.h>
+
+typedef enum {
+	FORMAT_TABLE,
+	FORMAT_LONG
+} format_t;
+
+static struct jx_table queue_headers[] = {
+{"project",       "PROJECT", JX_TABLE_MODE_PLAIN, JX_TABLE_ALIGN_LEFT, -18},
+{"name",          "HOST",    JX_TABLE_MODE_PLAIN, JX_TABLE_ALIGN_LEFT, -21},
+{"port",          "PORT",    JX_TABLE_MODE_PLAIN, JX_TABLE_ALIGN_RIGHT, 5},
+{"tasks_waiting", "WAITING", JX_TABLE_MODE_PLAIN, JX_TABLE_ALIGN_RIGHT, 7},
+{"tasks_running", "RUNNING", JX_TABLE_MODE_PLAIN, JX_TABLE_ALIGN_RIGHT, 7},
+{"tasks_complete","COMPLETE",JX_TABLE_MODE_PLAIN, JX_TABLE_ALIGN_RIGHT, 8},
+{"workers",       "WORKERS", JX_TABLE_MODE_PLAIN, JX_TABLE_ALIGN_RIGHT, 7},
+{NULL,NULL,0,0,0}
+};
+
+static int work_queue_status_timeout = 30;
 
 static const char *catalog_host = 0;
 static int catalog_port = 0;
@@ -58,7 +79,12 @@ static int worker_timeout = 300;
 static int consider_capacity = 0;
 
 static char *project_regex = 0;
+static char *submission_regex = 0;
 static char *foremen_regex = 0;
+
+char *master_host = 0;
+int master_port = 0;
+int using_catalog = 0;
 
 static char *extra_worker_args=0;
 static const char *resource_args=0;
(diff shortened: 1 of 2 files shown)


nkremerh:
@btovar Here is a first pass. How does it look?

nkremerh:
@btovar looks like your comments finally showed up, that was a weird GitHub hiccup. Regardless, I believe I resolved your requested changes. Please take a look when you have time.

batch_job/src/work_queue_factory.c
```diff
@@ -782,6 +791,8 @@ enum { LONG_OPT_CORES = 255, LONG_OPT_MEMORY, LONG_OPT_DISK, LONG_OPT_GPUS, LONG
 
 static const struct option long_options[] = {
 	{"master-name", required_argument, 0, 'M'},
+	{"master-host", required_argument, 0, 'H'},
```
btovar: Dead code.

batch_job/src/work_queue_factory.c
```diff
@@ -930,6 +941,22 @@ int main(int argc, char *argv[])
...
+			show_help(argv[0]);
+			exit(1);
+		}
+
+		master_host = argv[optind];
+		master_port = atoi(argv[optind+1]);
+
+	}
+
+	if(!(master_host && master_port)) {
```
btovar: This conditional is unneeded/incorrect. Already covered in line 989 anyway.

btovar:
Nate, did you test it? It does not quite work...

You want to use host:port to ask for information (e.g., do a query ala line 648, but contacting host:port rather than the catalog). See how we do it for -T and -W in work_queue_status.

nkremerh:
Huh, I tested locally and it ran fine. Let me work on it.

btovar:
It is not getting information from the masters. It simply assigns the host:port to the workers, which is not what we want. (E.g., how do we know how many workers do we need?)

dthain:
@nkremerh is this one fixed or not?

nkremerh:
@dthain Not yet, I got a bit sidetracked. I will make this a priority.

nkremerh:
@btovar I believe this resolves the issue. Please let me know what you think when you get a chance.

batch_job/src/work_queue_factory.c
```diff
@@ -150,6 +176,44 @@ int master_workers_needed_by_resource(struct jx *j) {
...
+	struct list* master_list = list_create();
+
+	time_t stoptime = time(0) + work_queue_status_timeout;
+
+	if(!domain_name_cache_lookup(master_host,master_addr)) {
+		fprintf(stderr,"couldn't find address of %s\n",master_host);
+		return 0;
+	}
+
+	l = link_connect(master_addr,master_port,stoptime);
```
btovar: l is never freed.

batch_job/src/work_queue_factory.c
```diff
@@ -954,6 +1030,17 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	if(!project_regex) {
```
btovar: if(project_regex) {
using_catalog = 1;
else if(master_name ...)
else
help and exit

batch_job/src/work_queue_factory.c
```diff
@@ -980,10 +1067,10 @@ int main(int argc, char *argv[])
 			fprintf(stderr,"work_queue_factory: There were errors in the configuration file: %s\n", config_file);
 			return 1;
 		}
-	}
+	}	
 
-	if(!project_regex) {
-		fprintf(stderr,"work_queue_factory: You must give a project name with the -M option, or the master-name option with a configuration file.\n");
+	if(!project_regex && !(master_host && master_port)) {
```
btovar: Would be covered with the previous comment.

batch_job/src/work_queue_factory.c
```diff
@@ -625,7 +696,7 @@ current list of masters, compare it to the number actually submitted, then
 submit more until the desired state is reached.
 */
 
-static void mainloop( struct batch_queue *queue, const char *project_regex, const char *foremen_regex )
```
nkremerh: Rather than passing project_regex and foremen_regex to the mainloop (even if we do not use them), I chose to make them global. I can undo this if needed.

nkremerh:
@btovar How does this look?

batch_job/src/work_queue_factory.c
```diff
@@ -84,6 +84,7 @@ static char *foremen_regex = 0;
 
 char *master_host = 0;
 int master_port = 0;
+int using_catalog = -1;
```
btovar: = 0

batch_job/src/work_queue_factory.c
```diff
@@ -320,7 +321,7 @@ static int submit_worker( struct batch_queue *queue )
 {
 	char *cmd;
 
-	if(submission_regex) {
+	if(using_catalog) {
```
btovar: This is incorrect if using_catalog = -1. You want using_catalog = 0, or 1.

batch_job/src/work_queue_factory.c
```diff
@@ -1069,7 +1074,7 @@ int main(int argc, char *argv[])
 		}
 	}	
 
-	if(!project_regex && !(master_host && master_port)) {
+	if(using_catalog < 0) {
```
btovar: Move this to the if else above:

if(project_regex...)
else if((argc - optind ....
else
  fpritnf(stderr, ....

nkremerh:
@btovar I was trying to be unnecessarily clever. This should take care of things.

batch_job/src/work_queue_factory.c
```diff
@@ -1036,14 +1036,13 @@ int main(int argc, char *argv[])
 	}
 	else {
 		if((argc - optind) != 2) {
+			fprintf(stderr,"work_queue_factory: You must either give a project name with the -M option or master-name option with a configuration file, or give the master's host and port.\n");
```
btovar: I rather have this with an else all by itself. Then it is easier to follow the logic.

nkremerh: Sure, that makes sense. Would you rather I take that whole if statement and make it a separate case (if project_regex...else if argc - optind != 2...else...) or move the fprintf to a separate case? Is there a style preference?

nkremerh:
@btovar Commits squashed, how does it look?

btovar:
Ready to merge, if you are.

nkremerh:
@btovar Ready.

PR #1524, opened 2016-11-02
Let bindings

trshaffer:
Makeflow's JX representation was written to allow JX variables to be introduced at the file level and at the rule level, using special `"variables"` keys under the root object and rule objects. Unfortunately, this made evaluating the Makeflow complicated. Previously, Makeflow's parser built the contexts itself and `jx_eval`ed each rule object individually. Aside from being ugly, this precludes using a function or operator to build the list of rules, as you need to evaluate that down to an array. @nhazekam have been thinking about how to do this for some time and were coming up with quoting + multiple eval passes, some sort of partial or lazy evaluation, eval hooks, etc.

I realized that Makeflow has no business building and manipulating variable contexts, which should be part of the JX language. As of this PR, Makeflow only knows about `"environment"` mappings that it can put into the DAG. To allow local variable contexts, I added a new JX function,

    let(bindings, expr)

that binds the names in `bindings` into the context for evaluating `expr`. The bindings themselves get evaluated in the enclosing context. I think this is less surprising than the previous behavior, as there are no magic Makeflow evaluation rules; plain `jx_eval` gives the same result that Makeflow will use. Users also get more precise and straightforward control over what variables are in scope, and aren't restricted to awkward per-rule evaluation contexts.

This also brings us one function closer to Lisp...

### DIFF
--- a/doc/jx.html
+++ b/doc/jx.html
@@ -412,6 +412,25 @@ join(["a", "b", "c"], ", ")
 =&gt; "a, b, c"
 </code></pre>
 
+<h3 id="let">let<a class="sectionlink" href="#let" title="Link to this section.">&#x21d7;</a></h3>
+
+<blockquote>
+<pre><code>let(A, B) -&gt; C
+</code></pre>
+
+<p>where A = Object</p>
+</blockquote>
+
+<p>Evaluates an expression with local name bindings. The first argument is an object whose keys will be bound variable names when evaluating the second argument.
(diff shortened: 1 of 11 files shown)


nhazekam:
This does make changes to jx_parse. I feel this will have a limited affect on how we use it to this point, but wanted to comment to make this clear that if we merge and see odd behavior at the catalog (though with those constant now...) we know where to look first.

Issue #1503, opened 2016-10-12
use parrot_run options instead of environ vars in Umbrella


hmeng-19:
Currently, umbrella's parrot execution engine accepts the mount file and the ldso path via environment variables, `PARROT_MOUNT_FILE` and `PARROT_LDSO_PATH`. This may make it difficult to debug any potential problems in the future.

Solution:
Use the command options instead.

hmeng-19:
Fixed by 1534.

PR #1534, opened 2016-11-11
Use parrot_namespace to allow parrot-in-parrot

hmeng-19:
Use `-m` and `-l` instead of env variables to set options of parrot


hmeng-19:
This should fix #1503 .

PR #1538, opened 2016-11-16
Reset the input files of makeflow umbrella wrapper for each rule

hmeng-19:
Every rule in a makefile may have its own umbrella spec file.
To avoid `w->wrapper->input_files` accumulating umbrella spec files for different rules, first clean it up and then recreate it.


hmeng-19:
@nhazekam , PTAL.

PR #1540, opened 2016-11-17
Stop checking the existence of umbrella logfile

hmeng-19:
Problems:
Currently, makeflow always first check whether a umbrella logfile already exists before starting or recovering a makeflow. This makes it impossible for the users to retry their makeflows.

Failure cases:
A makeflow run may fail if some jobs failed on some bad workers, and the user may want to retrying the makeflow to rerun these failed jobs. Checking the existence of umbrella_logfile would cause makeflow retries fail.

Solutions:
Therefore, in this PR, we stop check the existence of umbrella_logfile here.


hmeng-19:
@nhazekam , PTAL.

Issue #1541, opened 2016-11-17
Take advantage of shared filesystems with Makeflow


annawoodard:
This is a feature request. I am running an HEP analysis with Makeflow+WQ+Condor. Part of my workflow runs over a large collection of input files which are several gigabytes each, stored in an HDFS filesystem which is accessible to all machines in our cluster. I would like to be able to tell Makeflow about these files by specifying them as inputs, so I can take advantage of nice Makeflow features like re-running downstream rules if one if its input files are modified. Running over this collection of inputs is an "embarrassingly parallel" problem, but parallelization is broken because absolute paths are forbidden-- WQ+condor must be run in transfer mode, so the input files are transferred in series. @dthain suggested you might be able to add a command line switch which would allow me to tell WQ that any file under a given path is globally accessible to the workers and does not need to be transferred. His idea has the advantage that it preserves the status of the Makeflow script as a complete description of the workflow which is independent of the batch system.

In case it's helpful, you can find an example Makeflow script for the workflow I'm referring to here: `/afs/crc.nd.edu/user/a/awoodard/www/ttV/6/1/1/Makeflow`.

annawoodard:
@hmeng-19 see the path above in case it is useful for the Umbrella test. Ping me if I can be helpful in any way.

hmeng-19:
@annawoodard , thanks. I will give it a try.

dthain:
What we need here is a command line argument like this:

```
makeflow -T condor --shared-fs /afs --shared-fs /hadoop
```

The effect of the argument will be quite simple: files inside directories that are indicated as shared filesystems will not be added to the input or output lists of a batch job, when it is constructed from a `dag_node`.  The effect is that makeflow will still "manage" the file at the DAG level, but not pass it down to the batch level.

dthain:
@trshaffer please take a look at this one, I think it is a quick fix and will have a big impact on Anna's work.

dthain:
@trshaffer please add to the manual and man pages, and discuss with @annawoodard to see if it meets her needs...

PR #1545, opened 2016-11-20
Add --shared-fs to Makeflow

trshaffer:
The new `--shared-fs` option can be specified multiple times and takes a directory that's accessible to the head node and all workers through some external means, e.g. HDFS. Makeflow will do all the usual file checks when building the DAG and scheduling tasks. Any input or output files living under a shared fs, however, aren't emitted to the batch system to be transferred. This allows files on a shared FS to be specified in the DAG and to take advantage of existence and age checks, but prevents the batch system from doing extra work on these (potentially large) files.

This is a really simple stab at #1541;  interested parties should have a look and see if this works.
@dthain @annawoodard @nhazekam


dthain:
It seems odd to return null from `makeflow_file_format`, since the caller (`makeflow_file_list_format`) uses the returned pointer in a number of ways and then frees it.  Does that actually work?

Why not do the check in `makeflow_file_list_format` and skip items that match?

trshaffer:
`string_combine()` and `free()` both do the right thing on null pointers, but I could see someone in the future getting tripped up and assuming non-null pointers. I moved that logic out into the loop so it's obviously part of the control flow

annawoodard:
Sorry about the delay. Thanks for putting this together, @trshaffer!

This is probably an error on my side, but this is what I see when I try to run, after picking up your changes:
```
[earth] ~/www/ttV/6/1/4 >makeflow -T wq -M ttV_FTW --wrapper ./w.sh --wrapper-input w.sh --shared-fs /hadoop
parsing ./Makeflow...
the --umbrella-binary option is not set, therefore an umbrella binary should be available on an execution node if umbrella is used to deliver the execution environment.
checking ./Makeflow for consistency...
2016/11/22 11:10:32.10 makeflow[1427617] error: Absolute paths are not supported on selected batch system. Rule 197.
```

The Makeflow specification is here: `/afs/crc.nd.edu/user/a/awoodard/www/ttV/6/1/4/Makeflow`

Note that the only difference between it and a working version are the `/hadoop*` input files. I've put a working version here in case that's helpful: `/afs/crc.nd.edu/user/a/awoodard/www/ttV/6/1/4/Makeflow.works`

trshaffer:
My mistake @annawoodard, I didn't test with Work Queue, so I hadn't noticed that error. I just pushed a commit to skip that check under `--shared-fs`. Let me know if this works for you

makeflow/src/makeflow.c
```diff
@@ -1523,6 +1538,14 @@ int main(int argc, char *argv[])
 				if(!wrapper) wrapper = makeflow_wrapper_create();
 				makeflow_wrapper_add_output_file(wrapper, optarg);
 				break;
+			case LONG_OPT_SHARED_FS:
+				assert(shared_fs);
+				s = string_combine(xxstrdup(optarg), "/");
+				t = realpath(s, NULL);
+				free(s);
+				if (!t) fatal("can't access shared filesystem %s: %s", optarg, strerror(errno));
+				list_push_head(shared_fs, string_combine(t, "/"));
```
annawoodard: @trshaffer 
It looks like you might be appending the trailing slash twice (here and 1543)? Changing 1547 to this makes my makeflow script work:
```
list_push_head(shared_fs, t);
```

trshaffer: Huh, that's odd...
The double append is actually intentional. The first one forces `realpath()` to resolve the path as a directory, which should give back an absolute path without a trailing `/`. The second append is to ensure that shared fs checks work under directories, not just under string prefixes, e.g. should match `/hadoop/foo` but not `/hadoop1`

What OS/libc are you using? Maybe there's some platform weirdness. If that's the case, I'm OK with removing the second append, so that `--shared-fs` matches `/hadoop*`

dthain: IMHO, it's best not to "massage" file names unless there is a really compelling reason, otherwise confusing things can happen.  It's probably simplest to just store the string given by the user, and compare it to the string names in the batch jobs.  (The individual files will have their existence checked anyhow.)

trshaffer: OK, makes sense. I just pushed a commit to rip that part out

annawoodard:
It's working perfectly for me! Thanks @trshaffer and @dthain, this new feature is really going to improve my workflow!

dthain:
@btovar please take a look and merge if you are satisfied.

makeflow/src/makeflow.c
```diff
@@ -1523,6 +1537,10 @@ int main(int argc, char *argv[])
 				if(!wrapper) wrapper = makeflow_wrapper_create();
 				makeflow_wrapper_add_output_file(wrapper, optarg);
 				break;
+			case LONG_OPT_SHARED_FS:
+				assert(shared_fs);
+				list_push_head(shared_fs, xxstrdup(optarg));
```
btovar: I suggest we check that shared_fs is an absolute path.

btovar:
Thanks Tim, ready to merge if you are.

trshaffer:
Just added that bit of debug logging, so I think so

dthain:
Complete!

Issue #1550, opened 2016-11-22
an issue about the http timeout setting in makeflow mount module 


hmeng-19:
Currently, the http timeout value in the makeflow mount module (used via the `--mounts` option) is hard-coded as 300 seconds, which is too short if a dependency is huge or the user runs multiple instances of a makeflow concurrently.

hmeng-19:
@btovar , @dthain , 
Currently, the mount module depends on `http_fetch_to_file`, and does not support https deps (I do not think `dttools` includes any library for fetching files via https.)
I have two options: libcurl and wget.
Using libcurl we need to either assume the machine has libcurl installed in a standard path or import libcurl into cctools somehow.
Using wget we need to assume `wget` is available on the machine.
Currently, I plan to use the `wget`. Let me know if you have any thoughts on this.

dthain:
Or possible `curl`.  Is one used more widely than the other?

hmeng-19:
@dthain , take a look at here https://daniel.haxx.se/docs/curl-vs-wget.html

hmeng-19:
@dthain , the thing I would like go with `wget` is:
"Its ability to recover from a prematurely broken transfer and continue downloading has no counterpart in curl."

dthain:
Ok, that's fine with me.

PR #1557, opened 2016-11-29
Use wget to download http(s) resources in the makeflow mount files

hmeng-19:
Use `wget` to download the http and https resources specified in the makeflow mount files.


hmeng-19:
The PR aims to fix #1550 .
@dthain , PTAL.

hmeng-19:
@dthain , as you suggested, I used `link_close` instead of `free`. Please take another look.

dthain:
Looks good.

PR #1565, opened 2016-12-08
use the new canonical external sofware paths (ND AFS)

btovar:


dthain:
Oh, I didn't think of the docker case.
You can ignore my suggestions when they don't work.  :)

Issue #1568, opened 2016-12-12
work_queue_status --capacity


dthain:
Please create an option `work_queue_status --capacity` that easily displays the current reported capacity in tasks, cores, memory, and disk.  Should help in initial diagnosis of scalability problems.

dthain:
Looks good, please add to the `work_queue_status` man page, and let's call it done.

PR #1570, opened 2016-12-13
Adds capacity to Work Queue Status

nkremerh:
Implements #1568 

@dthain Please let me know how this first pass looks.


dthain:
Close #1568

Issue #1571, opened 2016-12-13
Parrot Bind Test Failing


dthain:
It seems that the parrot bind test failed recently, though it's not immediately clear what commit caused it:
http://ccl.cse.nd.edu/software/autobuild/

The test fails on an attempt to bind a Unix domain socket.
Reproduce by logging into cclvm04.crc.nd.edu, building master and running this:
```
% cd parrot/test
% ./TR_parrot_bind.sh prepare
% ./TR_parrot_bind.sh run
[<stdin>:35] bind(master, (struct sockaddr *)&addr, sizeof(addr)): Operation not permitted
...
```
@trshaffer please take a look at this, since you most recently looked at Unix domain sockets in parrot.

trshaffer:
I get this error even if I'm not running under Parrot, so it appears to be a problem with the test itself. I just put in a PR to resolve it

dthain:
Ok, that's a relief

dthain:
Resolved by #1574

PR #1574, opened 2016-12-14
Don't test socket bind in the working directory

trshaffer:
This addresses #1571 

If the current working directory doesn't support sockets (which appears to be the case on some versions of AFS), the test fails. This patch switches to absolute pathnames that are extremely unlikely to be on weird filesystems

### DIFF
--- a/parrot/test/TR_parrot_bind.sh
+++ b/parrot/test/TR_parrot_bind.sh
@@ -26,7 +26,7 @@ prepare()
 		if ((expr) == -1) {\\
 			fprintf(stderr, "[%s:%d] ", __FILE__, __LINE__);\\
 			perror(#expr);\\
-			unlink("foo");\\
+			unlink(argv[1]);\\
 			exit(EXIT_FAILURE);\\
 		}\\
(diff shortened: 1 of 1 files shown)


parrot/test/TR_parrot_bind.sh
```diff
@@ -82,14 +86,14 @@ EOF
 run()
 {
 	set -e
-	parrot -- "$(pwd)/$exe"
-	parrot --work-dir=/http -- "$(pwd)/$exe" && return 1
+	parrot -- "$(pwd)/$exe" /tmp/foo
+	parrot -- "$(pwd)/$exe" /http/foo && return 1
```
dthain: I'm confused, what's the purpose of this line?

trshaffer: I think it's making sure that non-writable directories like `/http` don't allow sockets. That seemed a little odd, but it's how the test was written. Do you want to get rid of that?

dthain: Oh, I get it.

PR #1579, opened 2017-01-09
Add `mode` attr into umbrella specs

hmeng-19:
This is useful for the case when an executable is downloaded as a plain file instead of a tarball. 

By default, the python `urllib` library turns off the execute mode of a file, which can not be found by searching the `$PATH` env. 

Using the new `mode` attr allows the user to set the file mode of a plain file.


hmeng-19:
This PR aims to fix #1171 .

hmeng-19:
@dthain , @Baguage , please take a look.

dthain:
Makes sense to me

Issue #1559, opened 2016-12-01
Parrot doesn't know about magic symlinks under /proc/*/ns/


trshaffer:
On recent kernels, the process namespaces are visible as symlinks, for example on my RHEL6 workstation

    [tshaffe1@cclws19 ~]$ ll /proc/self/ns/
    total 0
    dr-x--x--x. 2 tshaffe1 dip 0 Dec  1 13:14 .
    dr-xr-xr-x. 8 tshaffe1 dip 0 Dec  1 13:14 ..
    lrwxrwxrwx. 1 tshaffe1 dip 0 Dec  1 13:14 ipc -> ipc:[4026531839]
    lrwxrwxrwx. 1 tshaffe1 dip 0 Dec  1 13:14 mnt -> mnt:[4026531840]
    lrwxrwxrwx. 1 tshaffe1 dip 0 Dec  1 13:14 net -> net:[4026531954]
    lrwxrwxrwx. 1 tshaffe1 dip 0 Dec  1 13:14 pid -> pid:[4026531836]
    lrwxrwxrwx. 1 tshaffe1 dip 0 Dec  1 13:14 uts -> uts:[4026531838]

These symlinks don't point to anything in the filesystem; the value returned by `readlink()` is just text referring to an inode number. A syscall that follows symlinks will nonetheless successfully resolve these.

    [tshaffe1@cclws19 ~]$ stat -L /proc/self/ns/ipc 
      File: `/proc/self/ns/ipc'
      Size: 0         	Blocks: 0          IO Block: 1024   regular empty file
    Device: 3h/3d	Inode: 4026531839  Links: 1
    Access: (0444/-r--r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
    Access: 2016-12-01 13:21:04.199953881 -0500
    Modify: 2016-12-01 13:21:04.199953881 -0500
    Change: 2016-12-01 13:21:04.199953881 -0500

As noted in #1548, Parrot tries to follow these symlinks and returns `ENOENT` from calls that should have succeeded.

dthain:
However, note this:
```
$ cat /proc/self/ns/ipc
cat: /proc/self/ns/ipc: Invalid argument
```

trshaffer:
[...]

PR #1583, opened 2017-01-10
Don't follow namespace symlinks

trshaffer:
This addresses #1559

Since the namespace symlinks have some unusual behaviour, Parrot shouldn't try to follow them by itself. Also, this pull request changes the handling of `/proc/self` and `/dev/fd`. Previously, these paths short circuited name resolution, so for example, mount lists simply wouldn't be applied to any path with those prefixes. This moves the substitution as late as possible, after name resolution and symlink following.

### DIFF
--- a/dttools/src/stringtools.c
+++ b/dttools/src/stringtools.c
@@ -141,7 +141,7 @@ int whole_string_match_regex(const char *text, const char *pattern)
 }
 
 
-int string_match_regex(const char *text, char *pattern)
+int string_match_regex(const char *text, const char *pattern)
 {
 	int ret = 0;
 	regex_t re;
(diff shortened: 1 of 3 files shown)


dthain:
@btovar please take a look at this one.
@trshaffer could this have any bearing on #1584?

trshaffer:
@dthain I don't think so. That issue seems to be related to `mmap`. Any real I/O operations on these magic files should just fail

btovar:
It looks good to me.

btovar:
@trshaffer: is this ready to be merged?

trshaffer:
@btovar I don't have any additional changes, so go ahead if you're ready

Issue #1582, opened 2017-01-10
python allocation test is failing for some platforms


btovar:
E.g.,

log: http://ccl.cse.nd.edu/a/806c4512/806c4512.x86_64-debian8.log
tests: http://ccl.cse.nd.edu/a/806c4512/806c4512.x86_64-debian8.tests.log

PR #1585, opened 2017-01-12
python allocation fix tests

btovar:
It was hard to get consistent samples across runs, so we do not check for a particular value anymore.


btovar:
Fixex #1582

PR #1589, opened 2017-01-17
fix mis-assignment of categories

btovar:
When switching to .MAKEFLOW directives, all the related variables were being assigned at the node level, rather than the dag. This meant that all nodes were assigned to the first category that was defined.

Since we do not use directives inside a node anyway, I simply deleted that from the code.

Also, we were assigning two times the values to a variable when it was both a directive and a special variable. The result was not an error, but it was inefficient.

### DIFF
--- a/makeflow/src/dag_visitors.c
+++ b/makeflow/src/dag_visitors.c
@@ -74,7 +74,7 @@ int dag_to_file_exports(const struct dag *d, FILE * dag_stream, const char *pref
 	set_first_element(vars);
 	for(name = set_next_element(vars); name; name = set_next_element(vars))
 	{
-		v = hash_table_lookup(d->default_category->mf_variables, name);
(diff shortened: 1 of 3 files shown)


nhazekam:
Does removing the directives from all nodes affect how umbrella uses? I believe the umbrella specification needs to be tied to the rule it applies to if its in the rule. @hmeng-19 will this at all affect how umbrella uses it? 

This should be revisited with a more comprehensive fix later, as I think having directives that can work on a node specifically may be useful.

btovar:
Currently, specifications cannot be in the rule, but just before it, if that is what you mean.

before the rule:
```make
SOMETHING=something
output: input
\tcommand
```

inside the rule:
```make
output: input
@SOMETHING=something
\tcommand
```

btovar:
@nhazekam, @hmeng-19: please make sure your stuff still works with this changes to makeflow.

These changes should be backwards compatible. If they are not, let me know.

PR #1590, opened 2017-01-18
configure: replace expr substr with native bash

kevinwern:
According to POSIX, `substr` in `expr` produces undefined results. In
some versions of `expr`, it has no special meaning. This causes
configure to fail on these expressions and assign the literal string
"$Format:%.*$" to config commit values, which causes the build to fail
because the closing quote is interpreted as a variable.

Change instances of `expr substr $VAR $index $len` to ${VAR:$index:$len},
which should be supported by all instances of bash.

### DIFF
--- a/configure
+++ b/configure
@@ -36,7 +36,7 @@ fi
 
 # `git archive` (man gitattributes(5)) fills this in, otherwise we do it manually...
 COMMIT='$Format:%H$'
-if [ "$(expr substr "$COMMIT" 1 1)" = '$' ]; then
+if [ "${COMMIT:0:1}" = '$' ]; then
 	if [ "$GIT" = y ]; then
 		COMMIT="$(git rev-parse HEAD)"
 	else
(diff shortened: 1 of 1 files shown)


batrick:
I believe the original reason I did this `substr` nonsense is that `configure` is an `sh` script, not `bash`. I think maybe reverting 1fb823f099 and fixing https://github.com/cooperative-computing-lab/cctools/blob/04344f19ad976e9cfc5852d894c8751244461e76/configure.tools.sh#L24-L36 (which apparently doesn't work on Blue Waters, hence 1fb823f099) is the right way to go.

kevinwern:
> I believe the original reason I did this substr nonsense is that configure is an sh script, not bash. 

I just checked, and the POSIX shell standard doesn't include my modification--I guess `bash --posix` is pretty unreliable. That makes sense.

I don't know what the issue could be with Blue Waters. Nothing jumps out to me in the code. '-0' shouldn't be passed to `head`, but that shouldn't arise based on how the code is called (or rather, was called before 1fb823f).

Closing this pull request because I don't think I can solve this without more information.

dthain:
Thanks for the attempt, this one has been bugging me for a while.

As a matter of procedure note that pull requests get built on all of our platforms, and it looks like this one triggered some oddity in debian and ubuntu:
http://ccl.cse.nd.edu/software/autobuild/

IMHO, its an accident inversion to be doing `git` operations inside of `configure`, since the software should work independently of the storage means.  I think the more robust solution is to take the git version detection out of configure, and simply pass in the revision information via a configure argument, when needed.

btovar:
See #1593.

Issue #1584, opened 2017-01-12
Git breaks when run with parrot


matz-e:
Per discussion in the meeting earlier today. On `crcfe01`, this unexpected behavior occurs:
```
crcfe01: which parrot_run
/afs/crc.nd.edu/group/ccl/software/x86_64/redhat7/cctools/current/bin/parrot_run
crcfe01: parrot_run git clone -vvv git@github.com:cms-ttH/ttH-TauMCGeneration.git
Cloning into 'ttH-TauMCGeneration'...
fatal: bad config file line 3 in /afs/crc.nd.edu/user/m/mwolf3/ttH-TauMCGeneration/.git/config
```
without parrot, git clones the repository successfully.

dthain:
Tim, please take a look at this one.

trshaffer:
Quick update on this issue

It looks like any operation that creates a new repo,  including `git clone` or `git init` is affected. Under Parrot, git is writing out a corrupted config file during repo initialization. I'm tracking down the cause now, but I suspect it's somewhere deep in Parrot's `mmap` code

dthain:
To be clear, is git attempting to write to an mmaped file?  Or just read?

trshaffer:
No, it reads/maps `.git/config`, writes to `.git/config.lock`, then renames to replace the config with the lock file. More careful invalidation on rename would probably fix this, but I suspect there would still be combinations of writes and mappings that are well defined but handled incorrectly.

dthain:
Maybe a up-to-date check when performing mmap would be more robust than
invalidate-on-close?  e.g. stat the open file to make sure the size, time,
inode match?

trshaffer:
I just put in a small pull request to get correct handling of renames. I tested it, and git seems to work again. The clone in the original post still won't work, but that's because ssh tries to use a `setuid` helper. If I clone over HTTP, everything's fine

dthain:
@trshaffer close if this is solved

PR #1594, opened 2017-01-18
Make existing channel anonymous on rename

trshaffer:
This is a minimal fix for #1584. Currently Parrot incorrectly handles channels during a rename. If there's an existing mapped entry under the new name, then a rename will result in multiple entries with the same name, only one of which is current. This patch makes any files being renamed over anonymous, that way they won't be returned from the cache. Once the last existing mapping is removed, they can be cleaned up as usual.


dthain:
I think that's a good conservative solution to the rename problem.

trshaffer:
I added a note about `mmap`, so I think this is ready to merge

PR #1595, opened 2017-01-18
Git out configure

### LINKED ISSUE #1593: Take git out of configure
Related to #1590

btovar:
Fixes #1593.

Git code was moved to configure.afs.

To ./configure, I added --build-label and --build-date. These are set from configure.afs (or manually, if a user so desires.)


dthain:
No objections here.

Issue #1591, opened 2017-01-18
CernVM Parrot/Docker container doesn't start under newer versions of Docker


jblomer:
Hi all,

happy new year (and apologies for starting it with an issue report)!

As you know, we have a [Docker container](https://cernvm.cern.ch/portal/docker) where parrot provides the init process and presents the OS environment from cvmfs (--> [download](https://cernvm.cern.ch/portal/downloads)).

This container works fine on older versions of docker but with newer versions (e.g. 1.12.5) parrot seems to be unable to trace children.  In the docker container, we use patchelf to set parrot_run's interpreter to /UCVM/lib/ld-linux-x86-64.so.2, but I'm not sure if this is related.

Below are two outputs, one from a new docker version and one from a working older version of docker.  Can you help us?

Thanks!
Jakob

# Docker 1.12.5
```
$ docker run -e CERNVM_DEBUG=1 -e PARROT_OPTIONS="-d all" -it cernvm /init
***********************************
*** Welcome to CernVM in Docker ***
***********************************

  -- based on uCernVM bootloader version 2.7-8.cernvm.x86_64

[INF] Mounting operating system /cvmfs/cernvm-prod.cern.ch/cvm3 with parrot
[DEBUG] link /bin to /cvmfs/cernvm-prod.cern.ch/cvm3/bin
[DEBUG] link /lib to /cvmfs/cernvm-prod.cern.ch/cvm3/lib
[DEBUG] link /lib64 to /cvmfs/cernvm-prod.cern.ch/cvm3/lib64
[DEBUG] link /opt to /cvmfs/cernvm-prod.cern.ch/cvm3/opt
[DEBUG] link /sbin to /cvmfs/cernvm-prod.cern.ch/cvm3/sbin
[DEBUG] create empty /srv
[DEBUG] create empty /tmp
[DEBUG] create empty /usr
[DEBUG] create empty /usr/local
[DEBUG] link /usr/bin to /cvmfs/cernvm-prod.cern.ch/cvm3/usr/bin
[DEBUG] link /usr/doc to /cvmfs/cernvm-prod.cern.ch/cvm3/usr/doc
[DEBUG] link /usr/etc to /cvmfs/cernvm-prod.cern.ch/cvm3/usr/etc
[DEBUG] link /usr/include to /cvmfs/cernvm-prod.cern.ch/cvm3/usr/include
[...]

PR #1598, opened 2017-01-19
Add helpful warning if ptrace attach fails with EPERM

trshaffer:
As a follow-up from #1591, I added a nicer warning message if `ptrace` appears to be disabled. This should give users some pointers rather than just looking like Parrot is broken


dthain:
That's the right idea.

trshaffer:
As far as I know, this is ready to merge

PR #1599, opened 2017-01-23
Enhance parrot_packaging_run

hmeng-19:
Currently, `parrot_package_run` does not work well on packages created by tools
other than `parrot_package_create` due to lacking of `common_mountlist` and
`env_list` files inside the packages.

This commit fixes this.


hmeng-19:
@dthain , PTAL.

dthain:
What other sort of packages are you intending to work with?

hmeng-19:
Any package including a rootfs image, for example, a docker image tarball. 
Currently, running `parrot_package_run` on a docker tarball fails because some parts of the rootfs, such as `/proc` and `/dev` are not mounted into the sandbox.

dthain:
I see, so you have some reasonable defaults if the user is just providing a plain old tar file.  Makes sense to me.

PR #1600, opened 2017-01-31
Add parrot features: fixed PID and warped PID

pivie:
Fixed PID was used to help bring at least one application closer to being deterministic.

These could be dangerous options if there are any incidental PID conflicts and important processes could get killed. Use at your own risk.


dthain:
The approach looks good, but you have a merge conflict, please rebase, merge, and push again.

Issue #1601, opened 2017-02-02
Link uses IPv6 by default


trshaffer:
The default behaviour when listening with no address specified is to use `IN6ADDR_ANY`, which accepts both IPv4 and IPv6 binds. This seems to cause problems if IPv6 is not available in the kernel. Pull request if forthcoming

dthain:
@btovar I think we need to make a point release for this PR (and any others you think suitable.)

PR #1602, opened 2017-02-03
Check CCTOOLS_IP_MODE when handling addresses

trshaffer:
This is for #1601. With no address specified, the default behaviour is IPv4 only, which can be overridden with an environment variable. @dthain this fixes the issue I was having on the CRC machines

### DIFF
--- a/dttools/src/address.c
+++ b/dttools/src/address.c
@@ -1,31 +1,45 @@
 #include "address.h"
+#include "debug.h"
 #include <assert.h>
 #include <string.h>
 #include <arpa/inet.h>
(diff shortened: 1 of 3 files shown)


dthain:
Yes, that's the right idea, but factor out the little bit of logic that does getenv and decides on the default mode.  We want that in one place (called from both link.c and address.c) so that the default is controllable from one place.

trshaffer:
I just pushed a refactored version

dthain:
Looks good, go ahead and merge when the tests pass.

PR #1603, opened 2017-02-03
Adds --snapshot-file=<file> to resource_monitor

btovar:
When <file> exists, the resource_monitor takes a snapshot of the
resources and removes <file>. The snapshot is written to
PREFIX.snapshot.NN, where PREFIX is the prefix used for the summary
file. The snapshots are also included in the final summary file, under
the key "snapshots".

Example:
myscript.sh
init()
echo "after_init" > signal.file
analysis()
echo "after_analysys" > signal.file

$ resource_monitor --snapshot-file=signal.file -Omysummary -- ./myscript.sh
$ ls
mysummary.snapshot.01
mysummary.snapshot.02
mysummary.summary

(as requested by @klannon, and @matz-e)

### DIFF
--- a/doc/man/resource_monitor.m4
+++ b/doc/man/resource_monitor.m4
@@ -2,7 +2,7 @@ include(manual.h)dnl
 HEADER(resource_monitor)
 
 SECTION(NAME)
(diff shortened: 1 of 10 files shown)


btovar:
I still need to integrate it to work queue...

PR #1609, opened 2017-02-08
Fix indentation to compile on newer GCC

trshaffer:
Newer versions of GCC (6.3.1 on my laptop) issue a warning about "misleading indentation" here. I had to switch to tabs to compile with `--strict`


trshaffer:
Trivial change and no objections, so merged

Issue #1610, opened 2017-02-13
Parrot /proc/self/exe


dthain:
Parrot test `TR_parrot_procexe.sh` is failing.  @trshaffer please take a look and see how this relates to the recent bug report on `/proc/self`.  Is the test itself doing something sensible?

PR #1611, opened 2017-02-14
Match /proc/self more carefully

trshaffer:
Old versions of `find` read links in an odd way. In the case of `/proc/self/exe`, `find` first opens `/proc/self`, then reads the link `exe`. Parrot's symlink following has been short circuiting on any resolved path with prefix `/proc/self/` (note the trailing slash), so opening `/proc/self` falls through to Parrot.

I wrote a more careful regex, and adjusted the footwork for dealing with symlinks and `/proc`, which I think is *more correct*. Testing on RHEL5-7, the failing test from #1610 passes with these changes.

### DIFF
--- a/parrot/src/pfs_table.cc
+++ b/parrot/src/pfs_table.cc
@@ -548,27 +548,27 @@ int pfs_table::resolve_name(int is_special_syscall, const char *cname, struct pf
 			follow_symlink(pname, mode, depth + 1);
 		}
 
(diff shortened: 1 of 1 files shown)


dthain:
The approach looks right to me.

trshaffer:
I'd like to get some yellow off the Autobuild board, so merging this now

PR #1613, opened 2017-02-16
Add local mode to GROW-FS

trshaffer:
This is a first draft of local mode for GROW-FS. The idea is to index a directory on the local system, and let Parrot handle metadata without touching the filesystem. To access the directory `/home/john` with this, first generate the index.

    $ make_growfs /home/john
    $ touch /home/john/foo

Note that we added a file after generating the index. Next, run your program under Parrot. `/grow/` should still work as before, except for the path `/grow/local`. Parrot will search the local filesystem rather than connect to a web server. Now you can look at metadata under `/home/john` without touching the disk.

    $ ls /grow/local/home/john
    $ cat /grow/local/john/foo          # ENOENT: the file is not in the index

Any reads are passed down to the local filesystem.

I haven't done extensive testing, but it seems to work as expected. Some limitations:
- This mode could shadow a GROW server. If someone actually needs to connect to a GROW server named `local`, it's easy to add a command line flag for this feature.
- Only one repo is supported. The first access to a local repo binds that path. For example, if you access `/grow/local/home/john/repoA`, you can no longer access `/grow/local/home/john/repoB`. The GROW service does this for a given host:port as well.

@dthain thoughts?

### DIFF
--- a/dttools/src/Makefile
+++ b/dttools/src/Makefile
@@ -95,6 +95,7 @@ SOURCES = \
 	sigdef.c \
 	sleeptools.c \
 	sort_dir.c \
+	stats.c \
 	string_array.c \
 	stringtools.c \
 	text_array.c \
--- a/dttools/src/stats.c
+++ b/dttools/src/stats.c
@@ -1,0 +1,127 @@
+/*
+Copyright (C) 2017- The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file COPYING for details.
+*/
+
+#include <assert.h>
+#include <inttypes.h>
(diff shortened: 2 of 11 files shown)


dthain:
I think it's the right idea; we may want multiple indexes later, but one will do for now.

dthain:
Ready to merge?

trshaffer:
I added a couple of small things; pending Autobuild, ready to merge

trshaffer:
I have commits for another PR hanging off this one, so if there are no objections, I'd like to merge

dthain:
Ok w/me

Issue #1612, opened 2017-02-15
Argument lists was messed up in Parrot when an symlink executable is executed


hmeng-19:
Supposed we have two files, `source.sh` and `target.sh`.
`target.sh` is a simple shell script:
```
#!/bin/sh
basename $0
```
`source.sh` is a symlink pointing to `target.sh`.

When we run `parrot_run ./source.sh`, the output is `target.sh`. However, the correct
output should be `source.sh`.

hmeng-19:
@trshaffer , please take a look. Thanks.

PR #1614, opened 2017-02-17
Pass the right argv[0]

trshaffer:
It looks like some previous fixes lined up to cause an issue for the specific case of `exec()`ing a symlink to an interpreted script. `decode_execve` was overwriting the passed-in path with the value of the symlink. This should leave `argv[0]` as passed in while still setting `/proc/self/exe` to the canonical path to the executable.

This seems to resolve the issue in #1612


dthain:
Thank you for adding a regression test.

trshaffer:
@hmeng-19 does this fix that issue with VirtualBox you mentioned?

hmeng-19:
@trshaffer , this PR fixes the issue #1612.  Good job.

trshaffer:
OK, then I'll merge this and close the issue

PR #1616, opened 2017-02-21
add preferred connection to bindings

btovar:
Adds the binding to specify 'by_ip' or 'by_hostname'.


nhazekam:
Sorry Ben, I did not realize this wasn't in the bindings! I just saw that
it was in the WQ API.

Nick

PR #1619, opened 2017-03-02
add x.x.x.x:pppp;y.y.y.:qqqq master specification

### LINKED ISSUE #1615: Configure which IP is bound/advertised
Hello!

I'm currently getting my feet wet w/cctools (and Lobster) at Vanderbilt, and I've stumbled into a difficulty in our firewalls.

Our interactive nodes are dual-homed with interfaces on `129.59.197.0/24` and `10.0.0.0/16`, and all of our worker nodes live within the private `10.0.0.0/16` range. Any host that is dual homed has its default route set to the external interface and its hostname is set to the public FQDN. If I run a simple lobster example then check the catalog server, I see the following (trimmed) output:

```
$ work_queue_status
PROJECT                      HOST                             PORT WAITING RUNNING COMPLETE WORKERS 
lobster_meloam_20170221      vmps13.accre.vanderbilt.edu      9000      10       0        0       0 
```

Unfortunately, our local security policy requires that we block all external access to the interactive nodes, so trying to connect to (e.g.) `vmps13.accre.vanderbilt.edu:9000` won't work. To be able to use WQ here, we'd want to somehow tell WQ/the catalog that the preferred address to connect to is the internal one, which we also have DNS configured for, if that helps:

```
$ dig +short vmps13.vampire
10.0.20.238
```

If the workers tried to connect to ` vmps13.vampire`/`10.0.20.238` instead of `vmps13.accre.vanderbilt.edu`/`129.59.197.183`, then the connection would work okay.

Would it be possible to tell WQ masters what IP address they should advertise? I think it would be most helpful if it could be configured via environment variables which could be set in login script instead of having to remember to add the command line option each time. Or, perhaps it would be possible for the WQ/catalog to advertise/track all the ip addresses on the host, then let the clients try each one in succession.

Thanks!

btovar:
First part to fix #1615. 

Allows several  master hosts to be specified per worker.

### DIFF
--- a/doc/man/work_queue_worker.m4
+++ b/doc/man/work_queue_worker.m4
@@ -8,6 +8,12 @@ dispatched through Work Queue
 SECTION(SYNOPSIS)
 CODE(BOLD(work_queue_worker [options] PARAM(masterhost) PARAM(port)))
 
+CODE(BOLD(work_queue_worker [options] PARAM(masterhost:port])))
+
+CODE(BOLD(work_queue_worker [options] "PARAM(masterhost:port;[masterhost:port;masterhost:port;...])))"
+
+CODE(BOLD(work_queue_worker [options] -M PARAM(projectname)))
+
 SECTION(DESCRIPTION)
 
 BOLD(work_queue_worker) is the worker process for executing tasks dispatched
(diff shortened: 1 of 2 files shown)


dthain:
Looks like the right idea.

Main question: What exactly is the fail-over behavior when one master cannot be contacted?

btovar:
All masters are kept on a (essentially circular) list. When a master cannot be contacted, the next one is tried, until all masters  are tried. Then we increase the back_off time, and try again.

dthain:
Ok, that sounds good.

PR #1620, opened 2017-03-03
Inform catalog about multiple interfaces

### LINKED ISSUE #1615: Configure which IP is bound/advertised

btovar:
Second part to fix #1615.

The master informs the catalog about multiple interfaces with a json field that looks like:

"network_interfaces":[{"family":"AF_INET","host":"129.74.153.171","interface":"eth0"},{"family":"AF_INET","host":"172.17.42.1","interface":"docker0"},{"family":"AF_INET6","host":"fe80::b6b5:2fff:fecf:a0d4%eth0","interface":"eth0"},{"family":"AF_INET6","host":"fe80::c025:bff:feaa:31f2%docker0","interface":"docker0"}]

Only works on linux, as it needs getifaddrs, which is not POSIX.  If the master runs on a non-linux machine, the field is not included.

### DIFF
--- a/configure
+++ b/configure
@@ -1288,14 +1288,15 @@ optional_function utime /usr/include/utime.h HAS_UTIME HAVE_UTIME
 optional_function utimensat /usr/include/sys/stat.h HAS_UTIMENSAT
 
 # sqlite3 uses define "HAVE_*"
-optional_include attr/xattr.h HAS_ATTR_XATTR_H
-optional_include fts.h HAS_FTS_H
-optional_include inttypes.h HAS_INTTYPES_H HAVE_INTTYPES_H
-optional_include stdint.h HAS_STDINT_H HAVE_STDINT_H
-optional_include sys/statfs.h HAS_SYS_STATFS_H
+optional_include attr/xattr.h  HAS_ATTR_XATTR_H
+optional_include fts.h         HAS_FTS_H
+optional_include ifaddrs.h     HAS_IFADDRS
+optional_include inttypes.h    HAS_INTTYPES_H   HAVE_INTTYPES_H
+optional_include stdint.h      HAS_STDINT_H     HAVE_STDINT_H
+optional_include sys/statfs.h  HAS_SYS_STATFS_H
 optional_include sys/statvfs.h HAS_SYS_STATVFS_H
-optional_include sys/xattr.h HAS_SYS_XATTR_H
-optional_include syslog.h HAS_SYSLOG_H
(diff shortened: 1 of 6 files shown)


dthain:
Wow, that's not as awful as I expected it to be.  :)

One thing to think about: how does this interact with IPV4/IPV6 mode selection?
Should the reporter leave out address modes?  Or should the worker?  Or something else?

(See `address_check_mode` in `domain_name` for example)

btovar:
I think the reporter should leave out the information, as the extra information is useless for the rest of the components.

Also, I realized that the previous pr is broken for ipv6, as we look for port numbers after a ':'.

btovar:
With the last commit, only the current mode is reported.

btovar:
E.g.:
"network_interfaces":[{"family":"AF_INET","host":"129.74.153.171","interface":"eth0"},{"family":"AF_INET","host":"172.17.42.1","interface":"docker0"}]

PR #1626, opened 2017-03-08
FUSE module for GROW-FS

trshaffer:
This is a first draft implementation of a FUSE module providing in-memory metadata operations and file caching. Currently, only the newly-added local mode is supported. I can look at adding HTTP down the road if there's demand.

I refactored the GROW parsing and metadata code into a new project/module, which builds the FUSE parts if the host system has FUSE available. We're still adding features and adjusting things, so when this code settles down after the paper I can work on documentation and tests.


dthain:
You are on the right track.

It would be helpful to isolate what the public interface of growfs is, in contrast to the implementation details.

I think it's something like this:

```
struct growfs_dirent * growfs_lookup( const char *path, struct growfs_dirent *root )
int growfs_open( struct growfs_dirent *d );
int growfs_read( int fd, void *data, int length );
int growfs_close( int fd );
```

And then limit the fuse and the parrot module to accessing only the public interface.

Issue #1629, opened 2017-03-20
Running cvmfs-enabled parrot_run on a server with cvmfs mounted by FUSE


hmeng-19:
Currently, running cvmfs-enabled parrot_run on a server (e.g., `earth.crc.nd.edu`) with cvmfs installed by FUSE does not behave elegantly.

Here is an example illustrating this problem:
```
hmeng@earth ~/test/cms_umbrella$ parrot_run ls /cvmfs/cms.cern.ch/
2017/03/20 11:38:21.50 parrot_run[1040246] <child:1040247> notice: CVMFS requires an http proxy.  None has been configured!
2017/03/20 11:38:21.50 parrot_run[1040246] <child:1040247> notice: CVMFS requires an http proxy.  None has been configured!
2017/03/20 11:38:21.50 parrot_run[1040246] <child:1040247> notice: CVMFS requires an http proxy.  None has been configured!
ls: cannot access /cvmfs/cms.cern.ch/: No such file or directory
```

There are several potential solutions to tackle this problem:
1) change the parrot_run code to detect whether /cvmfs is mounted by FUSE: if yes, parrot_run disbles its cvmfs module and use the cvmfs on the server directly; if no, parrot_run delivers cvmfs.

2) add a new option into parrot_run to allow the users to disable the cvmfs module, this should also silent the following warning when the user is running sth irrelavant to cvmfs.
```
2017/03/20 11:38:21.50 parrot_run[1040246] <child:1040247> notice: CVMFS requires an http proxy.  None has been configured!
```

3) ignore it. Let the user to suffer and figure out they would need two versions of parrot_run on a machine with cvmfs installed by FUSE. This one is not very nice.

hmeng-19:
@dthain , @trshaffer , @btovar , PTAL.

trshaffer:
I'd be in favor of adding an option to disable a given service, since I don't like trying to guess what the user wants to do. If there are no objections, I can put in a quick pull request on this

hmeng-19:
@trshaffer , adding a new option sounds good to me. 
It will be great if you have time to fix this. Thanks very much.

PR #1632, opened 2017-03-23
Add --disable-service to parrot

trshaffer:
Add an option to Parrot to disable a compiled-in service at runtime. This addresses #1629, so that it's possible to tell Parrot not to intercept `/cvmfs` and instead let things fall through to the FUSE module. @hmeng-19 let me know if this works for you


hmeng-19:
@trshaffer , just to clarify the usage of this new opiton.
To use the `--disable-service` option to disable cvmfs module, I should run:
```
parrot_run --disable-service cvmfs -- <cmd>
```
To disable multiple services, should I use multiple `--disable-service` options? Or the option can take a string including multiple servcies like "cvmfs irod"?

hmeng-19:
@trshaffer , this PR solves #1629 . Good job!
Before merging it, it would be nice if you can add the new option into the docs of parrot_run: `doc/parrot.html` and `doc/man/parrot_run.m4`.
Except for the docs, the PR looks good to me. Thanks.

trshaffer:
I added those docs, so this should be ready to merge

Issue #1647, opened 2017-04-27
wq_maker: -disk and -memory options


nathanweeks:
In cctools/apps/wq_maker/wq_maker, lines 381-387

```
	$task->specify_cores($cores);
	if($disk){
		$task->specify_cores($disk);
	}
	if($memory){
		$task->specify_cores($memory);
	}
```

Should that be:

```
	$task->specify_cores($cores);
	if($disk){
		$task->specify_disk($disk);
	}
	if($memory){
		$task->specify_memory($memory);
	}
```

nhazekam:
Thanks for letting us know. I'm currently working on a branch and well
address this asap. I'll mention this issue with the pull request tonight.

nhazekam:
This should have been addressed, please reopen if you have more comments.

PR #1648, opened 2017-05-01
Maker stats

nhazekam:
Update to print stats and properly set resources. This requires Error.pm and Error/Simple.pm, both of which are included with MAKER. However, the user needs to add it PERL5LIB path.


nhazekam:
Additionally addresses #1647

dthain:
Since @btovar is travelling and these changes are local to WQ-Maker, I'll approve.
(Keep the customer satisfied.)

PR #1653, opened 2017-05-15
Add --cms_siteconf opt into umbrella

### LINKED ISSUE #1645: allow Umbrella users to set SITECONF for cms apps
Background:
Run cms applications using parrot in Umbrella requires a `SITECONF` directory which includes the site configuration files (https://twiki.cern.ch/twiki/bin/view/CMSPublic/SiteConfInGitlab).

Problem:
Currently, umbrella does not allow the users to specify the `SITECONF` dir, instead it utilizes a frozen copy of `SITECONF`.

Solution:
Add a new option into umbrella to allow the users to set SITECONF for cms applications.

hmeng-19:
This option allows the users to specify a SITECONF for their cms applications instead of using the default one provided by umbrella.
The option should be a tar.gz local file path (e.g., /tmp/1.tar.gz) or url path (e.g., http://.../1.tar.gz) pointing to the site configuration files for cms applications, the SITECONF dir should be organized following the suggestions at https://twiki.cern.ch/twiki/bin/view/CMSPublic/SiteConfInGitlab

This PR aims to fix #1645 .

### DIFF
--- a/umbrella/src/umbrella.py
+++ b/umbrella/src/umbrella.py
@@ -3997,6 +3997,9 @@ To check the help doc for a specific behavoir, use: %prog <behavior> help""",
 	parser.add_option("--parrot_path",
 					action="store",
 					help="the path of parrot_run on the host machine",)
+	parser.add_option("--cms_siteconf",
+					action="store",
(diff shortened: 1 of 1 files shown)


hmeng-19:
@pivie , please take a look.

pivie:
Thanks for adding this feature. While it didn't fix the problem I was having, it seems to work and should be a helpful feature for non-ND researchers.

Issue #1657, opened 2017-05-20
seteuid not permitted when using ssh under parrot 


khurtado:
Hello,

I was testing an application that at some point executes some stuff on a different node via ssh (ssh <host> <command> ) and noticed this gives problems inside parrot. This is what I get using cctools 6.1.0 on RH6:

```
$ ./parrot_run ssh --help
PRIV_END: seteuid: Operation not permitted
```

Any ideas?

```
$ ssh -V
OpenSSH_5.3p1, OpenSSL 1.0.1e-fips 11 Feb 2013
$ cat /etc/redhat-release 
Scientific Linux Fermi release 6.9 (Ramsey)

```

dthain:
The short story here is that Parrot cannot support an application that tries to change its privilege level via the setuid (and related) system calls.  That's a fundamental limitation of ptrace on Linux, and we can't get around it.

Perhaps the larger question is, why is ssh trying to change its privilege level?

trshaffer:
That is the ssh client you're using, correct? And are you running as root? Even in that case, it seems odd to try to drop privileges

I can have a look at what SSH is trying to do here

dthain:
Thanks Tim

khurtado:
@dthain: Good question, I'm just to print the `--help` menu in the example above :)
@trshaffer: This is running the ssh client as a user. I just reproduced this on `crcfe02.crc.nd.edu` (in case it makes things easier to debug `ssh`) with an AFS-located parrot binary:

```
$ whoami
khurtado
$ hostname
crcfe02.crc.nd.edu
$ /afs/crc.nd.edu/group/ccl/software/x86_64/redhat7/cctools/6.1.0/bin/parrot_run ssh --help
PRIV_END: seteuid: Operation not permitted
```

trshaffer:
Hmmmm, it looks like SSH explicitly sets its EUID to its RUID. I'm not certain why they would want to drop privileges here, though. I can look into automatically allowing no-op UID changes in Parrot.

In the meantime, @khurtado, try running Parrot with the `--fake-setuid` option. I think it should be enough to get things working here.

khurtado:
@trshaffer : That did the trick, thank you!

PR #1659, opened 2017-05-22
Parrot Setuid No-Ops

trshaffer:
This addresses #1657 

I also refactored some of the id change code, which was duplicated and all over the place


dthain:
This all looks good to me.  Just make sure to add a comment to the code somewhere to indicate why it's necessary to support no-op setuids.

trshaffer:
I copied my commit message into the source to explain things

PR #1660, opened 2017-05-23
Makeflow JX format

trshaffer:
This PR documents Makeflow's JX mode. The code has been in place for some time, but it wasn't really publicized or documented. I added a section to the Makeflow manual documenting the format, and tweaked a few things. I also added some regression tests for the JX format. I decided not to include any  JX features in these tests, as langauge-related stuff is orthogonal to Makeflow and belongs in the tests for JX itself. There might be some adjustments on the JX end, but I think the Makeflow format should be usable for someone who wants to work with plain JSON.

@btovar @nhazekam PTAL


dthain:
Two main things here -
1 - Clearly distinguish between plain JSON format, describe that first, and then describe JX in a separate section.  (We may even want different parsing modes, so that you know when you are using strict JSON.)
2 - Include some examples in addition to the specification.  Maybe the JSON equivalent of the animation Makefile at the beginning?

trshaffer:
I might like to see a `--json` option as well as `--jx` for Makeflow. I made a point to only use JSON in Makeflow's format spec, so I could just change the manual to say JSON and mention that we also support applying JX stuff with the appropriate flag.

There's already a JX example in the `examples/` directory, I could add a flat JSON one.

trshaffer:
I've been working on a number of things for this:
- differentiate Makeflow's JSON format from JX
- add `--json` in addition to `--jx` to Makeflow to choose which to parse
- update JSON and JX examples and link from the Makeflow manual
- add comments (introduced by `#`) to JX
- actually report JX errors from Makeflow
- make Makeflow's files specified as objects only, with "path" and "execution_path"
- track line numbers in the JX parser
- give more helpful error messages in JX parse and eval

I still have a few things on my todo list:
- add parsing modes to JX, e.g. JSON or constant-only
- add a utility for expanding JX into JSON

The remaining stuff shouldn't block Makeflow's JSON format. Makeflow people should have a look at the [JSON format](https://htmlpreview.github.io/?https://github.com/trshaffer/cctools/blob/makeflow-jx/doc/makeflow.html#json) before merging

makeflow/src/makeflow.c
```diff
@@ -126,6 +126,7 @@ static int skip_file_check = 0;
 
 static int cache_mode = 1;
 
+static int json_input = 0;
 static int jx_input = 0;
 static char *jx_context = NULL;
```
nhazekam: If these variables are only used in main function, move these to the beginning of that function. If there is a foreseeable reason to have them as global statics just let me know.

makeflow/src/parser_jx.c
```diff
@@ -148,21 +137,21 @@ static int rule_from_jx(struct dag *d, struct jx *j) {
 	struct jx *inputs = jx_lookup(j, "inputs");
 	debug(D_MAKEFLOW_PARSER, "Parsing inputs");
 	if (!files_from_jx(n, 1, inputs)) {
-		debug(D_MAKEFLOW_PARSER, "Failure parsing inputs");
+		debug(D_MAKEFLOW_PARSER|D_NOTICE, "Failure parsing inputs");
```
nhazekam: See above.

makeflow/src/parser_jx.c
```diff
@@ -148,21 +137,21 @@ static int rule_from_jx(struct dag *d, struct jx *j) {
...
 		return 0;
 	}
 	struct jx *outputs = jx_lookup(j, "outputs");
 	debug(D_MAKEFLOW_PARSER, "Parsing outputs");
 	if (!files_from_jx(n, 0, outputs)) {
-		debug(D_MAKEFLOW_PARSER, "Failure parsing outputs");
+		debug(D_MAKEFLOW_PARSER|D_NOTICE, "Failure parsing outputs");
```
nhazekam: See above.

makeflow/test/TR_makeflow_json_local.sh
```diff
@@ -0,0 +1,26 @@
...
+prepare()
+{
+	clean $@
+}
+
+run()
+{
+	# slurm shouldn't be set up here, so this should only run
+	# if local_job works
+	cd json && ../../src/makeflow -T slurm --json local.json && exit 0
```
nhazekam: I like this as a means to ensure local is working. Would there be any point in having a noop batch system for testing things like this?

makeflow/src/parser_jx.c
```diff
@@ -93,36 +93,25 @@ static int resources_from_jx(struct hash_table *h, struct jx *j) {
...
-		debug(D_MAKEFLOW_PARSER, "Input must be a string or object");
+	if (!jx_istype(j, JX_OBJECT)) {
+		debug(D_MAKEFLOW_PARSER|D_NOTICE, "File must be specified as an object");
+		return 0;
+	}
+
+	const char *path = jx_lookup_string(j, "path");
+	const char *remote = jx_lookup_string(j, "execution_path");
+	if (!path) {
+		debug(D_MAKEFLOW_PARSER|D_NOTICE, "File must have a path");
```
nhazekam: I must have deleted this comment, should have been in the review.

Would it make sense to say "File must have key:value pair for 'path'" to clarify that this is JSON formatted values. This applies to the other comments in this file.

makeflow/src/parser_jx.c
trshaffer: For this and the output one, `files_from_jx` will print a more descriptive error message and return failure, so this just needs to say if it was inputs or outputs

PR #1663, opened 2017-05-26
Add line numbers in makeflow error messages

annawoodard:
For debugging errors in makeflow scripts, associating a problematic rule with its line number might not be straightforward in some cases (like long makeflow scripts with comments). This PR adds the line number to the error message.

before:
```2017/05/26 14:24:41.09 makeflow[45593] error: Absolute paths are not supported on selected batch system. Rule 244.```

after:
```2017/05/26 14:26:03.26 makeflow[46565] error: Absolute paths are not supported on selected batch system. Rule 244 (line 740).```

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -935,11 +935,11 @@ static int makeflow_check_batch_consistency(struct dag *d)
 
 		if(itable_size(n->remote_names) > 0 || (wrapper && wrapper->uses_remote_rename)){
 			if(n->local_job) {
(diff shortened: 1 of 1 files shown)


nhazekam:
This seems like a reasonable consideration. I will look into whether the parser is currently able to track our line numbers.

nhazekam:
Scratch what I just said. I thought this was an issue report. I'll review this as well.

nhazekam:
Merged as this was a minor change and the tests passed. Will revisit if needed.

btovar:
@annawoodard: thanks Anna!

PR #1666, opened 2017-05-31
remove redundant code

charleszheng44:
We declared variable `b` as `unit_64`, which is an unsigned number, thus `b = abs(b)` is not necessary any more.


btovar:
Yes, previous incantations used int_64, and probably it was then when the abs call was introduced.

btovar:
Thanks @charleszheng44

PR #1667, opened 2017-06-04
makeflow k8s mode

charleszheng44:
1. The kubectl is required

2. The working process of k8s mode,

1) For each task, a json configuration file is created to generate the
corresponding pod on the cluster
2) The batch_job_k8s_wait is called to keep polling each running pod
and check the status of each pod
3) If the pod is running, we launch the task inside the pod
4) At the same time, we keep polling the task state, if the task state
"job_done" we return the task id.

3. TODO: The batch_queue_k8s_free need to be implemented


charleszheng44:
This PR is deprecated, please refer to #1760

PR #1673, opened 2017-06-07
Makeflow Amazon Batch Support

Nekel-Seyew:


makeflow/src/makeflow.c
```diff
@@ -1233,6 +1233,7 @@ int main(int argc, char *argv[])
 
 	random_init();
 	debug_config(argv[0]);
+	debug_config_file_size(string_metric_parse("0"));//to set debug file size to "don't delete anything"
```
nhazekam: Is there any need to pass this through `string_metric_parse`? I believe you can either create an off_t or just pass 0, with no need to added a string conversion into the mix.

dthain:
Ok, my main suggestion for this body of code is to work on factoring out actions into separate functions that are no more than a few lines each.  That will makes this much easier to read, and then deal with failures that occur.

For example, write one function named `run_json_command` that just takes a command string, runs it through popen, parses the stream as JX ,and returns the jx object.  Set that at the top of your code.

Then, for each interaction with AWS, write a function named `aws_submit_job` or similar that constructs a command string, calls `run_json_command`, parses the output, deletes temporaries, and returns the item of interest.

Then, the main body of your code will be more comprehensible because it will consist of a few calls to some small functions.

I think if you put some effort into factoring that way, you will get better control over error handling and debugging...

That's all for now.

dthain:
Also, let me recommend that you work by making one localized improvement, testing it, committing the change, then pushing it.  That's usually more successful than accumulating huge changes in one commit.  And I can also help to point you in the right direction.

batch_job/src/Makefile
```diff
@@ -3,7 +3,7 @@ include ../../rules.mk
 
 LIBRARIES = libbatch_job.a
 
-PROGRAMS = work_queue_factory work_queue_pool
+PROGRAMS = work_queue_factory work_queue_pool amazon_batch_cleanup amazon_batch_initialize
```
dthain: A small thing, but important for keeping our namespace clean:
Call these scripts `makeflow_amazon_batch_setup` and `makeflow_amazon_batch_cleanup`

batch_job/src/amazon_batch_cleanup_script.sh
```diff
@@ -0,0 +1,83 @@
...
+#
+# File: amazon_batch_cleanup_script.sh
+# Author: Kyle Sweeney
+#
+# Created on Jul 17, 2017
+#
+
+configFile=$1
+
+ec2_vpc=$(python -c "import json; print json.load(open('$configFile','r'))['vpc'];")
```
dthain: That's a neat idea, it didn't occur to me to do it that way.
Does this work in both python 2 and python 3?

batch_job/src/amazon_batch_cleanup_script.sh
```diff
@@ -0,0 +1,83 @@
...
+ec2_security_group_id=$(python -c "import json; print json.load(open('$configFile','r'))['sec_group'];")
+env_name=$(python -c "import json; print json.load(open('$configFile','r'))['env_name'];")
+queue_name=$(python -c "import json; print json.load(open('$configFile','r'))['queue_name'];")
+bucket_name=$(python -c "import json; print json.load(open('$configFile','r'))['bucket'];")
+gateway=$(python -c "import json; print json.load(open('$configFile','r'))['gateway'];")
+
+#disabling jobqueues
+aws batch update-job-queue --job-queue $queue_name --state DISABLED
+
+queue_done_check_output="$(aws batch describe-job-queues --job-queues $queue_name)"
```
dthain: Suggest that you echo "Deleting queue X" etc as you go along, so the user sees what's happening.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,465 @@
...
+
+static struct internal_amazon_batch_amazon_ids{
+	char* aws_access_key_id;
+	char* aws_secret_access_key;
+	char* aws_region;
+	char* aws_email;
+	char* master_env_prefix;
+}initialized_data;
+
+static struct jx* run_command(char* cmd){
```
dthain: Good, you are getting the right idea.
Handle the case where popen fails.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,465 @@
...
+
+static void clean_str_array(int size, char*** array){
+	int i=0;
+	for(i=0; i< size; i++){
+		free((*array)[i]);
+	}
+	free(*array);
+}
+
+static void upload_input_files_to_s3(char* files){
```
dthain: Upload could fail, so this should return int.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,465 @@
...
+}
+
+
+static struct internal_amazon_batch_amazon_ids initialize(struct batch_queue* q){
+	if(initialized){
+		return initialized_data;
+	}
+	
+	FILE* config_file = fopen("./makeflow_amazon_batch.config","r");
+	if(!config_file){
```
dthain: I wouldn't do this automatically, since the script could also fail and produces important output.
Leave it to the user to do manually, so that they are on the hook for doing the cleanup step as well.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,465 @@
...
+	clean_str_array(files_split_num,&files_split);
+}
+
+
+static struct internal_amazon_batch_amazon_ids initialize(struct batch_queue* q){
+	if(initialized){
+		return initialized_data;
+	}
+	
+	FILE* config_file = fopen("./makeflow_amazon_batch.config","r");
```
dthain: Handle failure of fopen.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,465 @@
...
+	bucket_name                 = (char*)jx_lookup_string(config,"bucket");
+	vpc                         = (char*)jx_lookup_string(config,"vpc");
+	sec_group                   = (char*)jx_lookup_string(config,"sec_group");
+	queue_name                  = (char*)jx_lookup_string(config,"queue_name");
+	compute_env_name            = (char*)jx_lookup_string(config,"env_name");
+	subnet                      = (char*)jx_lookup_string(config,"subnet");	
+
+	//const char* aws_email = jx_lookup_string(config,"aws_email");
+
+	if(!aws_access_key_id)
```
dthain: Give the file name in the error message.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,465 @@
...
+	//copy out any external files
+	char* cpy_out = generate_s3_cp_cmds(output_files,"./",bucket);
+	cmd_tmp = string_format("%s\n%s\n",cmd_tmp,cpy_out);
+
+	//add headder
+	char* final_cmd = string_format("#!/bin/sh\n%s",cmd_tmp);
+	free(cmd_tmp);
+
+	//write out to file	
+	FILE* tmpfile = fopen("TEMPFILE.sh","w+");
```
dthain: fopen can fail

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,465 @@
...
+	system("chmod +x TEMPFILE.sh");
+	cmd_tmp = string_format("%s aws s3 cp ./TEMPFILE.sh s3://%s/COMAND_FILE_%i.sh",env_var,bucket_name,jobid);
+	system(cmd_tmp);
+	free(cmd_tmp);
+	remove("TEMPFILE.sh");	
+
+
+}
+
+static char* aws_submit_job(char* job_name, char* properties_string){
```
dthain: Split this into two functions: one to define the job and the other to submit it.

makeflow/src/makeflow.c
```diff
@@ -1118,6 +1118,7 @@ static void show_help_run(const char *cmd)
 	printf(" %-30s Advertise the master information to a catalog server.\n", "-a,--advertise");
 	printf(" %-30s Specify path to Amazon credentials (for use with -T amazon)\n", "--amazon-credentials");
 	printf(" %-30s Specify amazon-ami (for use with -T amazon)\n", "--amazon-ami");
+	printf(" %-30s Specify path to Container Image for Amazon-Batch (for use with -T amazon)\n", "--amazon-batch-image");
```
dthain: Do we also need an option for the config file generated by `makeflow_amazon_batch_setup`?

dthain:
I made a bunch of comments inline, but it's coming along nicely...

batch_job/src/batch_job.h
```diff
@@ -35,6 +35,7 @@ typedef enum {
 	BATCH_QUEUE_TYPE_LOCAL,	              /**< Batch jobs will run as local processes. */
 	BATCH_QUEUE_TYPE_CONDOR,              /**< Batch jobs will be sent to Condor pool. */
 	BATCH_QUEUE_TYPE_AMAZON,              /**< Batch jobs will be sent spun up Amazon ec2 instances */
+        BATCH_QUEUE_TYPE_AMAZON_BATCH,        /**< Batch jobs will be sent to Amazon Batch System */
```
dthain: Just FYI, I am updating the amazon batch type to work in a manner similar to that of your code.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,465 @@
+/*
```
dthain: Use the CCL license boilerplate as in batch_job.h

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,639 @@
...
+		fatal("Cannot open /dev/urandom");
+	union amazon_batch_ccl_guid guid;
+	size_t k = fread(guid.c,sizeof(char),8,ran);
+	if(k<8)
+		fatal("couldn't read 8 bytes from /dev/urandom/");
+	fclose(ran);
+	return guid.ul;	
+}
+static struct jx* run_command(char* cmd){
+	FILE* out = popen(cmd,"r");
```
dthain: What if popen fails?

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,639 @@
...
+
+static void clean_str_array(int size, char*** array){
+	int i=0;
+	for(i=0; i< size; i++){
+		free((*array)[i]);
+	}
+	free(*array);
+}
+
+static void upload_input_files_to_s3(char* files,char* jobname){
```
dthain: Uploads can and will fail. This function must detect those failures and return overall success or failure.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,639 @@
...
+	
+
+	initialized_data.aws_access_key_id = aws_access_key_id;
+	initialized_data.aws_secret_access_key = aws_secret_access_key;
+	initialized_data.aws_region=aws_region;
+	initialized_data.master_env_prefix = env_var;
+	return initialized_data;
+}
+
+/*static batch_job_id_t get_id_from_name(char* name){//name has structure of "QUEUENAME_id"
```
dthain: Eliminate dead code.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,639 @@
...
+	jx = run_command(tmp);
+	free(tmp);
+	char* jaid = strdup((char*)jx_lookup_string(jx,"jobId"));
+	if(!jaid)
+		fatal("NO JOB ID FROM AMAZON GIVEN");
+	jx_delete(jx);
+	return jaid;
+}
+
+//Return of 1 means: job exists, succeeded
```
dthain: Gah!  Use an enumeration!

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,639 @@
...
+				info->exit_code=1;//NEED TO FIX EVENTUALLY, and find how to ACTUALLY get the exit code
+				memcpy(info_out, info, sizeof(*info));
+				free(info);
+				return id;
+			}
+		}else{
+			continue;
+		}
+	}
+	/*while(!done){
```
dthain: Dead code

dthain:
Lots to fix before this code is ready to use..

dthain:
@Nekel-Seyew has this code progressed since the last review on Oct 11?

Nekel-Seyew:
The code has made progress. It's cleaned up a bunch and the issues raised here were addressed. Right now I'm testing to ensure that it still works and fixing the job_remove function when ctrl-c a makeflow run.

dthain:
Ok, please push your code so we can keep up...

dthain:
How is `sh_popen` different from `fast_popen`?

Nekel-Seyew:
`sh_open` uses execvp with enables programs in the path to be called, while `fast_popen` uses execv which requires that everything have an exact path. Additionally, `sh_open` is more of a re-implementation of popen since it will call sh and pipe the output, but `fast_popen` skips sh. SH is needed due to how we pass credentials to the CLI.

dthain:
That's fine, please module `sh_popen`and firends into its own module with some documentation so we know why it's there.  Also, please use `process_wait/waitpid` should be used instead of `wait/waitpid` throughout makeflow.  (There are other child processes and other code calling wait, and processes can get lost.)

dthain:
@Nekel-Seyew is this the "going back a few commits" that we discussed earlier this week?

Nekel-Seyew:
Yes, bringing the code into a fully functional, stable state without dealing with the terminal-takeover of CLI programs.

dthain:
First attempt to use it fails:

```
 dthain@cclws15~/work/cctools/batch_job/src$ ./makeflow_amazon_batch_setup 
created vpc: vpc-ae0ef0d5
created subnet: subnet-64357a5b
created security group: sg-81a5d7f6
created gateway: igw-0388697b
THis is the route table: rtb-55709529
{
    "Return": true
}
creating the environment

invalid literal for int() with base 10: ''
  File "<string>", line 3, in <module>
IndexError: list index out of range
  File "<string>", line 3, in <module>
IndexError: list index out of range
. . .
```

After reading the script code, I guess at this usage:

```
./makeflow_amazon_batch_setup 2 2 2
created vpc: vpc-a307f9d8
created subnet: subnet-8f4f00b0
created security group: sg-b094e6c7
created gateway: igw-348e6f4c
THis is the route table: rtb-c343a6bf
{
    "Return": true
}
creating the environment
. . .
```

But then it hangs forever.

If we except anyone else to be able to use this, it needs to detect errors, have a help text, and be described in the manual.  Please see `makeflow_ec2_setup` for a good example to emulate.

btovar:
Kyle, please fix the conflicts  in batch_job.c and makeflow.c.

batch_job/src/amazon_batch_cleanup_script.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/sh
+#
+# File: amazon_batch_cleanup_script.sh
```
btovar: Use the copyright header used in the other files (copyright to ND, GPL, etc.)

batch_job/src/amazon_batch_cleanup_script.sh
btovar: Since it depends in such a way in python, it is a shell script only in name. Why don't we make it a proper python script?

batch_job/src/amazon_batch_script.sh
```diff
@@ -0,0 +1,196 @@
...
+
+if [ -n "$4" ]; then
+	outputfile="$4"
+else
+	outputfile="makeflow_amazon_batch.config"
+fi
+
+if [ -e $outputfile ]; then
+	echo "File: $outputfile already exists! Please cleanup/delete first"
+	exit -12
```
btovar: Why negative numbers?
Negative usually means exit with a signal.

batch_job/src/amazon_batch_script.sh
```diff
@@ -0,0 +1,196 @@
...
+then
+	echo "config and credentials configured correctly, proceeding to setup"
+else
+	echo "configuration and credentials files not properly set, please reconfig aws and try again"
+	exit -33
+fi
+
+check_arn_iam=$(aws --output json iam get-role --role-name AWSBatchServiceRole)
+#good_iam=$(python -c "print 'True' if 'CreateDate' in '$check_arn_iam' else print 'False'")
+if [[ $check_arn_iam = *"CreateDate"*  ]]; then
```
btovar: Double [ is bash. Use a single [ for sh. 

That matching operator may not be available in sh, though, so you may need to find another solution.

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,129 @@
...
+		return 0;
+	
+	argv = malloc(sizeof(char*)*4);
+	argv[0]="sh";
+	argv[1]="-c";
+	argv[2] = (char*)command;
+	argv[3] = NULL;
+	argc = 4;
+
+	if(argc < 1)
```
btovar: Uses braces, even for a single statement.

batch_job/src/Makefile
```diff
@@ -43,6 +44,12 @@ work_queue_factory: work_queue_factory.o libbatch_job.a $(EXTERNAL_LIBRARIES)
 work_queue_pool: work_queue_factory
 	cp $< $@
 
+makeflow_amazon_batch_cleanup:
+	cp amazon_batch_cleanup_script.sh makeflow_amazon_batch_cleanup
```
btovar: Let's rename amazon_batch_cleanup_script.sh and amazon_batch_script.sh  to the names without .sh. That way there won't be confusion in the future that they are the same files.

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+
+static struct itable *process_table = 0;
+
+FILE *sh_popen(const char *command)
+{
+	pid_t pid;
+	int argc;
+	char **argv;
+	int fds[2];
+	char cmd[4096];
```
btovar: char *cmd;

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+	int result;
+
+	strcpy(cmd, command);
```
btovar: cmd = xxstrdup(command);

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+
+	if(string_split_quotes(cmd, &argc, &argv) < 1){
+		return 0;
+	}
+	
+	argv = malloc(sizeof(char*)*4);
+	argv[0]="sh";
```
btovar: /bin/sh

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+	argv[1]="-c";
+	argv[2] = (char*)command;
+	argv[3] = NULL;
+	argc = 4;
+
+	if(argc < 1){
```
btovar: Logic is weird. You just set argc to 4, why check if it is less than 1?

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+		return 0;
+	}
+
+	result = pipe(fds);
+	if(result < 0) {
```
btovar: If error, print a message. (e.g., with strerr(errno)).

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+		int i;
+
+		close(0);
+		i = dup2(fds[1], STDOUT_FILENO); if(i < 0) perror("Error Dup2 stdout from SH_popen.");
+		i = dup2(fds[1], STDERR_FILENO); if(i < 0) perror("Error Dup2 stderr SH_popen.");
+		close(fds[1]);
+		close(fds[0]);
+
+		i = execvp(argv[0], argv);
+		if (i<0) perror("Error in execvp from SH_popen");
```
btovar: Use debug(D_WARN and strerr(errno)

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+		return 0;
+	}
+}
+
+int sh_pclose(FILE * file)
+{
+	pid_t pid;
+	struct process_info* result;
+
+	pid = (PTRINT_T) itable_remove(process_table, fileno(file));
```
btovar: Do you mean (pid_t) ?

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+
+int sh_system(const char* command) {
+	int pid;
+	int res = 0;
+	pid = fork();
+	if (pid == 0) {//child
+		char* argv[4];
+		argv[0] = "sh";
+		argv[1] = "-c";
+		argv[2] = (char*)command;
```
btovar: Again casting from const.. I am surprised the compiler allows this call. You probably can make const char*argv[4].

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+		argv[3] = NULL;
+
+		res = execvp(argv[0], argv);
+		if (res < 0) {
+			perror("SH_system past execvp: ");
```
btovar: debug(D_WARN, etc.

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,131 @@
...
+		}
+	} else if (pid > 0) {//parent
+		struct process_info* pres;
+		while((pres = process_waitpid(pid, 0)) == 0);
+		if (WIFEXITED(pres->status)) {
+			return WEXITSTATUS(pres->status);
+		}
+	} else {
+		fprintf(stderr, "error in forking\n");
```
btovar: DWARN etc.

dttools/src/sh_popen.c
Nekel-Seyew: I coppied the functionality from fast_popen, so I'm not sure if it's an error or not, but I assume it's fine, since I copied it from that file.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+	describe_aws_job_success = 1, //job exists, succeeded
+	describe_aws_job_failed = 0, //job exists, failed
+	describe_aws_job_non_final = -1, //exists, but in non-final state
+	descibe_aws_job_non_exist = -2 //job doesn't exist, should treat as a failure
+};
+static int describe_aws_job(char* aws_jobid, char* env_var){
+	//fprintf(stderr,"aws_jobid:%s\n\n",aws_jobid);
+	char* cmd = string_format("%s aws batch describe-jobs --jobs %s",env_var,aws_jobid);
+	free(cmd);
+	cmd = string_format("%s aws batch describe-jobs --jobs %s",env_var,aws_jobid);
```
nhazekam: This line and the previous free seem redundant .

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+	if(!jaid)
+		fatal("NO JOB ID FROM AMAZON GIVEN");
+	jx_delete(jx);
+	return jaid;
+}
+enum{
+	describe_aws_job_success = 1, //job exists, succeeded
+	describe_aws_job_failed = 0, //job exists, failed
+	describe_aws_job_non_final = -1, //exists, but in non-final state
+	descibe_aws_job_non_exist = -2 //job doesn't exist, should treat as a failure
```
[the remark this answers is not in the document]
nhazekam: Also make enums all-caps.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+		}
+	}
+	jx_delete(jx);
+	return succeed;
+}
+
+static batch_job_id_t batch_job_amazon_batch_submit(struct batch_queue* q, const char* cmd, const char* extra_input_files, const char* extra_output_files, struct jx* envlist, const struct rmsummary* resources){
+	struct internal_amazon_batch_amazon_ids amazon_ids = initialize(q);
+	char* env_var = amazon_ids.master_env_prefix;
+	fprintf(stderr,"env_var: %s\n",env_var);
```
nhazekam: Is this a debug string or an actual error?

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+		cpus = cpus > 1? cpus:1;
+		mem = mem > 1000? mem:1000;
+		disk = disk > 1000 ? disk : 1000;
+	}
+	clock_t start_t = clock();
+	//upload files to S3
+	upload_input_files_to_s3((char*)extra_input_files,job_name);	
+	upload_cmd_file(bucket_name,(char*)extra_input_files,(char*)extra_output_files,(char*)cmd,jobid);
+	clock_t end_t = clock();
+	//fprintf(stderr,"\n\tTime to Upload Files: %lf\n",(double)(end_t-start_t)/CLOCKS_PER_SEC);
```
nhazekam: Remove dead code.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+	
+	//create the fmd string to give to the command
```
nhazekam: fmd seems like a typo.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+	free(job_name);
+	free(fmt_cmd);
+	
+	return jobid;
+	
+}
+
+static batch_job_id_t batch_job_amazon_batch_wait(struct batch_queue *q, struct batch_job_info *info_out, time_t stoptime){
+	struct internal_amazon_batch_amazon_ids amazon_ids = initialize(q);
+	//int i=0;
```
nhazekam: Dead code?

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+	int done  = 0;
+	char* env_var = amazon_ids.master_env_prefix;
+	itable_firstkey(amazon_job_ids);
+	char* jaid;
+	UINT64_T jobid;
+ 	while(itable_nextkey(amazon_job_ids,&jobid,(void**)&jaid)){
+		done = describe_aws_job(jaid,env_var);
+		char* jobname = string_format("%s_%u",queue_name,(unsigned int)jobid);
+		unsigned int id = (unsigned int)jobid;
+		if(done == 1){
```
nhazekam: Use enums from above.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+				struct batch_job_info* info = itable_remove(q->job_table, id);//got from batch_job_amazon.c
+				info->finished = time(0);//get now
+				info->exited_normally=1;
+				info->exit_code=0;//NEED TO FIX EVENTUALLY, and find how to ACTUALLY get the exit code
+				debug(D_BATCH,"copying over the data to info_out");
+				memcpy(info_out, info, sizeof(struct batch_job_info));
+				free(info);
+				return id;
+			}
+		}else if(done == 0 || done == -2){
```
nhazekam: Use enums from above.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+				//id is done, returning here
+				itable_insert(done_jobs,id+1,jobname);
+				itable_remove(amazon_job_ids,jobid);
+				
+				debug(D_BATCH,"Failed job: %i",id);
+				
+				struct batch_job_info* info = itable_remove(q->job_table, id);//got from batch_job_amazon.c
+				info->finished = time(0);//get now
+				info->exited_normally=0;
+				info->exit_code=1;//NEED TO FIX EVENTUALLY, and find how to ACTUALLY get the exit code
```
nhazekam: Is this an error from Amazon or a failure on our side?

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+				memcpy(info_out, info, sizeof(*info));
+				free(info);
+				return id;
+			}
+		}else{
+			continue;
+		}
+	}
+	//if we get to this point, we should probably clean up and exit....
```
nhazekam: Does returning -1 cleanup? Is this a mem-leak? Or is this unreachable code?

makeflow/src/makeflow.c
```diff
@@ -1157,6 +1157,8 @@ static void show_help_run(const char *cmd)
 	printf("\nBatch System Options:\n");
 	printf("    --amazon-config             Amazon EC2 config file from makeflow_ec2_setup.\n");
 	printf("    --lambda-config             Amazon Lambda config file from makeflow_lambda_setup.\n");
+    printf("    --amazon-batch-img          Specify amazon ECS Image (for use with -T amazon-batch)\n");
+    printf("    --amazon-batch-config       Amazon config file from makeflow_amazon_batch_setup.\n");
```
nhazekam: Properly indent these lines.

makeflow/src/makeflow.c
```diff
@@ -1420,6 +1428,9 @@ int main(int argc, char *argv[])
 		{"docker-opt", required_argument, 0, LONG_OPT_DOCKER_OPT},
 		{"amazon-config", required_argument, 0, LONG_OPT_AMAZON_CONFIG},
 		{"lambda-config", required_argument, 0, LONG_OPT_LAMBDA_CONFIG},
+		{"amazon-ami", required_argument, 0, LONG_OPT_AMAZON_AMI},
```
nhazekam: AMAZON_AMI is added here and in long_opts but I don't see it handled in the cases you added. Is this a preexisting case or should it be removed.

makeflow/src/makeflow_amazon_batch_cleanup
```diff
@@ -0,0 +1,103 @@
...
+
+#disabling jobqueues
+echo "Deleting queue"
+aws batch update-job-queue --job-queue $queue_name --state DISABLED
+
+queue_status="$(aws --output text --query 'jobQueues[0].status' batch describe-job-queues --job-queues $queue_name)"
+queue_state="$(aws --output text --query 'jobQueues[0].state' batch describe-job-queues --job-queues $queue_name)"
+
+while [ "$queue_status" != "VALID" ] || [ "$queue_state" = "ENABLED" ]
+#while ! [ "$queue_status" == "VALID" -a "$queue_state" != "ENABLED"  ]
```
nhazekam: Remove dead code.

makeflow/src/makeflow_amazon_batch_cleanup
```diff
@@ -0,0 +1,103 @@
...
+
+
+#disabling the environment
+echo "Deleting environments"
+aws batch update-compute-environment --compute-environment $env_name --state DISABLED
+
+env_status="$(aws --output text --query 'computeEnvironments[0].status' batch describe-compute-environments --compute-environments $env_name)"
+env_state="$(aws --output text --query 'computeEnvironments[0].state' batch describe-compute-environments --compute-environments $env_name)"
+
+#while ! [ "$env_status" == "VALID" -a "$env_state" != "ENABLED"  ]
```
nhazekam: Remove dead code.

makeflow/src/makeflow_amazon_batch_setup
```diff
@@ -0,0 +1,186 @@
...
+fi
+
+if [ -z "$2" ]; then
+	echo "Empty number of CPUs requested."
+	echo "Correct Ussage: ./makeflow_amazon_batch_setup [desired_num_cpus] [min_number_cpus] [max_number_cpus] (config-output)"
+	exit 33
+fi
+
+if [ -z "$3" ]; then
+	echo "Empty number of CPUs requested."
```
nhazekam: Since you are checking for specific empty fields make the first line of each error more specific or just use a single catch all case for code clarity.

makeflow/src/makeflow_amazon_batch_setup
```diff
@@ -0,0 +1,186 @@
...
+if [ $cpus -gt $max_cpus ]; then
+	echo "Desired number of cpus must be less than or equal to maximum number of cpus"
+        echo "Correct Ussage: ./makeflow_amazon_batch_setup [desired_num_cpus] [min_number_cpus] [max_number_cpus] (config-output)"
+        exit 42
+fi
+
+if [ $max_cpus -lt $min_cpus ]; then
+	echo "Minimum number of cpus must be less than or equal to maximum number of cpus"
+        echo "Correct Ussage: ./makeflow_amazon_batch_setup [desired_num_cpus] [min_number_cpus] [max_number_cpus] (config-output)"
+        exit 143
```
nhazekam: Spacing seems off by a space in a couple of these previous lines.

makeflow/src/makeflow_amazon_batch_setup
```diff
@@ -0,0 +1,186 @@
...
+fi
+
+outputfile=${4:-makeflow_amazon_batch.config}
+
+#if [ -n "$4" ]; then
+#	outputfile="$4"
+#else
+#	outputfile="makeflow_amazon_batch.config"
+#fi
```
nhazekam: Remove dead code.

makeflow/src/makeflow_amazon_batch_setup
```diff
@@ -0,0 +1,186 @@
...
+echo "\"aws_id\":\"$aws_id\"," >> $outputfile
+echo "\"aws_key\":\"$aws_key\"," >> $outputfile
+echo "\"aws_reg\":\"$aws_reg\"," >> $outputfile
+
+#echo create the vpc
+ec2_vpc="$(aws --output text --query 'Vpc.VpcId' ec2 create-vpc --cidr-block 10.0.0.0/16)"
+echo "created vpc: $ec2_vpc"
+
+aws --output json ec2 modify-vpc-attribute --vpc-id $ec2_vpc --enable-dns-hostnames
+#aws --output json ec2 modify-vpc-attribute --vpc-id $ec2_vpc --enable-dns-support
```
nhazekam: Dead code.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+	char* cmd_tmp = string_format("%s\n%s\n%s\n",cpy_in,chmod,cmd);
+	free(cpy_in);
+	//copy out any external files
+	char* cpy_out = generate_s3_cp_cmds(output_files,"./",bucket);
+	cmd_tmp = string_format("%s\n%s\n",cmd_tmp,cpy_out);
+
+	//add headder
+	char* final_cmd = string_format("#!/bin/sh\n%s",cmd_tmp);
+	free(cmd_tmp);
+	//char* final_cmd = cmd_tmp;
```
nhazekam: Remove dead code.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+	return initialized_data;
+}
+
+static char* generate_s3_cp_cmds(char* files, char* src, char* dst){
+	char* env_var = initialized_data.master_env_prefix;
+	int i;
+	int files_split_num = 0;
+	char** files_split;
+	split_comma_list(files, &files_split_num, &files_split);
+	//char* new_cmd=string_format("sleep 0");
```
nhazekam: Remove dead code.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,564 @@
...
+	char* aws_secret_access_key = (char*)jx_lookup_string(config, "aws_key");
+	char* aws_region            = (char*)jx_lookup_string(config,"aws_reg");
+	bucket_name                 = (char*)jx_lookup_string(config,"bucket");
+	vpc                         = (char*)jx_lookup_string(config,"vpc");
+	sec_group                   = (char*)jx_lookup_string(config,"sec_group");
+	queue_name                  = (char*)jx_lookup_string(config,"queue_name");
+	compute_env_name            = (char*)jx_lookup_string(config,"env_name");
+	subnet                      = (char*)jx_lookup_string(config,"subnet");	
+
+	//const char* aws_email = jx_lookup_string(config,"aws_email");
```
nhazekam: Remove dead code.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,567 @@
...
+		*size += 1;
+		ta = strtok(0,",");
+	}
+	
+	*output = malloc(sizeof(char*) * (*size));
+	int i = 0;
+	tmp = strdup(in);
+	ta = strtok(tmp,",");
+	while(ta != NULL){
+		if(strlen(ta) < 1) continue;
```
btovar: Same as above, possible infinite loop.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,567 @@
...
+}
+
+static char* generate_s3_cp_cmds(char* files, char* src, char* dst){
+	char* env_var = initialized_data.master_env_prefix;
+	int i;
+	int files_split_num = 0;
+	char** files_split;
+	split_comma_list(files, &files_split_num, &files_split);
+
+	char* new_cmd=malloc(sizeof(char)*1);
```
btovar: Why do you need to malloc here?

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,567 @@
...
+		for(i=0; i<files_split_num; i++){
+			char* tmp = string_format("%s %s/%s %s",copy_cmd_prefix, src, files_split[i], dst);
+			char* tmp2 = string_format("%s\n%s\n",new_cmd,tmp);
+			free(new_cmd);
+			free(tmp);
+			new_cmd = tmp2;
+		}
+	}
+	clean_str_array(files_split_num,&files_split);
+	return new_cmd;
```
btovar: If no files need to be copied, perhaps it is better to return NULL.

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,567 @@
...
+	fwrite(final_cmd,sizeof(char),strlen(final_cmd),tmpfile);
+	fclose(tmpfile);
+	free(final_cmd);
+	
+	//make executable and put into s3
+	sh_system("chmod +x TEMPFILE.sh");
+	cmd_tmp = string_format("%s /usr/bin/time -f \"Submitting cmd_file Job %s_%u: %%e\" -a -o 'BatchJobAmazonBatchTimings.txt' aws s3 cp ./TEMPFILE.sh s3://%s/COMAND_FILE_%u.sh",env_var,queue_name,jobid,bucket_name,jobid);
+	sh_system(cmd_tmp);
+	free(cmd_tmp);
+	remove("TEMPFILE.sh");
```
btovar: Use tmpfile instead. (man 3 tmpfile)

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,567 @@
...
+	free(cmd_tmp);
+	remove("TEMPFILE.sh");	
+
+}
+
+static char* aws_submit_job(char* job_name, char* properties_string){
+	char* queue = queue_name;
+	char* env_var = initialized_data.master_env_prefix;
+	//submit the job-def
+	char* tmp = string_format("%s /usr/bin/time -f \"Registering Job %s: %%e\" -a -o 'BatchJobAmazonBatchTimings.txt' aws batch register-job-definition --job-definition-name %s_def --type container --container-properties \"%s\"",env_var,job_name,job_name, properties_string);
```
btovar: why /usr/bin/time?

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,567 @@
...
+	if(strstr((char*)jx_lookup_string(first_item,"status"),"SUCCEEDED")){
+		succeed = DESCRIBE_AWS_JOB_SUCCESS;
+	}
+	if(strstr((char*)jx_lookup_string(first_item,"status"),"FAILED")){
+		succeed = DESCRIBE_AWS_JOB_FAILED;
+	}
+
+	//start and stop
+	if(succeed == DESCRIBE_AWS_JOB_SUCCESS || succeed == DESCRIBE_AWS_JOB_FAILED){	
+		int64_t created_string = (int64_t) jx_lookup_integer(first_item,"createdAt");
```
btovar: I don't think you need these casts.

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,129 @@
...
+	pid_t pid;
+	int argc;
+	char** argv;
+	int fds[2];
+	char* cmd;
+	int result;
+
+	cmd = xxstrdup(command);
+
+	if(string_split_quotes(cmd, &argc, &argv) < 1){
```
btovar: Why are you calling these function when you overwrite argc and argv below?

dttools/src/sh_popen.c
```diff
@@ -0,0 +1,129 @@
...
+	} else if (pid > 0) {//parent
+		struct process_info* pres;
+		while((pres = process_waitpid(pid, 0)) == 0);
+		if (WIFEXITED(pres->status)) {
+			return WEXITSTATUS(pres->status);
+		}
+	} else {
+		debug(D_ERROR, "error in forking\n");
+	}
+	return -17;
```
btovar: Why a negative number here?

makeflow/src/makeflow_amazon_batch_setup
```diff
@@ -0,0 +1,179 @@
...
+
+load_fieldJSON()
+{
+	grep $1 $2 | cut -d ":" -f 2 | tr -d ' ' | tr -d ','
+}
+
+if [ -z "$1" ]; then
+	echo "Empty number of CPUs requested."
+	echo "Correct Ussage: ./makeflow_amazon_batch_setup [desired_num_cpus] [min_number_cpus] [max_number_cpus] (config-output)"
+	exit 32
```
btovar: Usually we use the same exit code for the same kind of thing. So, a missing argument would be something like exit 1 You don't need exit 32, 33, 34 as all of the errors are about missing arguments.

makeflow/src/makeflow_amazon_batch_setup
```diff
@@ -0,0 +1,179 @@
...
+	echo "Correct Ussage: ./makeflow_amazon_batch_setup [desired_num_cpus] [min_number_cpus] [max_number_cpus] (config-output)"
+	exit 34
+fi
+
+cpus=$1
+min_cpus=$2
+max_cpus=$3
+time=$(date +%s)
+
+if [ $cpus -lt $min_cpus ]; then
```
btovar: Same as before, no need to specialize the error code to the particular error.

dttools/src/sh_popen.c
Nekel-Seyew: it was an easier way to check and ensure that something resembling a command line command is being passed.

makeflow/src/makeflow_amazon_batch_setup
Nekel-Seyew: addressed with previous comment

batch_job/src/batch_job_amazon_batch.c
Nekel-Seyew: the return is jx_int_t and I felt it safe to be explicit about the casting, so the compiler wouldn't complain.

batch_job/src/batch_job_amazon_batch.c
Nekel-Seyew: By having a malloc here, it makes the logic inside the for loop cleaner without needing a special case for the first time it tries to append to the string.

batch_job/src/batch_job_amazon_batch.c
Nekel-Seyew: That doesn't work for my purposes. tmpfile gives you a temporary file, but I need an actual file location so I can use the awscli to upload the file itself. there's no standard way in C to get the filename of a FILE* .

batch_job/src/batch_job_amazon_batch.c
```diff
@@ -0,0 +1,567 @@
...
+	}
+	pclose(out);
+	return jx;
+}
+
+static void split_comma_list(char* in, int* size, char*** output){
+	*size = 0;
+	char* tmp = strdup(in);
+	char* ta = strtok(tmp,",");
+	while(ta != NULL){
```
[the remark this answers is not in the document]
Nekel-Seyew: Makeflow won't run if the input files is listed simply as ","

batch_job/src/batch_job_amazon_batch.c
Nekel-Seyew: Also, testing with "," as the output file didn't cause an infinite loop

batch_job/src/batch_job_amazon_batch.c
Nekel-Seyew: needed to ensure specific form of time so it will accept that format string.

dttools/src/sh_popen.c
btovar: If so, you need to free argv before the malloc below.

batch_job/src/batch_job_amazon_batch.c
Nekel-Seyew: that's a fair enough assessment, but I think it logically flows better to always return an allocated string, even if it's an empty one. That way the code which calls it knows that it's always safe to use and delete the string, even if that string is just empty. It reduces the number of if statements in the rest of the code.

PR #1681, opened 2017-06-15
DON'T do remote renaming on shared fs

### LINKED ISSUE #1680: shared-fs + remote file renaming fails
Hi experts! Sorry if I'm missing something obvious here, but is it possible to mix remote file renaming and the (extremely useful!) `--shared-fs` option?  I'd like to be able to use renaming and avoid having to copy files around when I can. Of course I could add a clause to the command, but I'd prefer to have Makeflow handle it if possible. You can find a minimal example in case it's helpful at `/afs/crc.nd.edu/user/a/awoodard/shared-fs-test`. The first two rules are successful and the third fails. 

```
[earth] ~/shared-fs-test >cat Makeflow
1.txt->remote.txt:
	echo foobar > remote.txt

/afs/crc.nd.edu/user/a/awoodard/shared-fs-test/2.txt:
	echo foobar > /afs/crc.nd.edu/user/a/awoodard/shared-fs-test/2.txt

/afs/crc.nd.edu/user/a/awoodard/shared-fs-test/3.txt->remote.txt:
	echo foobar > remote.txt
```

```
[earth] ~/shared-fs-test >makeflow -T wq -M ttV_FTW --shared-fs '/afs/crc.nd.edu/user/a/awoodard/'
parsing ./Makeflow...
checking ./Makeflow for consistency...
./Makeflow has 3 rules.
recovering from log file ./Makeflow.makeflowlog...
starting workflow....
listening for workers on port 9000.
submitting job: echo foobar > remote.txt
submitted job 1
submitting job: echo foobar > /afs/crc.nd.edu/user/a/awoodard/shared-fs-test/2.txt
submitted job 2
submitting job: echo foobar > remote.txt
submitted job 3
warning: this work queue master is visible to the public.
warning: you should set a password with the --password option.
job 1 completed
echo foobar > remote.txt did not create file /afs/crc.nd.edu/user/a/awoodard/shared-fs-test/3.txt
job 2 completed
job 3 completed
workflow failed.
```

trshaffer:
When using Makeflow with `--shared-fs`, also check if the file is remotely renamed. If so, pass it to the batch system anyway. This could lead to poor performance if used on large files.

resolves #1680

### DIFF
--- a/doc/makeflow.html
+++ b/doc/makeflow.html
@@ -679,7 +679,7 @@ filesystem like NFS or HDFS, and so, to be safe, it copies all necessary depende
 be used to efficiently access files without making remote copies.
 Makeflow must be told where the shared filesystem is located, so that it
 can take advantage of it.  To enable this, invoke Makeflow with the 
-<code>--shared-fs</code> option, indicating the path where the shared
+<tt>--shared-fs</tt> option, indicating the path where the shared
 filesystem is mounted.  (This option can be given multiple times.)
 Makeflow will then verify the existence of input and output files in these
 locations, but will not cause them to be transferred.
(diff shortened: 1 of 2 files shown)


nhazekam:
This pull request forces Makeflow to check if there is a remote rename of a file in the shared-fs. If so, we exit after printing that it is not allowed. This enforces that shared-fs bypasses the batch submission and is not subject to remote renaming. 

If there is need, this can be revisited, but the only clear alternative is to ignore the shared-fs behavior in the presence of remote renaming which may lead to poor performance. Other approaches would require more drastic redesign.

PR #1684, opened 2017-06-22
Save outputs of failed Makeflow rules

### LINKED ISSUE #1451: On resource exhaustion, makeflow should not delete the resource summary file
When a rule fails, it deletes all of its outputs. This includes the resource summary file. This is not advantageous, as the summary provides information about the failure (e.g., which resources where exhausted).

trshaffer:
This PR changes Makeflow's default behavior for failed tasks. Assuming there isn't an input file or an existing file not managed by Makeflow with a conflicting name, any outputs produced by a failed rule are copied to a `makeflow.failed.n` directory, where `n` is the node ID, and print a message for the user. Running `makeflow -c` will remove failed directories. This should make debugging failures much easier, as logs and such can be brought back as long as they're specified as outputs. I tested locally, on Condor, and with Work Queue.

resolves #1451

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -93,6 +93,7 @@ an example.
 */
 
 #define MAX_REMOTE_JOBS_DEFAULT 100
+#define FAIL_DIR "makeflow.failed.%d"
 
 static sig_atomic_t makeflow_abort_flag = 0;
 static int makeflow_failed_flag = 0;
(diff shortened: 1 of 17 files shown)


dthain:
Thanks Tim, Per our discussion, two changes:

1 - Make a single log entry regarding the failure directory.  This allows Makeflow to treat it as an opaque object regardless of the contents.
2 - Move the files into the failure directory via rename, so as to avoid the expense of copying large files.

(Actually, when you put these two together, it allows you to generate a single log message for each case of #2, so you avoid the problem of making multiple entries atomic.)

makeflow/src/makeflow.c
```diff
@@ -713,6 +714,91 @@ int makeflow_node_check_file_was_created(struct dag_node *n, struct dag_file *f)
...
+			debug(D_MAKEFLOW_RUN,
+					"skipping %s since it already exists",
+					path);
+			return NULL;
+		}
+		return dag_file_lookup_or_create(d, path);
+	}
+}
+
+static int makeflow_prep_fail_dir(
```
nhazekam: Also rename to `makeflow_clean_prep_fail_dir`.

nhazekam:
Before merging this into master, I just want to make note that is will become default behavior in Makeflow. I think this is a good idea, but wanted to clarify that first.

dthain:
I'm ok with that as long as (A) the effect is clearly explained on stdout
and (B) the directory is removed during a clean.

PR #1697, opened 2017-07-13
Makeflow Local Resources

dthain:
@btovar please take a look at this PR.  I believe the overall structure is right, except that it seems like resources are not propagated to dag nodes.  Am I misunderstanding something about how resources are handled?

For example, this workflow should (correctly) hang if run with `makeflow --local-cores 4`

```
CORES=8
output:
    echo hello > output
```

If you set gdb to break on `makeflow_local_resources_subtract` and then examine `n->resources_requested` then all the fields are undefined.  Any ideas?

### DIFF
--- a/doc/man/makeflow.m4
+++ b/doc/man/makeflow.m4
@@ -50,6 +50,10 @@ OPTIONS_BEGIN
 OPTION_TRIPLET(-B, batch-options, options)Add these options to all batch submit files.
 OPTION_TRIPLET(-j, max-local, #)Max number of local jobs to run at once. (default is # of cores)
 OPTION_TRIPLET(-J, max-remote, #)Max number of remote jobs to run at once. (default is 1000 for -Twq, 100 otherwise)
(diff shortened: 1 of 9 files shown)


btovar:
On it.

btovar:
Resources are read to the correct category, and they correctly appear as an argument to makeflow_node_submit_retry, and batch_job_submit.  They may be resolved to late to make a decision locally. Exploring that possibility...

btovar:
The issue is that   n->resources_requested  only captures resources asked for that particular node, as:

output:
CORES=4
       echo hello > output

but we need to combine with resources explicitly set for the category, e.g.:

MEMORY=1024
output:
CORES=4
       echo hello > output

and also, for resources per category that are dynamically computed (slices of the cake).

So, instead of using   n->resource_requested,  we need to use  dag_node_dynamic_label(n).

Making the changes now.

btovar:
Almost there. Local jobs are being added to the remote queue for some reason. I think the is_local_job is clashing with n->local_job in some places.

dthain:
Supseded by #1698

PR #1699, opened 2017-07-14
compile statically parrot_run and cvmfs (parrot_cvmfs_static_run)

btovar:
Not made by default. Use: parrot_cvmfs_static_run

Still needs to be run on the same kernel version and glibc it was
compiled with.


batrick:
I'm not privy to the details of this but wouldn't making it static mean glibc on the host machine is unused?

As for the kernel, a given Parrot binary should have no trouble running on different kernel versions. There are run-time checks everywhere for the different kernel versions.

btovar:
One quirk of glibc is that you cannot have true static executables. Some network related functions are always used dynamically from the host glibc.  It may actually be that parrot is ok if does not call any of those functions, but I did not check too deeply. (Note that this is particular to glibc, as for example, libmusl does what would you expect with static executables.)

As for the kernel, when I compiled with 3.10, and tried to run it on a 2.6.32, it failed telling me that the kernel was too old. Perhaps it was an overzealous check?

batrick:
huh, TIL.

This code? 

https://github.com/cooperative-computing-lab/cctools/blob/9f50cdfff573cb1c8a51c176b2067d7b3619dfbe/parrot/src/pfs_main.cc#L200-L201

2.6.32 should have worked...

btovar:
It is a different executable for cases where it might be needed. I am not sure we want to include it in make install, as it is quite optional.

PR #1711, opened 2017-08-02
Adds Work Queue capacity calculation

nkremerh:
@btovar Code was overhauled after a lengthy rebase, please take a look.

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -246,8 +246,6 @@ struct blacklist_host_info {
 	time_t release_at;
 };
 
-int master_capacity_weighted = 0;
-
 static void handle_worker_failure(struct work_queue *q, struct work_queue_worker *w);
 static void handle_app_failure(struct work_queue *q, struct work_queue_worker *w, struct work_queue_task *t);
 static void remove_worker(struct work_queue *q, struct work_queue_worker *w, worker_disconnect_reason reason);
@@ -1459,7 +1457,6 @@ static void fetch_output_from_worker(struct work_queue *q, struct work_queue_wor
 
 	// Start receiving output...
 	t->time_when_retrieval = timestamp_get();
-	debug(D_BJ, "Task %d time_when_retrieval.\n", t->taskid);
 
 	if(t->result == WORK_QUEUE_RESULT_RESOURCE_EXHAUSTION) {
 		result = get_monitor_output_file(q,w,t);
@@ -1475,7 +1472,6 @@ static void fetch_output_from_worker(struct work_queue *q, struct work_queue_wor
 	if(result == WORKER_FAILURE) {
 		// Finish receiving output:
 		t->time_when_done = timestamp_get();
-		debug(D_BJ, "Task %d time_when_done.\n", t->taskid);
 
 		return;
 	}
@@ -1494,7 +1490,6 @@ static void fetch_output_from_worker(struct work_queue *q, struct work_queue_wor
 
 	// Finish receiving output.
 	t->time_when_done = timestamp_get();
-	debug(D_BJ, "Task %d time_when_done.\n", t->taskid);
 
 	work_queue_accumulate_task(q, t);
 
@@ -1789,7 +1784,6 @@ static work_queue_result_code_t get_result(struct work_queue *q, struct work_que
 
 	execution_time = atoll(items[3]);
 	t->time_workers_execute_last = observed_execution_time > execution_time ? execution_time : observed_execution_time;
-	debug(D_BJ, "Task %d time_workers_execute_last.\n", t->taskid);
 
 	t->time_workers_execute_all += t->time_workers_execute_last;
 
@@ -2249,8 +2243,7 @@ static struct jx * queue_to_jx( struct work_queue *q, struct link *foreman_uplin
 	jx_insert_integer(j,"capacity_memory",info.capacity_memory);
 	jx_insert_integer(j,"capacity_disk",info.capacity_disk);
 	jx_insert_integer(j,"capacity_instantaneous",info.capacity_instantaneous);
-	jx_insert_integer(j,"capacity_weighted",master_capacity_weighted);
-	debug(D_BJ, "FLAGGED: %d\t%d\n", info.capacity_instantaneous, master_capacity_weighted);
+	jx_insert_integer(j,"capacity_weighted",info.capacity_weighted);
 
 	jx_insert_string(j,"master_preferred_connection",q->master_preferred_connection);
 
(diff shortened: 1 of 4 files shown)


work_queue/src/work_queue.c
```diff
@@ -3094,13 +3100,17 @@ Used for computing queue capacity below.
...
 
 	// Create a new report object and add it to the list.
 	tr = calloc(1, sizeof(struct work_queue_task_report));
 
 	tr->transfer_time = (t->time_when_commit_end - t->time_when_commit_start) + (t->time_when_done - t->time_when_retrieval);
+	debug(D_WQ, "tr->transfer_time %"PRId64" = (%"PRId64" - %"PRId64") + (%"PRId64" - %"PRId64")\n", tr->transfer_time, t->time_when_commit_end, t->time_when_commit_start, t->time_when_done, t->time_when_retrieval);
 	tr->exec_time     = t->time_workers_execute_last;
-
+	debug(D_WQ, "tr->exec_time = %"PRId64"\n", t->time_workers_execute_last);
+	tr->master_time = (s.time_status_msgs - s.prev_time_status_msgs) + (s.time_internal - s.prev_time_internal);
```
btovar: The above is per task. master_time is global.  Further, they do not correspond to the time the master dedicated to this task, but to the time between calls to add_task_report. 

If this is to be taken as some sort of average of master_time across tasks, then there should better/clearer ways to get it.

work_queue/src/work_queue.c
```diff
@@ -4689,6 +4715,8 @@ struct work_queue *work_queue_create(int port)
 	q->time_last_wait = timestamp_get();
 
 	debug(D_WQ, "Work Queue is listening on port %d.", q->port);
+	time_t ts;
+	time(&ts);
```
btovar: These two lines do not seem to have any effect.

nkremerh: You found some holdover debugging code. I will get rid of that.

work_queue/src/work_queue.c
```diff
@@ -5153,6 +5181,13 @@ static work_queue_task_state_t change_task_state( struct work_queue *q, struct w
 			push_task_to_ready_list(q, t);
 			break;
 		case WORK_QUEUE_TASK_DONE:
+			s = malloc(sizeof(struct work_queue_stats));
+			work_queue_get_stats(q, s);
+			//if(t->result == WORK_QUEUE_RESULT_SUCCESS) {
+			//	compute_capacity(q, s);
+			//}
+			log_queue_stats(q);
+			break;
```
btovar: Move the line log_queue_stats(q) to after the switch.  Remove anything that deals with s (including the definition above), as those lines do not have any effect. Remove the commented code.

work_queue/src/work_queue.c
```diff
@@ -5533,6 +5567,10 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
 {
 	int events = 0;
 
+	q->stats->prev_time_application = q->stats->time_application;
```
btovar: Instead of having prev_* fields,  another field to struct work_queue,  stats_prev. Here simply do a memcpy.

work_queue/src/work_queue.h
```diff
@@ -226,6 +226,10 @@ struct work_queue_stats {
 	timestamp_t time_internal;     /**< Total time the queue spents in internal processing. */
 	timestamp_t time_polling;      /**< Total time blocking waiting for worker communications (i.e., master idle waiting for a worker message). */
 	timestamp_t time_application;  /**< Total time spent outside work_queue_wait. */
+	timestamp_t prev_time_status_msgs;
```
btovar: As above. Remove these prevs. We do not want to modify work_queue_stats unless we really need to. Add a prev_stats to struct work_queue.

work_queue/src/work_queue_graph_log
```diff
@@ -66,7 +66,7 @@ def read_log_entries(file, fields):
                         epoch = numbers[i]/1000000
                     record['timestamp'] = numbers[i]/1000000 - epoch
                 elif time_field_p(field):
-                    record[field] = numbers[i]/1000000
```
btovar: why?

work_queue/src/work_queue_graph_log
```diff
@@ -75,29 +75,12 @@ def read_log_entries(file, fields):
...
-            record['bytes_sent'] /= unit_factors['GB']
-            record['bytes_received'] /= unit_factors['GB']
-            record['bytes_transfered'] = record['bytes_sent'] + record['bytes_received']
-
-            try:
-                record['capacity_tasks'] = (record['time_execute'] + record['time_polling']) / record['time_master']
-            except ArithmeticError:
-                record['capacity_tasks'] = 0
-
+            record['time_master']   = record['time_send'] + record['time_receive']
```
btovar: This time_master is different from the one computed in work_queue. Either make them the same, or give more appropriate names.

work_queue/src/work_queue_graph_log
```diff
@@ -75,29 +75,12 @@ def read_log_entries(file, fields):
...
             if(prev_line is None):
-                record['time_master_weighted']  = record['time_master']
-                record['time_execute_weighted'] = record['time_execute_good'] + record['time_polling']
+#                record['time_master_weighted']  = record['time_master']
```
btovar: Why?
Also, remove code if it is not used.

work_queue/src/work_queue_graph_log
```diff
@@ -120,6 +103,29 @@ def read_log_entries(file, fields):
                 record['stack_time_application'] = 0
                 record['stack_time_other']       = 0
 
+            print "Execute: %d  I/O: %d  Capacity: %d\n" % (delta['time_execute_good'], delta['time_master'], record['capacity_weighted'])
+            record['bytes_sent'] /= unit_factors['GB']
+            record['bytes_received'] /= unit_factors['GB']
+            record['bytes_transfered'] = record['bytes_sent'] + record['bytes_received']
+
+#            try:
```
btovar: Do not include dead code.

work_queue/src/work_queue_graph_log
nkremerh: Capacity calculation was moved to runtime at the master, so calculating all these weighted and unweighted times is unnecessary. I will continue cleaning up Work Queue Graph Log, it needs some work. I have not looked at it in awhile.

work_queue/src/work_queue.c
nkremerh: This should be the time the master spends on the specific task for the current task report (remember we use transfer_time + master_time as the denominator for capacity). I will look into a better way of getting this information.

nkremerh:
@btovar I believe I have addressed your concerns. Please let me know what you think of the updated master_time calculation.

work_queue/src/work_queue.c
```diff
@@ -3094,13 +3099,16 @@ Used for computing queue capacity below.
...
+	struct work_queue_stats s;
+	work_queue_get_stats(q, &s);
 
 	// Create a new report object and add it to the list.
 	tr = calloc(1, sizeof(struct work_queue_task_report));
 
 	tr->transfer_time = (t->time_when_commit_end - t->time_when_commit_start) + (t->time_when_done - t->time_when_retrieval);
 	tr->exec_time     = t->time_workers_execute_last;
-
+	tr->master_time   = (((t->time_when_done - t->time_when_commit_start) - tr->transfer_time) - tr->exec_time);
```
btovar: I like this definition much better. Please remove the dangling commented code below.

nkremerh: It is removed.

btovar:
Nate,
When starting the workflow, the capacity reported is 0 ("capacity_weighted":0,"capacity_instantaneous":0)

instead of WORK_QUEUE_DEFAULT_CAPACITY_TASKS.

Thus, the factory does not create any workers (if using -w 0).

Issue #1712, opened 2017-08-02
Update inline JX Expressions


nkremerh:
The JX expression on line 278 of Work Queue Status is out of date since the syntax has changed.

dthain:
Ah, I didn't realize the recent JX changes deprecated `&&` in favor or `and`.
Could this break anything else?

trshaffer:
Since JX is a Python subset now, it uses `and` and `not` and such. I didn't realize there are snippets of JX hard-coded places. I can search around the codebase and see if anything else is out of date

PR #1720, opened 2017-08-11
Replace inline JX strings with type-checked expressions

trshaffer:
Recent changes to JX syntax broke a few places internal to the code that parsed hard-coded JX expressions from C code. I hadn't noticed them until something cropped up in #1712. The immediate fix was just to update the query strings. After that, I searched around the codebase and replaced the hard-coded strings with expressions to build up JX structures in memory. This is safer for wonky inputs, and lets the compiler do more to check the code. As far as I know, these are all the hard-coded JX expressions in CCTools.

### DIFF
--- a/chirp/src/chirp_global.c
+++ b/chirp/src/chirp_global.c
@@ -125,7 +125,11 @@ static int server_table_load(time_t stoptime)
 
 	debug(D_CHIRP, "querying catalog at %s", CATALOG_HOST);
 
-	struct jx *jexpr = jx_parse_string("type==\"chirp\"");
+	struct jx *jexpr = jx_operator(
+		JX_OP_EQ,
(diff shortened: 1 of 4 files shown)


dthain:
@trshaffer this approach is fine in the code.

But I am concerned that others are likely to get tripped up since `&& || !` are so widely used in other languages.  Would it break something for us to support them as synonyms for `and or not` ?

trshaffer:
@dthain making them synonyms would be easy technically. The main reason to
switch rather than keep both is that JX wouldn't be a subset of any one
language. We could add it as a compatibility option though

nhazekam:
In autobuild, the osx build failed from a timeout. I just built and ran the test suite, which had 0 failures.

Issue #1670, opened 2017-06-05
Makeflow GC options 


nhazekam:
Makeflow still has the functionality to utilize garbage collection, but the options are not in the help message or manpage.  We reference them in the User Guide, so there is some level of information.

btovar:
@nhazekam: is this one done?

nhazekam:
Resolved in #1724

PR #1724, opened 2017-08-21
Updated makeflow options to include GC

nhazekam:
Addressed #1670  reintroduces garbage collection options to the help message.

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -1205,6 +1205,9 @@ static void show_help_run(const char *cmd)
 	printf(" -A,--disable-afs-check         Disable the check for AFS. (experts only.)\n");
 	printf("    --cache=<dir>               Use this dir to cache downloaded mounted files.\n");
 	printf(" -X,--change-directory=<dir>    Change to <dir> before executing the workflow.\n");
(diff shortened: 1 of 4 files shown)


makeflow/src/makeflow.c
```diff
@@ -1405,7 +1408,6 @@ int main(int argc, char *argv[])
 		{"gc", required_argument, 0, 'g'},
 		{"gc-size", required_argument, 0, LONG_OPT_GC_SIZE},
 		{"gc-count", required_argument, 0, 'G'},
-		{"wait-for-files-upto", required_argument, 0, LONG_OPT_FILE_CREATION_PATIENCE_WAIT_TIME},
```
dthain: On purpose?

nhazekam: Yes. For some reason probably due to an old code change this entry in is the struct twice. The first is on line 1404/1407.

dthain: Got it

nhazekam:
Need to update test prior to merge.

dthain:
Ready to go?

nhazekam:
Yep, should be.

PR #1731, opened 2017-09-08
monitor does not kill task when it goes over stated limits

btovar:
adds --without-limits
adds q.tune("rmonitor-do-not-enforce-limits-use-carefully")

MEANT FOR DEBUGGING. 

The monitor does not terminate a task when it goes above the specified limits. The exhausted limits are still recorded, and the monitor still reports exit status 2. 

If using with wq, a task is still marked as a failure with WORK_QUEUE_RESULT_RESOURCE_EXHAUSTION. However, as many output files as possible are transferred back from the worker.


dthain:
@btovar is this one that you want to finish and merge, or is it just for testing?

btovar:
Not needed anymore.

Issue #1732, opened 2017-09-13
Large UDP Updates Sometimes Lost


btovar:

dthain:
It's not yet clear from observations whether makeflow is *sending* port zero or something is lost in transmission and the catalog *receives* port zero.  (Does port zero even have a useful meaning?)

btovar:
I think it is a red-herring.

Specifying  a project name with -M, but not using -Twq, causes makeflow to report to the catalog using port 0.

During virtualclusters testing, I indeed forgot -Twq more than twice.

btovar:
Nevermind. port 0 also appears with -Twq.

dthain:
Keep in mind that with `-Twq`, makeflow will send two updates: a makeflow-level report, which will have port 0, and also a workqueue level report, which will have port not zero.  Does that explain the behavior?

dthain:
@nhazekam is looking at this too

btovar:
I think it does, yes.

dthain:
Ok, so let me summarize my understanding so far:

- Receiving a message with port=0 at the makeflow level is not a problem per se.  That's just the high-level status message available via makeflow_status.
- When `-T wq` is given, the catalog should *also* receive a message from the work queue layer, with a valid port number in the JX document.
- The port number in the JX document may not necessarily match the source port number given by `datagram_recv` due to network translation issues.    This is expected.
- So, the problem that Nick is observing is really that when he runs makeflow with `-T wq` from Jetstream, it seems that the makeflow layer packet arrives, but the WQ layer packet does not.

@nhazekam is that correct so far?

@btovar are you observing a bug not explained by the above?

nhazekam:
This is correct.

As mentioned above the port is not determined by where the message is located, but by what the message indicates the port is in the JX. The Makeflow message does not provide a port, so the port is 0 by default.

After running a small test last night (and after a suggestion from @nkremerh and @trshaffer), we found that if we replace the larger work queue message with a small test message, aka `test`, the packet goes through. Is it possible that the UDP packet is just being dropped in transit due to its size? The makeflow message is well under 1024 bytes and the wq message if almost 2048.

dthain:
That's certainly possible.  Try using the `catalog_update` tool to send JX of larger sizes, and see if that gets dropped as well.  If that's the case, then we should look into paring down the WQ message.

(For a variety of reasons, it's also probably time that we allow for TCP updates to the catalog)

[...]

PR #1734, opened 2017-09-14
Catalog Server Accepts Compressed Updates

dthain:
Partially addresses #1732. @nhazekam you will need to test with your half of the solution.

### DIFF
--- a/dttools/src/catalog_server.c
+++ b/dttools/src/catalog_server.c
@@ -27,6 +27,7 @@ See the file COPYING for details.
 #include "daemon.h"
 #include "getopt_aux.h"
 #include "change_process_title.h"
+#include "zlib.h"
 
 #include <stdlib.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


dthain:
Go ahead and test them together, then merge when you are satisfied.

dttools/src/catalog_server.c
```diff
@@ -216,19 +223,37 @@ static void make_hash_key(struct jx *j, char *key)
...
 	char key[LINE_MAX];
 	int port;
-	int result;
+	unsigned long data_length, raw_data_length;
 	struct jx *j;
 
 	while(1) {
-		result = datagram_recv(update_port, data, DATAGRAM_PAYLOAD_MAX, addr, &port, 0);
-		if(result <= 0)
+		raw_data_length = datagram_recv(update_port, raw_data, DATAGRAM_PAYLOAD_MAX, addr, &port, 0);
```
nhazekam: `datagram_recv` returns -1 if there is no data.

Since `raw_data_length` is unsigned it never terminates the loop.

dthain:
Quite right, how about this?

nhazekam:
That change works great.

PR #1735, opened 2017-09-14
Catalog Client Sends Compressed Updates

nhazekam:
Compresses data for catalog updates using zlib.

### DIFF
--- a/dttools/src/catalog_query.c
+++ b/dttools/src/catalog_query.c
@@ -242,7 +242,7 @@ int catalog_query_send_update(const char *hosts, const char *text)
 	char* compress_data= malloc(compress_data_length);
 
 	/* Do not compress if env var is set to off.*/
-	if(!do_compress || strcmp(do_compress, "off")) {
+	if(do_compress && !strcmp(do_compress, "on")) {
 		success = compress((Bytef*)compress_data+1, &compress_data_length, (const Bytef*)text, data_length);
 		/* Prefix the data with 0x1A (Control-Z) to indicate a compressed packet. */
 		compress_data[0] = 0x1A;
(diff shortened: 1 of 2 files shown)


dttools/src/catalog_query.c
```diff
@@ -227,12 +229,28 @@ int catalog_query_send_update(const char *hosts, const char *text)
...
+	data_length = strlen(text)+1;
+
+	/* Estimates the bounds for the compressed data. */
+	compress_data_length = compressBound(data_length);
+	char* compress_data= malloc(compress_data_length);
+
+	/* Prefix the data with 0x1A (Control-Z) to indicate a compressed packet. */
+	compress_data[0] = 0x1A;
+
+	int success = compress((Bytef*)compress_data+1, &compress_data_length, (const Bytef*)text, data_length);
```
dthain: compress_data_length-1?

dttools/src/catalog_query.c
```diff
@@ -227,12 +229,28 @@ int catalog_query_send_update(const char *hosts, const char *text)
...
+
+	if(success!=Z_OK) {
+		debug(D_DEBUG,"warning: Unable to compress data for update.\n");
+	}
```
dthain: If this fails, you will send garbage data.
Fall back by sending the data uncompressed..

dttools/src/catalog_query.c
nhazekam: compress_data_length is an estimated value from compressBound and is then set by the compress function. I don't modify it here as I will modify it when we send it.

dttools/src/catalog_query.c
nhazekam: Sounds good.

dthain:
Now that I think about it, we should have a means of turning compression off in a pinch via an environment variable.  It's quite possible that someone will deploy new WQ code while relying on an old (local) catalog.  (We will update the global catalogs, of course.)

nhazekam:
I'll add a env var to add this.

nhazekam:
@dthain @btovar Please take a look. I can change the environment variable name and expected behavior if needed. Also, where should this be documented?

dthain:
`doc/catalog.html`

dthain:
Merge?

nhazekam:
I am making a slight adjustment to the code since the compressed path of execution is no longer the default so we were doing excess work.

PR #1736, opened 2017-09-15
Use queue_lean_to_jx to update the catalog

btovar:


btovar:
Testing, don't merge.

PR #1742, opened 2017-09-20
adding join function

AndrewLitteken:
Adding the join function to JX which operates like the join function in Python where a list of strings can be combined into a single string with a delimiter if it is specified.

### DIFF
--- a/dttools/src/jx_function.c
+++ b/dttools/src/jx_function.c
@@ -170,3 +170,63 @@ struct jx *jx_function_range(struct jx *args) {
 
 	return result;
 }
+
+
+struct jx *jx_function_join(struct jx *orig_args) {
+	const char *funcname = "join";
+	const char *err = NULL;
+
+	struct jx *args = jx_copy(orig_args);
+	
+	int length = jx_array_length(args);
+	if(length>2){
+		err = "too many arguments to join";
+		goto FAILURE;
+	}
+	else if(length<=0){
+		err = "too few arguments to join";
+		goto FAILURE;
+	}
+	
+	struct jx *list = jx_array_shift(args);
+	if (!jx_istype(list, JX_ARRAY)){
+		err = "A list must be the first argument in join";
+		goto FAILURE;
+	}
+	
+	struct jx *delimeter=NULL;
+	if (length==2){
+		delimeter  = jx_array_shift(args);
+		if(!jx_istype(delimeter, JX_STRING)){
+			err = "A list must be the first argument in join";
+			goto FAILURE;
+		}
+	}
+	
+	char *result=xxstrdup("");	
+	int list_length = jx_array_length(list);
+	int location = 0;
+	struct jx *value=jx_array_shift(list);
+	while(value){
+		result = string_combine(result, value->u.string_value);
+		if(location<list_length-1){	
+			if(delimeter) result = string_combine(result, delimeter->u.string_value);
+			result = string_combine(result, " ");
+		}
+		value = jx_array_shift(list);
+		++location;
+	}
+	
+	jx_delete(args);
+	jx_delete(list);
+	jx_delete(delimeter);
+	struct jx *j = jx_string(result);
+	free(result);
+	return j;
+	
+	FAILURE:
+	    jx_delete(args);
+	    free(result);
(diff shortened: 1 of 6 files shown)


dttools/src/jx_function.c
```diff
@@ -170,3 +170,63 @@ struct jx *jx_function_range(struct jx *args) {
...
+	if (!jx_istype(list, JX_ARRAY)){
+		err = "A list must be the first argument in join";
+		goto FAILURE;
+	}
+	
+	struct jx *delimeter=NULL;
+	if (length==2){
+		delimeter  = jx_array_shift(args);
+		if(!jx_istype(delimeter, JX_STRING)){
+			err = "A list must be the first argument in join";
```
trshaffer: Looks like the wrong error message here?

dttools/src/jx_function.c
```diff
@@ -170,3 +170,63 @@ struct jx *jx_function_range(struct jx *args) {
...
+			goto FAILURE;
+		}
+	}
+	
+	char *result=xxstrdup("");
```
trshaffer: You should define this variable at the top of the function (before the `goto`s) and initialize it to `NULL`. That should fix the autobuild issue

dttools/src/jx_function.c
```diff
@@ -170,3 +170,63 @@ struct jx *jx_function_range(struct jx *args) {
...
+			goto FAILURE;
+		}
+	}
+	
+	char *result=xxstrdup("");	
+	int list_length = jx_array_length(list);
+	int location = 0;
+	struct jx *value=jx_array_shift(list);
+	while(value){
+		result = string_combine(result, value->u.string_value);
```
trshaffer: You need to check that each item is really a string before calling `string_combine()`. If someone passes a number, for example, it'll segfault trying to dereference it as a pointer.

dttools/src/jx_function.c
```diff
@@ -170,3 +170,63 @@ struct jx *jx_function_range(struct jx *args) {
...
+		if(location<list_length-1){	
+			if(delimeter) result = string_combine(result, delimeter->u.string_value);
+			result = string_combine(result, " ");
```
trshaffer: You should put this behind `else` so it's possible to do things like

    join(["a", "b", "c"], "") => "abc"

dttools/src/jx_function.c
```diff
@@ -170,3 +170,63 @@ struct jx *jx_function_range(struct jx *args) {
...
+	jx_delete(list);
+	jx_delete(delimeter);
+	struct jx *j = jx_string(result);
+	free(result);
+	return j;
+	
+	FAILURE:
+	    jx_delete(args);
+	    free(result);
+	    FAIL(funcname, JX_BUILTIN_FORMAT, orig_args, err);
```
trshaffer: You should also call `jx_delete()` on `list` and `delimiter` here, as they could have been set before failure. You can move both declarations to the top and set them to `NULL` to start with. It's safe to pass null pointers to `jx_delete()`.

trshaffer:
You should also add `join()` unit tests. Try to test all the syntax variations and error cases in `dttools/test/jx.input`. You'll need to run

    ../src/jx_test < jx.input > jx.output

to update the expected result. After adding some unit testing and doing those small fixups, this should be ready to merge!

dttools/src/jx_function.c
```diff
@@ -170,3 +170,71 @@ struct jx *jx_function_range(struct jx *args) {
...
+			goto FAILURE;
+		}
+	}
+	
+	result=xxstrdup("");	
+	int list_length = jx_array_length(list);
+	int location = 0;
+	struct jx *value=jx_array_shift(list);
+	while(value){
+		/*if (!jx_istype(value->u.string_value, JX_STRING)){
```
trshaffer: If you get rid of `->u.string_value` you should be able to uncomment this

dttools/test/jx.expected
```diff
@@ -450,6 +450,33 @@ value:      false
 expression: [1,2,[3,4]]==[1,2,[3,4]]
 value:      true
 
+expression: join([])
+value:		true
```
trshaffer: Are these supposed to be strings?

Issue #1745, opened 2017-09-22
Error importing Work Queue in Python


pivie:
I get the following error when trying to import work_queue in python with the latest master (9d055f67900d368b1bde16cd939ef7dc280fc3da).

Traceback (most recent call last):
  File "/afs/crc.nd.edu/user/p/pivie/work_queue_example.py", line 12, in <module>
    from work_queue import *
  File "/afs/crc.nd.edu/user/p/pivie/cctools/lib/python2.7/site-packages/work_queue.py", line 28, in <module>
    _work_queue = swig_import_helper()
  File "/afs/crc.nd.edu/user/p/pivie/cctools/lib/python2.7/site-packages/work_queue.py", line 24, in swig_import_helper
    _mod = imp.load_module('_work_queue', fp, pathname, description)
ImportError: /afs/crc.nd.edu/user/p/pivie/cctools/lib/python2.7/site-packages/_work_queue.so: undefined symbol: compress

PR #1746, opened 2017-09-22
add -lz to swig bingings

btovar:
Fix for #1745 

@pivie, @nhazekam  can you corroborate it works for you?


pivie:
Yes, that fixed it!

Issue #1714, opened 2017-08-02
Resource Monitor Snapshots


klannon:
It would be very helpful if the resource monitor could take snapshots whenever either (a) a particular file appeared or (b) a line matching some kind of regexp appeared in a given file.  The basic idea is that if you could do that, then the resource monitor could watch your output directories and use the log files to decide when to take snapshots.  That way, you could not only detect when programs are starting, but also when they have changed from one phase to another (e.g. looking for the phrase, "Initialization complete" in a log file and snapshotting then).

If I were implementing this, I would pass the resource monitor a text file in which each line represents a file that the resource monitor should look for and optionally a regex inside the file to watch for.  The resource monitor would start on line one and watch for the file specified there to appear.  If a regexp is also specified, it would wait until it sees a line in the specified file that matches the regexp.  Once a match to a line has been seen, it would move to the next line.

With this example implementation, if I wanted to tell the resource monitor to take a snapshot when it sees `setup.log`, `prepare.log`,`running.log`, a line with `Initialization complete; Start processing` in `running.log`, and `transferoutput.log`, the text file passed to resource monitor would look something like the following:

```
setup.log
prepare.log
running.log
running.log "Initialization complete; Start processing"
transferoutput.log
```

Of course, I don't care how its implemented, as long as the basic functionality is available.  Let me know if you have questions.

btovar:
In #1757,  for task t, you would say:

t.specify_snapshot_file("file")

where "file" would looks:
```json
{
    "setup.log":{
        "events":[
            {
                "label":"setup_started",
                "on-create":true,
                "count":1
            }
        ]
    },
    "prepare.log":{
        "events":[
            {
                "label":"prepare_started",
                "on-create":true,
                "count":1
            }
        ]
    },
    "running.log":{
        "events":[
            {
                "label":"setup_started",
                "on-create":true,
                "count":1
            },
            {
                "label":"processing_started",
                "pattern":"Initialization complete; Start processing",
                "count":1
            }
        ]
    },
    "transferoutput.log":{
        "events":[
            "label":"txoutput_started",
            "on-create":true,
            "count":1
        ]
    }
} 
```

The "count" 1 is simply there for efficiency, to stop checking after the first time an event is detected. The default is -1, which means  no limit in the number of times an event may happen.

Note that this changes specify_snapshot_file from the queue to task, and that the file is a json-document you provide, rather than a file the task produces.

[...]

PR #1757, opened 2017-09-28
resource_monitor snapshots when watching (log) files

btovar:
adds --snapshot-events=file to resource_monitor

file looks like:

```json
{
    "one.log":
    {
        "events":
        [
            {
                "label":"start",
                "pattern":"# START",
                "count":1
            }
        ]
    },
    "two.log":
    {
        "events":[
            {
                "label":"anyline",
                "pattern":"^.*$",
                "on-truncate":true
            }
        ]
    },
    "three.log":
    {
        "delete-if-found":true,
        "events":[
            {
                "label":"manualsnapshot",
                "on-creation":true
            }
        ]
    }
}

```

The monitor watches for changes in the files, and matches the regular
expressions. If there is a match, a snapshot is generated.

Still working on it... @klannon, is this what you had in mind for #1714?

### DIFF
--- a/doc/man/resource_monitor.m4
+++ b/doc/man/resource_monitor.m4
@@ -258,7 +258,7 @@ LONGCODE_BEGIN
 }
 LONGCODE_END
 
-Generate snapshots everytime a line is added to "my.log":
+Generate snapshots every time a line is added to "my.log":
 
 LONGCODE_BEGIN
 {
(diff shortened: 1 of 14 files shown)


btovar:
Still cleaning it up. Missing documentation and wq bindings.

btovar:
Ready for review.

Issue #1729, opened 2017-08-31
Add Ability to Pass Environmental Variables Into the Worker From the Factory


klannon:
See https://github.com/matz-e/lobster/issues/600, in particular @matz-e's last comment.

It would be nice to have some way to pass a list of environmental variables into the worker environment from the factory.  Here is one example use case:  Suppose the task uses an environmental variable to determine the squid proxy location.  If you could tell each factory to set the environmental variable to the appropriate value, it would allow configure different factories for different resource environments (e.g. different squid servers).

@khurtado might also want to comment on whether I have the right idea here.

I know there's a lot of pressure with other deadlines, so no rush here.  No work is stuck.  I just wanted to file it before I forgot about it.

dthain:
Should be pretty straightfoward, since the batch job interface has an explicit place to pass environment variables along to the underlying job.

dthain:
@nkremerh Please implement this one.  Suggest that you allow the factory to have multiple arguments of the form `--env X=Y` that just add elements to JX dictionary, and then pass that through to `batch_job_submit`.  (@btovar has plenty to do right now.)

PR #1758, opened 2017-09-29
Adds environment variable passing option for batch_job_submit

nkremerh:
Addresses #1729.

@btovar @dthain Thoughts?


dthain:
Looks like the right idea to me.
How did you test it?

nkremerh:
@dthain 
I ran a makeflow with -T wq and only one rule:
_worker_env.txt:
    env > worker_env.txt_

I then passed into the factory _--env TESTING123=test_123_ and _--env TESTING321="test"_. Both environment variables were at the worker when it ran the task.

dthain:
Fixes #1729

PR #1760, opened 2017-10-02
Makeflow kubernetes support

charleszheng44:
1. Merge k8s_mode with latest cctools
(i.e. commit 04b5542534a985300324d8585c2897e92d2579ca)
2. How does it works
(1) `batch_job_k8s_submit()` create pods on k8s cluster using
`kubectl create` command
(2) `batch_job_k8s_wait()` keep pulling the k8s cluster and find
running pods, then transfer inputs to the running pods, execute
the command in the target pods and transfer outputs back to the
user end
3. Compare to v2, add function `batch_job_k8s_handle_complete_task()`
for dealing with succeed and failed tasks


dthain:
Add something to `doc/makeflow.html` and `doc/man/makeflow.m4` please!

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+		if(envlist) {
+			jx_export(envlist);
+		}
+
+		char *pod_id = string_format("%s-%d", mf_uuid->str, job_id);
+		char *k8s_config_fn = string_format("%s.json", pod_id);
+
+		FILE *fd = fopen(k8s_config_fn, "w+");
+		
+		if(resources) {
```
dthain: A makeflow job could specify just the CPU or just the RAM, right?  So those default needs to be applied independently.

Also, is there a reason to have the default resource size in the k8s driver?  Why not have a default resource selection across all of makeflow?

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+		}
+
+		fprintf(fd, k8s_config_tmpl, mf_uuid->str, pod_id, pod_id, k8s_image, 
+				resources_block, job_id, pod_id);
+
+		fclose(fd);
+
+		if(access(k8s_script_file_name, F_OK | X_OK) == -1) {
+			debug(D_BATCH, "Generating k8s script...");
+			FILE *f = fopen(k8s_script_file_name, "w");
```
dthain: What if fopen fails?

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+    if(pid > 0) {
+
+    	int status;
+
+    	debug(D_BATCH, "Trying to remove task %d by deleting pods %s.", 
+			(int)jobid, pod_id);
+		
+		// wait child process to complete
+
+		if(waitpid(pid, &status, 0) == -1) {
```
dthain: I think this needs to use process_waitpid, otherwise we have a garbage collection problem in the process module.

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+		k8s_job_info *curr_k8s_job_info, int exited_normally, 
+		struct batch_job_info *info_out, struct list *running_pod_lst,
+		struct batch_queue *q)
+{
+	struct batch_job_info *info;
+	int timeout = 5;
+	debug(D_BATCH, "%d is failed to execute.", job_id);
+	info = itable_remove(q->job_table, job_id);
+	info->exited_normally = exited_normally;
+	// TODO get exact exit_code
```
dthain: Yikes!

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+	if (exited_normally == 0) {
+		info->exit_code = 1;
+	}
+	memcpy(info_out, info, sizeof(*info));
+	free(info);
+
+	list_remove(running_pod_lst, pod_id);
+	batch_job_k8s_remove(q, job_id);
+
+	if(curr_k8s_job_info->is_running == 0) {
```
dthain: This seems very strange.
You need to check the return value of process_wait, otherwise you can't tell if anything exited or not.
And the returned structure needs to be deleted.

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+
+	while(1) {
+	
+		// generate the list of running pod	
+		if(running_pod_lst != NULL) {
+			list_free(running_pod_lst);	
+			list_delete(running_pod_lst);
+		}
+		
+		running_pod_lst = list_create();
```
dthain: Break this into a separate function called get_running_pod_list or the like.

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+			list_delete(running_pod_lst);
+		}
+		
+		running_pod_lst = list_create();	
+		
+		char *cmd = string_format("kubectl get pods -l app=%s | awk \'{if (NR != 1) {print $1\" \"$3}}\' 2>&1 ", 
+				mf_uuid->str);
+		FILE *cmd_fp;
+		char pod_info[128];
+		cmd_fp = popen(cmd, "r");
```
dthain: What if popen fails?

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+		// iterate the running pods
+		char *curr_pod_id;
+		list_first_item(running_pod_lst);	
+		while((curr_pod_id = (char *)list_next_item(running_pod_lst))) {
+			
+			char *get_log_cmd = string_format("kubectl exec %s -- tail -1 %s.log", 
+					curr_pod_id, curr_pod_id);
+			FILE *cmd_fp;
+			char log_tail_content[64];
+			cmd_fp = popen(get_log_cmd, "r");
```
dthain: Again!

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+				return curr_job_id_int;
+			}
+		}
+		pclose(cmd_fp);
+
+		// iterate the running pods
+		char *curr_pod_id;
+		list_first_item(running_pod_lst);	
+		while((curr_pod_id = (char *)list_next_item(running_pod_lst))) {
+
```
dthain: Factor the interior of this loop into a separate function get_pod_status or similar

batch_job/src/batch_job_k8s.c
```diff
@@ -0,0 +1,463 @@
...
+		if(WIFEXITED(status)) {
+			debug(D_BATCH, "Successfully delete pods %s", pod_id);
+		}
+		
+		free(pod_id);
+		return 0;
+
+    } else if(pid < 0) {
+
+		fatal("couldn't create new process: %s\n", strerror(errno));
```
dthain: Why fatal?  Just return failure.

dthain:
Charles, the overall structure looks pretty good, but there are a number of places where we aren't handling failures.  And the logic would be clearer if several complicated bits were broken into smaller functions.

dthain:
Please rebase on the current master to solve conflicts, and then let's get this one finished and committed.

charleszheng44:
Deprecated. Please refer to #1838

PR #1764, opened 2017-10-06
Issue1762

### LINKED ISSUE #1762: jx_eval memory leaks
Reproduce with:   echo '{"key" : "value"} ' | valgrind --leak-check=full ./jx2json 

There are at least three memory leaks:

1) In jx_check_errors.  I think it is caused by both key and value having JX_ERROR type. The jx_copy for the key is leaked.

2) In jx_eval, if a context was not provided as an argument, the locally created context is not freed.

3) In jx_parse_binary, I am not sure what is leaking. It might be fixed by the above.

trshaffer:
Resolves #1762

### DIFF
--- a/dttools/src/jx2json.c
+++ b/dttools/src/jx2json.c
@@ -43,6 +43,7 @@ int main(int argc, char *argv[]) {
 	FILE *stream = stdin;
 	struct jx *ctx = NULL;
 	struct jx *body = NULL;
+	struct jx *tmp = NULL;
 	void (*print_stream)(struct jx *, FILE *) = jx_print_stream;
 
 	int c;
@@ -85,7 +86,9 @@ int main(int argc, char *argv[]) {
 		if (!ctx) return 1;
 	}
 
+	tmp = ctx;
 	ctx = jx_eval(ctx, NULL);
+	jx_delete(tmp);
 	if (jx_istype(ctx, JX_ERROR)) {
 		printf("invalid context\n");
 		print_stream(ctx, stdout);
@@ -106,9 +109,13 @@ int main(int argc, char *argv[]) {
 	fclose(stream);
 	if (!body) return 1;
 
+	tmp = body;
 	body = jx_eval(body, ctx);
+	jx_delete(tmp);
 	print_stream(body, stdout);
 	printf("\n");
+	jx_delete(body);
+	jx_delete(ctx);
 	return 0;
 }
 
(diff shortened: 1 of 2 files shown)


btovar:
Tim, I don't think this solves the leaks in general. Try:

cd resource_monitor/src
valgrind --leak-check=full  -- ./resource_monitor -Omon --snapshot-events file_watch_example.json -- ls

That one still gets:

```text
==19225== 2,341 (48 direct, 2,293 indirect) bytes in 1 blocks are definitely lost in loss record 235 of 235
==19225==    at 0x4C2B975: calloc (vg_replace_malloc.c:711)
==19225==    by 0x41125C: jx_create (jx.c:47)
==19225==    by 0x4114AE: jx_object (jx.c:115)
==19225==    by 0x4143FF: jx_parse_atomic (jx_parse.c:587)
==19225==    by 0x4148FD: jx_parse_postfix (jx_parse.c:742)
==19225==    by 0x414BE2: jx_parse_unary (jx_parse.c:823)
==19225==    by 0x414C05: jx_parse_binary (jx_parse.c:833)
==19225==    by 0x414C1F: jx_parse_binary (jx_parse.c:835)
==19225==    by 0x414C1F: jx_parse_binary (jx_parse.c:835)
==19225==    by 0x414C1F: jx_parse_binary (jx_parse.c:835)
==19225==    by 0x414C1F: jx_parse_binary (jx_parse.c:835)
==19225==    by 0x414C1F: jx_parse_binary (jx_parse.c:835)
==19225== 
==19225== LEAK SUMMARY:
==19225==    definitely lost: 48 bytes in 1 blocks
==19225==    indirectly lost: 2,293 bytes in 69 blocks
==19225==      possibly lost: 0 bytes in 0 blocks
==19225==    still reachable: 15,744 bytes in 225 blocks
==19225==         suppressed: 0 bytes in 0 blocks
```

trshaffer:
That looks just like two of the leaks in the original issue. Valgrind is complaining because the JX isn't getting freed by the caller. fd1f89d just shuffled pointers around to delete the structures before exiting. I hadn't been doing that because the program is exiting. This leak goes away if I add `jx_delete(j)` to the end of `set_snapshot_watch_events()`. I'm not sure if that will cause a segfault later, though.

Anyway, I'm sure there are still plenty of leaks in JX, but I haven't been too concerned with these ones in programs' `main()`s that get cleaned up on exit.

btovar:
Thanks Tim, you are right, that leak is from the monitor.

PR #1767, opened 2017-10-09
adding c style boolean operations and tests

AndrewLitteken:


dttools/src/jx_parse.c
```diff
@@ -270,6 +270,10 @@ static jx_token_t jx_scan( struct jx_parser *s )
 	} else if(c=='!') {
 		char d = jx_getchar(s);
 		if(d=='=') return JX_TOKEN_NE;
```
dthain: Let's just make this a simple `else` and avoid the problem of understanding all the valid characters that may follow..
  (For example, an open paren could come after `!`)

dttools/src/jx_parse.c
```diff
@@ -367,8 +371,12 @@ static jx_token_t jx_scan( struct jx_parser *s )
 					return JX_TOKEN_FALSE;
 				} else if(!strcmp(s->token,"or")) {
 					return JX_TOKEN_OR;
+				} else if (!strcmp(s->token,"||")) {
```
dthain: This is kind of strange to group `||` and `&&` with the keyword tokens.
Please give these their own cases as `<=` and `>=` above.

dttools/src/jx_eval.c
```diff
@@ -77,10 +77,16 @@ static struct jx *jx_eval_boolean(struct jx_operator *op, struct jx *left, struc
 			return jx_boolean(a!=b);
 		case JX_OP_AND:
```
dthain: Combine each pair of equivalent cases!

dthain:
@AndrewLitteken please double check the output of `jx_test < ../test/jx.input`.
I think you accidentally removed a bunch of stuff from `jx.expected`

trshaffer:
Is there a reason to introduce distinct `JX_OP_C_*` operators? I had assumed this could be done as a small change to the parser.

AndrewLitteken:
Because `jx_print` Prints based off of the the operator and not the token, I added the operators to print the correct expression.

trshaffer:
I'd be OK with always printing `and` rather than introducing more operators. That way, you wouldn't have to modify anything outside `jx_parse.c`. The new operators currently have different precedence, but I think the Python-style and C-style operators should be equivalent after parsing.

dttools/src/jx_print.c
```diff
@@ -70,7 +70,6 @@ const char * jx_operator_string( jx_operator_t type )
 		case JX_OP_NOT:	return " not ";
 		// note that the closing bracket/paren is in jx_print_subexpr
 		case JX_OP_LOOKUP: return "[";
-		case JX_OP_CALL: return "(";
```
nhazekam: Should not be removed.

trshaffer:
I think this is just about ready. Few things to fix:
- `jx_print.c` shows up in the diff, but you didn't really touch that file. Just to keep this PR clean, it would be good to check out the previous version of this file.
- Looks like your `jx.expected` is out of date, which is causing a failing test.
- Merge conflict. Re-generating `jx.expected` might fix this, or you might need a rebase.

If you want to go over the git stuff needed, feel free to stop by!

Issue #1149, opened 2016-02-09
Makeflow should have an example workflow for shipping local executables


batrick:
We should expand on `makeflow/example/example.categories.makeflow` to also package up the submission site's `convert` and `curl` (`curl` is also necessary for `-T chirp`). Then, each rule should add dependencies for these executables. This is something we tell our users to do so we need to set a good example.

This issue came from discussion in #1146 and #1147.

dthain:
Good to have an example like this.

I suspect few people see the examples in makeflow/examples, but would be more likely to see it if it is discussed clearly in the manual.

dthain:
After playing around with this, I have come to the conclusion that `convert` is not a great tool to use in our default example workflow.  It has many sub-components installed all over the filesystem and so cannot be easily deployed to remote machines.

A really good example would use `starch` to create a portable executable and then send it along.

dthain:
Added to makeflow examples.

PR #1770, opened 2017-10-10
Adds new Makeflow example using packaged executable

nkremerh:
Addresses #1149.

This example makeflow  has four general steps:
locally curl needed text files, starch the local /usr/bin/perl, run a perl script over the text files, gather results to output final result.

@btovar, mind taking another look at this?

### DIFF
--- a/makeflow/example/shakespeare/Makeflow
+++ b/makeflow/example/shakespeare/Makeflow
@@ -1,0 +1,64 @@
+#This is a sample Makeflow script which retrieves a sampling of William Shakespeare's plays (raw text modified from source at shakespeare.mit.edu),
+#runs an analysis of the amount of dialogue for each character in the sampled plays, and returns the character with the most dialogue and the
+#number of times that character spoke.
+#
+#For each text retrieved (Henry IV Part 1, Henry IV Part 2, Henry VI Part 1, Henry VI Part 2, and Henry VI Part 3) Makeflow will retrieve the text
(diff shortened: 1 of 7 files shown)


dthain:
How about the JX equivalent?

dthain:
Sorry, I just noticed that this is a pull against cctools.
Why not add this to makeflow-examples instead?
https://github.com/cooperative-computing-lab/makeflow-examples

btovar:
I'm fine with either/both.

nkremerh:
I was under the impression this was intended to replace the current Makeflow example in CCTools proper. Please let me know the most appropriate place for this. I will add in a JX version soon.

dthain:
Good, I think that does a good job of conveying how JX can help with the representation.

One thing I don't understand -- it looks like you are using starch multiple times to combine the perl script with the perl executable _and_ the individual input files.  

Why not just use starch once to combine the script with perl, and send the input files individually?  (That way the starched perl can be cached.)

nkremerh:
@dthain, good point! It was my impression starch required a specific command to create a .sfx archive. Since each task specified a unique file to be read by the perl script, I figured I would need to create a unique .sfx for each task which uses perl. I will take another look at it.

nkremerh:
@dthain @btovar Do we want this in CCTools or the Makeflow Examples repository?

dthain:
Makeflow examples, please.
I don't think anyone finds the examples in the current location, and so we should work on moving those as well...

nkremerh:
I will close this pull request and move this to the examples repository.

PR #1781, opened 2017-10-24
Skip "make clean" on re-configure.

dthain:
@btovar what do you think?


btovar:
One reason for that reconfigure is to keep up to date the version and github commit numbers shown with --version in all the executables.

If this is done for fast development, I would rather add a --do-not-clean option to configure.

dthain:
Ok, never mind then

PR #1783, opened 2017-10-26
modified work_queue_factory to integrate runos into the factory. 

Nekel-Seyew:
Not super portable right now. This pull request is more for demonstrating an idea, to be discussed at HEP meetings.

### DIFF
--- a/batch_job/src/work_queue_factory.c
+++ b/batch_job/src/work_queue_factory.c
@@ -45,7 +45,7 @@ See the file COPYING for details.
 #include <unistd.h>
 #include <signal.h>
 
-#define CCTOOLS_RUNOS_PATH /afs/crc.nd.edu/group/ccl/software/runos/runos.py
+#define CCTOOLS_RUNOS_PATH "/afs/crc.nd.edu/group/ccl/software/runos/runos.py"
 
 typedef enum {
 	FORMAT_TABLE,
@@ -385,7 +385,7 @@ static int submit_worker( struct batch_queue *queue )
 	
 	if(runos_os){
 		char* vc3_cmd = string_format("./vc3-builder --require cctools-statics -- %s",cmd);
-		char* temp = string_format("python CCTOOLS_RUNOS_PATH %s %s",runos_os,vc3_cmd);
+		char* temp = string_format("python %s %s %s",CCTOOLS_RUNOS_PATH,runos_os,vc3_cmd);
 		free(vc3_cmd);
 		free(cmd);
 		cmd = temp;
@@ -1280,15 +1280,16 @@ int main(int argc, char *argv[])
 			fprintf(stderr, "work_queue_factory: Could not Access specified worker_queue_worker binary.\n");
 			exit(EXIT_FAILURE);
 		}
+		free(cmd);
 	}else{
 		cmd = string_format("cp \"$(which work_queue_worker)\" '%s'",scratch_dir);
 		if (system(cmd)) {
 			fprintf(stderr, "work_queue_factory: please add work_queue_worker to your PATH.\n");
 			exit(EXIT_FAILURE);
 		}
+		free(cmd);
 	}
 	
-	free(cmd);
 	
 	if(runos_os) {
 		cmd = string_format("cp \"$(which vc3-builder)\" '%s'", scratch_dir);
@@ -1300,8 +1301,9 @@ int main(int argc, char *argv[])
 	}
 
 	if(password_file) {
-		sprintf(cmd,"cp %s %s/pwfile",password_file,scratch_dir);
+		cmd = string_format("cp %s %s/pwfile",password_file,scratch_dir);
 		system(cmd);
+		free(cmd);
 	}
 
 	if(chdir(scratch_dir)!=0) {
--- a/batch_job/src/work_queue_factory.c
+++ b/batch_job/src/work_queue_factory.c
@@ -45,6 +45,8 @@ See the file COPYING for details.
 #include <unistd.h>
 #include <signal.h>
 
+#define CCTOOLS_RUNOS_PATH /afs/crc.nd.edu/group/ccl/software/runos/runos.py
+
 typedef enum {
 	FORMAT_TABLE,
 	FORMAT_LONG
@@ -383,7 +385,7 @@ static int submit_worker( struct batch_queue *queue )
 	
 	if(runos_os){
 		char* vc3_cmd = string_format("./vc3-builder --require cctools-statics -- %s",cmd);
-		char* temp = string_format("python /afs/crc.nd.edu/group/ccl/software/runos/runos.py %s %s",runos_os,vc3_cmd);
+		char* temp = string_format("python CCTOOLS_RUNOS_PATH %s %s",runos_os,vc3_cmd);
 		free(vc3_cmd);
 		free(cmd);
 		cmd = temp;
--- a/batch_job/src/work_queue_factory.c
+++ b/batch_job/src/work_queue_factory.c
@@ -98,7 +98,7 @@ static char *wrapper_command = 0;
 static char *wrapper_input = 0;
 static char *worker_command = 0;
 
-static char *os = NULL;
+static char *runos_os = 0;
 
 /* -1 means 'not specified' */
 static struct rmsummary *resources = NULL;
@@ -331,7 +331,7 @@ static void set_worker_resources_options( struct batch_queue *queue )
 static int submit_worker( struct batch_queue *queue )
 {
 	char *cmd;
-	const char *worker = "work_queue_worker";
+	const char *worker = runos_os ? "work_queue_worker" : "./work_queue_worker";
 
 	if(using_catalog) {
 		cmd = string_format(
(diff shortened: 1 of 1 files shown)


batch_job/src/work_queue_factory.c
```diff
@@ -373,6 +375,14 @@ static int submit_worker( struct batch_queue *queue )
 		free(files);
 		files = newfiles;
 	}
+	
+	if(os){
+		char* temp = string_format("python /afs/crc.nd.edu/group/ccl/software/runos/runos.py %s %s",os,cmd);
```
btovar: I think you want to have runos in your PATH, copy it to scratch_dir, and send it as we send the worker. (See the line that has "which work_queue_worker".

batch_job/src/work_queue_factory.c
```diff
@@ -360,7 +362,7 @@ static int submit_worker( struct batch_queue *queue )
 		cmd = newcmd;
 	}
 
-	char *files = string_format("work_queue_worker");
+	char *files = os ? string_format("") : string_format("work_queue_worker");
```
btovar: I would make it:

files = os ? string_format("runos.py") : string_format("work_queue_worker");

Ideally, the factory should not send anything that depends on a global path.

batch_job/src/work_queue_factory.c
```diff
@@ -322,7 +324,7 @@ static void set_worker_resources_options( struct batch_queue *queue )
 static int submit_worker( struct batch_queue *queue )
 {
 	char *cmd;
-	const char *worker = "./work_queue_worker";
+	const char *worker = os ? "work_queue_worker": "./work_queue_worker";
```
btovar: This assumes the worker is always in the PATH in os. Is this always true?

batch_job/src/work_queue_factory.c
```diff
@@ -100,6 +100,8 @@ static char *wrapper_command = 0;
 static char *wrapper_input = 0;
 static char *worker_command = 0;
 
+static char *os = NULL;
```
btovar: I suggest to rename this variable to something more explicit, e.g., runos_opsys.

batch_job/src/work_queue_factory.c
btovar: That is, it should work as if we passed --wrapper-file runos.py to the factory.

batch_job/src/work_queue_factory.c
Nekel-Seyew: if it's in the case of being inside of singularity (as all cctools are in the path of my custom images), or now as being inside the wrapper of vc3-builder, then yes.

btovar:
This seems very specific to ND.

I can deploy it so that ND CMS can use in the meanwhile, but I don't think it should be merged in its current form to master.

batch_job/src/work_queue_factory.c
```diff
@@ -1161,6 +1187,13 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	memset(cmd,'\0',1024*sizeof(char));
```
btovar: bzero(cmd, sizeof(*cmd));

Nekel-Seyew: Isn't bzero a gnu specific function in string.h?

batch_job/src/work_queue_factory.c
```diff
@@ -98,6 +98,8 @@ static char *wrapper_command = 0;
 static char *wrapper_input = 0;
 static char *worker_command = 0;
 
+static char *os = NULL;
```
btovar: change to *runos_os = 0;

batch_job/src/work_queue_factory.c
```diff
@@ -329,7 +331,7 @@ static void set_worker_resources_options( struct batch_queue *queue )
 static int submit_worker( struct batch_queue *queue )
 {
 	char *cmd;
-	const char *worker = "./work_queue_worker";
+	const char *worker = "work_queue_worker";
```
btovar: This change breaks work_queue_factory when not using runos.

batch_job/src/work_queue_factory.c
```diff
@@ -365,7 +367,7 @@ static int submit_worker( struct batch_queue *queue )
 		cmd = newcmd;
 	}
 
-	char *files = string_format("work_queue_worker");
+	char *files=string_format("%s","");// = os ? string_format("%s","") : string_format("work_queue_worker");
```
btovar: instead use: xxstrudp("");

batch_job/src/work_queue_factory.c
```diff
@@ -378,6 +380,23 @@ static int submit_worker( struct batch_queue *queue )
 		free(files);
 		files = newfiles;
 	}
+	
+	if(os){
```
btovar: if(runos_os) {

batch_job/src/work_queue_factory.c
```diff
@@ -378,6 +380,23 @@ static int submit_worker( struct batch_queue *queue )
...
+		char* vc3_cmd = string_format("./vc3-builder --require cctools-statics -- %s",cmd);
+		char* temp = string_format("python /afs/crc.nd.edu/group/ccl/software/runos/runos.py %s %s",os,vc3_cmd);
+		free(vc3_cmd);
+		free(cmd);
+		cmd = temp;
+		temp = string_format("%s,%s",files,"vc3-builder");
+		free(files);
+		files = temp;
+	}else{
+		char* vc3_cmd = string_format("./vc3-builder --require cctools-statics -- %s",cmd);
```
btovar: If we are not using runos, we do not want to use the builder.

batch_job/src/work_queue_factory.c
btovar: Yes, I did not know it was deprecated... I liked it had less confusing order of arguments.

Do remove the 1024. Instead use sizeof(cmd);

batch_job/src/work_queue_factory.c
```diff
@@ -1263,6 +1288,14 @@ int main(int argc, char *argv[])
 		}
 	}
 
+	memset(cmd,'\0',1024*sizeof(char));
+	sprintf(cmd,"cp \"$(which vc3-builder)\" '%s'",scratch_dir);
+	int k = system(cmd);
+	if(k){
+		fprintf(stderr,"can't copy vc3-builder!: %i\n",k);
```
btovar: If so, say that the builder should be in the PATH.

batch_job/src/work_queue_factory.c
```diff
@@ -381,21 +381,19 @@ static int submit_worker( struct batch_queue *queue )
...
 		char* vc3_cmd = string_format("./vc3-builder --require cctools-statics -- %s",cmd);
-		char* temp = string_format("python /afs/crc.nd.edu/group/ccl/software/runos/runos.py %s %s",os,vc3_cmd);
+		char* temp = string_format("python /afs/crc.nd.edu/group/ccl/software/runos/runos.py %s %s",runos_os,vc3_cmd);
 		free(vc3_cmd);
 		free(cmd);
 		cmd = temp;
 		temp = string_format("%s,%s",files,"vc3-builder");
 		free(files);
 		files = temp;
 	}else{
```
btovar: Where is cmd defined for the regular case?

batch_job/src/work_queue_factory.c
```diff
@@ -381,21 +381,19 @@ static int submit_worker( struct batch_queue *queue )
 		files = newfiles;
 	}
 	
-	if(os){
+	if(runos_os){
 		char* vc3_cmd = string_format("./vc3-builder --require cctools-statics -- %s",cmd);
-		char* temp = string_format("python /afs/crc.nd.edu/group/ccl/software/runos/runos.py %s %s",os,vc3_cmd);
+		char* temp = string_format("python /afs/crc.nd.edu/group/ccl/software/runos/runos.py %s %s",runos_os,vc3_cmd);
```
btovar: make a #define CCTOOLS_RUNOS_PATH /afs/... at the top of this file, and then use "python " CCTOOLS_RUNOS_PATH " %s %s", etc

batch_job/src/work_queue_factory.c
Nekel-Seyew: cmd is only ever wrapped around if it's changed. So the root cmd is defined first by "if catalog else" statements, and then only modified from there. The else statement here for the runos_os if statement simply verifies that we're not using runos or vc3-builder, so to use the normal work_queue_worker.

batch_job/src/work_queue_factory.c
```diff
@@ -1248,21 +1271,30 @@ int main(int argc, char *argv[])
...
-		sprintf(cmd,"cp \"$(which work_queue_worker)\" '%s'",scratch_dir);
+		cmd = string_format("cp \"$(which work_queue_worker)\" '%s'",scratch_dir);
 		if (system(cmd)) {
 			fprintf(stderr, "work_queue_factory: please add work_queue_worker to your PATH.\n");
 			exit(EXIT_FAILURE);
 		}
 	}
 
+	if(runos_os) {
+		cmd = string_format("cp \"$(which vc3-builder)\" '%s'", scratch_dir);
```
btovar: free(cmd), since it was allocated in the if-else block above.

Nekel-Seyew: Done! Also changed it in other places to reflect that change.

dthain:
@btovar Merge if you are satisfied.

PR #1784, opened 2017-10-26
Fix zlib when static linking.

btovar:
It broke when compressed updates to the catalog where added.


dthain:
Should this be merged?

Issue #1773, opened 2017-10-18
Makeflow use of set data structure is incorrect


btovar:
set works by unique ids, usually pointers. In makeflow, it is used with strings, thus the same string appears several times. 

It is used to see which files are input/outputs and which variables to export. Current implementation works, but it is inefficient.

PR #1788, opened 2017-10-31
Fix var sets

nhazekam:
@btovar please take a look. I still need to do a bit of testing, but I figure it can still be removed. Addressed #1773


dthain:
@btovar What do you think?

Issue #1796, opened 2017-11-09
Add option to makeflow to send local environment to workers


btovar:
(generated from #1790)

PR #1797, opened 2017-11-10
adds --send-environment

btovar:
Sends local environment variables where the makeflow is running to the
remote execution sites.

Fix for #1796


btovar:
It is defined in unistd.h. That is why we have to use the 'extern' declaration.

PR #1802, opened 2017-11-23
Enable cvmfs repository switching by default on new libcvmfs

jblomer:
Enables multi-repository functionality without the need to set `PARROT_ALLOW_SWITCHING_CVMFS_REPOSITORIES`


dthain:
I think we have had enough experience with repo switching to use it by default now.
Thanks!

PR #1807, opened 2017-12-05
First pass at fixing node level resources

nhazekam:
@btovar The previous category resource specifications removed local node level resource specification support. This should add it back.

@trshaffer You are add due to changes in the `parser_jx.c` 

TODO: Add tests for resource resolution
TODO: Test more thoroughly with categories.


nhazekam:
@btovar @trshaffer Please take a look and review. I would prefer if this were merged prior to the hook migration.

PR #1808, opened 2017-12-11
Submakeflows

AndrewLitteken:
Adding support for submakeflows


dthain:
@nhazekam I'm looking at the bit where a sub-makeflow node is transformed into a command execution of makeflow.  Could this capability be expressed using a hook?

(Not asking you to do it at this point, just want your advice on whether it fits.)

dthain:
After discussing with @nhazekam , here's what we need to do to integrate this into the latest code form:

Choice 1: Modify `dag_node_to_batch_task` to create a suitable task for a makeflow, given a node as an input.  Right now, this function assumes that every task has a `command` which is not the case here.

Choice 2: Create a hook module (always on) that looks for nodes of type `makeflow` and then, before submission, converts them into makeflow command lines.

dthain:
See #2102

PR #1809, opened 2017-12-13
Added functions to convert data struct to jx

nhazekam:
Functions that return a Node/Task as a JX object.

Will be used later for wrappers and in Batch Job.

### DIFF
--- a/makeflow/src/dag_file.c
+++ b/makeflow/src/dag_file.c
@@ -32,6 +32,24 @@ struct dag_file * dag_file_create( const char *filename )
 	return f;
 }
 
+/* Return JX object containing name, remote name, and size. */
+
+struct jx * dag_file_to_jx( struct dag_file *f, struct dag_node *n)
+{
+	struct jx *file = jx_object(0);
+
+	jx_insert(file, jx_string("filename"), jx_string(f->filename));
+
+	const char *remote_name = dag_node_get_remote_name(n, f->filename);
+	if(remote_name){
+		jx_insert(file, jx_string("remote_filename"), jx_string(remote_name));
+	}
+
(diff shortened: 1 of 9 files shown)


dthain:
Thanks for making a first incremental step.
We seem to have a variety of names for the local/remote filename distinction used in batch_job, makeflow, makeflow_jx, wrappers, and now this code.
Could you please summarize the various names we have used here, so we can zero in on a consistent nomenclature?
@trshaffer @btovar

nhazekam:
After looking at the existing naming structure and in conjunction with the batch_file concept,
this branch uses the names:
name_on_submission : which was analogous to filename/path.
name_on_execution : which was analogous to remote_name.

dthain:
We seem to be accumulating names for this concept.
Could you please look at the nomenclature currently used for this in batch_job/dag/work_queue/confuga and paste them here?
Let's try to synchronize on one set of names.

nhazekam:
batch_job:
name_on_submission
name_on_execution
* (collection of name occasionally referenced by specific batch_queue, not shared)

dag:
filename
remote_name

dag (parse_jx)
path
execution_path

work_queue:
local_name
payload
cached_name
remote_name

confuga_file: contains a chirp_file
path (contained in both confuga_file and chirp_file)
fid (file id)
host(possibly, in chirp_file)

dthain:
In batch_job_chirp (confuga):
- task_path
- serv_path

In batch_job_amazon:

- localname
- remotename
-

nhazekam:
As per our team discussion yesterday, we will be moving toward:
outer_name : inner_name in the code
dag_name : task_name in the documentation

To reflect this, jx uses the keys dag_name and task_name. This corresponds to filename and remote_name currently, but will be outer_name and inner_name when makeflow is updated to reflect the decided upon naming.

PR #1810, opened 2017-12-13
Makeflow parse formats

nhazekam:
The goal of this is to create a single entry point for creating a dag from a file. Hopefully this will force us to keep other apps current as makeflow_analyze and makeflow_viz did not support makeflow. 

There are several files with significant changes though this is a result of renaming parser.c to parser_make.c.


makeflow/src/makeflow_viz.c
```diff
@@ -207,6 +235,42 @@ int main(int argc, char *argv[])
 			case LONG_OPT_DOT_FILE:
 				file_attr = xxstrdup(optarg);
 				break;
+			case LONG_OPT_JSON:
+				dag_syntax = DAG_SYNTAX_JSON;
+				break;
+			case LONG_OPT_JX:
+				dag_syntax = DAG_SYNTAX_JX;
+				break;
+			case LONG_OPT_JX_ARGS:
```
trshaffer: It might be better to refactor this stuff, since handling these options shows up a few times in this PR, and it's also in `jx2json`

PR #1819, opened 2018-01-11
Hook header

nhazekam:
Initial commit to show the structure and organization of MAKEFLOW HOOKS.

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -119,6 +119,16 @@ static batch_queue_type_t batch_queue_type = BATCH_QUEUE_TYPE_LOCAL;
 static struct batch_queue *local_queue = 0;
 static struct batch_queue *remote_queue = 0;
 
+struct batch_queue * makeflow_get_remote_queue(){
+	return remote_queue;
+}
+
+struct batch_queue * makeflow_get_local_queue(){
+	return local_queue;
+}
+
+
+
 static struct rmsummary *local_resources = 0;
 
 static int local_jobs_max = 1;
@@ -156,8 +166,6 @@ static struct makeflow_monitor *monitor = 0;
 static struct makeflow_wrapper *enforcer = 0;
 static struct makeflow_wrapper_umbrella *umbrella = 0;
 
-static struct list *makeflow_hooks = NULL;
-
 static int catalog_reporting_on = 0;
 
 static char *mountfile = NULL;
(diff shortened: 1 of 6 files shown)


makeflow/src/makeflow_hook.h
```diff
@@ -0,0 +1,322 @@
...
+ *
+ * All methods are optional. 
+ *
+ * Here are list the immutable features and structures for each
+ * high level struct:
+ * DAG:
+ * Node:
+ * File:
+ *
+ * Almost all operations take a path which can be of any length.
```
trshaffer: Copy/paste error?

makeflow/src/makeflow_hook.h
```diff
@@ -0,0 +1,322 @@
...
+	 * This hook occurs when nodes are checked for execution. This
+	 * allows hooks to veto submission based on internal qualifiers.
+	 * Example would be the storage allocation, or job limits.
+	 *
+	 * @param dag_node The dag_node that is being checked.
+	 * @param batch_job_feature A strucuture that describes supported
+	 *             features of used batch_job system.
+	 * @return 1 if successful, 0 if not.
+	 */
+	int (*node_check)    (struct dag_node *, struct hash_table *);
```
trshaffer: Since the doc comments refer to parameters, it might be good to add dummy names in the header definitions, e.g.

    int (*node_check) (struct dag_node *d, struct hash_table *f);

so the names in the comments line up. Here, `batch_job_feature` isn't obviously connected with a `struct 
hash_table *`.

makeflow/src/makeflow_hook.h
```diff
@@ -0,0 +1,322 @@
...
+	 * @return 1 is successful, 0 if not.
+	 */
+	int (*file_deleted)  (struct dag_file *);
+	
+};
+
+typedef enum {
+    MAKEFLOW_HOOK_SUCCESS = 0,
+    MAKEFLOW_HOOK_FAILURE
+} makeflow_hook_return;
```
trshaffer: Should the hook functions return `makeflow_hook_return` rather than `int`? (And just being picky: if so, I think something like `makeflow_hook_result` would read better)

trshaffer:
Based on discussion looks like it will be necessary to make the DAG and batch queue available throughout. I would prefer not to add them as parameters to every hook. Instead, I'd like to add functions like `struct dag *makeflow_hook_get_dag();` to `makeflow_hook.h`. `makeflow.c` could provide the implementations. Since these are cross-cutting and should be used sparingly/phased out if possible, I'd prefer to put them behind functions.

makeflow/src/makeflow.c
```diff
@@ -172,6 +175,27 @@ static struct makeflow_alloc *storage_allocation = NULL;
...
+		makeflow_hooks = list_create(); \
+	list_first_item(makeflow_hooks); \
+	for (struct makeflow_hook *h; (h = list_next_item(makeflow_hooks));) { \
+		int rc = MAKEFLOW_HOOK_SUCCESS; \
+		if (h->hook_name) \
+			rc = h->hook_name(__VA_ARGS__); \
+		if (rc !=MAKEFLOW_HOOK_SUCCESS) \
+			fatal("hook %s:" #hook_name " returned %d",h->module_name?h->module_name:"", rc); \
+	} \
+} while (0)
```
dthain: Let's not fill up the main loop with macro expansions, that can lead to crazy error messages and difficult debugging.
In the main loop of makeflow, just make simple function calls to `makeflow_hook_x()` etc at the key points.
Then, define `makeflow_hook_x()` in `makeflow.c` and let that handle the iteration over the various hook objects.
You can use macros there if you like.

makeflow/src/makeflow.c
```diff
@@ -172,6 +175,27 @@ static struct makeflow_alloc *storage_allocation = NULL;
...
+
+static void register_hook(struct makeflow_hook *hook) {
+	assert(hook);
+	if (!makeflow_hooks) makeflow_hooks = list_create();
+	struct makeflow_hook *h = xxmalloc(sizeof(*h));
+	memcpy(h, hook, sizeof(*h));
+	list_push_head(makeflow_hooks, h);
+}
+
```
dthain: Hide this function in `makeflow_hook.c`

makeflow/src/makeflow_hook.h
```diff
@@ -0,0 +1,320 @@
...
+
+	/* Destroy/Clean up hooks.
+	 *
+	 * Call when Makeflow is about to exit. This is to clean up any memory
+	 * or structs left during execution. Most will be removed when exiting.
+	 * This should only be used to clean up internal strucutures.
+	 *
+	 * @return 1 if successfully destroyed, 0 if not.
+	 */
+	int (*destroy)       ();
```
dthain: May need to pass in the dag here?

trshaffer:
It looks like some of the Makeflow tests are failing

PR #1824, opened 2018-01-15
Batch task

nhazekam:
Initial setup of batch_task to be used with wrapper and hooks in Makeflow. Common interface with batch_job. Does not interface with submit yet.

### DIFF
--- a/batch_job/src/batch_file.c
+++ b/batch_job/src/batch_file.c
@@ -18,7 +18,7 @@ See the file COPYING for details.
  **/
 struct batch_file *batch_file_create(struct batch_queue *queue, char * name_on_submission, char * name_on_execution)
 {
-	struct batch_file *f = malloc(sizeof(*f));
+	struct batch_file *f = calloc(1,sizeof(*f));
     f->name_on_submission = xxstrdup(name_on_submission);
 
 	if(batch_queue_supports_feature(queue, "remote_rename") && name_on_execution){
@@ -37,6 +37,9 @@ struct batch_file *batch_file_create(struct batch_queue *queue, char * name_on_s
  **/
 void batch_file_delete(struct batch_file *f)
 {
+	if(!f)
+		return;
+
 	free(f->name_on_submission);
 	free(f->name_on_execution);
 
@@ -67,10 +70,7 @@ char * batch_files_to_string(struct batch_queue *queue, struct list *files )
 
     char * file_str = strdup("");
 
-	/* This could be set using batch_queue feature or option 
-	 * to allow for batch system specific separators. */
-	char * separator = ",";
-	int location = 0;
+	char * separator = "";
 
     if(!files) return file_str;
 
@@ -81,12 +81,15 @@ char * batch_files_to_string(struct batch_queue *queue, struct list *files )
 			continue;
 
 		/* Only add separator if past first item. */
-		if(location == 0)
-			file_str = string_combine(file_str,separator);
+		file_str = string_combine(file_str,separator);
 
         char *f = batch_file_to_string(queue, file);
         file_str = string_combine(file_str,f);
-		location++;
+	
+		/* This could be set using batch_queue feature or option 
+		 * to allow for batch system specific separators. */
+		separator = ",";
+
         free(f);
     }
 
--- a/batch_job/src/batch_file.c
+++ b/batch_job/src/batch_file.c
@@ -9,22 +9,22 @@ See the file COPYING for details.
 #include "xxmalloc.h"
 
 /**
- * Create batch_file from host_name and exe_name.
+ * Create batch_file from name_on_submission and name_on_execution.
  * Host name indicates the name that will be on the host/submission side.
  *  This is equivalent to the filename in Makeflow.
  * Exe name indicates the name that will be used for execution.
- *  IF no exe_name is given, or the specified batch_queue does not support
- *  remote renaming the host_name will be used.
+ *  IF no name_on_execution is given, or the specified batch_queue does not support
+ *  remote renaming the name_on_submission will be used.
  **/
-struct batch_file *batch_file_create(struct batch_queue *queue, char * host_name, char * exe_name)
+struct batch_file *batch_file_create(struct batch_queue *queue, char * name_on_submission, char * name_on_execution)
 {
-	struct batch_file *f = malloc(sizeof(struct batch_file));
-    f->host_name = xxstrdup(host_name);
+	struct batch_file *f = malloc(sizeof(*f));
+    f->name_on_submission = xxstrdup(name_on_submission);
 
-	if(batch_queue_supports_feature(queue, "remote_rename") && exe_name){
-		f->exe_name = xxstrdup(exe_name);
+	if(batch_queue_supports_feature(queue, "remote_rename") && name_on_execution){
+		f->name_on_execution = xxstrdup(name_on_execution);
 	} else {
-		f->exe_name = f->host_name;
+		f->name_on_execution = xxstrdup(name_on_submission);
 	}
 
 	f->batch_sys_include = BATCH_FILE_INCLUDE;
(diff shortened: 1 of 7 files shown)


batch_job/src/batch_file.c
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Delete batch_file, including freeing host_name and exe_name/
+ **/
+void batch_file_delete(struct batch_file *f)
+{
+	free(f->host_name);
+	f->host_name = NULL;
+	if(f->exe_name)
```
btovar: No need to check for NULL. free(NULL) is a no-op.

batch_job/src/batch_file.c
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Given a file, return the string that identifies it appropriately
+ * for the given batch system, combining the local and remote name
+ * and making substitutions according to the node.
+ **/
+char * batch_file_to_string(struct batch_queue *queue, struct batch_file *f )
+{
+	if(f->batch_sys_include == BATCH_FILE_EXCLUDE) return strdup("");
```
btovar: Isn't NULL better here? Better to have logic that handles exclude later explicitly.

batch_job/src/batch_file.c
```diff
@@ -0,0 +1,93 @@
...
+{
+    struct batch_file *file;
+
+    char * file_str = strdup("");
+
+    if(!files) return file_str;
+
+    list_first_item(files);
+    while((file=list_next_item(files))) {
+        char *f = batch_file_to_string(queue, file);
```
btovar: Aren't you missing commas, or spaces?

batch_job/src/batch_file.c
```diff
@@ -0,0 +1,93 @@
...
+        file_str = string_combine(file_str,f);
+        free(f);
+    }
+
+    return file_str;
+}
+
+/**
+ * Indicate if file is to be included or not in files string
```
btovar: Do you mean set?

batch_job/src/batch_file.c
```diff
@@ -0,0 +1,93 @@
...
+ **/
+void batch_file_inclusion(struct batch_file *f, batch_file_inclusion_t inclusion)
```
btovar: batch_set_file_inclusion may be better.

batch_job/src/batch_file.h
```diff
@@ -0,0 +1,61 @@
...
+#include "batch_job.h"
+#include "list.h"
+
+typedef enum {
+    BATCH_FILE_INCLUDE,   /* This file is included in files string. (Default) */
+    BATCH_FILE_EXCLUDE,   /* This file is excluded from files string. */
+} batch_file_inclusion_t;
+
+struct batch_file {
+	char *host_name;
```
btovar: See above first comment. host_name sounds to much like a name of a machine. exe_name sounds to much like the name of an executable.

batch_job/src/batch_file.h
```diff
@@ -0,0 +1,61 @@
...
+*/
+
+#ifndef BATCH_FILE_H
+#define BATCH_FILE_H
+
+#include "batch_job.h"
+#include "list.h"
+
+typedef enum {
+    BATCH_FILE_INCLUDE,   /* This file is included in files string. (Default) */
```
btovar: A comment describing why the distinction of these cases would be nice.

batch_job/src/batch_file.h
```diff
@@ -0,0 +1,61 @@
...
+struct batch_file *batch_file_create(struct batch_queue *queue, char *host_name, char *exe_name );
+
+/** Delete batch_file struct.
+ This includes freeing host_name and exe_name if defined.
+@param file A batch_file struct to be deleted.
+*/
+void batch_file_delete(struct batch_file *f);
+
+/** Flatten batch_file to string.
+@param queue The batch_queue that this is being flattened for.
```
btovar: What does flatten mean here? Document the format of the output string.

batch_job/src/batch_file.h
```diff
@@ -0,0 +1,61 @@
...
+@param file A batch_file struct to be stringified. 
+@return pointer to char * representing the flattened list.
+*/
+char * batch_file_to_string(struct batch_queue *queue, struct batch_file *f );
+
+/** Flatten batch_file list to string.
+@param files A list struct containing batch_file structs. 
+@param queue The batch_queue that this is being flattened for. 
+@return pointer to char * representing the flattened list.
+*/
```
btovar: Document how filenames separated in the output string.

batch_job/src/batch_file.h
```diff
@@ -0,0 +1,61 @@
...
+char * batch_files_to_string(struct batch_queue *queue, struct list *files );
+
+/** Signal that this batch_file is ignored by batch_queue.
```
btovar: Set? signal sounds more dynamic.

batch_job/src/batch_file.h
```diff
@@ -0,0 +1,61 @@
...
+char * batch_files_to_string(struct batch_queue *queue, struct list *files );
+
+/** Signal that this batch_file is ignored by batch_queue. 
+ Used in batch_file_to_string to exclude file from string list, but
+ acknowledge the necessity of this file.
+@param file The batch_file to be ignored/enabled.
+@param ignored If BATCH_FILE_INCLUDE this file will be included in files string,
+	if BATCH_FILE_EXCLUDE this file is ignored for files string.
+*/
+void batch_file_inclusion(struct batch_file *f, batch_file_inclusion_t inclusion);
```
btovar: batch_file_set_inclusion_mode?

batch_job/src/batch_task.c
```diff
@@ -0,0 +1,90 @@
...
+See the file COPYING for details.
+*/
+
+#include "batch_task.h"
+#include "stringtools.h"
+#include "xxmalloc.h"
+
+struct batch_task *batch_task_create(struct batch_queue *queue)
+{
+	struct batch_task *t = malloc(sizeof(struct batch_task));
```
btovar: Given the 0 and NULL to follow (and on fields that may be added later), I suggest:

struct batch_task *t = calloc(1,sizeof(*t));

Then you can remove the = 0 and = NULL.

batch_job/src/batch_task.c
```diff
@@ -0,0 +1,90 @@
...
+	t->resources = NULL;
+
+	t->envlist = NULL;
+
+	return t;
+}
+
+void batch_task_delete(struct batch_task *t)
+{
+	if(t->command)
```
btovar: No need of if. free(NULL) is a no-op.

batch_job/src/batch_task.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef BATCH_TASK_H
+#define BATCH_TASK_H
+
+#include "batch_file.h"
+#include "list.h"
+#include "jx.h"
+#include "rmsummary.h"
+
+struct batch_task {
+	int taskid;
```
btovar: Document the difference between taskid and jobid.

batch_job/src/batch_file.c
nhazekam: Commas are currently output in the above function. I can move them down for more explicit formatting.

batch_job/src/batch_file.c
nhazekam: Would you prefer this logic in batch_files_to_string instead, or are you looking for something even higher in the logic?

btovar: Yes, for example. Otherwise, in some other future uses, we may end up with weird "".

batch_job/src/batch_file.c
btovar: I see. The very last comma may be annoying for parsing later.

PR #1830, opened 2018-01-23
Builder for wrapper scripts

trshaffer:
This is a library function for generating wrapper scripts. The same functionality is repeated in a few places in Makeflow, so this gives a common implementation to simplify the transition to modules.

### DIFF
--- a/batch_job/src/batch_task.c
+++ b/batch_job/src/batch_task.c
@@ -4,9 +4,23 @@ This software is distributed under the GNU General Public License.
 See the file COPYING for details.
 */
 
+#include <assert.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+
 #include "batch_task.h"
 #include "stringtools.h"
 #include "xxmalloc.h"
+#include "path.h"
+#include "debug.h"
+
+struct batch_task_wrapper {
+	struct list *pre;
+	struct list *post;
+	struct list *cmd;
+};
 
 /** Creates new batch_task and initializes file lists. */
 struct batch_task *batch_task_create(struct batch_queue *queue)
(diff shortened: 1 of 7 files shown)


btovar:
Don't we want all pre's to be push_head (most recent added executed first), and all post's to be push_tail (most recent added executed last)?

batch_job/src/batch_task.c
```diff
@@ -133,4 +147,120 @@ void batch_task_set_info(struct batch_task *t, struct batch_job_info *info)
...
+	list_delete(w->post);
+	list_free(w->cmd);
+	list_delete(w->cmd);
+	free(w);
+}
+
+void batch_task_wrapper_pre(struct batch_task_wrapper *w, const char *cmd) {
+	assert(w);
+	assert(cmd);
+	assert(!w->cmd);
```
nhazekam: Are you requiring that `w->cmd` is not specified prior to the pre or post? Is there a purpose to asserting this here? Also in post at line 206.

nhazekam:
@btovar I think by using list_push_tail, writing a wrapper lets the user specify them in the order they want them run. I think these are going to be nested, so we can separate wrapper levels and the order of pre and post are as added. @trshaffer let me know if this understanding is off.

batch_job/src/batch_task.c
```diff
@@ -4,9 +4,23 @@ This software is distributed under the GNU General Public License.
...
 #include "batch_task.h"
 #include "stringtools.h"
 #include "xxmalloc.h"
+#include "path.h"
+#include "debug.h"
+
+struct batch_task_wrapper {
+	struct list *pre;
+	struct list *post;
+	struct list *cmd;
```
nhazekam: Should a batch_task_wrapper contain a batch_task_wrapper to allow for nesting? Also how do we want to handle wrapping? We can talk more in depth later, but my initial thinking on this code is that we have functionality like a linked list for wrappers.

trshaffer:
@nhazekam you're right, each layer of a big wrapped command would use one of these builders. Calling write generates a shell script for a single layer of the wrapper, so we should keep multiple pre/post commands in the order they're provided. As an example, this task wrapper

    struct batch_task_wrapper *w = batch_task_wrapper_create();
    batch_task_wrapper_pre(w, "./check_environment.sh");
    batch_task_wrapper_pre(w, "./wrapper_setup.sh");
    batch_task_wrapper_argv(w, orig_argv);
    batch_task_wrapper_post(w, "rm -rf ./output");
    char *wrapper_name = batch_task_wrapper_write(w, t, "example_");
    batch_task_wrapper_delete(w);

would generate the following script and add as an input to the batch task.

    #!/bin/sh
    set -e
    CLEANUP_76tnb43rr7 () {
    eval "rm -rf ./output"
    }
    trap CLEANUP_76tnb43rr7 EXIT INT TERM
    eval "./check_environment.sh"
    eval "./wrapper_setup.sh"
    "./run" "-arg" "foo"

batch_job/src/batch_task.c
trshaffer: That's basically because of laziness. I have uses in mind for multiple pre commands followed a single argv. Rather than adding in structures to support arbitrary combinations of commands, I just implemented the former usage and put in those asserts so we could add the full functionality (if it's needed down the road) without breaking any existing code. I can add a comment explaining that this is a limited subset of functionality.

trshaffer:
The changes I just pushed give a more complete view of how I'm thinking about using this. If Makeflow has a JX wrapper, it can apply it to a node like so.

    {
        "pre": "export LD_PRELOAD=whatever.so",
        "cmd": {
            "argv": ["original_command", "--from-makeflow"],
            "post": ["./cleanup.sh"]
        }
    }

This PR adds infrastructure to recursively generate wrapper scripts. The innermost one, `temp_0.sh`, would look roughly like this:

    #!/bin/sh
    cleanup_fn () {
    ./cleanup.sh
    }
    trap cleanup_fn EXIT INT TERM
    "original_command" "--from-makeflow"

Then the outer layer, `temp_1.sh`, would be

    #!/bin/sh
    export LD_PRELOAD=whatever.so
    ./temp_0.sh

These two scripts are added as inputs of the batch task, and the task's command is set to `./temp_1.sh`. The goal is to create an organized, uniform way to create these wrapper scripts (which is currently done in a number of places in Makeflow). Both the JSON command spec mentioned here and the API that would be used from a Makeflow hook offer equivalent functionality. The JX handling here is also a good example for module writers on how to nest these wrapper scripts.

batch_job/src/batch_wrapper.c
```diff
@@ -0,0 +1,324 @@
...
+			goto FAIL;
+		}
+	}
+
+	if (commands != 1) {
+		debug(D_NOTICE|D_BATCH, "a command is required to generate a wrapper");
+		goto FAIL;
+	}
+
+	result = batch_wrapper_write(w, t, "./wrapper_");
```
nhazekam: Can we parse out a wrapper name instead of just using wrapper? I think wrapper is an adequate default, but would prefer being able to name it something more descriptive if possible.

PR #1832, opened 2018-01-24
condor scratch may not be accessible inside a container

btovar:
As we found out using singularity, sometimes _CONDOR_SCRATCH_DIR points to a directory that is not accessible outside the container. This causes trouble with the work_queue_worker.

This pr makes it so that the work_queue_worker and starch only use a temp directory if that directory is accessible.


dthain:
Makes sense to me.

dthain:
@Nekel-Seyew Please talk with Ben so that you understand what's going on here.  This may or may not fix your problem with runos.  (And in any case, you could avoid the bug by unsetting some environment variables.)

PR #1838, opened 2018-01-31
K8s release v3

charleszheng44:
Merge the latest cctools (commit c29b87cd7f729148b9785f7734da2e5e2f3cd1cc)


batch_job/src/work_queue_factory.c
```diff
@@ -344,17 +346,30 @@ static int submit_worker( struct batch_queue *queue )
...
-		worker,
-		master_host,
-		master_port,
-		worker_timeout,
-		catalog_host,
-		password_file ? "-P pwfile" : "",
-		resource_args ? resource_args : "",
-		extra_worker_args ? extra_worker_args : ""
-		);
+		if(k8s_worker_image) {
```
dthain: I see what you are trying to here, but structuring the code this way is going to lead to confusion down the road and difficulty in merging, since these if-branches are practically identical.

How about just this instead:

```
if( kubernetes mode ) {
   worker = "work_queue_worker"
}
```

And then leave the existing string_format as-is.

batch_job/src/batch_job_mesos.c
```diff
@@ -81,13 +81,28 @@ static void start_mesos_scheduler(struct batch_queue *q)
...
 		char *ld_preload_str = NULL;
+		char *python_path = NULL;
 	
 		if(mesos_preload) {
 			ld_preload_str = string_format("LD_PRELOAD=%s", mesos_preload);
-		} 
-		char *envs[] = {ld_preload_str, NULL};
-	
-		char *mesos_python_path = xxstrdup(mesos_py_path);
+		}
```
dthain: Is this fixing a bug in mesos only, or is this related to Kubernetes somehow?

charleszheng44: It is only related to Mesos. We used to setup Mesos on disc cluster, which requires users to specify the 
share library to use Mesos. So I made `ld_preload` a mandatory flag. When I tried to setup Mesos on AWS, I can expose all Mesos libs to the `PATH`, which does not require me to pass any args to `ld_preload`. So I think it is more reasonable to make it an Optional flag

batch_job/src/work_queue_factory.c
charleszheng44: Yes. It is more reasonable to change the value of  `char *worker` .

charleszheng44:
@dthain I PR should be ready to be merged

PR #1839, opened 2018-01-31
Hook register

nhazekam:
Updates the register functionality to allow flexibility about adding hooks and duplicate hooks.

### DIFF
--- a/makeflow/src/makeflow_hook.c
+++ b/makeflow/src/makeflow_hook.c
@@ -33,38 +33,38 @@ struct list * makeflow_hooks = NULL;
 
 struct dag_file *makeflow_hook_add_input_file(struct dag *d, struct batch_task *task, const char * name_on_submission_pattern, const char * name_on_execution_pattern)
 {
-    char *id = string_format("%d",task->taskid);
-    char * name_on_submission = string_replace_percents(name_on_submission_pattern, id);
-    char * name_on_execution = string_replace_percents(name_on_execution_pattern, id);
+	char *id = string_format("%d",task->taskid);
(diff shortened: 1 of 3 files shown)


makeflow/src/makeflow_hook.c
```diff
@@ -33,47 +33,68 @@ struct list * makeflow_hooks = NULL;
...
+		list_first_item(makeflow_hooks);
+		while((h = list_next_item(makeflow_hooks))){
+			if(!strcmp(h->module_name, hook->module_name)){
+				rc = MAKEFLOW_HOOK_SKIP;
+				break;
+			}
+		}
+	}
+
+	if(rc == MAKEFLOW_HOOK_SUCCESS){
```
trshaffer: Would we want to check for failure here? `MAKEFLOW_HOOK_SKIP` should fall through, but modules might want to return failure and abort Makeflow if there's a very serious incompatibility.

makeflow/src/makeflow_hook.h
```diff
@@ -406,7 +435,7 @@ struct dag_file * makeflow_hook_add_output_file(struct dag *d, struct batch_task
  Example of use see above.
 @param hook The new hook to register.
 */
-void makeflow_hook_register(struct makeflow_hook *hook);
+int makeflow_hook_register_hook(struct makeflow_hook *hook);
```
trshaffer: Could this be shortened to `makeflow_hook_register`?

makeflow/src/makeflow_hook.c
nhazekam: This was addressed, git hub did not detect to difference as the change is several lines later.

PR #1842, opened 2018-02-01
Hook storage allocation

nhazekam:
Migrates the storage allocation code out of Makeflow main into a module. This also removes several unnecessary hooks.

A change of note is that this continues moving all exits success or failure to using gotos and properly cleaning up. As a result some logging events needed to check for null to avoid segfaults. Let me know if this is an unacceptable change.

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -1309,6 +1309,8 @@ int main(int argc, char *argv[])
 	struct jx *jx_args = jx_object(NULL);
 	
 	struct jx *hook_args = jx_object(NULL);
+	extern struct makeflow_hook makeflow_hook_example;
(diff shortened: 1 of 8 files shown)


makeflow/src/makeflow.c
```diff
@@ -1812,14 +1734,25 @@ int main(int argc, char *argv[])
 				list_push_head(shared_fs_list, xxstrdup(optarg));
 				break;
 			case LONG_OPT_STORAGE_TYPE:
-				storage_type = atoi(optarg);
+				{
+					extern struct makeflow_hook makeflow_hook_storage_allocation;
```
btovar: I would just make this declaration once at the top of this file.

PR #1846, opened 2018-02-05
Improved list structure

trshaffer:
As a quality-of-life improvement, I wanted to rework the basic list structure used internally throughout CCTools. The main motivation was the way iteration is handled, specifically, that there's global iteration state on each list. If there is any chance that a function called during iteration might also try to look at the original list, strange things ensue. I try to use JX lists, or write as little code as possible that touches `struct list` stuff. @nhazekam makes copies of lists before looking at them. @btovar mentioned running into a bug in Work Queue that caused workers to drop off because of unexpected list iteration behavior.

This PR replaces `struct list` with a new core list structure built around iteration. This structure is more complicated that the existing basic linked list structure, so I wrote it as a minimal core API that's easier to reason about and test. There's documentation in the header, but this core API is low-level and very focused on iteration. I reimplemented the existing list functions as convenience functions on top of this new API. This new list implementation works as a drop-in replacement for the existing code. Going forward, I would like to avoid the current list iteration style.

    list_first_item(list);
    while ((item = list_next_item(list))) {
        // do something
    }

This is how iteration looks under this PR.

    struct list_cursor *cur = list_cursor_create(list);
    for (list_seek(cur, 0); list_get(cur, &item); list_next(cur)) {
        // do something
    }
    list_cursor_destroy(cur);

Note that this is safe even with recursion or calls to other functions that iterate over the same list. This ended up being a pretty substantial PR, so it needs plenty of testing. I wrote a unit test to verify core behavior, and ran Valgrind to make sure there are no leaks. Since this PR uses the new structure under the existing list functions, all existing tests need to pass, too.


dthain:
I like the new API, it is definitely cleaner than what came before.
But, a thread-safe interface doesn't mean you have permission to multi-thread things.  :)

dttools/src/list.c
```diff
@@ -1,406 +1,633 @@
...
+	assert(item->refcount > 0);
+	if (--item->refcount == 0) {
+		if (item->prev != item)
+			list_item_unref(item->prev);
+		list_item_unref(item->next);
+		free(item);
 	}
+}
+
+static void list_cursor_relax(struct list_cursor *cur) {
```
btovar: What's this function about?

dttools/src/list.c
```diff
@@ -1,406 +1,633 @@
...
+	assert(cur);
+	if (!list_cursor_moved(cur)) return;
 
-	if(!bottom->head) {
-		list_delete(bottom);
-		return top;
+	struct list_item *old = cur->target;
+	struct list_item *tmp = old;
+	while (tmp && tmp->prev == tmp) {
```
btovar: Isn't this a cycle?

dttools/src/list.c
```diff
@@ -1,406 +1,633 @@
...
 
-struct list *list_splice(struct list *top, struct list *bottom)
-{
-	if(!top->head) {
-		list_delete(top);
-		return bottom;
+static void list_item_unref(struct list_item *item) {
+	if (!item) return;
+	assert(item->refcount > 0);
+	if (--item->refcount == 0) {
```
btovar: I rather have this as two statements. First decrement, then the if.

dttools/src/list.c
```diff
@@ -1,406 +1,633 @@
...
+
 #include "list.h"
 
-#include <stdlib.h>
+struct list {
+	unsigned refcount;
+	unsigned length;
+	struct list_item *head;
+	struct list_item *tail;
+	struct list_cursor *iter;
```
btovar: Add a comment that this the global iterator, used for backwards compatibility, etc.

dttools/src/list.c
```diff
@@ -1,406 +1,633 @@
...
+#include <math.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
 #include "list.h"
 
-#include <stdlib.h>
+struct list {
+	unsigned refcount;
```
btovar: Comment why refcount is important now, ref count controlled by the cursors, etc.

dttools/src/list.c
```diff
@@ -1,406 +1,633 @@
...
+	unsigned length;
+	struct list_item *head;
+	struct list_item *tail;
+	struct list_cursor *iter;
+};
+
+struct list_item {
+	unsigned refcount;
```
btovar: I'm not sure I like refcounts also for items. It will get messy when two cursors try to modify the list structure at the same time. 
If modifying the list is important, we can have a unique writing cursor. A list may have only one such cursor.

dttools/src/list.c
```diff
@@ -1,406 +1,633 @@
...
-{
-	struct list_node *n;
+int list_iterate_reverse(struct list *l, list_op_t operator, const void *arg) {
+	void *item;
 	int alltheway = 1;
 
-	for(n = l->tail; n; n = n->prev) {
-		if(!operator(n->data, arg)) {
+	struct list_cursor *cur = list_cursor_create(l);
+	for (list_seek(cur, -1); list_get(cur, &item); list_prev(cur)) {
```
btovar: why -1?

dttools/src/list.c
trshaffer: So switching over to Rust's ownership model? :) I wouldn't mind having a `list_cursor_mut`, and it would simplify book keeping

dttools/src/list.c
trshaffer: This comes back to the per-item refcounts. `list_drop()` flags items as dead by tying their `->prev` pointer back to themselves and removes a reference. Rather than keeping a list of pointers to update, this function gets called whenever someone touches a pointer. It checks if the item is dead and fixes things up. Changing to 1 writer xor n readers would eliminate all this.

dttools/src/list.c
trshaffer: Marker for dead items

dttools/src/list.c
trshaffer: Iterating in reverse. Negative indices are taken from the back a la Python, so index -1 is the tail, and the cursor moves by `list_prev()`.

trshaffer:
@nhazekam and I have been talking about how to handle multiple readers and writers, and agree that *1 writer* xor *n readers* is an improvement over the per-item refcounts in this PR. I'm not sure how we should carry this out, though.

Type Level
---------------

Introduce another cursor type, `list_cursor_mut`. We enforce the invariants that a writer cannot be created while readers exist, and readers cannot be created while a writer exists. The functions that modify the list would take a mutable cursor. We provide two functions (a writer version and a reader version) whenever an API action could work for either, such as moving around. Alternatively, we could have `cursor_upgrade()`/`cursor_downgrade()` that let you change modes only when you need to. This is nice in that it's very clear at the type level what is allowed by the API. It's also more verbose, since you need to either upgrade/downgrade all the time, or duplicate parts of the API.

Refcount-based Enforcement
-------------------------------------

Keep a refcount on the list tracking the number of cursors. There's no need for different types of cursors, write operations just fail if there is more than one cursor out there. This implicit handling is less noisy than type level distinctions. It's also less clear about what's required when calling a function, and might make it difficult to figure out whose fault it is when things go wrong.

------------

Since this is low-level plumbing, I'm inclined to go with the type level approach. I think @nhazekam was leaning toward implicit because of the added noise with upgrading/downgrading cursors all the time.

Thoughts? @dthain @btovar

dthain:
@trshaffer I think you are going way too deep on this.
What is the concurrency problem you are trying to solve?
Can it be adequately addressed by simply having one user of the list at a time?

btovar:
The problem is when we try to iterate the list more than once at the same time (say a O(n2) where we compare each item to each other). Then there is more than one user at once. In the same function this is not so bad, but it gets ugly when the list is an argument to different functions (a list_firstitem that resets another one.)

That said, I think multiple readers is a plus, but I am not convinced even that we need one writer. I rather just modify the list with push/pop/insert_priority. On a list, anything more fancy than that gets difficult to debug.

dthain:
Try to find the simplest solution that addresses our needs...

trshaffer:
I simplified the bookkeeping to partially address @btovar's concerns, but didn't go into the more complicated ownership stuff. I absolutely a structure that supports concurrent access from multiple callers. I also like the addition of linked list functionality. The existing queue interface still works as before, so @btovar can continue to use it.

btovar:
Let's preserve list_push/pop_priority, creating a temp cursor inside them.

trshaffer:
After talking with @btovar, `list_push_priority()` should be OK. If there are no other objections, I'd like to get this merged in. We just had to fix some weirdness in Makeflow related to the list of hooks, so it'd be nice to start using this API.

PR #1851, opened 2018-02-07
Detect magic symlinks under /proc and don't try to follow them

### LINKED ISSUE #1847: Parrot issues when using bash process substitution
I've been trying to use a bash script that uses bash process substitution in a similar way to this:

```
#! /bin/bash
while read input
do
  echo "$input"
done < <(echo "Hello world")
```
This is what I get when running the script above with and without parrot:

```
$ ./parrot_test.sh 
Hello world
$ ./afs/crc.nd.edu/group/ccl/software/x86_64/redhat7/cctools/current/bin/parrot_run ./parrot_test.sh 
cat: /dev/fd/63: No such file or directory
```

A simpler way to test this is:
```
-bash-4.2$ bash -c 'cat <(date)'
Mon Feb  5 16:56:11 EST 2018
-bash-4.2$ /afs/crc.nd.edu/group/ccl/software/x86_64/redhat7/cctools/current/bin/parrot_run -- bash -c 'cat <(date)'
cat: /dev/fd/63: No such file or directory
-bash-4.2$ 
```

trshaffer:
Fixes #1847

This is basically the same issue as #1559. Rather than adding more special cases as we find them, this PR tries to detect any occurrences of magic symlinks. These symlinks have similar-looking targets, so when in `/proc` Parrot won't follow this kind of symlink. I don't think this will cause issues in not following normal symlinks that happen to have this form. Since they're always relative, there would have to be entries under `/proc` with unusual filenames (containing `:`, `[`, and `]`). Everything under `/proc` is managed by the kernel, so I think this should be a more reliable approach.

### DIFF
--- a/parrot/src/pfs_table.cc
+++ b/parrot/src/pfs_table.cc
@@ -421,6 +421,7 @@ void pfs_table::follow_symlink( struct pfs_name *pname, mode_t mode, int depth )
 		 * to follow symlinks of this form.
 		 */
 		if (in_proc && string_match_regex(link_target, "^[a-z]+:\\[[0-9]+\\]$")) return;
+		if (in_proc && string_match_regex(link_target, "^anon_inode:\\[?[a-zA-Z_0-9]+\\]?$")) return;
 
 		const char *basename_start = path_basename(pname->logical_name);
 		size_t dirname_len = basename_start - pname->logical_name;
--- a/parrot/src/pfs_table.cc
+++ b/parrot/src/pfs_table.cc
@@ -395,24 +395,33 @@ void pfs_table::follow_symlink( struct pfs_name *pname, mode_t mode, int depth )
 	char link_target[PFS_PATH_MAX];
 	char link_parent[PFS_PATH_MAX];
 	struct pfs_name new_pname = *pname;
+	int in_proc = false;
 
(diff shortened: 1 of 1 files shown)


dthain:
@trshaffer could you elaborate with some examples?  I can see cases where /proc does have valid symlinks, and this might break it in the opposite direction.  (for example, /proc/self/root)

trshaffer:
This PR doesn't skip *all* symlink handling under `/proc`. Cases like `/proc/self/root`, `/proc/self/exe`, etc. are currently handled correctly. These are symlinks out to real entries in the filesystem, and are always represented as absolute paths. There are also some valid relative symlinks in `/proc`. The magic symlinks of #1847 and #1559, however, never point to valid entries in the filesystem. From `proc(5)`,

    For file descriptors for pipes and sockets, the entries will
    be symbolic links whose content is the file type with the
    inode.  A readlink(2) call on this file returns a string in
    the format:

        type:[inode]

    For example, socket:[2248868] will be a socket and its inode
    is 2248868.  For sockets, that inode can be used to find more
    information in one of the files under /proc/net/.

The namespace symlinks follow a similar format. These are always dangling relative symlinks. Since there are no filenames under `/proc` of the form `type:[inode]`, we should be able to stop following in this case to fix both issues.

dthain:
Ok, that makes sense, merge if you are satisfied.

PR #1560, opened 2016-12-06
Makeflow Support for AWS Lambda

ConradBailey:
Just like the commit says. If you want to get it up and running yourself the documentation and links therein should walk you through setting things up. batch_job_lambda_setup.sh and batch_job_lambda_teardown.sh are used to automate the process of creating and destroying the necessary AWS environment. I'm not sure if those are good names, or generally how to deal with them, but they were suggested by Dr. Thain in our research meetings. As of right now every job uploads all of its input files every time because within batch_job_submit one cannot distinguish immutable inputs; in our meetings we thought this may be a deficiency in the API.

PR #1852, opened 2018-02-08
Makeflow Amazon Lambda Support

dthain:
Merged Conrad's branch with master, making a new PR to fix up a few issues for consistency.


dthain:
Supersedes PR #1560

dthain:
@nhazekam @Nekel-Seyew please take a look and see if anything is amiss.

makeflow/src/makeflow_lambda_setup
```diff
@@ -0,0 +1,225 @@
...
+	exit 1
+fi
+
+# Create the role that will be attributed to the Lambda function. This
+# gives the Lambda function 'permission' to access S3 and to execute
+
+echo -n "Creating role policy file $ROLE_POLICY_FILE..."
+
+# The skeleton of the role, to which we will attach the policies
+cat > "$ROLE_POLICY_FILE" <<- EOF
```
Nekel-Seyew: The user needs permissions to create the roll and add it to themselves. Additionally, should the user already have these privileges before they can access the tools? That way if it's a student, the professor knows who has what permissions, and we're not adding powers to the users that they're not cleared to have.

dthain: AFAIK, no special privileges are required b/c this is creating a fresh role unrelated to exisiting roles.  The user is simply granting to this function the ability to access this particular S3 bucket, so it's all self contained, I think.

Maybe see if you can run it yourself and run into any problems.

dthain:
@nkremerh we need someone with a fresh account that can try this out from a new user perspective.
Please check out this branch, read the (brief) docs about Amazon Lambda support, set up an AWS account, and then see if you are able to get a simple makeflow to run by following the instructions.

makeflow/src/makeflow_lambda_setup
dthain: Ah, I didn't read Conrad's code closely enough.  This creates a policy that let's the function access anything in S3 (that the user can access).  It should instead just grant access to a single bucket, e.g.:

https://aws.amazon.com/blogs/security/writing-iam-policies-how-to-grant-access-to-an-amazon-s3-bucket/

dthain:
@nkremerh please try again and see if this solves your python issue.

dthain:
@Nekel-Seyew take a look at how I downgraded from json to nvpair in the setup/cleanup/batch modules.  This way, the user doesn't require a specific version of python to make things work.  (@nkremerh crashed due to the usual python 2/3 conflicts.)  You probably want to do this in Amazon Batch as well.

nkremerh:
@dthain Ran BWA again with the changes, worked perfectly.

Nekel-Seyew:
Ah, didn't think about that. The json module normally works pretty well. Ok, will look into changing that.

Issue #1850, opened 2018-02-07
When using -p, parrot sets the HTTP_PROXY variable


btovar:
We use -p to pass the proxy information to libcvmfs. However, we also set HTTP_PROXY.

This may cause issue for programs running inside parrot. For example:

% parrot_run -p ";DIRECT" bash
... run my program ...
[DEBUG] 2018-02-07 10:54:00,592 vc3-glidein.download_tarball(): [Errno url error] invalid proxy for http: ';:DIRECT'

I am inclined not to set the HTTP_PROXY, or to set something like PARROT_HTTP_PROXY.

PR #1853, opened 2018-02-08
do not set HTTP_PROXY if using -p, instead set PARROT_HTTP_PROXY

btovar:
With regards to #1850

### DIFF
--- a/parrot/src/pfs_main.cc
+++ b/parrot/src/pfs_main.cc
@@ -1178,7 +1178,8 @@ int main( int argc, char *argv[] )
 	}
 
 	if (http_proxy)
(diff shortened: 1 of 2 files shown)


dthain:
Careful, CVMFS tests are failing on this PR

btovar:
I was afraid of that. It was likely some other code assumed HTTP_PROXY was set.

btovar:
Actually only in rh5. I wonder if it is a transient.

btovar:
I think I know. Since we did not use rh5 in autobuild for a long time, the libcvmfs used there is a little behind and different from the other platforms.

btovar:
Now rh5 passes the cvmfs tests.

PR #1854, opened 2018-02-08
Hook sandbox

nhazekam:
Added sandbox option to Makeflow.

### DIFF
--- a/makeflow/src/makeflow_module_sandbox.c
+++ b/makeflow/src/makeflow_module_sandbox.c
@@ -22,12 +22,15 @@
 
 static int makeflow_module_sandbox_node_submit(struct dag_node *node, struct batch_task *task){
 	struct batch_wrapper *wrapper = batch_wrapper_create();
+	char *cmd = string_format("task_%d", task->taskid);
+	batch_wrapper_prefix(wrapper, cmd);
+	free(cmd);
 
 	/* Save the directory we were originally working in. */
 	//batch_wrapper_pre(wrapper, "export CUR_WORK_DIR=$(pwd)");
 
 	/* Create sandbox. This should probably have a hex or random tail to be unique. */
-	char *cmd = string_format("mkdir task_%d_sandbox", task->taskid);
+	cmd = string_format("mkdir task_%d_sandbox", task->taskid);
 	batch_wrapper_pre(wrapper, cmd);
 	free(cmd);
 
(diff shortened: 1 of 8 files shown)


makeflow/src/makeflow_module_sandbox.c
```diff
@@ -0,0 +1,90 @@
...
+static int makeflow_module_sandbox_node_submit(struct dag_node *node, struct batch_task *task){
+	struct batch_wrapper *wrapper = batch_wrapper_create();
+	char *wrap_name = string_format("task_%d_sandbox", task->taskid);
+	batch_wrapper_prefix(wrapper, wrap_name);
+
+	/* Save the directory we were originally working in. */
+	batch_wrapper_pre(wrapper, "export CUR_WORK_DIR=$(pwd)");
+
+	/* Create sandbox. This should probably have a hex or random tail to be unique. */
+	char *cmd = string_format("mkdir %s", wrap_name);
```
trshaffer: This is probably fine, but it wouldn't be too hard to change this line to

    char *cmd = string_format("export SANDBOX=$(mktemp -d %s_XXXXXX)", wrap_name);

Then subsequent commands could just refer to `$SANDBOX`.

makeflow/src/makeflow_module_sandbox.c
```diff
@@ -0,0 +1,90 @@
...
+		batch_wrapper_pre(wrapper, cmd);
+		free(cmd);
+	}
+	/* Enter into sandbox_dir. */
+	cmd = string_format("cd %s", wrap_name);
+	batch_wrapper_pre(wrapper, cmd);
+	free(cmd);
+
+	/* Execute the previous levels commmand. */
+	cmd = string_format("sh -c \"%s\"", task->command);
```
trshaffer: You don't need `sh -c` here, as it's already in a shell script and `batch_wrapper_cmd()` takes care of quoting

makeflow/src/makeflow_module_sandbox.c
```diff
@@ -0,0 +1,90 @@
...
+	batch_wrapper_cmd(wrapper, cmd);
+	free(cmd);
+
+	/* Once the command is finished go back to working dir. */
+	batch_wrapper_post(wrapper, "cd $CUR_WORK_DIR");
+
+	list_first_item(task->output_files);
+	while((f = list_next_item(task->output_files))){
+		/* Copy out results to expected location. */
+		cmd = string_format("cp %s/%s %s", wrap_name, f->inner_name, f->inner_name);
```
trshaffer: If one of the outputs didn't get created, this will fail and prevent the subsequent cleanup steps. These should be postfixed with `|| true` to ensure the script doesn't die

Issue #1857, opened 2018-02-14
Resource Limits for Torque


dthain:
@nhazekam please take a break from the makeflow re-write to work on this one:

Mailing list thread:
https://groups.google.com/forum/#!topic/cctools-nd/VfLsGOM_nd4

Please fix up the Makeflow-Torque driver to generate the right options corresponding to the resources specified by a job.  For example, `MEMORY=16 CORES=8` should yield `-l nodes=1:ppn=8;mem=16mb`  Since we know that Torque options can vary, allow the user to override it: if `BATCH_OPTIONS` is set, it should totally overwrite the string.

This should be straightforward, please put out a fix today so it can be tested..

stemangiola:
Hello,

currently if I write 

```
MEMORY=**8000**
CORES=**2**
export BATCH_OPTIONS="-l nodes=1:ppn=2,mem=8gb"
<JOB>

MEMORY=**32000**
CORES=**16**
export BATCH_OPTIONS="-l nodes=1:ppn=2,mem=8gb"
<JOB>
```

I always get the first (or maybe the last, my makeflow file is long) BATCH_OPTIONS executing all the time.

nhazekam:
Hi @stemangiola 

I am glad you have given it a try. Can you try it without the export `BATCH_OPTIONS`?  The Memory and cores should be translated into the command without `BATCH_OPTIONS` now.

Also can you run makeflow with debug options? 
`makeflow <other options> -d all -o makeflow.debug`
and send this to me at nhazekam@nd.edu

nhazekam:
Also, I am making changes now that only specify resources when you are not using BATCH_OPTIONS.

stemangiola:
Hello,

I will update the repository and install it again. I will let you know.

Thanks!

Stefano Mangiola



*Data analyst (MSc, MPhil, PhD candidate) @ Melbourne University *

*Hovens lab - Papenfuss lab*
*Royal Melbourne Hospital - WEHI*
*+61 (0)450135954*

2018-02-20 13:40 GMT+11:00 Nick Hazekamp <notifications@github.com>:

stemangiola:
Hello,

I have the same error as before

Here the log file

`2018/02/21 12:30:55.57 makeflow[2837] makeflow: COLON
2018/02/21 12:30:55.57 makeflow[2837] makeflow: LITERAL: benchmark/pure-N52_ARNA_input.RData
[...]

PR #1858, opened 2018-02-14
Add resource parsing to Torque

nhazekam:
Should address #1857. Untested by virtue of not having access to a Torque cluster. @btovar please take a look.

### DIFF
--- a/batch_job/src/batch_job_cluster.c
+++ b/batch_job/src/batch_job_cluster.c
@@ -109,7 +109,7 @@ static char *cluster_set_resource_string(struct batch_queue *q, const struct rms
 		char *mem = string_format(",mem=%" PRId64 "mb", resources->memory);
 		char *disk = string_format(",file=%" PRId64 "mb", resources->disk);
 		cluster_resources = string_format(" -l nodes=1:ppn=%" PRId64 "%s%s ", 
-			resources->cores ? resources->cores : 1,
+			resources->cores>0 ? resources->cores : 1,
 			resources->memory>0 ? mem : "",
 			resources->disk>0 ? disk  : "");
 		free(mem);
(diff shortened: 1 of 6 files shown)


dthain:
The idea looks about right to me, but the main thing is to run it by the person who is experiencing the problem...

nhazekam:
This only touches a single file and should be able to be cherry picked to wherever we want. I don't know if testing this on master is necessarily the best option. @btovar is there a prefered branch I should base this off?

nhazekam:
Rudimentary testing was done to make sure that changes to the command line were being made, -dall does print out the updated line with resources. 

I agree it is just up to the user for testing at this point.

dthain:
Please reach out and invite him to try it.

btovar:
Looks good to me, assuming that arguments can be repeated and the last overrides the previous. Let's wait to see when it is tested in a torque system.

nhazekam:
@btovar Can you please take a look? I am still going to add a note about what options are used and why, but the functional code is there.

batch_job/src/batch_job_cluster.c
```diff
@@ -99,6 +100,40 @@ static int setup_batch_wrapper(struct batch_queue *q, const char *sysname )
...
+static char *cluster_set_resource_string(struct batch_queue *q, const struct rmsummary *resources)
+{
+	char *cluster_resources = NULL;
+	const char *ignore_mem = hash_table_lookup(q->options, "ignore-mem-spec");
+
+	if(q->type == BATCH_QUEUE_TYPE_TORQUE || q->type == BATCH_QUEUE_TYPE_PBS){
+		char *mem = string_format(",mem=%" PRId64 "mb", resources->memory);
+		char *disk = string_format(",file=%" PRId64 "mb", resources->disk);
+		cluster_resources = string_format(" -l nodes=1:ppn=%" PRId64 "%s%s ", 
+			resources->cores ? resources->cores : 1,
```
btovar: Probably we want a > 0 here too, as resources->cores may be -1.

doc/makeflow.html
```diff
@@ -410,14 +410,36 @@ <h2>Table of Contents</h2>
 
 <p>Use the <tt>-T pbs</tt> option to submit jobs to the <a href=http://www.arc.ox.ac.uk/content/pbs>Portable Batch System</a> or compatible systems.</p>
 
+<p>This will add the values for cores, memory, and disk. These values will be added onto <tt>qsub</tt> in this format:<br>
+<tt>-l nodes=1:ppn=$CORES,mem=$MEMORY,file=$DISK</tt></p>
```
btovar: should it be 

mem=${MEMORY}mb  (same for file).

We also want to be explicit in that measurements are in MB.

PR #1859, opened 2018-02-14
Hook singularity

nhazekam:
Migrate the singularity wrapper to a makeflow module that utilizes batch_wrappers.


makeflow/src/makeflow_module_singularity.c
```diff
@@ -0,0 +1,68 @@
...
+
+	return MAKEFLOW_HOOK_SUCCESS;
+}
+
+static int node_submit(struct dag_node *n, struct batch_task *t){
+	struct batch_wrapper *wrapper = batch_wrapper_create();
+	batch_wrapper_prefix(wrapper, CONTAINER_SINGULARITY_SH);
+
+	/* Assumes a /disk dir in the image. */
+	char *cmd = string_format("singularity exec --home $(pwd) --bind $(pwd):/disk %s %s", singularity_image, t->command);
```
Nekel-Seyew: If you bind the cwd to /disk inside the image, part of the command should be cd-ing into /disk when the container starts. That's why I generally say `singularity exec --blah --blah img sh -c "<actual command>"`

Issue #1855, opened 2018-02-12
Another parrot bug with process substitution and redirection


khurtado:
Hi, when I try to run script in [1] in a wrapper [2], with parrot and nohup + redirecting the stdout/err as in:

```
nohup ./parrot_cvmfs_static_run ./bug_wrapper.sh>nohub1 2>&1 &
```
I get the dump of some SQLite 3 stuff rather than the "Hello" at the end of the script:

```
nohup: ignoring input
LibCvmfs version 2.4, revision 25
SQLite format 3
ø^Qßâ<95>­MÇ»^Y^HÛ^M<91>^Q<8d>^B
...
```
Some notes:
1. Not using nohub+redirection actually makes the script work. 
2. Running `bug.sh` without the wrapper, even with nohub+redirection also seems to work.
I tried to isolate the issue as much as possible. This should be reproducible without any CMS credential.

[1]
```
#!/bin/bash

for FS in \
   oasis.opensciencegrid.org \
   singularity.opensciencegrid.org \
; do
    if [ -e /cvmfs/$FS/. ]; then
        REV_VAL=`/usr/bin/attr -q -g revision /cvmfs/$FS/. 2>/dev/null`
    fi
done

TS_VAL=`(cat /cvmfs/oasis.opensciencegrid.org/osg/update.details  | egrep '^Update unix time:' | sed 's/.*: //') 2>/dev/null`

source "/cvmfs/cms.cern.ch/cmsset_default.sh"

tmpname=`mktemp --tmpdir=$PWD installed_cms_software.XXXXXX`
SCRAM_ARCH="slc6_amd64_gcc481"
scramv1 list -c CMSSW | grep CMSSW | awk '{print $2}' | sort | uniq | grep ^CMSSW >> $tmpname
rm $tmpname

# Do process substitution
while read line
do
    echo $line
done < <(echo "Hello")
```
[2]
```
#! /bin/bash

./bug.sh
```

trshaffer:
This was a tricky bug to track down, as it is only likely to show up when using bash process substitution and CVMFS. Big thanks to @khurtado for finding a case that reliably reproduces the issue!

[...]

PR #1860, opened 2018-02-14
Rewrite /proc/self and /dev/fd before following symlinks

trshaffer:
For #1855 

This doesn't fix the issue in general, so we would still see buggy behavior with symlinks that have `/proc/self` as an intermediate path component. This PR just adjusts existing path resolution that handles `/dev/fd` as a special case. I think this resolution order is more correct than what's there now.


khurtado:
@trshaffer Thanks a lot! 
@btovar : Could you make another static binary for parrot with this PR?

btovar:
@khurtado: It is ready, in the usual location.

trshaffer:
@khurtado reports that this closes #1855, so merging

PR #1863, opened 2018-02-15
hardcoded vc3-builder location fix for prof lannon

Nekel-Seyew:


Nekel-Seyew:
@btovar  I've fixed it Sorry!

batch_job/src/work_queue_factory.c
```diff
@@ -1316,10 +1316,10 @@ int main(int argc, char *argv[])
 	
 	
 	if(runos_os) {
-		cmd = string_format("cp \"$(which vc3-builder)\" '%s'", scratch_dir);
+		cmd = string_format("cp '/afs/crc.nd.edu/group/ccl/software/vc3-builder' '%s'", scratch_dir);
```
btovar: Please update to 
/afs/crc.nd.edu/group/ccl/software/vc3-builder-src/vc3-builder

I already created that location.

btovar: Also use a #define with the path, as with runos.

PR #1864, opened 2018-02-15
Add .clang-format

trshaffer:
I want to autoformat, but I don't want to reformat the **entire** codebase. I'm looking at setting things up to autoformat only the parts of the code I touch while committing. To that end, I'd like to add a `.clang-format` for CCTools. The one in this PR is an [example](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#examples) from the LLVM folks that's "similar to the Linux Kernel style". It looks close enough to our current stuff, and I'd rather avoid coming up with a complicated format that's matched to my taste.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,6 @@
+BasedOnStyle: LLVM
+IndentWidth: 8
+UseTab: Always
+BreakBeforeBraces: Linux
+AllowShortIfStatementsOnASingleLine: false
+IndentCaseLabels: false


btovar:
Tim, as long as nothing happen automatically, and it is only applied to the lines of code you modify, it should be ok (e.g., only after a selection in visual mode in vim). Some time ago we tried autoformating when saving files/committing, and the pain was not worth it.

trshaffer:
Yeah, I still see those commits when I do `git blame`, I don't like that. clang can do per-line autoformatting, so it works well as a git subcommand, to clean up only the changes in your index (stuff you `git add`ed).

PR #1868, opened 2018-02-21
Pull Makeflow command line args from a JSON file

trshaffer:
Following discussion about handling sub-Makeflows, I wanted to add support for specifying Makeflow's command line via JSON rather than a list of strings in `argv`. This PR adds a new option, `--argv`, that gives the path of a JSON file. Makeflow interprets each key:value pair as a long option name and its argument. Both the normal `getopt()` stuff and the JSON options go through the same code path. It's still a little rough around the edges, and I'm writing some tests for this functionality as well.
@nhazekam have a look


dthain:
This is fun, but what problem does it solve?  :)

nhazekam:
This provides a portable and reproducible way to pull options into makeflow. This is not a problem so much at the highest level of invoking a makeflow, but as we discussed passing arguments to sub-makeflows and being concise about how to invoke them this seemed like a flexible, but consistent approach. If/when this is merged the #1808 pull request will be rebased on this and arguments will be parsed into a argv file in JSON format.

dthain:
I want to make sure that we clearly separate two different things:

1 - The "logical arguments" to the *workflow* that specific filenames, scientific parameters, number of splits, etc.  Things that affect the actual output of the dag.  This definitely needs to be JSON so that it can be easily referred to from within the dag itself.  It is expressed as the `jx_context` argument to makeflow, or as a JSON document within a parent makeflow.

2 - The "technical arguments" to the workflow *coordinator* that explain what type of batch system, how many cores to use, where to put the debug file, etc.  These are chosen by the user in the parent DAG, but could conceivably be selected automatically for a child dag.

nhazekam:
This should hold true. The argv only parses information that is set using getopt. Since our commandline options only affect "technical args", argv only affects them as well. You can set where to find jx_context using argv, but not what is in jx_context.

makeflow/src/makeflow.c
```diff
@@ -177,6 +178,74 @@ static struct list *shared_fs_list = NULL;
 
 static int did_find_archived_job = 0;
 
+static struct list *jx_argv = NULL;
+static struct jx *jx_optarg = NULL;
+
+static void push_jx_argv(struct jx *j) {
+	assert(j);
+	if (!jx_argv)
```
btovar: Please use braces, even for single statement ifs. We had problems in the past where what looked like a single statement was a macro expansion, and that was fun to debug.

makeflow/src/makeflow.c
```diff
@@ -177,6 +178,74 @@ static struct list *shared_fs_list = NULL;
...
+	case JX_DOUBLE:
+		return string_format("%g", j->u.double_value);
+	case JX_STRING:
+		return xxstrdup(j->u.string_value);
+	default:
+		return NULL;
+	}
+}
+
+static int makeflow_getopt(int argc, char *const argv[], const char *optstring, const struct option *longopts, int *longindex) {
```
btovar: All these are option related functions seem generic. I would suggest to move them to a jx_options.c in dttools. We can then use them somewhere else (e.g. to not worry about quoting when wrapping the resource_monitor).

makeflow/src/makeflow.c
```diff
@@ -1891,17 +1980,19 @@ int main(int argc, char *argv[])
 
 	makeflow_hook_create(hook_args);
 
-	if((argc - optind) != 1) {
+	if((argc - optind) == 1) {
+		if (dagfile)
+			fatal("only one dagfile can be specified");
```
btovar: Why?

makeflow/src/makeflow.c
nhazekam: Agreed!

makeflow/src/makeflow.c
trshaffer: If multiple dagfiles are specified, I'd rather not choose by some unspecified ordering, and just error out instead.

PR #1869, opened 2018-02-21
Umbrella Module

nhazekam:
Move umbrella support in Makeflow to a module. This is currently compiling, but we are working on testing. To this point have not been able to successfully run umbrella.

Note there are some small changes to other files that seem unrelated, but are used for debugging this hook/wrapper. If there is interest in moving to separate pull, I will do that.


nhazekam:
This will need to be rebased after #2021 for a correct history. #2021 is also need so that umbrella in Makeflow will work.

nhazekam:
@btovar @trshaffer This has been updated and works from master. Let me know if there is anything that needs to be updated/fixed for this to be merged. 

As it stands the examples in the `makeflow_umbrella/README` file work, with the exception of those using wrapper. I removed these as support and behavior are not currently defined and I don't believe there is an active use case for these. If this is not the case, this or a future change can be made to add them and support that behavior.

nhazekam:
@btovar @trshaffer Please take a look when you get the chance. I think this is close/ready to merge.

dthain:
@nhazekam let's get this one finished.

Please rebase on master, then I'll be ready to merge it.

PR #1871, opened 2018-02-21
Hook shared fs

nhazekam:
Move the shared-fs support out of makeflow's main and into a module. This is still being tested, but should have the same behavior as the previously released functionality.

### DIFF
--- a/makeflow/src/makeflow_module_shared_fs.c
+++ b/makeflow/src/makeflow_module_shared_fs.c
@@ -36,7 +36,7 @@ static int prefix_match(void *stem, const void *filename) {
 
 static int batch_file_on_sharedfs( const char *filename )
 {
-    return !list_iterate(shared_fs_list,prefix_match,filename);
+	return !list_iterate(shared_fs_list,prefix_match,filename);
 }
 
 static int create(struct jx *args){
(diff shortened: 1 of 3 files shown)


makeflow/src/makeflow_module_shared_fs.c
```diff
@@ -0,0 +1,191 @@
...
+	shared_fs_list = list_create();
+	shared_fs_saved_inputs  = itable_create(0);
+	shared_fs_saved_outputs = itable_create(0);
+
+	//JX ARRAY ITERATE
+	if(jx_lookup(args, "shared_fs_list")){
+		struct jx *array = jx_lookup(args, "shared_fs_list");
+		struct jx *item = NULL;
+		while((item = jx_array_shift(array))) {
+			list_push_head(shared_fs_list, xxstrdup(item->u.string_value));
```
trshaffer: `jx_array_shift()` gives ownership of the items, so you should delete them after use. Also, this will segfault if someone passes in an array with a non-string member.

makeflow/src/makeflow_module_shared_fs.c
```diff
@@ -0,0 +1,191 @@
...
+	
+	return MAKEFLOW_HOOK_SUCCESS;
+}
+
+static int destroy(){
+	char * fs = NULL;
+
+	list_first_item(shared_fs_list);
+	while((fs = list_next_item(shared_fs_list))){
+		free(fs);
```
trshaffer: `list_free()` does this

makeflow/src/makeflow_module_shared_fs.c
```diff
@@ -0,0 +1,191 @@
...
+}
+
+static int create(struct jx *args){
+	shared_fs_list = list_create();
+	shared_fs_saved_inputs  = itable_create(0);
+	shared_fs_saved_outputs = itable_create(0);
+
+	//JX ARRAY ITERATE
+	if(jx_lookup(args, "shared_fs_list")){
+		struct jx *array = jx_lookup(args, "shared_fs_list");
```
trshaffer: Two lookups?

PR #1883, opened 2018-03-19
Coarse short processes

btovar:
The resource monitor stops processes to measure them just before they exit. For very short children processes, say a system("touch ..."), this adds a noticeable overhead without any real gain in accuracy.

This pr makes it so that very short running processes (by default 0.25s) are not stopped. In all other respects they are accounted in the same way.

To restore the previous behavior, the option --accurate-short-processes has been added.

The overhead was observed in lobster using madgraph, in which a task ran for hours without termination instead of the expected 30 min without the monitor and parrot.  The monitor would get so busy managing the small processes one by one that it would eventually trigger load suspension from condor, and the tasks would be eventually evicted. With this fix + parrot +generating detailed logs, the task ran for about 54 minutes.

(@klannon)


btovar:
Also, the change only applies to children processes. The first parent process is always stopped at the end.

dthain:
This approach makes sense to me.
Add a note to the docs so it doesn't surprise anyone..

PR #1885, opened 2018-03-22
Jx2env

btovar:
Small utility to help in writing shell scripts that need values from a json document. Given keys in a document, it produces export or setenv lines with the corresponding values.

Example document mydoc.json:
```json
{
    "hello":"goodbye",
    "plate": { "fork" : { "knife" : { "spoon" : "tea spoon" } } },
    "tag"  : "it"
}
```
```shell
jx2env  mydoc.json   VARNAME=hello   MYOTHERVARNAME=plate.fork.knife.spoon
export VARNAME="goodbye"
export MYOTHERVARNAME="tea spoon"
```

Usually you would call it inside an eval, as:

```shell
eval $(jx2env mydoc.json   hello plate.fork.knife.spoon)
echo "${VARNAME}"
echo "${MYOTHERVARNAME}
```

This should help with scripts so that we don't have to call python repeatedly to read a value from a json document.

### DIFF
--- a/dttools/src/jx2env.c
+++ b/dttools/src/jx2env.c
@@ -1,0 +1,159 @@
+/*
+  Copyright (C) 2018- The University of Notre Dame This software is
+  distributed under the GNU General Public License.  See the file
+  COPYING for details.
+
+
+  Given a JSON object in a file, it prints...
+
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+
+#include "getopt.h"
+#include "int_sizes.h"
+#include "jx.h"
+#include "jx_parse.h"
+#include "jx_print.h"
+#include "stringtools.h"
+#include "xxmalloc.h"
+
+void show_help(const char *exe) {
+	fprintf(stderr, "Usage:\n%s [--csh] input-file NAME=json.key.path [NAME=json.key.path ...]\n", exe);
+}
+
+static const struct option long_options[] =
+{
+	{"csh", no_argument, 0, 'c'},
+	{0, 0, 0, 0}
+};
+
+void dots_to_underscores(char *spec) {
+	char *c = strchr(spec, '.');
+
+	while(c) {
+		*c = '_';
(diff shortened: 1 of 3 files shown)


dthain:
This is handy to have, but I would prefer to solve the problem by simplifying things away.
For example, see how batch_job_lambda writes config data in plain name-value pairs,
rather than json.  Then, the cleanup script can just use awk/cut.

btovar:
In general, I agree. But like in the script included with the makeflow amazon batch support pr, sometimes the information is given to us in json. @Nekel-Seyew got away with greps and cuts, but that assumes the json is nicely formatted, and that keys are unique across objects in the document.

dthain:
Indeed, this  approach is better than grepping json...

dttools/src/jx2env.c
```diff
@@ -0,0 +1,159 @@
...
+		*next_dot = '\0';
+		next_dot++;
+
+		struct jx *l = jx_lookup_guard(j, spec_mod, &found);
+		char *value;
+
+		if(found) {
+			value = value_of_dotted(l, next_dot);
+		} else {
+			value = xxstrdup("x");
```
trshaffer: Does `"x"` mean something special here?

dttools/src/jx2env.c
```diff
@@ -0,0 +1,159 @@
...
+#include "getopt.h"
+#include "int_sizes.h"
+#include "jx.h"
+#include "jx_parse.h"
+#include "jx_print.h"
+#include "stringtools.h"
+#include "xxmalloc.h"
+
+void show_help(const char *exe) {
+	fprintf(stderr, "Usage:\n%s [--csh] input-file NAME=json.key.path [NAME=json.key.path ...]\n", exe);
```
trshaffer: When I read this, it looks like you're binding a name by path rather than doing the underscore escape

btovar: I am thinking of doing that to avoid name clashes. Let me submit the next commit.

PR #1887, opened 2018-04-04
Update resource monitor module

nhazekam:
Add handling to specify resource monitor and also enable measuring a directory. 

When measuring the directory in a Makeflow task we use $PWD always. This allows
the directory to reflect wherever it is executed and not some assumed name.


btovar:
Looks good.

PR #1888, opened 2018-04-04
Nest hooks

nhazekam:
Add the required structures to nest hooks. This allows for mutliple instance of the same type of hook, but with different arguments and ordering to when they are applied.

General Idea is that each hook has a args list in JX and an Instance struct if needed.
There are 3 cases for hooks nesting:
Not allowed: This means that there can only be one global instance of the hook. Most are like this notably containers and shared_fs.
Separated: Multiple are allowed, but they must have additional hooks between them such as with resource monitor. This allows differentiation of wrapper arguments.
Always nest: There are no arguments so it is always a new instance. We see this with sandbox.

This does touch most of the hooks as I wanted the instances to be well formed and consistent, though strictly speaking it was not necessary as many of them can not be nested. Overall I think it lends to a cleaner organization with less global variables.

This is rebased on #1887 so there are some changes that below to that pull request, but I wanted to push this out for review. I will rebase off master once the other is merged.


makeflow/src/makeflow_hook.c
```diff
@@ -55,45 +72,93 @@ struct dag_file * makeflow_hook_add_output_file(struct dag *d, struct batch_task
...
 
-		list_push_tail(makeflow_hooks, h);
+		debug(D_MAKEFLOW_HOOK, "Hook %s:registered",hook->module_name?hook->module_name:"");
 	} else if(rc == MAKEFLOW_HOOK_FAILURE){
-		debug(D_MAKEFLOW_HOOK, "Hook %s:register failed",h->module_name?h->module_name:"");
-	} else if(rc == MAKEFLOW_HOOK_SKIP){
-		debug(D_MAKEFLOW_HOOK, "Hook %s:register skipped",h->module_name?h->module_name:"");
+		/* Args are NULL to prevent other hooks from modifying an
+		 * unintended arg list. */
+		args = NULL;
```
trshaffer: Did you mean to dereference this (and delete the JX)? This will only clear the local copy of the pointer, not the value.

makeflow/src/makeflow_hook.c
```diff
@@ -55,45 +72,93 @@ struct dag_file * makeflow_hook_add_output_file(struct dag *d, struct batch_task
...
-		list_first_item(makeflow_hooks);
-		while((h = list_next_item(makeflow_hooks))){
-			if(!strcmp(h->module_name, hook->module_name)){
+		rc = hook->register_hook(hook, makeflow_hooks, args);
+	}
+
+	if(!(hook->register_hook) || rc == MAKEFLOW_HOOK_SKIP) {
+		struct list_cursor *cur = list_cursor_create(makeflow_hooks);
+		struct list_cursor *acur = list_cursor_create(makeflow_hook_args);
+		// This should be a reverse traversal so that the most recent
```
trshaffer: I wasn't clear about this comment. Is this a TODO, or a description of how things work now?

nhazekam: Changing to a reverse traversal with a break when we find the value.

Issue #1873, opened 2018-02-27
Allowing edits to the makeflow file "on the go"


stemangiola:
Hello,

while enjoying your makeflow, I am thinking more and more to the limitations due to this feature.

For pipeline developing, is often necessary to insert a new algorithm of combination of parameters for an algorithm within the already executed pipeline. In my case for example, I test an algorithm with a combination of parameters:

a = 1:100
b = 1:10
c="mode1", "mode2", "mode3"

100*10*3 combinations 

It is pretty common after having executed the whole pipeline, to be wanting to test for b = 11

Now,  your team already suggested to produce many makefiles instead of pissing together a high number of commands. Although this can be more modular and robust generally speaking, in a grid of parameters like this (and even more complex bioinfomatics pipelines) deciding where to modularize is non easy, and sometime even not feasible.

To gain momentum and adoption by researchers like me, Makeflow should be able in my opinion (like GNU make) to be robust to changes in the makefile after full or partial execution, simply updating the existing dependencies. 

Some genomics pipelines (is not my case) take 3 weeks to complete (mine takes 2 days), and there cannot be the risk of having to run from the top if a secondary small change is required.

Makeflow is easy to use and almost no brainier for queuing systems. I hope you can consider this request, as I would like to stick to your system, however at the moment is requiring a lot of time and mental energies to go around this hard needs of makeflow.

Thanks.

dthain:
@stemangiola we had a long discussion about this at our team meeting this week.

There are two underlying issues here:

1 - If the goal is to re-run what "changes" then we need to be careful about the whole definition of the job, including the files, the text of the command, and possibly even the relevant environment variables.  Standard make just looks at the modification times of the files, which isn't complete.

2 - The transaction log refers to job by their ordinal position in the makeflow file.  This means that log is no longer valid if any change is made to the contents of the makeflow file.  We need a way of referring to jobs in a way that allows for modification of the file.

To fix this, we should do the following:

1 - Change the makeflow job id from a simple integer into a string in the makeflow core.
2 - Compute the makeflow job id as a "content based identifier" by computing a consistent checksum of the relevant job text.
3 - When recovering the transaction log, assume that the makeflow file may have changed, so it could contain new rules not previously mentioned in the log (or vice versa).

[...]

PR #1893, opened 2018-04-27
Compute a hash identifier for each DAG node

trshaffer:
This PR adds a hash to each `struct dag_node` based on its rule definition. I'm only hashing the following fields.

+ command
+ inputs
+ outputs

The hash is derived based on textual names, not file contents. Thus we can assign hash IDs while parsing. Archive mode would need to take more into account, such as file contents, environment, etc. These hashes are insensitive to ordering, both in the input/output lists and among the rules themselves. Adding a new rule is safe since it produces a fresh hash that's distinct from the existing rules. In addition, modifications to a rule will also result in a fresh hash.

This PR only defines a hashing scheme required for #1873. There are no changes to the log or to Makeflow's behavior. I imagine there will be some discussion about what should be covered by the hash, so I wanted to get this sorted out and merged before moving on.

CC @nhazekam @dthain @btovar

### DIFF
--- a/dttools/src/stringtools.c
+++ b/dttools/src/stringtools.c
@@ -24,6 +24,13 @@ See the file COPYING for details.
 #include <stdlib.h>
 #include <string.h>
 
+int string_compare(const void *p1, const void *p2) {
+	/* The actual arguments to this function are "pointers to
+	 * pointers to char", but strcmp(3) arguments are "pointers
+	 * to char", hence the following cast plus dereference */
+	return strcmp(* (char * const *) p1, * (char * const *) p2);
+}
+
 /*
  * Based on opengroup.org's definition of the Shell Command Language (also gnu's)
  * In section 2.2.3 on Double-Quoted Strings, it indicates you only need to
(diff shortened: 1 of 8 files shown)


btovar:
1. We are making a hash to identify the line in the file, not the rule as an abstract entity, so I think the current hashing is enough.

dthain:
Ok, I agree that it is sufficient for that purpose.  Except that the hash is being used in the archive feature.  :)  So perhaps we need to develop a different hashing function for that purpose.  @nhazekam what do you think?

dthain:
I like your distinction between hashing the batch_task vs hasing the dag_node, that makes a lot of sense.

Re hashing the batch task, I would like the archive directory format to be as transparent as possible so that the user can see what's going on, and perhaps other tools can process the same data without makeflow.  So, md5sum on the task_info file should give the expected hash.  That doesn't work now b/c we load the info into memory, rearrange the data, and then compute the hash.

I think the better solution is to write out the task_info file as exactly the information to be hashed, in the form of a json file that has been canonicalized.

nhazekam:
I think that creating the hash from a json file would work well. Then the info and hashes of the files would also be saved in the json, which is similar to how we are currently hashing. Using our json print function we could easily output the json document, much like is done for task info now in archive.

The only potential danger is that if the json printing function changes for any reason in the future, older archives would be invalid. Not sure if this is likely, by possible.

trshaffer:
This needs re-done/re-examined now that archive mode is in

Issue #1891, opened 2018-04-25
Makeflow Archive Mode Segfault


nhazekam:
As reported on the forums, using archive mode on torque resource branch causes a segfault. 

- Need to see if this is an issue in master (likely).
- If not need to address the changes causing it prior to merging resources.

Command that segfaulted:
$ makeflow -T torque -B "-l nodes=1:ppn=2,mem=4gb" -N bench --max-remote 100 --do-not-save-failed-output --archive=Makeflow_archive Makeflow
parsing Makeflow...
local resources: 4 cores, 15887 MB memory, 202637821 MB disk
max running remote jobs: 100
max running local jobs: 4
checking Makeflow for consistency...
Makeflow has 6186 rules.
recovering from log file Makeflow.makeflowlog...
starting workflow....
Segmentation fault (core dumped)

Command that run normally:
$ makeflow -T torque -B "-l nodes=1:ppn=2,mem=4gb" -N bench --max-remote 100 --do-not-save-failed-output  Makeflow

nhazekam:
I am currently working on this. I'm glad you're interested. I will link to this issue as soon as it's ready for use. It's currently my highest priority.

stemangiola:
Hello,

I received some notifications about updates on github code, however its not clear to me if the issue has been solved and tested?

Thanks!

nhazekam:
Thanks for checking in. The code is working well, but still in testing. We are currently working on the release schedule so that we can have this available to you soon. I will update this both when it is merged with our master branch and when a formal release container the code is created.

stemangiola:
Hello,

this Thursday I will present your Makeflow tool at the meeting for our torque HPC (cluster) infrastructure. I would be glad to talk about the features:

1) tracking of inside changes of pipelines (without the need to rerun the hole makeflow tasks)
2) the ability of makeflow to use locally defined job attributes (e.g., time, memory and cores requrements)

As far as I know none of these two features are available (of fully tested) for torque system.

Can you confirm?

Thanks a lot.

dthain:
@trshaffer can you please work with @stemangiola on this?
@nhazekam is out on an internship this summer.

dthain:
@trshaffer Is this one fixed?

trshaffer:
@dthain The archive mode changes are merged in. We never heard back with new problems, so as far as I know this is resolved.

PR #1898, opened 2018-04-30
Hook archive

nhazekam:
Migrate archive to hook/module.

Updated @piercecunneen code to utilize hook structure.

This is in testing currently. Resolves segfault occurring in original code as reported in #1891 
@stemangiola feel free to give this a try.

Until this is merged into master, there are no guarantees that the hashing will not change.

That said, the hashing should not change going forward, though environment variables may be added later differentiating tasks that do/don't specify environment.

This has been tested with several systems, but more comprehensive testing is underway.


dthain:
With apologies to @nhazekam let's not ask @stemagnolia to try it out until we have a little experience testing and using it within our team.

makeflow/src/makeflow_module_archive.c
```diff
@@ -0,0 +1,503 @@
...
+			return 0;
+		}
+	}
+	list_cursor_destroy(cur);
+
+	return 1;
+}
+
+
+//static int node_submit( void * instance_struct, struct dag_node *n, struct batch_task *t){
```
Nekel-Seyew: dead code

makeflow/src/makeflow_module_archive.c
```diff
@@ -0,0 +1,503 @@
...
+		fprintf(fp, "EXIT_CODE : %d\n", t->info->exit_code);
+		fprintf(fp, "EXIT_SIGNAL : %d\n", t->info->exit_signal);
+	}
+	fclose(fp);
+	free(task_info);
+
+	return 0;
+}
+
+/* writes the file symlink that links to the archived job that created it */
```
Nekel-Seyew: I would update this comment to be more reflective of the function, or change the name of the function, or do both! The function does checksum, yes, but it also does a more, but the checksum isn't really talked about in the comment.

makeflow/src/makeflow_module_archive.c
```diff
@@ -0,0 +1,503 @@
...
+	debug(D_MAKEFLOW_HOOK, "archiving task %d to %s", t->taskid, archive_directory_path);
+
+	int dir_create = 0;
+	/* We create all the sub directories upfront for convenience */
+	dir_create = makeflow_archive_create_dir(archive_directory_path, "/output_files/");
+	dir_create += makeflow_archive_create_dir(archive_directory_path, "/input_files/");
+	dir_create += makeflow_archive_create_dir(archive_directory_path, "/descendants/");
+	dir_create += makeflow_archive_create_dir(archive_directory_path, "/ancestors/");
+
+	if(dir_create){
```
Nekel-Seyew: This is just being pedantic, but I don't think this reads well. It's clever, but I think it needs a better name. Otherwise, it reads as "if dir_created" when it should be "if error_dir_created"

makeflow/src/makeflow_module_archive.c
```diff
@@ -0,0 +1,503 @@
...
+		result = 0;
+		goto FAIL;
+	}
+
+	/* Log the task info in the task directory */
+	if(makeflow_archive_task_info(a, n, t, archive_directory_path)){
```
Nekel-Seyew: same here. I don't think this reads well....

makeflow/src/makeflow.c
```diff
@@ -2153,7 +2160,11 @@ int main(int argc, char *argv[])
 	}
 
 	printf("starting workflow....\n");
-	makeflow_hook_dag_start(d);
+	rc = makeflow_hook_dag_start(d);
+	if(rc != MAKEFLOW_HOOK_SUCCESS){
+		debug(D_ERROR, "Failed dag start hooks");
```
nkremerh: Stylistically, do we want to write DAG in all caps in the error message for clarity? This may not matter since we can assume if a user is reading through the debug log, they probably know what a DAG is.

makeflow/src/makeflow_module_archive.c
```diff
@@ -0,0 +1,472 @@
...
+		fprintf(fp, "EXIT_CODE : %d\n", t->info->exit_code);
+		fprintf(fp, "EXIT_SIGNAL : %d\n", t->info->exit_signal);
+	}
+	fclose(fp);
+	free(task_info);
+
+	return 1;
+}
+
+/* Archive the specified file.
```
trshaffer: If there is an error or failure while archiving a file/task, will it corrupt the archive?

makeflow/src/makeflow_module_archive.c
```diff
@@ -0,0 +1,472 @@
...
+	struct batch_file *f;
+	char *task_info = string_format("%s/task_info", archive_path);
+
+	FILE *fp = fopen(task_info, "w");
+	if (fp == NULL) {
+		free(task_info);
+		debug(D_ERROR|D_MAKEFLOW_HOOK, "could not create task_info for node %d archive", n->nodeid);
+		return 0;
+	} else {
+/* task_info :
```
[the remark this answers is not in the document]
nhazekam: Good Point. I will move this over to JSON.

makeflow/src/makeflow.c
```diff
@@ -1750,17 +1740,28 @@ int main(int argc, char *argv[])
 				k8s_image = xxstrdup(optarg);
 				break;
 			case LONG_OPT_ARCHIVE:
-				should_read_archive = 1;
-				should_write_to_archive = 1;
-				set_archive_directory_string(&archive_directory, optarg);
+				if (makeflow_hook_register(&makeflow_hook_archive, &hook_args) == MAKEFLOW_HOOK_FAILURE)
+					goto EXIT_WITH_FAILURE;
+				jx_insert(hook_args, jx_string("archive_read"), jx_integer(1));
```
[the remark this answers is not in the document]
nhazekam: I will update this to use a boolean.

makeflow/src/makeflow_module_archive.c
nhazekam: Thats a good point. Currently it will be an incomplete/corrupt entry. I will think about removing partial entries on failure.

batch_job/src/batch_wrapper.c
```diff
@@ -114,7 +114,8 @@ char *batch_wrapper_write(struct batch_wrapper *w, struct batch_task *task) {
 	assert(w);
 	assert(task);
 
-	char *name = string_format("%s_XXXXXX", w->prefix ? w->prefix : "./wrapper");
+	char *id = batch_task_generate_id(task);
```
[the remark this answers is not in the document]
nhazekam: Probably. This change should also not be in this pull request. I will remove this and make a separate pull request later.

PR #1899, opened 2018-04-30
Add functions to set up struct dag_node

trshaffer:
In my other PR, I had to awkwardly make changes throughout both Makeflow parsers. Some places (setting the command/submakeflow and adding the node to the DAG) were repeated between the two, and in fact differed in behavior. This PR abstracts that logic into functions on `struct dag_node` that both parsers can call into. Rebasing onto this will also clean my other PR up significantly.


trshaffer:
Yes, that only existed in the Make parser to prefix the subcommand. That feature doesn't appear to be documented.

PR #1900, opened 2018-04-30
Add VC3 Builder Module

nhazekam:
Added VC3 Builder Module to makeflow.

### DIFF
--- a/makeflow/src/makeflow_module_vc3_builder.c
+++ b/makeflow/src/makeflow_module_vc3_builder.c
@@ -1,0 +1,128 @@
+/*
+ Copyright (C) 2018- The University of Notre Dame
+ This software is distributed under the GNU General Public License.
+ See the file COPYING for details.
+ */
+
+#include "path.h"
+#include "stringtools.h"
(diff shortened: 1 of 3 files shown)


makeflow/src/makeflow_module_vc3_builder.c
```diff
@@ -0,0 +1,128 @@
...
+		executable = string_format("./%s", path_basename(v->exe));
+	} else {
+		// Else just use executable in path
+		executable = string_format("%s", v->exe);
+	}
+
+	/* Assumes a /disk dir in the image. */
+	char *log = string_format("%s_%d", v->log, t->taskid);
+	char *task_cmd = string_escape_shell(t->command);
+	char *cmd = string_format("%s --home $PWD %s %s > %s", executable, v->opt, task_cmd, log);
```
btovar: I suggest adding a -- before the command:

"%s --home $PWD %s -- %s > %s"

nhazekam: Good suggestion. Will do.

PR #1902, opened 2018-05-08
Guard against time_application being out of step when plotting.

btovar:


dthain:
@Nekel-Seyew does this fix the problem you observed?

Nekel-Seyew:
Yes, this does fix the problem we were having when graphing. Thank you Ben!

PR #1903, opened 2018-05-09
Cleanup of Master-Worker Capacity

nkremerh:
This is a cleanup and revision of the capacity model to reflect changes made between revisions of the corresponding paper on master-worker capacity.

### DIFF
--- a/batch_job/src/work_queue_factory.c
+++ b/batch_job/src/work_queue_factory.c
@@ -78,6 +78,9 @@ static int tasks_per_worker = -1;
 static int autosize = 0;
 static int worker_timeout = 300;
 static int consider_capacity = 0;
+int capacity_weighted = 0;
+int tr = 0;
+int td= 0;
 
 static char *project_regex = 0;
 static char *submission_regex = 0;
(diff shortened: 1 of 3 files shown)


nkremerh:
@btovar This is an attempt to clean up how capacity is managed. Feel free to take a look when you have time.

work_queue/src/work_queue.h
```diff
@@ -244,9 +244,8 @@ struct work_queue_stats {
 	int capacity_cores;     /**< The estimated number of workers' cores that this master can effectively support.*/
 	int capacity_memory;    /**< The estimated number of workers' MB of RAM that this master can effectively support.*/
 	int capacity_disk;      /**< The estimated number of workers' MB of disk that this master can effectively support.*/
-	int capacity_instantaneous; /**< The estimated number of tasks that this master can support considering only the most recently completed task. */
```
btovar: Changes in names of fields, or order of fields should be avoided in wq_stats.

batch_job/src/work_queue_factory.c
```diff
@@ -78,6 +78,9 @@ static int tasks_per_worker = -1;
 static int autosize = 0;
 static int worker_timeout = 300;
 static int consider_capacity = 0;
+int capacity_weighted = 0;
```
btovar: Could we move these three values inside the function they are used?

work_queue/src/work_queue.h
nkremerh: I removed capacity_instantaneous since it is never used. We had anticipated we would use this with the capacity code, but that never came to fruition. I am alright with keeping it in just in case we would like to use it later. Thoughts?

batch_job/src/work_queue_factory.c
nkremerh: I kept these three values outside the function for the purpose of debugging. I can get creative with logging them though to keep them all inside the same function. Thoughts?

work_queue/src/work_queue.h
nkremerh: capacity_instantaneous is still kept currently, but it is renamed.

btovar: Don't rename it. We don't want abbreviated names in the stats.

batch_job/src/work_queue_factory.c
btovar: Please do. Switches may live outside functions for convenience, but it gets ugly when the state of a function lives outside it.

nkremerh:
@btovar Let me know if I addressed your requested changes.

batch_job/src/work_queue_factory.c
```diff
@@ -281,33 +310,14 @@ static int count_workers_needed( struct list *masters_list, int only_waiting )
 		}
 
 		debug(D_WQ,"%s %s:%d %s %d %d %d",project,host,port,owner,tasks,capacity,need);
-
+		debug(D_WQ,"capacity: %d %d %d %d %d", capacity, need, count_workers_connected(masters_list), td, tr);
```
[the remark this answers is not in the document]
nkremerh: You make a good point. I collapsed the redundant debug message into the first debug message.

dthain:
Ready to merge?

nkremerh:
@dthain I believe I have addressed all of @btovar's concerns.

nkremerh:
@dthain This is ready to merge.

PR #1904, opened 2018-05-21
factory to have N workers to wait connection at any time.

btovar:
Before, the factory would wait until all workers of previous cycles to
connect before submitting new workers. If a worker had a problem, this
would halt future submissions.

Now the factory would have workers_per_cycle waiting at any time. (Until
the target number of workers is reached.)

### DIFF
--- a/batch_job/src/work_queue_factory.c
+++ b/batch_job/src/work_queue_factory.c
@@ -865,16 +865,17 @@ static void mainloop( struct batch_queue *queue )
 			workers_needed = workers_min;
 		}
 
-		int new_workers_needed = workers_needed - workers_submitted;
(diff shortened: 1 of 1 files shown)


dthain:
I'm not understanding why this takes into account "workers waiting to connect", which is an unstable quantity.

Don't we just want to submit the difference between the desired workers and the workers currently submitted?

btovar:
It is throttling.  We don't want hundreds of workers suddenly connecting to the master as they would likely time-out before the master can get to them. Without this, the factory blindly submits workers_per_cycle every cycle, which actually defeats some of the purpose of workers_per_cycle.

The number is unstable, yes, but the number of submissions is always limited by 0 and workers_per_cycle.

dthain:
Ok, makes sense to me.

PR #1905, opened 2018-05-25
Amazon Batch Integration

Nekel-Seyew:
Updated Version of Ben's Amazon Batch pull request with fixes, specifically for keeping directories and proper sh_popen/sh_pclose matching.


dttools/src/sh_popen.c
```diff
@@ -0,0 +1,130 @@
...
+		char* argv[4];
+		argv[0] = "/bin/sh";
+		argv[1] = "-c";
+		argv[2] = command;
+		argv[3] = NULL;
+
+		res = execvp(argv[0], argv);
+		if (res < 0) {
+			debug(D_ERROR,"SH_system past execvp: %s",strerror(errno));
+		}
```
dthain: Always _exit(1) after a failed exec, otherwise you will end up with excess processes.

PR #1906, opened 2018-05-30
Use a counter for job names submitted via cluster

trshaffer:
This is to work around a buggy qsub wrapper in place at the CRC.

cc @dthain


dthain:
Looks simple and (hopefully) foolproof.

PR #1911, opened 2018-06-15
S3 Archiving Code

npotteig:
This code modifies makeflow_module_archive.c so that it can upload files to a S3 bucket and download files from there. This code is integrated with the already functioning local archive mechanic so they both run at the same time. Unfortunately there is a bug in the code at the moment that I am trying to fix that has to do with the archive's inability to handle a directories for input and output files. Nicholas Hazekamp wanting me to create this request so he could review my code and help figure out a solution to the problem.


nhazekam:
Simple misunderstanding. I only meant for you to push to your repo so it was available online for review. I don't think this is ready for a full pull request. I can view your code from your repo. I am going to close this and will touch base with you on the code via email.

PR #1918, opened 2018-06-26
JX Embedded Context

dthain:
Under the existing JX code, there is no obvious translation of our example "convert" workflow into a JX form, because one must always have a separate file in order to define values for substitution.  This PR allows a JX document to have a "define" section which is merged into the context for evaluation, allowing you to have a self-contained document.

Also, the context is evaluated against itself, which allows various definitions to build upon each other, or to use non-atomic values such as list comprehensions..

I'm not entirely sure about the appropriate name of this section.
Maybe "let" or "define" or "constants" or ...

For example:

```
{
  "define": {
    "CONVERT":"/usr/bin/convert",
    "CURL":"/usr/bin/curl",
    "URL":"http://ccl.cse.nd.edu/images/capitol.jpg",
    "RANGE":range(0,360,90)
  },
  "rules": [
    {
      "command": CONVERT+" -delay 10 -loop 0 capitol.jpg capitol.90.jpg capitol.180.jpg capitol.270.jpg capitol.360.jpg capitol.270.jpg capitol.180.jpg capitol.90.jpg capitol.anim.gif",
      "local_job":true,
      "inputs": [
        "capitol.jpg",
        format("capitol.%d.jpg",N) for N in RANGE,
        CONVERT,
      ],
      "outputs": [ "capitol.anim.gif" ]
    },    
    {
      "command":format("%s -swirl %d capitol.jpg capitol.%d.jpg",CONVERT,N,N),
      "inputs": [ "capitol.jpg", CONVERT ],
      "outputs": [ format("capitol.%d.jpg",N) ]
    } for N in RANGE,
      
    {
      "command":CURL+" -o capitol.jpg http://ccl.cse.nd.edu/images/capitol.jpg",
      "local_job":true,
      "inputs" : [ CURL ],
      "outputs": [ "capitol.jpg" ]
    }
  ]
}
```

Now, I do realize we are slowly stretching the "pure" JX into something more when used with workflows. @btovar @nhazekam @trshaffer @nkremerh any thoughts on this?

### DIFF
--- a/makeflow/src/parser.c
+++ b/makeflow/src/parser.c
@@ -63,9 +63,6 @@ struct dag *dag_from_file(const char *filename, dag_syntax_type format, struct j
 	FILE *dagfile = NULL;
 	struct jx *dag = NULL;
 	struct jx *jx_tmp = NULL;
-	struct jx *defines = NULL;
-	struct jx *context = NULL;
-	struct jx *econtext = NULL;
 	struct dag *d = NULL;
 
 	// Initial verification of file existence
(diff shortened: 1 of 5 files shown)


makeflow/src/parser.c
```diff
@@ -93,10 +98,28 @@ struct dag *dag_from_file(const char *filename, dag_syntax_type format, struct j
...
+			econtext = jx_eval(context,context);
+			if(!econtext) {
+			  fprintf(stderr,"makeflow: error in context definitions:\n");
+				jx_print_stream(context,stderr);
+				printf("\n");
+				return 0;
+			}
+
+			jx_tmp = jx_eval(dag,econtext);
+			jx_delete(context);
```
nhazekam: Do we need to also free the evaluated context, `econtext`?

nhazekam:
I think that define fits well.

trshaffer:
The two other approaches we considered for doing this ended up being problematic, so this seems like a fine middle ground

PR #1919, opened 2018-06-26
JX String Concatenation

dthain:
Overload the + operator to allow for string concatenation between unequal types.

### DIFF
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -431,6 +431,19 @@ void jx_delete( struct jx *j )
 	free(j);
 }
 
+int jx_isatomic( struct jx *j )
+{
+	switch(j->type) {
+		case JX_BOOLEAN:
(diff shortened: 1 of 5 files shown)


dthain:
Ah, this fails one existing test which expects "abc"+[1] to fail with a type error.
This PR causes it to evaluate to the string "abc[1]".
That is expected but perhaps not desirable.

So, for those of your following along at home, should we allow *any* type to concatenate with a string?  Or only atomic types?  Or something else?

@nhazekam @nkremerh @trshaffer

nhazekam:
I think this would be useful in keeping command creation and file formatting more concise. I am not sure if this will have consequences elsewhere, but since this is limited to "+", I am less worried.

dthain:
Ok, I'm going with concatenation only on atomic types.

PR #1920, opened 2018-06-26
copy_dir_real fix in copy_tree.c

npotteig:
This error would always produce a logic error where only one entry in the directory is being copied over due to breaking from the while loop when something other than 0 returning from copy_direntry. This function will not necessarily always return 0 when it successfully runs but it will return -1 on error. The change in the conditional statement may not be necessary but was added for clarity on when the while loop will terminate.

### DIFF
--- a/dttools/src/copy_tree.c
+++ b/dttools/src/copy_tree.c
@@ -95,7 +95,7 @@ int copy_dir_real(const char *source, const char *target) {
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


npotteig:
I ran make test on my branch and found that one test failed which was 
chirp/test/TR_chirp_job.sh

dthain:
Test failures appear to be intermittent unrelated issues...

PR #1921, opened 2018-06-27
Updated JX Documentation

dthain:
Added more up-front example and discussion to Makeflow manual.
Added workflow-oriented tutorial on JX.
Added quick reference card on JX.

Still needs some cleanup to reconcile end of makeflow manual with JX reference page.


dthain:
Depends on #1918 and #1919

PR #1924, opened 2018-06-28
Catalog Updates via TCP

dthain:
Combines update compression and TCP transport.
Merged code has not been tested yet.
Supersedes #1739


dthain:
Tested and works: udp/tcp and raw/compressed with no leaks at either end.

Issue #1915, opened 2018-06-26
Improved JX Documentation


dthain:
While we have a nice reference regarding the JX language itself (JSON + operators + comprehensions), we don't have a clear statement of how JX is used to encode a makeflow, in a way that a new user could get going easily.

As a starting point, I would like to just write down a schema for a dag expressed in JX.
Here is a first stab at it, with (illegal) comments. Are we missing anything important?

```
# Whole document must be a dictionary.
{
  # Required: rules must be an array of jobs, each one a dictionary.
  "rules" : [
                  {
                        # Required: A job must have either be a command or a makeflow
			"command" : "./simulation.py input.dat > output.txt",
			"makeflow" : { "path" : "test.mf", "cwd" : "." },

                        # Required: Inputs and outputs must be a list of files.
                        # Each file can either be a simple string, or a dictionary with "dag_name" and "task_name"
                        "inputs" : [ "input.dat", "simulation.py" ],
			"outputs" : [ "output.txt", { "dag_name":"output.10.txt", "task_name":"output.txt"],
                        # Optional: local_job is a boolean
			"local_job" : true|false,
                        # Optional: environment is a dictionary that overrides values in the global environment.
			"environment" : [ ],
                        # Optional: category names an element in the categories dictionary.
			"category" : "simulation",
                        # Optional: resources needed for execution (Q: What are the units?)
			"resources" : { "cores" : 4, "disk": 10, "memory": 20 }
                        # Optional: First allocation strategy.
			"allocation" : "first|max|error"
		},
	]

  # Optional: Named categories describe resources and environment for a set of jobs.
  "categories" : {
		"simulation" : { "resources" : { "cores":4, "memory":12, "disk":16 }, "environment": { "PATH":"/usr/local/bin"} }
	},

  # Optional give name of default category
  "default_category" : "simulation",

  # Optional: give top level environment to be shared by all jobs
  "environment" : { "PATH" : "/usr/local/bin" },
}
```

dthain:
@btovar @nhazekam @trshaffer your thoughts?

dthain:
See PRs #1916 #1918 #1919

dthain:
@btovar can *anything* go in a categories clause, or only environment and resources?

btovar:
In Makeflow and Work Queue we allow to set the mode for the first allocation computation (min waste or max throughput), which is part of the category definition. 

Thus, not only resources and environments, but do we want to allow everything?

[...]

PR #1925, opened 2018-06-29
resources and categories

btovar:
For #1915


dthain:
Yes, that's just what I had in mind. Could you extend to discuss how to do allocation/monitoring when you (the user) doesn't know what resources to pick?

dthain:
Looking at this with fresh eyes, it seems like the term “first” may not be clear to new users.  How about we add “auto” as an alias for “first”?

PR #1932, opened 2018-07-03
Move old json.[ch] to Chirp module.

dthain:
Localizing this code to the chirp module, which is the last place it is used (extensively) in the `chirp_job` capability.  Everything else is now using JX for JSON support.  (Would be lots of work to update `chirp_job` to use JX with no external benefit.)


dthain:
Not essential for 7.0

dthain:
Any objections to merging?

btovar:
Not as such, but there are conflicts with this branch.

dthain:
Ok, let me take a look.

dthain:
Rebased on master.

PR #1933, opened 2018-07-03
Added simple call graph generator.

dthain:
Added simple call graph generator, to use in our various orientation and debugging documents.
Try it like this:

```
./make_call_graph.py parrot/src/*.o | dot -T pdf > parrot.pdf
```


dthain:
Not essential for 7.0

dthain:
Whoops, I messed up the cherry-pick

dthain:
There it is now.

PR #1935, opened 2018-07-06
add wall_time as makeflow jx resource

btovar:


dthain:
Don't forget to add it  to the tutorial, quick reference, and full
reference pages...

btovar:
Ah yes, thanks for the reminder!

PR #1937, opened 2018-07-09
Fix documentation errors in JX.

dthain:


dthain:
@btovar These are dumb but small syntactic errors that are likely to throw off new users of JX.
We probably need to get out a 7.0.1 to avoid egg on the face.

btovar:
Sounds good. I have a couple of epsilon fixes to add too.

PR #1938, opened 2018-07-09
Makeflow S3 Archive Tool and Directory Hashing

npotteig:
This branch implements functionality to upload, download, and check for files in an aws s3 bucket using the c library curl. The local archive code has been updated to accommodate for directories when they are included in the input and output files of a task too. Hashing directories and checking to see if something is a directory has been implemented into the file batch_file.c. Most of the code that has been implemented has been into the file makeflow_module_archive.c. Flag options have been added to the command line so the user can run the s3 archive code or choose which s3 bucket they would want to upload/download their files from.


trshaffer:
Closing per @npotteig , re-opening from another branch

PR #1939, opened 2018-07-09
Makeflow S3 archive and Directory Hashing

npotteig:
This branch implements functionality to upload, download, and check for files in an aws s3 bucket using the c library curl. The local archive code has been updated to accommodate for directories when they are included in the input and output files of a task too. Hashing directories and checking to see if something is a directory has been implemented into the file batch_file.c. Most of the code that has been implemented has been into the file makeflow_module_archive.c. Flag options have been added to the command line so the user can run the s3 archive code or choose which s3 bucket they would want to upload/download their files from.

### DIFF
--- a/batch_job/src/batch_file.c
+++ b/batch_job/src/batch_file.c
@@ -8,6 +8,15 @@ See the file COPYING for details.
 #include "sha1.h"
 #include "stringtools.h"
 #include "xxmalloc.h"
+#include "path.h"
+#include "hash_table.h"
+#include <time.h>
+#include <sys/time.h>
+#include <sys/types.h>
+#include <dirent.h>
+
+struct hash_table *check_sums = NULL;
+double total_checksum_time = 0.0;
 
 /**
  * Create batch_file from outer_name and inner_name.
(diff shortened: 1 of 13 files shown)


dthain:
Hmm, I didn't realize there was a license incompatibility.  We should try to avoid that.
@npotteig now that you know how to use libcurl directly, could you just write three simple functions of your own to get/put/check files in s3?

npotteig:
I am not using most of the aws4c library anyway so I would be able to just delete the functions that are unnecessary and keep the ones that I have already worked with.

dthain:
Another point:
In the other places where we use AWS services, we assume the user has done an "aws configure" to set up the various secret keys and so forth.  This code should continue to use that information and load its info from `$HOME/.aws/credentials` which looks like this:

[default]
aws_access_key_id = XXXXXXXX...
aws_secret_access_key = YYYYYYYY...
```

npotteig:
Yep on it

btovar:
Merged with commit 11842d450483141fa3b4c1255200141eb320846e

PR #1940, opened 2018-07-09
JX eval errors

### LINKED ISSUE #1922: Improve JX Error Messages
Right now, if the user has a bug in their  JX workflow, the error messages they get back are rather difficult to understand.  There are really three levels of error handling:

1 - In `dttools/src/jx_parse.c` where we parse text to see if it is valid JX.
2 - In `dttools/src/jx_eval.x` where we evaluate the JX to resolve values.
3 - In `makeflow/src/parser_jx.c` where we examine the valid JX to see if it has the expected structure for makeflow.

@btovar, please take a pass at improving level 3 first. What I would like to see here for every possible error is that we print a message to stderr that explicitly states what was expected, and what was actually found.  (debug() isn't the right thing here b/c that is for internal developer info, not first-class results.)  Something like this:

```
fprintf(stderr,"makeflow: line %d: expected an XXX but got %s instead\n",line,jx_print_string(relevant_thing));
```

Let's take a crack at that first and then work our way down into the stack.

trshaffer:
Resolves #1922

This replaces the existing JX error messages (hard to read) with simpler function-looking messages. Errors are reported as normal strings.

See the new [expected test](https://github.com/trshaffer/cctools/blob/c1687324df43e34471b529e215138a711aecc6a1/dttools/test/jx.expected#L181) for how things look now.

### DIFF
--- a/dttools/src/.gitignore
+++ b/dttools/src/.gitignore
@@ -28,3 +28,4 @@ worker
 worker_condor_submit
 libforce_halt_enospc.so
 jx_count_obj_test
+jx2env
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -7,6 +7,7 @@ See the file COPYING for details.
 #include "jx.h"
 #include "stringtools.h"
 #include "buffer.h"
+#include "xxmalloc.h"
 
 #include <assert.h>
 #include <stdarg.h>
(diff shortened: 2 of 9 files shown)


dthain:
Yes, these errors look much more readable.

trshaffer:
If there are no objections, I'd like to merge this

PR #1942, opened 2018-07-16
Add jx_hash function

trshaffer:
This is a function to compute the SHA1 hash of a JX object. Since this operates on objects in memory, it avoids issues with whitespace when reading JSON. It also canonicalizes out things like object key orderings. Note that I only considered the JSON types as hashable, so any JX syntax must be evaluated out.

### DIFF
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -7,12 +7,17 @@ See the file COPYING for details.
 #include "jx.h"
 #include "stringtools.h"
 #include "buffer.h"
+#include "sha1.h"
+#include "xxmalloc.h"
 
 #include <assert.h>
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdbool.h>
+#include <stdint.h>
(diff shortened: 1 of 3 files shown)


dthain:
@trshaffer this is a clever and fast way to checksum, but I think we need to do it the more slow and direct way: convert it to a canonical json string, then checksum the string.

The reason for this is that we want to be able to externalize these data structures in a way that other tools can deal with them.  For example, to generate a task as a JSON object and then write it to an S3 bucket named with the hash of the string.  In that form, it's much easier to explain (and implement and check) that the checksum came from sha1(canonicalize(tostring(json)))

trshaffer:
I was avoiding that approach because it ties our `jx_print` code into the hashing, so a minor change to how we print JX down the road could invalidate all the old hashes. As long as we're aware of that, it's straightforward to do the canonicalization.

dthain:
That's certainly a good point.  I think you want to have a `jx_canonical_string` or something like that to clarify the purpose as distinct from general printing.

trshaffer:
Closing with #1893

PR #1944, opened 2018-07-17
Worker always clean on disk exhaustion

btovar:
Before we would try to keep the worker around when a task filled the
disk alloted to the worker. However, this only works for inputs that are
not cached (e.g., not makeflow). CJ found that the master would keep sending tasks to a
worker, and those tasks eventually would fill the worker's disk. The
tasks were terminated, but the cached was not cleared, thus disk usage
at the worker kept increasing, and the tasks kept failing.

Since there is no easy way to tell at the worker which files are cached,
this commit makes it so that the worker disconnects.


dthain:
This is the right short term fix.

In principle, the master knows the contents of the worker cache and could select which items to remove.  (Although this is not an easy decision to make.)

btovar:
Yes, I think the same way.

PR #1954, opened 2018-08-07
rename resource to feature in function implementation

btovar:
As reported by @annawoodard, python3 -mworkqueue would fail as the function was not defined (only declared).


annawoodard:
Reported to me by @MalcolmUnWang 😄

MalcolmUnWang:
This patch solves the installation error on Ubuntu. Thank you very much!

btovar:
Excellent, thanks for letting us know.

PR #1957, opened 2018-08-24
Add LDFLAGS for correct compilation

btovar:
To compile cctools on stampede:

CC=gcc CXX=g++ LDFLAGS='-fstack-protector' ./configure --without-static-libgcc

### DIFF
--- a/configure
+++ b/configure
@@ -134,6 +134,8 @@ config_irods_path=no
 config_openssl_path=no
 config_xrootd_path=no
 
+config_static_libgcc=yes
+
 while [ $# -gt 0 ]
 do
 	opt=$1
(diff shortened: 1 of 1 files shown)


dthain:
Is it realistic to consider providing instructions for a build gcc of perl/python?

btovar:
Well, I am having trouble compiling parrot even with gcc. There seems to be a mismatch with the pthread library.

btovar:
Ok, the problem for parrot seems to be the  -static-libgcc  flag.

btovar:
With 

CC=gcc CXX=g++ LDFLAGS='-fstack-protector' ./configure --without-static-libgcc

all cctools can be compiled.

PR #1958, opened 2018-08-27
Work queue: Consider workers connected from other sources

rlmv:
I ran into this situation the other day: I started a work queue, and a work queue factory to handle workers. After letting the computation run for a while, I realized that I had underestimated the amount of time my tasks would take, and also underspecified the worker limits of the factory. I started another factory.

However, multiple factories don't play well together. I got to a point where all the tasks were distributed to workers but, because the factory doesn't check whether the number of connected workers meets the task needs, each factory was constantly spinning up new workers to meet a need that didn't exist. 

This patch fixes that by doing a sanity check to see whether there are workers connected from other sources, and considering that number when it computes `new_workers_needed`.

Cheers!

### DIFF
--- a/batch_job/src/work_queue_factory.c
+++ b/batch_job/src/work_queue_factory.c
@@ -876,6 +876,7 @@ static void mainloop( struct batch_queue *queue )
 
 		int new_workers_needed         = MAX(0, workers_needed - workers_submitted);
 		int workers_waiting_to_connect = MAX(0, workers_submitted - workers_connected);
+		int workers_from_elsewhere     = MAX(0, workers_connected - workers_submitted);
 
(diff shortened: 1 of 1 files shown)


btovar:
Thanks Bo!

At first glance, the patch looks good.

rlmv:
Thanks Benjamin. When I was investigating this I also noticed that https://github.com/cooperative-computing-lab/cctools/blob/6c57f60399976a123097f9bc78e171bad777bf8e/batch_job/src/work_queue_factory.c#L908 seems like a dead branch. Since nothing ever subtracts from `new_workers_needed` it will never be `< 0`. Not sure that needs changing, but just wanted to draw your attention to it.

btovar:
Indeed, thanks. It seems like a leftover safeguard. Thanks again!

PR #1961, opened 2018-08-27
Fix existing new workers calculations

rlmv:
Fixes bug introduced in #1958.

Existing workers need to be applied *before* the max workers per cycle constraint, otherwise `new_workers_needed` can be reduced to 0 when the number of workers from other sources is greater than `workers_per_cycle`.

Sorry about that, @btovar. This seems to be working as expected now.


btovar:
Yes, good catch. Thanks again!

Issue #1947, opened 2018-07-23
Support for archives/disk images in Parrot


trshaffer:
There's been occasional discussion for some time about using Parrot to "mount" a tarball or similar file and present it to the user as a directory tree. There seems to be some interest now in the HEP community about using this kind of approach for distributing self-contained software. I'm opening this issue to track discussion on this feature.

An initial question is what format(s) to support. A few suggestions:
- **TAR**: Likely be the simplest to implement and widely used, but poorly structured for use as a mountable filesystem.
- **SquashFS**: More complicated structure to support directory indexes and seeking, as it's designed as a performant, mountable filesystem.
- **ISO 9660**: Simpler to implement than SquashFS, structured with proper directory trees (unlike TAR). Has some legacy limitations on file size, paths, etc.
- **ext[2-4]**: Default filesystems for many Linux systems. Used with Shifter. There is a library interface included in e2fsprogs.

IIRC Docker images are structured as tarballs of tarballs. @Nekel-Seyew mentioned that Singularity supports SquashFS for compressed disk images. On the desktop side of things, Ubuntu's Snap format also uses SquashFS. Are there other technologies to consider? How are HEP people planning to do things?

btovar:
TIm, as I understand it, SquashFS is now the default in singularity, that is, we should expect most singularity images to use SquashFS.

nhazekam:
@jblomer May have more insight into this, but the above suggested formats sound reasonable. The work I am currently doing is targeting both TAR and SquashFS as outputs for the packaged CVMFS repository. If there was Parrot support for these, the combination could be used in a location where you don't have access to CVMFS (possibly due to network) and you are not able to mount.

jblomer:
I think that would be a super useful feature, if only for my personal use!  I've always wanted a lightweight way to look in VM images, or at least ext2-4 file system blobs.  I'm not sure if there is a user-land library for ext.  Ext image files are, as far as I understand, used with the Shifter container system.

PR #1962, opened 2018-08-30
Parrot support for mounting ext images

trshaffer:
This PR adds ext2fs as an optional dependency so that Parrot can mount ext[234] disk images. This could be expanded to support read/write access, but for the moment I kept things RO. Users specify the path to an image file which Parrot loads and exposes as a directory tree under `/ext_%d`. I might adjust how images are specified on the command line and the paths they get to improve ergonomics. I've done enough testing to make sure basic functionality works (look around directories, read files, exec programs, follow symlinks).

See also #1947

### DIFF
--- a/parrot/src/pfs_main.cc
+++ b/parrot/src/pfs_main.cc
@@ -289,7 +289,7 @@ static void show_help( const char *cmd )
 	printf( " %-30s Make flock a no-op.\n", "--no-flock");
 	printf("\n");
 	printf("Filesystem Options:\n");
-	printf( " %-30s Mount an ext[234] disk image read-only at /ext_n.\n", "--ext-image=<image>");
+	printf( " %-30s Mount a read-only ext[234] disk image.\n", "--ext <image>=<mountpoint>");
 	printf("FTP / GridFTP options:\n");
 	printf( " %-30s Enable data channel authentication in GridFTP.\n", "-C,--channel-auth");
 	printf("\n");
--- a/parrot/src/pfs_main.cc
+++ b/parrot/src/pfs_main.cc
@@ -72,6 +72,8 @@ extern "C" {
 #include <string.h>
 #include <time.h>
 
+#include <vector>
+
 #define SIG_ISSTOP(s) (s == SIGTTIN || s == SIGTTOU || s == SIGSTOP || s == SIGTSTP)
 
 FILE *namelist_file;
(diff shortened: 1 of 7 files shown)


dthain:
@trshaffer can you post a quick how-to here for creating and mounting an image?
@nhazekam please give this a quick try and see if it works for you.

trshaffer:
Since this feature gives RO access, you need to use a machine with root access to make an image.

    truncate -s 64M /tmp/test.img
    mkfs.ext2 /tmp/test.img
    losetup -f /tmp/test.img

At this point, you'll need to figure out which loop device you got, e.g. using `losetup -a`.

    mount /dev/loopn /mnt

Now you can put some stuff in `/mnt`. To clean up the mount,

    umount /mnt
    losetup -d /dev/loopn

Now Parrot can mount you image.

    parrot_run --ext /tmp/test.img -- bash

The image will appear under `/ext_0`.

trshaffer:
@nhazekam I just changed the command line interface somewhat. The commit message for 
636b914 explains it. I updated the little howto above (mounts to `/ext` rather than `/ext_0`).

dthain:
Any reason that you can't make it just `--ext image=mountpoint` ?

trshaffer:
@dthain the current thing is just slightly simpler to implement. But now that you mention it, mirroring `mount (1)` is more intuitive for users. I'll switch that over.

batrick:
Just wanted to say this looks like a really cool feature! I imagine this will be used for sandboxing/quotas?

If you are considering using this with HTCondor to transmit inputs/outputs, you might want to check if you can shrink-to-fit images (grow on starting Parrot and shrink when exiting?) before transmitting over the network.

trshaffer:
@batrick this is actually related to some work @nhazekam is doing on delivering CVMFS to HPC sites. Next in line is reading SquashFS images through Parrot.

parrot/src/pfs_main.cc
```diff
@@ -282,6 +288,8 @@ static void show_help( const char *cmd )
 	printf( " %-30s Disable the given service.\n", "--disable-service");
 	printf( " %-30s Make flock a no-op.\n", "--no-flock");
 	printf("\n");
+	printf("Filesystem Options:\n");
+	printf( " %-30s Mount an ext[234] disk image read-only at /ext_n.\n", "--ext-image=<image>");
```
nhazekam: I believe this option has also changed. I am mounting it using `--ext /tmp/test.img=/cvmfs`

PR #1969, opened 2018-09-19
Enable Travis CI

trshaffer:
This PR will enable Travis on the main repo. I've been testing on my fork, and things seem to be working well. This configuration tests on Travis' default Linux and Mac environments, and in Docker environments we currently use for Autobuild. A couple of points might require discussion:
- The behavior of `run_all_tests.sh` changes so that failed tests result in non-zero exit status. This will likely break the current Autobuild setup. I'm of the opinion that any failed tests should be considered an overall failure, but we might need to drop this commit and keep the current setup with different levels of failure for compatibility with Autobuild.
- I removed the test for WQ Foreman. From a quick glance at the Autobuild display, this test fails about a third of the time, and is ignored in practice.

### DIFF
--- a/run_all_tests.sh
+++ b/run_all_tests.sh
@@ -101,6 +101,10 @@ echo ""
 echo "Test Results: ${FAILURE} of ${TOTAL} tests failed (${SKIP} skipped) in ${ELAPSED} seconds."
 echo ""
 
-exit $FAILURE
+if [ "$FAILURE" -eq 0 ]; then
+	exit 0
+else
+	exit 1
+fi
 
 # vim: set noexpandtab tabstop=4:
--- a/run_all_tests.sh
+++ b/run_all_tests.sh
@@ -95,5 +95,6 @@ echo ""
 echo "Test Results: ${FAILURE} of ${TOTAL} tests failed (${SKIP} skipped) in ${ELAPSED} seconds."
 echo ""
 
+exit $FAILURE
 
 # vim: set noexpandtab tabstop=4:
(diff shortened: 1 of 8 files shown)


run_all_tests.sh
```diff
@@ -95,5 +101,6 @@ echo ""
 echo "Test Results: ${FAILURE} of ${TOTAL} tests failed (${SKIP} skipped) in ${ELAPSED} seconds."
 echo ""
 
+exit $FAILURE
```
btovar: Let's make this one: 

if  [ "$FAILURE" = 0 ]
then
 exit 0
else
 exit 1
fi

or something of the sort. We don't want the exit status to vary according to the number of failures.

travis_build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -ex
+
+COMMAND="./configure --strict && make && make test"
```
btovar: don't you need to set the paths for cvmfs, python3, etc?

trshaffer: My bad, forgot to add those. It looks like the current setup specifies

    --with-cvmfs-path /opt/libcvmfs --with-uuid-path /opt/uuid

correct?

btovar: Yes, I think that should work.

For python3 we set it with configure.afs. I am not certain how to do extend that to this case, but perhaps python3 is automatically recognized in all of the containers?

trshaffer: For the distros that use Python 3, it seems to be picking it up automatically. For the ones that use Python 2 (Centos 6 & 7), do we want to try to change it?

btovar: For sure I know it picks python2 automatically for all of them, so let's try by only setting explicitly cvmfs and its dependencies (uuid).

trshaffer:
Those failures are also my fault. I hate Docker...

trshaffer:
Looks like everything's green, so I don't have any other changes in mind

btovar:
All right, let's merge it.

I am still using autobuild to deploy to the ND campus, but this pr changes do not seem to affect that.

PR #1970, opened 2018-09-20
Misc doc fixes for linking work queue.

dthain:


dthain:
@btovar look ok?

btovar:
They do. I manually tried the gcc line and it worked for me. (I also checked that the -lz was indeed needed.)

PR #1977, opened 2018-10-30
WorkQueue Golang API

charleszheng44:
Use SWIG to generate Golang API for work queue.


charleszheng44:
@btovar @dthain 
Should Golang api be generated automatically when making cctools?

btovar:
The new bindings do not add any dependencies. I don't see why not.

charleszheng44:
@btovar I added a new env variable `CCTOOLS_WORKQUEUE_SWIG_BINDINGS` which is `${CCTOOLS_SWIG_BINDINGS} golang`, for building golang API just for work_queue.

charleszheng44:
where can I see the autobuild error message?

charleszheng44:
For redhat 5/6/7, centos 6/7, the swig version is old and doesn't support golang (doesn't recognize `-cgo` option), for ubuntu 16.04, debian 9.2, opensuse42.3 I got error message

```sh
cp work_queue.pm work_queue.so       /var/condor/execute/dir_6600/cctools-db2734a0-x86_64-opensuse42.3/lib/perl5/site_perl/5.18.2/
mkdir -p              /var/condor/execute/dir_6600/cctools-db2734a0-x86_64-opensuse42.3/lib/perl5/site_perl/5.18.2/Work_Queue
cp Work_Queue.pm      /var/condor/execute/dir_6600/cctools-db2734a0-x86_64-opensuse42.3/lib/perl5/site_perl/5.18.2
cp: cannot stat 'Work_Queue.pm': No such file or directory
Makefile:31: recipe for target 'install' failed
```
when I make locally on ubuntu16.04, I didn't see these problems. Do you know how to work around them? @btovar

configure
```diff
@@ -1246,6 +1247,8 @@ fi
 if [ $found_swig != yes ]
 then
 	swig_bindings=""
+else
+	workqueue_swig_bindings="$swig_bindings golang"
```
btovar: something like:

```sh
workqueue_swig_bindings="$swig_bindings"

if ${swig} -cgo 2>&1 | grep -- '-cgo' > /dev/null
then
      echo "*** skipping golang workqueue support"
else
      workqueue_swig_bindings="$workqueue_swig_bindings golang"
fi
```

work_queue/src/perl/Work_Queue.pm
```diff
@@ -1,1053 +0,0 @@
-######################################################################
```
btovar: This file should not have been deleted.

configure
```diff
@@ -84,6 +84,7 @@ include_package_wavefront="wavefront"
 include_package_weaver="weaver"
 
 swig_bindings=""
+workqueue_swig_bindings=""
```
btovar: Could you generate the bindings for chirp too? That way we don't have to manage two different swig configurations.

charleszheng44: Will do

charleszheng44:
Add Golang API for chirp

charleszheng44:
refer to #1990

PR #1952, opened 2018-07-29
Mpi worker Code

Nekel-Seyew:
This pull request will ultimately solve issue issue #1755  when completed.

At present moment, mpi_worker, mpi_starter, and mpi_submitter all work. mpi_submitter needs more thorough testing, but for slurm and sge, it submits the mpi_starter and gets makeflow and workers running and communicating with one another. The MPI batch_job module for makeflow still needs debugging, but the core of the code is there.

PR #1986, opened 2018-11-15
MPI worker and batch mode

btovar:
Enable with e.g., ./configure --with-mpicc-path=$(which mpicc)

See doc/makeflow.html for running instructions.

Supersedes #1952


btovar:
ok to merge?

Nekel-Seyew:
@btovar I think so!

Issue #1998, opened 2018-12-22
Installation of work queue on Mac OS 10.14


yudongqiu:
Dear Developers,

I was trying to install the work queue program on Mac OS 10.14 but encountering some issues during compilation. I used python 3.6 installed with anaconda. I have also installed the command line tools via `xcode-select --install`

The source code I used was `http://www3.nd.edu/~ccl/software/files/cctools-6.2.10-source.tar.gz`

Once downloaded and decompressed, I used this command to configure:
```
./configure --prefix $HOME/opt/cctools-6.2.10 --with-python3-path /Users/qyd/codes/anaconda3/envs/forcebalance --with-swig-path /Users/qyd/codes/anaconda3/envs/forcebalance --with-perl-path no --with-globus-path no
```
And it finished without any issue.

However, when I follow that to run `make`, the following error pops up:
```
ld: warning: text-based stub file /System/Library/Frameworks//CoreFoundation.framework/CoreFoundation.tbd and library file /System/Library/Fram
eworks//CoreFoundation.framework/CoreFoundation are out of sync. Falling back to library file for linking.                                    
ld: warning: text-based stub file /System/Library/Frameworks//CoreFoundation.framework/CoreFoundation.tbd and library file /System/Library/Fram
eworks//CoreFoundation.framework/CoreFoundation are out of sync. Falling back to library file for linking.                                    
ld: -stack_size option can only be used when linking a main executable
clang: error: linker command failed with exit code 1 (use -v to see invocation)                                                               
make[3]: *** [_work_queue.so] Error 1
make[2]: *** [python3] Error 2
rm work_queue_priority_test.o work_queue_test_watch.o work_queue_test.o work_queue_status.o work_queue_example.o work_queue_worker.o          
make[1]: *** [src] Error 2
make: *** [work_queue] Error 2
```

I have attached the full configuration log and the make log. Could you please take a look? Thanks!

[config.mk.txt](https://github.com/cooperative-computing-lab/cctools/files/2705630/config.mk.txt)
[make.log](https://github.com/cooperative-computing-lab/cctools/files/2705631/make.log)

dthain:
Hello Yudong -

Thanks for your interest, but you caught us just as the Christmas holiday
begin.
We'll have a chance to look at this early in the new year.

(The error message suggests some problem with linking basic system
libraries,
are you able to build a simple C program outside of cctools?)

Best wishes -
Prof. Thain

yudongqiu:
[...]

PR #2002, opened 2019-01-04
WorkQueue with python 3.6 from anaconda

btovar:
./configure included some anaconda  compilation flags that caused an error when compiling bindings for 3.6.

This error did not occur with pythons not compiled with anaconda.

Compilation against python 3.7 still does not work, but it seems related to the C code generated by swig.


btovar:
Partially fixes #1998

PR #2003, opened 2019-01-07
Adds runos to CCTools

btovar:
Originally written by @Nekel-Seyew.

This version removes the dependency on the vc3-builder, and  sends the runos executable along the batch job (no fixed path dependency anymore).

The example configuration file (which is found automatically) works for Notre Dame path for images and workers.


btovar:
Not needed anymore.

PR #2005, opened 2019-01-11
add measure of master load

btovar:
Adds q->stats->master_load

When close to 1, the master is at full load dealing with the transfer of tasks, and cannot accept connections of new workers.
When close to 0, the master is mostly waiting for something to happen.

It can be seen as a measure of capacity taken from the behavior of the master, independent of any time measures.

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -472,6 +472,7 @@ static void log_queue_stats(struct work_queue *q)
 	buffer_printf(&B, " %d", s.capacity_disk);
 	buffer_printf(&B, " %d", s.capacity_instantaneous);
 	buffer_printf(&B, " %d", s.capacity_weighted);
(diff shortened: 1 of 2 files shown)


work_queue/src/work_queue.c
```diff
@@ -6440,7 +6464,7 @@ int work_queue_specify_log(struct work_queue *q, const char *logfile)
 			// bandwidth:
 			" bytes_sent bytes_received bandwidth"
 			// resources:
-			" capacity_tasks capacity_cores capacity_memory capacity_disk capacity_instantaneous capacity_weighted"
+			" capacity_tasks capacity_cores capacity_memory capacity_disk capacity_instantaneous capacity_weighte master_load"
```
nkremerh: Typo here on "capacity_weighte" at end of line.

PR #2009, opened 2019-01-26
Catalog/DeltaDB Merge Event

dthain:
In an effort to reduce the size of catalog logs, this PR introduces a new event type "M" (merge) which captures multiple changes in a single merge event.  (In short, each UDP update now corresponds to a single log event.)

### DIFF
--- a/deltadb/src/Makefile
+++ b/deltadb/src/Makefile
@@ -4,7 +4,7 @@ include ../../rules.mk
 EXTERNAL_DEPENDENCIES = ../../dttools/src/libdttools.a
 LIBRARIES = libdeltadb.a
 OBJECTS = $(SOURCES:%.c=%.o)
-PROGRAMS = deltadb_query
+PROGRAMS = deltadb_query deltadb_upgrade_log
 SCRIPTS =
 SOURCES = deltadb_stream.c deltadb_reduction.c
 TARGETS = $(LIBRARIES) $(PROGRAMS)
--- a/deltadb/src/deltadb_query.c
+++ b/deltadb/src/deltadb_query.c
@@ -279,6 +279,32 @@ int deltadb_delete_event( struct deltadb *db, const char *key )
 	return 1;
 }
 
+int deltadb_merge_event( struct deltadb *db, const char *key, struct jx *update )
+{
+	struct jx *current = hash_table_remove(db->table,key);
+	if(!current) {
+		/* If the key is not found, it was filtered out; skip the update. */
+		jx_delete(update);
+		return 1;
+	}
+
+	struct jx * merged = jx_merge(update,current);
+
+	hash_table_insert(db->table,key,merged);
+
+	if(display_mode==MODE_STREAM) {
+		display_deferred_time(db);
+		char *str = jx_print_string(update);
+		printf("M %s %s\n",key,str);
+		free(str);
+	}
+
+	jx_delete(update);
+	jx_delete(current);
+
+	return 1;
+}
+
 int deltadb_update_event( struct deltadb *db, const char *key, const char *name, struct jx *jvalue )
 {
 	struct jx * jobject = hash_table_lookup(db->table,key);
--- a/deltadb/src/deltadb_query.c
+++ b/deltadb/src/deltadb_query.c
@@ -279,18 +279,40 @@ int deltadb_delete_event( struct deltadb *db, const char *key )
 	return 1;
 }
 
+/*
+Merge all the values of the update object into the current object,
+replacing where they exist.  We previously used jx_merge here,
+but the O(n^2) nature of the function and the heavy reliance on
+malloc/free resulted in poor performance.  This function avoids
+many malloc/frees by popping pairs off of the update in order,
+finding matches in current, if needed, and then pushing the pair
+on to the head of the current list.
+*/
+
+static void jx_merge_into( struct jx *current, struct jx *update )
+{
+	while(1) {
+		struct jx_pair *p = update->u.pairs;
+		if(!p) break;
+
+		update->u.pairs = p->next;
+
+		struct jx *oldvalue = jx_remove(current,p->key);
+		if(oldvalue) jx_delete(oldvalue);
+
+		p->next = current->u.pairs;
+		current->u.pairs = p;
+	}
+}
+
 int deltadb_merge_event( struct deltadb *db, const char *key, struct jx *update )
 {
-	struct jx *current = hash_table_remove(db->table,key);
+	struct jx *current = hash_table_lookup(db->table,key);
 	if(!current) {
 		/* If the key is not found, it was filtered out; skip the update. */
 		jx_delete(update);
 		return 1;
 	}
-
-	struct jx * merged = jx_merge(update,current,0);
-
-	hash_table_insert(db->table,key,merged);
 
 	if(display_mode==MODE_STREAM) {
 		display_deferred_time(db);
@@ -299,8 +321,9 @@ int deltadb_merge_event( struct deltadb *db, const char *key, struct jx *update 
 		free(str);
 	}
 
+	jx_merge_into(current,update);
+
 	jx_delete(update);
-	jx_delete(current);
 
 	return 1;
 }
--- a/deltadb/src/deltadb_query.c
+++ b/deltadb/src/deltadb_query.c
@@ -288,7 +288,7 @@ int deltadb_merge_event( struct deltadb *db, const char *key, struct jx *update 
 		return 1;
 	}
 
-	struct jx * merged = jx_merge(update,current);
+	struct jx * merged = jx_merge(update,current,0);
 
 	hash_table_insert(db->table,key,merged);
 
--- a/deltadb/src/deltadb_stream.c
+++ b/deltadb/src/deltadb_stream.c
@@ -39,7 +39,8 @@ int deltadb_process_stream( struct deltadb *db, FILE *stream, time_t starttime, 
 		if(line[0]=='C') {
 			n = sscanf(line,"C %s %[^\n]",key,value);
 			if(n==1) {
-				/* backwards compatibility with old log format */				struct nvpair *nv = nvpair_create();
+				/* backwards compatibility with old log format */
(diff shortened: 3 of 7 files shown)


dthain:
The idea here is to reduce the catalog history logs by changing a sequence of update events like this:
```
T 1548562439
U 127.0.0.1:9094:localhost load1 2.1084
U 127.0.0.1:9094:localhost load5 2.10059
U 127.0.0.1:9094:localhost load15 2.14893
```

Into a single merge (M) event that reflects all the changes in a single update.
```
T 1548562439
M 127.0.0.1:9094:localhost {"load1":2.1084,"load5":2.10059,"load15":2.14893}
```

dthain:
@btovar please take a look at this one.
I have done some local testing, but caution is warranted.
If it looks good, please deploy this on the backup catalog server, and let's let it run for a week or so to be sure.

dthain:
Next step will be to go back and reprocess old archived logs to reduce the data size somewhat.

btovar:
I see how you can read M events from the log, but how are you generating them as updates come to the catalog?

dthain:
When the `catalog_server` calls `jx_database_update`:
https://github.com/cooperative-computing-lab/cctools/pull/2009/files#diff-a76df345799d9df1b3841f88c968e90cR251

dthain:
Hmm the plot thickens:

Converting U records to M records decreases storage, but increases query time!

A quick test on `history/2019/23.log`:

```
./deltadb_upgrade_log 23.log new.log
ls -la *.log
-rw-r--r--  1 dthain  staff  35250538 Jan 27 12:13 23.log
-rw-r--r--  1 dthain  staff  14195960 Jan 28 10:02 output.log

time ./deltadb_query --file 23.log --from 2019-01-01 > /dev/null
real	0m2.316s
user	0m2.261s
sys	0m0.039s

time ./deltadb_query --file new.log --from 2019-01-01 > /dev/null
real	0m2.810s
user	0m2.751s
sys	0m0.035s
```

dthain:
So, let me back up on that: don't deploy it yet, until we understand the performance better.
This suggest our JSON parser could use a tune up.

btovar:
Some of it may be our jx object implementation. Instead of keeping the pairs in a list, we could have a hash table instead. 

Also, jx_merge turns out to be quadratic in the number of updates. (E.g., if there are n updates, the first entry is copied n times across n jx objects.) We could make it linear if the current object (if it exists) is the result object.

btovar:
How did you generated new.log from 23.log?  Once I have new.log I can check what gprof tells us.

dthain:
See `deltadb_upgrade_log` which was just added.

btovar:
time ~/portal/src/cctools/deltadb/src/deltadb_query --file=23.log --from 2019-01-01 > /dev/null
3.16s user 0.03s system 99% cpu 3.189 total

time ~/portal/src/cctools/deltadb/src/deltadb_query --file=new.log --from 2019-01-01 > /dev/null
 2.54s user 0.01s system 99% cpu 2.557 total

The opposite form yours. Got similar numbers from several tries. I'll do a prof now.

dthain:
Interesting!  What machine?

btovar:
cclws16 from /tmp

btovar:
```text
new.log most used functions:
  %   cumulative   self              self     total          
 time   seconds   seconds    calls   s/call   s/call  name   
 27.12      0.32     0.32 57118012     0.00     0.00  jx_equals
  9.32      0.43     0.11  1684638     0.00     0.00  jx_remove
  8.47      0.53     0.10 23259042     0.00     0.00  jx_scan
  5.93      0.60     0.07  2047211     0.00     0.00  jx_parse_binary
  5.08      0.66     0.06 20727734     0.00     0.00  jx_getchar
  5.08      0.72     0.06  3867434     0.00     0.00  jx_copy
  3.81      0.77     0.05  7611640     0.00     0.00  jx_delete


23.log:
  %   cumulative   self              self     total          
 time   seconds   seconds    calls  ms/call  ms/call  name   
 19.74      0.15     0.15 16995292     0.00     0.00  jx_scan
 14.47      0.26     0.11 17078278     0.00     0.00  jx_getchar
 13.82      0.37     0.11  1610123     0.00     0.00  jx_parse_binary
 10.53      0.45     0.08  1067586     0.00     0.00  jenkins_hash
  5.26      0.49     0.04  1066618     0.00     0.00  hash_table_lookup
  5.26      0.53     0.04  1049726     0.00     0.00  jx_parser_create
  2.63      0.55     0.02 14813708     0.00     0.00  jx_unscan
```

That many jx_equals seem to come from jx_remove. The new version has more jx_remove's and those seem to come from jx_merge.
If there is a problem, it does not seem to be the lower level of the parser, as it shows similar numbers in both.

dthain:
Added a more efficient `jx_merge_into` which provides a dramatic improvement, at least on my macbook:

# Process log with U events.
time ./deltadb_query --file 23.log --from 2019-01-01 > /dev/null
**real	0m2.376s**
user 0m2.318s
sys	0m0.038s

# Process same log converted to M events:
./deltadb_upgrade_log 23.log new.log
time ./deltadb_query --file new.log --from 2019-01-01 > /dev/null
**real	0m1.295s**
user	0m1.267s
sys	0m0.020s

btovar:
Do we expect keys in objects not to be constant? (i.e., not string, integer, nor double)

If keys are always constant, then we can represent a jx_object with as table, not as list of pairs as we do now.  Looking for keys seems to be the bottleneck now.

dthain:
I believe JSON requires keys to be strings, and I'm inclined to make the same assertion for JX.
But that optimization can wait for another PR.  If we can go back and cut down the catalog history
data (and query time) by a factor of two using this, I'll be quite happy...

dthain:
Ah, there is still one missing piece to discuss here:
Is there an effective way to represent a removed item in an "M" record?

Under the old method, a removed item would result in an "R" record being logged in the database.
The new code, as written, reflects the removal of an item by giving it the value `null` in the M record.
This work, but it then makes it impossible to record an actual value of null, if that's important.

I see a few ways forward:
1 - Use `null` to represent removed data, as written.
2 - Use more obscure JSON expression than `null`, perhaps `[ ]`
3 - Use a JX-specific value, perhaps `undef`, which is clear but not portable.
4 - Continue using R records as before to reflect removals. 

@btovar what do you think?

btovar:
I don't like 1, as null is a valid json keyword.

I don't like 2, as [ ] it is a valid json value.

3 is only a problem when producing output to be consumed outside JX. Thus, any field with undef can be dropped from the output. No client should have the need to submit a field with 'undef'.

I don't mind 4 too much. We could represent an M record with an object and a list. The object has all the fields with their last known proper values, and the list the names of the fields that have been removed. In this way U and R are records are processed in a similar way.

dthain:
Ok, I'm going with #4, esp given the relatively small number of Rs in the data.
Let's keep it simple...

dthain:
An initial test conversion of the 2019 (so far) data from the old format to the new format, running on crcfe01:
```
# old format size:
1.9G	    /afs/crc.nd.edu/group/ccl/data/catalog.deltadb/2019
# new format size:
715M    /afs/crc.nd.edu/group/ccl/data/test.deltadb/2019

# Time to perform a full scan over the 2019 data

# Old data format:
time ./deltadb_query --db /afs/crc.nd.edu/group/ccl/data/catalog.deltadb --filter 'type=="wq_master"' --from 2019-01-01 > /tmp/wq.data
real	1m49.246s
du --si /tmp/wq.data
537M /tmp/wq.data

# New data format:
time ./deltadb_query --db /afs/crc.nd.edu/group/ccl/data/test.deltadb --filter 'type=="wq_master"' --from 2019-01-01 > /tmp/wq.data
real	0m42.143s
du --si /tmp/wq.data
135M	/tmp/wq.data
```

dthain:
Cancel that, the quick look at the data shows the conversion didn't work as desired...

dthain:
Let's try this again, with a properly working `deltadb_upgrade_log`:
```
# Size of old format data:
du --si /afs/crc.nd.edu/group/ccl/data/catalog.deltadb/2019
1.9G	/afs/crc.nd.edu/group/ccl/data/catalog.deltadb/2019

# Time to filter all wq data into a separate stream:
time ./deltadb_query --db /afs/crc.nd.edu/group/ccl/data/catalog.deltadb --from 2019-01-01 --filter 'type=="wq_master"' > /tmp/wq.data
real	1m59.434s

# Size of the filtered stream:
du --si /tmp/wq.data
322M	/tmp/wq.data

# Time to run a query on the filtered stream:
time ./deltadb_query --file /tmp/wq.old.data --from 2019-01-01 --output 'SUM(tasks_running)' --every 1d
real	0m18.185s

# Same thing with new format data:
du --si /afs/crc.nd.edu/group/ccl/data/test.deltadb/2019
972M	/afs/crc.nd.edu/group/ccl/data/test.deltadb/2019

# Time to filter all wq data into a separate stream:
time ./deltadb_query --db /afs/crc.nd.edu/group/ccl/data/test.deltadb --from 2019-01-01 --filter 'type=="wq_master"' > /tmp/wq.data
real	1m5.536s

# Size of filtered data:
du --si /tmp/wq.data 
168M	/tmp/wq.data

# Time to query filtered data:
time ./deltadb_query --file /tmp/wq.data --from 2019-01-01 --output 'SUM(tasks_running)' --every 1d
real	0m9.676s
```

So, we are getting a pretty substantial savings in both time and storage.

However, also note that the WQ-only data is quite a bit smaller than the total database, which has chirp, factories, catalogs, etc...  In the long run, we would get more benefit out of partitioning data by the `type` field so that it is not necessary to wander through every chirp load average update in order to get WQ statistics.

dthain:
Does the backup catalog look healthy at this point?

btovar:
It does.

dthain:
This last commit just addresses a memory leak when processing the final event during a log conversion.

dthain:
I think it's time to put it into production, do you see any show stoppers?

btovar:
I don't. I'll generate a new stable release including it.

PR #2011, opened 2019-01-29
Unary minus in JX

### LINKED ISSUE #2007: jx parses +number, -number as jx_operator instead of jx_NUMBER
This could be observed with jx_constant, where all jx_operator were marked as not constant.
Hot fix #2006 fixes jx_constant for these cases, but the correct solution is to change the parser.

trshaffer:
Fixes #2007 

This reverts the hotfix in #2006 and makes the changes in the parser. For the specific case of a unary minus preceding a numeric constant, we carry out the evaluation at parse time. JSON doesn't allow a preceding plus, but we could allow that if someone wants it.

### DIFF
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -621,27 +621,6 @@ static int jx_pair_is_constant(struct jx_pair *p) {
 		&& jx_pair_is_constant(p->next);
 }
 
-//hot fix for unary + and - with numbers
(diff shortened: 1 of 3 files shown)


btovar:
Tim it works for '-'. Could you extend to '+' too?

trshaffer:
That should do it

btovar:
Excellent, thanks Tim.

PR #2017, opened 2019-02-05
Add function to canonicalize JX

trshaffer:
Picking up from #1942, this PR introduces a `jx_canonicalize` function for rendering a JX object as a string in canonical form. This is intended for both generating a hash of some JX or for producing a stable representation. I allowed a conservative subset of JX:
- Only plain JSON types
- Object keys must be strings
- Object keys must be unique

This is to ensure that it's straightforward to decide how things should canonicalize without saying anything about future language changes. This canonical form is not great for general printing because of the aforementioned restrictions, lack of support for whitespace padding, and weird formatting of floats. It should, however, produce a stable JSON representation suitable for hashing or use as a dictionary key or what have you. For hashing uses like in #1893, the intended procedure is
1. Make a JX object and populate it with the fields identifying whatever you're hashing.
2. Call `jx_canonicalize` to get a string in canonical form.
3. Take the SHA1 or whatever hash.

For archive mode uses, the procedure is similar, but the canonical string itself is written to a file, which can then be hashed or viewed. The hash should remain the same across both uses.

### DIFF
--- a/dttools/src/Makefile
+++ b/dttools/src/Makefile
@@ -64,6 +64,7 @@ SOURCES = \
 	jx_parse.c \
 	jx_print.c \
 	jx_pretty_print.c \
+	jx_canonicalize.c \
 	jx_table.c \
 	jx_export.c \
 	jx_eval.c \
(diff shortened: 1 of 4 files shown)


dthain:
Looks good to me!

trshaffer:
Then if @btovar or @nhazekam don't have any changes in mind, I'll merge this

PR #2018, opened 2019-02-08
Python work_queue library that allows for asynchronous wait

btovar:
It should work as drop-in replacement for the classic work_queue
module.

HIGHLY EXPERIMENTAL CODE, IT MAY NOT WORK AT ALL

BEHAVIOUR DIFFERENCE: queue transfers/receives tasks even when wait() is not called.

INTERFACE DIFFERENCE: q.wait(0) returns immediately if no task to return is ready. Also, the new method q.receive() is the same as q.wait(0)

It mainly captures submit and wait calls. Tasks submitted get added to a
queue for later real submission. Returned tasks are put into a queue to
be retrieved at the client's convenience. Any other call is made with a
lock (that is, no two work queue calls can be made at the same time.)

### DIFF
--- a/work_queue/src/bindings/python/Makefile
+++ b/work_queue/src/bindings/python/Makefile
@@ -33,6 +33,6 @@ test:
 install: all
 	mkdir -p $(CCTOOLS_PYTHON_PATH)
 	chmod 755 work_queue_example.py
-	cp work_queue.py $(WQPYTHONSO) $(CCTOOLS_PYTHON_PATH)/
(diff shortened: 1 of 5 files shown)


dthain:
@btovar I like this approach, perhaps we can think of this as the first of (several) possible wrappers around the WQ interface that (cleanly) provide additional semantics.  Another example might be to have a wrapper that provides duplication of straggling tasks near the end of a workflow.

dthain:
Are there any obvious hazards to making this available as the non-default option?
e.g. Document this as "If plain work_queue doesn't fit your model, then use work_queue_async instead"

btovar:
Sounds good. Added documentation and an example, and will merge shortly.

PR #2033, opened 2019-02-28
add r api for work_queue

charleszheng44:


btovar:
Charles, please add an example on how it can be used.

charleszheng44:
will do

dthain:
Also needs updates to the manual in `doc/workqueue.html` and the example should go on the main WQ webpage alongside the C/Perl/Python examples.

charleszheng44:
@btovar @dthain can we move the definition of `struct work_queue` to work_queue.h. swig will not create R class for struct defined in ".c" file, which will cause an error when we want to call `work_queue_create` later in the R script

dthain:
Hmm, the problem here is that the `work_queue` structure is deliberately opaque.  We don't want clients of WQ messing around with the structure, hence we only return a typed pointer without returning the structure itself.

The other language bindings for WQ don't seem to need this.  Is this a problem particular to R?

charleszheng44:
I guess so.

charleszheng44:
I move the `struct work_queue` to work_queue.h just for now, and added the work_queue_example.R you can try out

dthain:
A couple of things:
1 - @btovar Can you help us to understand the problem with `struct work_queue` here?
2 - @charleszheng44 is it necessary to have a separate `work_queue.i` in this directory?  I think we want to have a common SWIG input for all of our language bindings?
3 - @charleszheng44 unless you have a specific use case in mind, please remove the bindings for Chirp; I don't think we have the bandwidth to support those as well.

Issue #2030, opened 2019-02-26
Synchronize Help/Options/Manual


dthain:
While looking into some recent user issues, I realized some of our help info is getting rather unsynchronized and difficult to read.

Everyone listed here, please take **one** of the following tools.  (Add a comment here so we know which one you are working on)

@trshaffer @nkremerh @nhazekam @charleszheng44 @btovar

- makeflow
- makeflow_viz
- work_queue_factory
- work_queue_status
- catalog_server

And sync up the documentation as follows:

1 - Fix up the --help text display so that it fits cleanly in a standard 80-column window.
2 - Make sure the --help display matches the actual options parsed by getopt.
3 - Make sure the man page matches the help display.

Make a PR and link it back to this issue.

nkremerh:
I will take work_queue_factory.

nhazekam:
I will take makeflow.

trshaffer:
I'll take the `catalog_server`

charleszheng44:
I will take wq

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

dthain:
@charleszheng44 @nhazekam Please finish up your doc updates today so we can wrap this one up today and move forward...

PR #2038, opened 2019-03-11
Makeflow usage

nhazekam:
Update the options to reflect which options are parsed and limit their length. Also update the corresponding man page.


nhazekam:
Addresses #2030

PR #2045, opened 2019-03-27
do not allow resumission of active task structures

btovar:
Fix issue where an active task structure is resubmitted, e.g.,  two q.submit(task) with the same task, as protection for naive task cloning.


dthain:
That looks like the right idea to me, although as a user, I'm not sure I would understand the error message given back from work_queue_submit.  Maybe "Task %d was submitted twice"  or something like that?

PR #2047, opened 2019-03-29
R WQ bindings

btovar:
Supersedes #2033


dthain:
This was a good idea at the time, but I think our understanding has evolved, and we should use the json-server approach going forward to provide new language bindings.

PR #2048, opened 2019-04-03
allow work queue to drain workers on given host

charleszheng44:
Implement the worker draining feature.

1. By calling `work_queue_drain_worker(struct work_queue *q, const char *hostname)`, workers on the given host will be drained (process to finish running task and disconnect from master)
2.  Add `draining_flag` to `work_queue_worker`
3. Master doesn't dispatch tasks to workers that have `draining_flag` been set to 1

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -213,6 +213,7 @@ struct work_queue_worker {
 	char addrport[WORKER_ADDRPORT_MAX];
 	char hashkey[WORKER_HASHKEY_MAX];
 	int  foreman;                             // 0 if regular worker, 1 if foreman
+	int  draining_flag;                       // 0 if not, 1 if worker is being drained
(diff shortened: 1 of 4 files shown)


work_queue/src/work_queue.c
```diff
@@ -6002,6 +6032,24 @@ int work_queue_shut_down_workers(struct work_queue *q, int n)
 	return i;
 }
 
+void work_queue_drain_worker(struct work_queue *q, const char *hostname)
```
btovar: Since you added this function, we don't need work_queue_draininglist_*, nor q->worker_draininglist.

Add a draining_flag to work_queue_worker instead, and use this flag in  check_hand_against_task instead of hash_table_lookup(q->worker_draininglist, w->hostname) != NULL

charleszheng44:
@btovar I have changed the code based on your request

btovar:
See #2063

PR #2050, opened 2019-04-08
Fix jx_merge()

trshaffer:
I ran into a bug when trying to override JX variables on the command line in Makeflow. Turns out `jx_merge()` was reversing the order of the list of pairs in an object, which led to old values mysteriously reappearing. This PR replaces that logic, doing a wholesale copy of each object then splicing the pieces together.

### DIFF
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -603,15 +603,16 @@ struct jx  *jx_copy( struct jx *j )
 struct jx *jx_merge(struct jx *j, ...) {
 	va_list ap;
 	va_start (ap, j);
-	struct jx *result = jx_object(NULL);
+	struct jx_pair *result = NULL;
 	for (struct jx *next = j; jx_istype(next, JX_OBJECT); next = va_arg(ap, struct jx *)) {
-		for (struct jx_pair *p = next->u.pairs; p; p = p->next) {
-			jx_delete(jx_remove(result, p->key));
-			jx_insert(result, jx_copy(p->key), jx_copy(p->value));
-		}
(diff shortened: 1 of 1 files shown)


dthain:
Hmm, this bug description makes we worry that we have other issues going on.  JX objects are by definition un-ordered, so if we have code sensitive to that, it sounds like a problem.  Also, JX objects shouldn't have duplicate keys, so that problem needs to be addressed as well.  (Though not necessarily at `jx_insert`).  Is this solving the right problem?

dthain:
I would be worried about losing an O(1) `jx_insert`.  Can we separate the use cases (check for unique vs known to be unique) into two different APIs?

trshaffer:
Do we have any applications that actually rely on non-unique or non-string keys? Non-string keys aren't allowed in JSON at all. Duplicated keys are one of the warts on JSON that implementations often forbid, anyway. I would be in favor of @btovar's suggestion to switch to a hash table of strings rather than making the API more awkward to support these features.

dthain:
I don't object to switching to all-string keys.  That would eliminate a number of other problems an unnecessary overheads.

I'm not so sure about hash tables.  The typical JX object is relatively small, and I wonder if the complexity is worth the trouble.  If you do go that way, it will be necessary to measure the performance accurately.

trshaffer:
Since this is a bug that has affected @nhazekam too, I propose merging this PR and opening a separate issue to track changes to the JX object implementation, as that will likely involve a couple pull requests and more discussion.

dthain:
OK

trshaffer:
Continued in #2051

PR #2063, opened 2019-05-10
adds work_queue_specify_draining_by_hostname

btovar:
Supersedes #2048 without the need to modify work_queues communication protocol. It also adds perl and python bindings.


dthain:
Looks like the worker is failing in single-shot mode (-S) when WQ tests are run.

Issue #437, opened 2014-03-24
WQ bindings can't handle off_t


batrick:
```
$ python2
Python 2.7.6 (default, Feb 26 2014, 12:07:17) 
[GCC 4.8.2 20140206 (prerelease)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from work_queue import *
>>> cctools_debug_config_file_size(1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: in method 'cctools_debug_config_file_size', argument 1 of type 'off_t'
```

We need to map off_t to a type python can handle. Incoming PR that fixes this.

nhazekam:
This created a patch for the issue, but limits the size of `off_t` to `INT_MAX`.
If the bindings are used and a file piece starts or ends at a location higher than `INT_MAX`, the
bindings will throw an `OverflowError'.

```
Traceback (most recent call last):
...
    t.specify_file_piece(file.parent.data_name, file.data_name, file.start_byte, abs(file.end_byte -1), WORK_QUEUE_INPUT)
  File "/afs/crc.nd.edu/user/n/nhazekam/cctools/lib/python2.7/site-packages/work_queue.py", line 1696, in specify_file_piece
    return work_queue_task_specify_file_piece(self._task, local_name, remote_name, start_byte, end_byte, type, flags)
OverflowError: in method 'work_queue_task_specify_file_piece', argument 5 of type 'off_t'
```

The size of end_byte for this was 2795680124.
This can be addressed by handling `off_t` as something larger than an integer, but I am not sure what is suitable.

btovar:
It seems we need to check whether a compile-time macro definition before we can switch to off64_t. I am investigating how to that safely.

nhazekam:
Re-addressed in #2070. `off_t` is now converted to `long long int` from python instead of `int`.

PR #2070, opened 2019-06-04
fix off_t in work_queue.i

btovar:
Fix for #437


dthain:
@nhazekam does this work for you?

PR #2074, opened 2019-06-17
makeflow ec2 setup updated

epyifany:
Thanks to Prof. Thain and Ben's suggestions, I have updated the makeflow_ec2_setup in cctools to a python2/3 script that is able to handle and parse json responses from the AWS. This is an updated and more sustainable version of the script and will be compatible for different users under different enviornments. -Yifan Yu

### DIFF
--- a/makeflow/src/makeflow_ec2_setup
+++ b/makeflow/src/makeflow_ec2_setup
@@ -1,94 +1,126 @@
-#!/bin/sh
+#!/usr/bin/env python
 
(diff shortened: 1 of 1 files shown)


dthain:
@epyifany please change the shebang line to use cctools_python, which will check for and use a specific python version.  Take a look at the makeflow_monitor for an example of how to use it correctly.

dthain:
Looks good, thank you.

PR #2076, opened 2019-06-24
Add flag to use previous logic for job names

trshaffer:
This PR adds a flag to switch to pre f892d75f behavior. This is helpful for figuring out which job/command is running. This is not the default, as we shouldn't assume that the batch system will handle it.


dthain:
That looks about right.

trshaffer:
It's showing up as green and I tested that this works as expected, so merging

PR #2088, opened 2019-07-09
Updated conda build recipe

eamonlm:
I've added support for tests in the meta.yaml, and changed the build script to exclude weaver when installing with Python 3 as the python version.


btovar:
Looks good, thanks @eamonlm

Issue #2016, opened 2019-02-05
Catalog: Split History by Type?


dthain:
Historical queries of the catalog are often expensive b/c of the need to scan data irrelevant to the current query.  e.g. There are lots of chirp loadavg/storage updates that dominate the wq_master updates by an order of magnitude.

One solution to this is to partition data by the static "type" field of the advertisement, and then store these in separate instances of `jx_database`, so the directory structure becomes `history/{type}/{year}/{day}.[log|ckpt]`

This would require several steps of evolution:
1 - Implement a new parent object `jx_multi_database` which creates and writes`jx_database` instances on demand as new type fields are observed, and merges states on reads.
2 - Modify the catalog server to use `jx_multi_database` instead of `jx_database`.
3 - Write a tool to convert historical data to the new format, by reading it from a `jx_database` and inserting into a `jx_multi_database`

It may not be necessary to modify `deltadb_query` b/c it will still work against the single database format, which remains unmodified.

Builds on PR #2009

btovar:
@dthain, was this implemented in #2093?

dthain:
Yes, however, #2093 still needs a conversion routine from old to new before I'm prepared to merge it.

dthain:
Explored this idea, implemented it, evaluated it, and the payoff wasn't worth the trouble.  Live and learn!

PR #2093, opened 2019-07-16
JX Multi Database

dthain:
Splits catalog history into multiple databases on the fly, one for each record type.
Still needs a conversion routine to move the old format into the new format.
Addresses #2016


dthain:
WIP

PR #2095, opened 2019-07-17
Python monitoring

btovar:
adds function decorator to monitor resource usage in python functions

work in progress


dthain:
Great timing.  :)

Issue #2094, opened 2019-07-16
Catalog Update Processing Bug


dthain:
We have periodic instances of garbage data in the catalog server, starting on May 14th and continuing through July.  It appears to happen to several different kinds of updates, both wq_master and chirp, where we get lots of duplicated JSON fields, almost as if the parser/printer is stuck in an infinite loop.

JX per se hasn't changed much recently, except for the jx_merge PR, which was merged in April:

https://github.com/cooperative-computing-lab/cctools/pull/2050/files

Can you take a very careful look at jx_merge (and how it is used) to make sure we aren't missing something?

Here is the first instance of corrupt data, notice that the "categories" field keeps repeating:

/afs/crc.nd.edu/group/ccl/data/catalog.deltadb/2019/132.log

```
...
T 1557769005
[...]

PR #2098, opened 2019-07-17
Hotfix for Catalog Server

trshaffer:
This is a fix for #2094 

Putting aside any changes to JX objects for the moment, I simply reverted to the original `jx_merge()` implementation and added a workaround for the motivating problem. I tested against the minimal case in the linked issue and didn't observe any corruption.

### DIFF
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -603,16 +603,15 @@ struct jx  *jx_copy( struct jx *j )
 struct jx *jx_merge(struct jx *j, ...) {
 	va_list ap;
 	va_start (ap, j);
-	struct jx_pair *result = NULL;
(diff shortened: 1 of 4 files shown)


dthain:
@btovar please take a quick look at this -- we want to get out a quick fix without causing any further rippples...

dthain:
@trshaffer please also add a test case that exercises jx_merge in several ways, so we can avoid a regression in the future...

btovar:
New version has been up for about two hours now (since 7:10am this morning). Catalog seems to be working as usual.

trshaffer:
@btovar did you try killing and restarting, to see if log replay works correctly?

dthain:
@btovar if this version works correctly after a restart, then please merge and issue a release.

PR #2099, opened 2019-07-18
 JX Binary Representation

dthain:
An experiment: `jx_binary_read` / `jx_binary_write` encode JX to stdio streams in a binary representation.  Does it pay to do so?

### DIFF
--- a/dttools/src/Makefile
+++ b/dttools/src/Makefile
@@ -154,7 +154,7 @@ PROGRAMS = $(MOST_PROGRAMS) catalog_query
 
 SCRIPTS = cctools_gpu_autodetect cctools_python
 TARGETS = $(LIBRARIES) $(PRELOAD_LIBRARIES) $(PROGRAMS) $(TEST_PROGRAMS)
-TEST_PROGRAMS = auth_test disk_alloc_test jx_test microbench multirun jx_count_obj_test histogram_test category_test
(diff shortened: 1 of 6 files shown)


dthain:
Sample results on 465KB of JSON obtained from `catalog_query`:

```
./jx_binary_test test.json test.bson test.out
text    read 20069 us
binary write 13453 us
binary read 5164 us
text  write 35005 us
```

Writing binary format is about 2x faster, and reading it is about 3x faster, since parsing is much simpler.  However, the binary data is only marginally smaller than the text representation:

```
[dthain@cclws15 src]$ ls -la test.json test.bson test.out
-rw-r--r-- 1 dthain campus 465191 Jul 17 15:44 test.json
-rw-r--r-- 1 dthain campus 413405 Jul 18 09:13 test.bson
-rw-r--r-- 1 dthain campus 463832 Jul 18 09:13 test.out
```

dthain:
This is similar in spirit to BSON (bsonspec.org) but note that BSON has many elements with no corresponding representation in JSON, which makes for some puzzling compatibility problems.

trshaffer:
This looks more like [MessagePack](https://msgpack.org/), which is similar to BSON without introducing a lot of extra types.

dthain:
Aha, MessagePack looks more thought out than BSON.

dthain:
This isn't hooked up to anything yet, but making it available for future uses...

PR #2102, opened 2019-07-18
Nested Workflows

dthain:
Clean implementation of nested workflows.
Supersedes #1808, since much has changed in Makeflow since then.
WIP


dthain:
@nhazekam please take a look at this and see if it fits the new structure well.  The main problem remaining is that the temporary file created to store the jx args never gets deleted.  Would this be better implemented as a hook, or should I simply flag the file as deletable in some way?

Two things still need to be implemented:
1 - Passing the resources down to the makeflow via --local-cores, etc.
2 - Detecting whether the sub-makeflow is make/jx so as to get the command-line arguments right.

dthain:
While the Make syntax obviously can't express everything desirable, I would like it to work as far as it can.  (There is a test case for both make and jx in there, although I will rename for clarity.)

I changed the keywords in both cases to be "workflow" instead of "makeflow" so that we aren't confusing brands with functionality.

Although now that you bring it up, I see that we have a problem detecting whether the workflow is in make or jx form.  I will have to add an option to control that...

dthain:
@nhazekam One more problem: this code generates a temporary file to store the jx_args for the workflow, and the file doesn't get cleaned up.  What's the right way to ensure that the file gets removed, regardless of the job disposition?

nhazekam:
This is the current dilemma that exists in the intersect of `dag_node` and `batch_task`. If the file is added only to the `batch_task` it is not tracked by makeflow, or the `dag` more specifically. The way this is handled with hooks right now is that it is added to the `batch_task` and `dag` as a `DAG_FILE_TYPE_TEMP`. Then temp files are removed when the node is completed.

I think to make this cleaner it would be better to clean all files from the returned `batch_task` that are not tracked by the `dag` as well. Then files added by hooks that are temporary would not be added to the `dag` at all. We still have the nebulous issue of should the `dag` track global outputs that are the result of a hook. An example of this would be the resource outputs of the tasks that may need to analyzed later.

One issue with this approach is that the possible existence of these file is not tracked, so if the makeflow fails mid run the files not written to the log are unknown and will never be cleaned by makeflow.

dthain:
Build failures are due to network timeouts in Travis.

PR #2104, opened 2019-07-19
Hook enforcement

nhazekam:
Converted parrot_run file enforcement from wrapper to module, using hooks.

Notable change from previous version:
Previously: File list was written to enforcement script, which was written to the mountlist at the site. From what I understand, this was done so the environment variable were evaluated on site just prior to run time.
Proposed: The mount file is full written on the master with environment variable for PWD and HOME in the list. This is then updated on site using `envsubst` and a temporary file. I think this could also be done with `sed` or `awk`, but `envsubst` reads cleaner.

`--enforcement` is superfluous at this point as specifying `parrot_path` is required. This could be changed to look for `parrot_run` locally or the option could be removed. Open to suggestions.

This is not a change, but just an observation. In this version and the previous version the user only specifies where `parrot_run` can be found, but are not allowed to change the local executable or prefix values for the enforcement script, mount list, and tmp directory name. That would only require minor changes if desired.

Prior to merging (once any changes to the command line options are decided) I will updated the documentation.


nhazekam:
After discussing with @trshaffer I will move the list translation to `sed` or `awk` as they are more readily available than `envsubst`.

makeflow/src/makeflow_module_enforcement.c
```diff
@@ -0,0 +1,225 @@
...
+
+static int create( void ** instance_struct, struct jx *hook_args ){
+	struct parrot_enforce_instance *p = parrot_enforce_instance_create();
+	*instance_struct = p;
+
+	if(jx_lookup_string(hook_args, "parrot_path")){
+		p->parrot_path = xxstrdup(jx_lookup_string(hook_args, "parrot_path"));
+		debug(D_MAKEFLOW_HOOK, "setting Parrot binary path to %s\n", p->parrot_path);
+	} else {
+		debug(D_NOTICE, "parrot_path must be set for parrot enforcement");
```
trshaffer: I wouldn't be opposed to checking `$PATH` if `--parrot-path` isn't given

makeflow/src/makeflow_module_enforcement.c
```diff
@@ -0,0 +1,225 @@
...
+{
+	struct parrot_enforce_instance *p = (struct parrot_enforce_instance*)instance_struct;
+	struct batch_wrapper *enforce = batch_wrapper_create();
+	batch_wrapper_prefix(enforce, p->enforce_prefix);
+
+	char *mountlist_path = string_format("%s%d", p->mountlist_prefix, n->nodeid);
+	char *tmp_path = string_format("%s%d", p->tmp_prefix, n->nodeid);
+
+	/* make an invalid mountfile to send */
+	int mountlist_fd = open(mountlist_path, O_WRONLY|O_CREAT, S_IRUSR|S_IWUSR);
```
trshaffer: Does this need to be an `open()` followed by an `fdopen()`?

makeflow/src/makeflow_module_enforcement.c
```diff
@@ -0,0 +1,225 @@
...
+
+	struct dag_file *df = makeflow_hook_add_input_file(n->d, t, mountlist_path, mountlist_path, DAG_FILE_TYPE_TEMP);
+	makeflow_log_file_state_change(n->d, df, DAG_FILE_STATE_EXISTS);
+
+	/* and generate a wrapper script with the current nodeid */
+	char *prefix = string_format("export MOUNTFILE='%s'", mountlist_path);
+	batch_wrapper_pre(enforce, prefix);
+	free(prefix);
+
+	batch_wrapper_pre(enforce, "envsubst < $PWD/$MOUNTFILE > $PWD/mount_tmp_file");
```
trshaffer: Are we guaranteed to have C running to do this as a library function?

makeflow/src/makeflow.c
```diff
@@ -1434,6 +1436,7 @@ int main(int argc, char *argv[])
 		{"jx-define", required_argument, 0, LONG_OPT_JX_DEFINE},
 		{"enforcement", no_argument, 0, LONG_OPT_ENFORCEMENT},
 		{"parrot-path", required_argument, 0, LONG_OPT_PARROT_PATH},
+		{"envreplace", no_argument, 0, LONG_OPT_ENVREPLACE},
```
trshaffer: `no_argument` here?

dttools/src/envreplace.c
```diff
@@ -0,0 +1,103 @@
...
+	int var_index = 0;
+	int valid_var = 0;
+
+	char c = fgetc(INPUT);
+	while (c != EOF)
+	{
+		if (c == '$'){
+			valid_var = 1;
+		} else if (valid_var && (
+				(c >= (char)65 && c <= (char)90) ||  // A-Z
```
trshaffer: Is this intended to only allow upper case? More generally, I think it would be better to use `isalpha()` and friends rather than raw ASCII codes

dthain:
Suggest that you demote envreplace to a library function `env_replace( const char *infile, const char *outfile )`.  Let's not introduce new potential failure modes for finding things in the PATH, etc...

dttools/src/envreplace.c
nhazekam: I will change this. I am open to either just upper or all alpha characters. Going by :

http://pubs.opengroup.org/onlinepubs/9699919799/

Environment variables for use in Shell consists of uppercase, underscore, and digit (when not first). I am open to other options.

nhazekam:
If this is used as a library function, where will this be called from on the execution site? I think this is what @trshaffer is referring to in an earlier comment. I may just be misunderstanding this though.

dthain:
Ah, I missed that part.  Ok, carry on.

dttools/src/envreplace.c
trshaffer: Interesting, I did not know POSIX only requires uppercase. I'm fine with leaving the character set as is, but still prefer `isupper()` and `isdigit()`

nhazekam: I agree. I have the code changed. I am just doing a couple of other cleanup options.

dthain:
@nhazekam regardless of what the standard says, both Mac and Linux allow for non-capitalized environment variables, so I think we ought to support them.  :+1:  for using isalpha and friends.

PR #2109, opened 2019-07-23
Python Environment Packaging Scripts

tjdasso:
This PR includes several scripts that analyze Python scripts and create isolated Conda environments with the necessary modules needed to run the script. Descriptions of each script as well as an example test can be found in the README.md file, but the scripts can be summarized as follows:

python_package_analyze: Takes a Python script, analyzes its Python version and all module imports, and generates a JSON file with the necessary data to create a Conda environment.
python_package_create: Takes the JSON file and creates a Conda environment with the appropriate Python interpreter and all the necessary modules installed within the environment. Packs the environment so that it is relocatable and can be sent to a different machine.
python_package_run: Wrapper script that unpacks and activates the Conda environment and runs the given command within the Conda environment.


dthain:
@btovar Is something wonky with TravisCI?  When I click through to the CI details, I get a blank page...  This PR isn't ooked into the build, so I wouldn't expect it to cause a failure...

btovar:
It works for me. The checks fail because this (unrelated?) test is failing:

--- Testing makeflow/test/TR_makeflow_alloc_output.sh ...         failure 0s

dthain:
@nhazekam it looks like we are getting intermittent build failures on all PRs, due to `TR_makeflow_alloc_output.sh`, which is your department.  The test works fine when I run it locally.

A hypothesis: could this test be generating more data than there is space available in the Travis container?  Please take a look and see what you can figure out.

dthain:
@tjdasso This looks pretty good overall.  Please work on the documentation a little more, and write it for the benefit of a completely new user who wants to run it.  e.g. Begin by explaining what problem this solves, then give some complete examples on how to use it, give a non-trivial example of the metadata listing, and then put the ideas for improvement at the end.

tjdasso:
I expanded on the documentation a bit more for each of the three scripts, hopefully its a little more clear to new users now.

dthain:
Looks good, thank you.

PR #2112, opened 2019-07-29
Remove wrappers

nhazekam:
This removes old/dead wrapper code. Moves wrapper implementation to hook/module setup.

I am testing this with different wrapper, but if there are wrapper of interest let me know.

One outstanding issue is how to guarantee that remote file names do not collide between wrappers and other hooks. If this is a problem in execution, interspersing `sandbox` layers will prevent collisions, but this is not always detected preemptively.


dthain:
Ready to merge?

nhazekam:
@dthain I think this is ready to go. I was planning on touching base with @nkremerh to see if there were wrappers I could test with to verify actual use cases, but the code should be good to go.

nhazekam:
I will also rebase to reconcile with `makeflow.c`.

Issue #2096, opened 2019-07-17
custom id field for records in the catalog 


btovar:
Records in the catalog server are identified by a key constructed from the address and port of the client. This limits the use of deltaDB as a client cannot submit and query different types of records. The particular desired use case is a client sending resource information per function. In the current implementation there is not a way to separate historical information per function.

For this use case, the client should be allowed to generate a different key per records histories. The client could include a special field in the record (such as 'id') which can be added to the construction of the record's key.

btovar:
Fixed by #2113

PR #2113, opened 2019-07-29
adds 'uuid' to identify records in catalog updates

btovar:
Fix for #2096

### DIFF
--- a/dttools/src/catalog_server.c
+++ b/dttools/src/catalog_server.c
@@ -209,7 +209,7 @@ static void update_all_catalogs()
 
 static void make_hash_key(struct jx *j, char *key)
 {
-	const char *name, *addr;
+	const char *name, *addr, *uuid;
 	int port;
 
 	addr = jx_lookup_string(j, "address");
@@ -222,7 +222,13 @@ static void make_hash_key(struct jx *j, char *key)
 	if(!name)
 		name = "unknown";
 
-	sprintf(key, "%s:%d:%s", addr, port, name);
+	uuid = jx_lookup_string(j, "uuid");
+	sprintf(key, "%s:%d:%s%s%.128s",
+			addr,
+			port,
+			name,
+			uuid ? ":" : "",
+			uuid ? uuid : "");
 }
 
 static void handle_update( const char *addr, int port, const char *raw_data, int raw_data_length, const char *protocol )
--- a/resource_monitor/src/resource_monitor.c
+++ b/resource_monitor/src/resource_monitor.c
@@ -127,6 +127,7 @@ See the file COPYING for details.
 #include <sys/types.h>
 
 #include "buffer.h"
+#include "catalog_query.h"
 #include "cctools.h"
 #include "copy_stream.h"
 #include "create_dir.h"
@@ -145,6 +146,8 @@ See the file COPYING for details.
 #include "xxmalloc.h"
 #include "elfheader.h"
 #include "domain_name_cache.h"
+#include "uuid.h"
+#include "random.h"
 
 #include "rmonitor.h"
 #include "rmonitor_poll_internal.h"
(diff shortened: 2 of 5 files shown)


dttools/src/catalog_server.c
```diff
@@ -222,7 +222,13 @@ static void make_hash_key(struct jx *j, char *key)
 	if(!name)
 		name = "unknown";
 
-	sprintf(key, "%s:%d:%s", addr, port, name);
+	custom_id = jx_lookup_string(j, CATALOG_RECORD_ID_FIELD);
+	sprintf(key, "%s:%d:%s%s%.128s",
```
dthain: Probably want to separate this with a colon as well.

dttools/src/catalog_server.c
```diff
@@ -222,7 +222,13 @@ static void make_hash_key(struct jx *j, char *key)
...
+			addr,
+			port,
+			name,
+			custom_id ? ":" : "",
```
dthain: Never mind, I see that you did.

dttools/src/catalog_query.h
```diff
@@ -20,6 +20,10 @@ Query the global catalog server for server descriptions.
 #define CATALOG_HOST (getenv("CATALOG_HOST") ? getenv("CATALOG_HOST") : CATALOG_HOST_DEFAULT )
 #define CATALOG_PORT (getenv("CATALOG_PORT") ? atoi(getenv("CATALOG_PORT")) : CATALOG_PORT_DEFAULT )
 
+/* Name for the automatically generated ids for records generated by
+ * non-services (e.g. the resource_monitor). */
+#define CATALOG_RECORD_ID_FIELD "cctools_catalog_record_id"
```
dthain: Suggest just keeping it `record_id` for consistency throughout.

btovar: I like better your idea of simply calling it 'uuid'. What do you think?

btovar:
> By default, the catalog server _observes_ the ip address and port of the sender, and then does a reverse DNS lookup to get the host name. If that fails, then it uses the `name` field from the JSON record. The constructed hash key is then `addr:port:name[:record_id]`, which uniquely identifies a thing to be reported.
> 
> The sender is currently generating a `uuid` as the `record_id`, as well as a human-readble `project` name and `task` name.
> 

Yes, that's it. I was thinking of using  'project'  as way to group tasks from a workflow, and 'task' to be a tag a user can set for a particular task.

dthain:
Ok, let's make it `uuid` then.

btovar:
According to me this one is ready to merge.

dttools/src/catalog_query.h
```diff
@@ -20,6 +20,10 @@ Query the global catalog server for server descriptions.
 #define CATALOG_HOST (getenv("CATALOG_HOST") ? getenv("CATALOG_HOST") : CATALOG_HOST_DEFAULT )
 #define CATALOG_PORT (getenv("CATALOG_PORT") ? atoi(getenv("CATALOG_PORT")) : CATALOG_PORT_DEFAULT )
 
+/* Name for the automatically generated ids for records generated by
```
dthain: Suggest just dropping this additional symbol and use "uuid" throughout, since it's going to be a strong convention for backwards compatibility..

dttools/src/catalog_server.c
```diff
@@ -209,7 +209,7 @@ static void update_all_catalogs()
 
 static void make_hash_key(struct jx *j, char *key)
 {
-	const char *name, *addr;
+	const char *name, *addr, *custom_id;
```
dthain: Call it uuid.

resource_monitor/src/resource_monitor.c
```diff
@@ -2221,6 +2282,11 @@ int main(int argc, char **argv) {
 		    {"snapshot-file",   required_argument, 0, LONG_OPT_SNAPSHOT_FILE},
 		    {"snapshot-events", required_argument, 0, LONG_OPT_SNAPSHOT_WATCH_CONF},
 
+		    {"catalog-task-name", required_argument, 0, LONG_OPT_CATALOG_RECORD_ID},
```
dthain: This is going to cause confusion -- the same concept is called `task-name` in the command line, `RECORD_ID` in the code, and `uuid` in the JSON.  Let's call it UUID throughout.

Issue #2090, opened 2019-07-10
Resource Monitor Update to Catalog


dthain:
Update the resource monitor so that it can send a periodic UDP message to a catalog server to report the resource consumption of a task over time.  Then, demo using Parsl + WQ.

dthain:
Nice demo, please add this as an option to the C resource monitor so we can use it from all environments.

btovar:
Done in #2114.

PR #2114, opened 2019-07-30
catalog reporting to the resource monitor

btovar:
works in conjuction with #2113.

Records are tagged with --catalog-record-id.
Additionally, they can be tagged with a project name --catalog-project.

--catalog indicates which catalog to use (usual defaults).
--catalog-interval how often in seconds an update is done (30s default).

Fix for #2090


btovar:
Included in #2113.

PR #2117, opened 2019-08-08
JX f-strings

### LINKED ISSUE #2107: JX String Formatting
cc @dthain @btovar @nhazekam 

In using JX for some of our workflows, the current string formatting is sort of a rough patch. Especially with highly-parametric workflows, things can get ugly for users (see the [BWA example](https://github.com/cooperative-computing-lab/makeflow-examples/blob/master/bwa/bwa.jx)). We currently have a few ways to work with strings:
- `format()` function
- Concatenation with `+`
- Automatic coercions from numbers, etc.

Python also has a number of string formatting facilities (`%` operator, `.format()`, etc.). Many of the awkwardness with strings has shown up there, so the Python people introduced f-strings in 3.6 as a more flexible option. The full description is given in [PEP 498](https://www.python.org/dev/peps/pep-0498/). To summarize, f-strings
- appear as string literals prefixed with `f` or `F`. This is not the first case of a prefix on a string in Python, but would be the first in JX.
- are expressions and not *actually* literals. They are evaluated at run time, not parse time. They can be evaluated multiple times in a loop body, etc. If there are side effects from the expressions, the evaluation order is defined left to right.
- support arbitrary expressions between curly braces. Literal curlies in f-strings are doubled as `{{` and `}}`.
- introduce some limitations on contents. Comments (with `#`) are not allowed in expressions (but are fine in the enclosing string). Backslash escapes are also not allowed.
- are somewhat awkward with strings for dictionary lookups, etc. Python sort of avoids this issue since they support both single and double quotes.

I'll give a few examples for JX here, and see if there are any other suggestions.

Current Syntax
-------------------

1. `format(“String %s, number %d”, VAR, 42 + i)`
2. `“file-” +  n + “.dat”`
3. `format("%s %s %s %s.%d > %s.%d.sam 2> %s.%d.err", BWA,ARGS,REF,INPUT_PRE,i,INPUT_PRE,i,INPUT_PRE,i)`

This syntax is very clear about how potentially complex expressions are handled, and fits within the current syntax without weird restrictions or incantations. For more complex substitutions, it can be a burden to work with (see example 3). It gets annoying when writing (and especially modifying) workflows to count over `%` fields and make sure the letter specifiers line up. This syntax is pretty familiar to C people or those using older Python.

F-strings
-----------

1. `f“String {VAR}, number {42 + i}”`
2. `f“file-{n}.dat”`
3. `f"{BWA} {ARGS} {REF} {INPUT_PRE}.{i} > {INPUT_PRE}.{i}.sam 2> {INPUT_PRE}.{i}.err"`

This syntax is much friendlier, especially for longer commands like example 3. Python's version allows for arbitrary expressions as in example 1. We could instead allow a restricted form that only supports bare variable names, to avoid some of the issues above with quoting and evaluation. As mentioned in the meeting, the prefix indicates that something weird is happening, and makes this functionality opt-in without changing the behavior or normal strings. This is somewhat foreign-looking syntax for me, but it's easy to explain to users who know Python (though this could be a liability if we don't support the full syntax). 

Implementation wise, these are a little awkward. I would be inclined to treat `f` as a unary prefix operator. The parser would validate that the string is well formed and we could print it back correctly, without intermediate sugar leaking out. If we only allow bare variables, then we can simply use the string as the operand and look up the variables at evaluation time. Allowing full expressions would either require additional parsing passes at eval time (ick) or some awkward in-memory representation as a list of string literals with expressions interspersed. We would need plenty of comments explaining how things need to look in memory, but none of this is user-visible.

New Format Function
-------------

1. `f(“String {VAR}, number {42 + i}”)`
2. `f(“file-{n}.dat”)`
3. `f("{BWA} {ARGS} {REF} {INPUT_PRE}.{i} > {INPUT_PRE}.{i}.sam 2> {INPUT_PRE}.{i}.err")`

Another suggestion was to introduce a new function that formats strings with substitutions rather than using a prefix. This would have many of the same advantages as f-strings, and we could still choose between variables only and arbitrary expressions. As a user, the function syntax throws me a bit. We haven't elaborated JX functions to a great extent, but we haven't given users any reason to think that functions don't work as usual, i.e. values go in as arguments, and a value comes back. This syntax implicitly uses the *calling* context, like C macros. I'm not sure if users would care, but I'm hesitant about something looking like a function but pulling in arguments/context implicitly. For explaining to new users, I wouldn't want to teach this as Python f-strings but written as functions, because I don't have a good reason (that doesn't involve digging around in our implementation) for why we didn't use an `f` prefix. If we're doing a more restricted form with bare variables only and will never allow expressions, this could be helpful so users don't expect things to act like Python.

Implementation wise, this is very simple. We would add another function called `f()` or what have you, without special parsing and printing required. Supporting arbitrary expressions without introducing special printing and such might require eval-time parsing. But this would be a fairly self-contained addition.

trshaffer:
This PR adds limited f-strings to JX. I also added a number of test cases and updated the docs.

Rather than allowing arbitrary expressions, this PR only allows variable names. I implemented f-strings as a unary prefix operator that the parser only allows on string literals. At evaluation time, invalid f-strings produce helpful error messages. I had considered adding a new fundamental type `JX_FSTRING` rather than doing an operator. The former is more flexible but requires more cross-cutting changes. Since I don't see much interest in generalizing this feature to arbitrary expressions, I did the latter.

Resolves #2107

### DIFF
--- a/doc/jx-quick.html
+++ b/doc/jx-quick.html
@@ -127,7 +127,7 @@ true | false
 <h2>Operators</h2>
 <pre>
 a["b"]   func(x)
-* % /
+* % /    f"{x}.dat"
 + -
 == != < <= > >=
 not
--- a/doc/jx-tutorial.html
+++ b/doc/jx-tutorial.html
@@ -202,16 +202,15 @@ instance of the job:
 
 Note that the value of <tt>N</tt> is substituted into both the commands
 string and the output list by using the plus sign to indicate string
-concatenation.  If you prefer a more <tt>printf</tt> style, you can
-use the <tt>format()</tt> function to insert values into strings
-into places indicate by percent codes:
+concatenation.  If you prefer a more compact style, you can
+use Python-style f-strings to insert values into strings:
 
 <code>{
    "rules": [
       {
-         "command" : <b>format("./simulate.py -n %d > output.%d.txt",N,N)</b>
+         "command" : <b>f"./simulate.py -n {N} > output.{N}.txt"</b>
          "inputs" : [ "simulate.py" ],
-         "outputs" : [ "output."+N+".txt" ],
+         "outputs" : [ <b>f"output.{N}.txt"</b> ],
       } for N in [ 1, 2, 3 ]
    ]
 }
@@ -250,7 +249,7 @@ Of course, it would be better to generate the list automatically.
 The list of output files is easy: just generate them using
 a list comprehension.  For example, this expression:
 
-<code>[ "output."+N+".txt" for N in [1,2,3] ]
+<code>[ f"output.{N}.txt" for N in [1,2,3] ]
 </code>
 
 Would evaluate to this array:
@@ -273,8 +272,8 @@ Evaluates to this string:
 We can put all of those bits into a single job, like this:
 
 <code>{
-   "command" : "cat "+join(["output."+N+".txt" for N in [1,2,3]])+" >output.all.txt",
-   "inputs" : [ "output."+N+".txt" for N in [ 1, 2, 3 ] ],
+   "command" : "cat "+join([f"output.{N}.txt" for N in [1,2,3]])+" >output.all.txt",
+   "inputs" : [ f"output.{N}.txt" for N in [ 1, 2, 3 ] ],
    "outputs" : [ "output.all.txt" ],
 }
 </code>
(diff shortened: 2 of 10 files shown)


dthain:
@btovar what do you think?

btovar:
The pr looks good to me.

Since 'f' can only be applied to string literals, we eliminate some of the problems that appear with shell quoting and variable expansion (i.e., there is not a way to create an 'f' string programatically).

The only possible issue I see is the use of 'f', as someone may expect that it evaluates full expressions as in python. But they would get an appropriate error message, and choosing a different letter from 'f' may be more confusing.

Issue #2035, opened 2019-03-01
Test suite outside compilation process


btovar:
Develop a test suite that runs outside the building process of CCTools. These test should run on the result of make install.

dthain:
Some initial thoughts:
1 - Move all of the `*/test` files into `/test`.
2 - Modify the tests to rely on the existing PATH and PYTHONPATH, rather than relative paths.
3 - Take the various test programs (`auth_test`,`jx_test`,etc) out of the source tree, and put them in the test tree to be built upon demand, not installed.
4 - Set up the top-level test script to perform a make install, then set the PATH/PYTHONPATH and run the tests on the installed version.

PR #2119, opened 2019-08-12
WIP: External Test Setup

dthain:
This is a first pass at cleaning up the test suite to operate external to the build system.  I didn't convert all the tests in one pass, but did a handful to test the idea.  I'm attempting to minimize what each component knows about the whole system, so it ought to be possible to run the test suite on a build that came via pip, conda, spack, etc without knowing the config.mk that came out of that build.

Here is the idea:

1 - Each test operates independently and expects that the appropriate `PATH`, `PYTHONPATH`, etc has already been set to point to the target test installation.  The test only uses assets in the cwd and relies on binaries, etc already being in the path. If it needs to compile something, it uses `CCTOOLS_COMPILE` or `CCTOOLS_COMPILE_STDIN` to invoke the compiler appropriately.

2 - `test_runner_common.sh` expects only `CCTOOLS_INSTALL_DIR` to be set, and uses that to construct an appropriate `PATH`, `PYTHONPATH`, etc for each test.  `CCTOOLS_COMPILE` is defined to use `CCTOOLS_INSTALL_DIR` to find `include`, `lib`, etc.

3 - `run_all_tests.sh` does not attempt to parse `config.mk` but just relies on `CCTOOLS_INSTALL_DIR` being set, and then invokes all of the tests.  (Still need to pass in what subset of tests to run somehow.)

@btovar @trshaffer  does this seem like a sensible approach?
Anything missing here?

### DIFF
--- a/test/TR_jx.sh
+++ b/test/TR_jx.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+. test_runner_common.sh
+
+prepare()
+{
+	$CCTOOLS_COMPILE jx_test.c -o jx_test -ldttools -lm
+}
+
+run()
+{
+	./jx_test < jx.input > jx.output
+	diff jx.output jx.expected
+	return $?
+}
+
+clean()
+{
+	rm -f jx.output jx_test
+	return 0
+}
+
+dispatch "$@"
+
+# vim: set noexpandtab tabstop=4:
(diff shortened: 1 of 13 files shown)


btovar:
I seems sensible to me. Perhaps my only issue is that all tests are in a flat hierarchy "test/". We may want to separate them in a directory per tool so that they are easier to manage.

dthain:
#2035

PR #2122, opened 2019-08-15
Fix LOCAL in Parrot mountlist

### LINKED ISSUE #2121: LOCAL access control in parrot's mountlists has no effect
We can either remove it in favor of --disable-filesystem, or make it work exactly as --disable-filesystem. 

I like the latter, as it shields a mountlist against services included when compiling parrot.

trshaffer:
The current path resolution code in Parrot seems to more or less ignore the `LOCAL` access control. This PR adds a new resolve outcome, `PFS_RESOLVE_LOCAL`, that causes subsequent opens to skip service lookup, forcing use of the local filesystem. Aside from that, this return code behaves almost the same as `PFS_RESOLVE_UNCHANGED`.

Fixes #2121

### DIFF
--- a/parrot/src/pfs_table.cc
+++ b/parrot/src/pfs_table.cc
@@ -516,7 +516,9 @@ int pfs_table::resolve_name(int is_special_syscall, const char *cname, struct pf
 		char tmp[PFS_PATH_MAX];
 		path_split(pname->path,pname->service_name,tmp);
 		pname->service = pfs_service_lookup(pname->service_name);
-		if (result == PFS_RESOLVE_LOCAL) pname->service = NULL;
+		if (result == PFS_RESOLVE_LOCAL) {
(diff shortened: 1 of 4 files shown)


parrot/src/pfs_table.cc
```diff
@@ -516,6 +516,7 @@ int pfs_table::resolve_name(int is_special_syscall, const char *cname, struct pf
 		char tmp[PFS_PATH_MAX];
 		path_split(pname->path,pname->service_name,tmp);
 		pname->service = pfs_service_lookup(pname->service_name);
+		if (result == PFS_RESOLVE_LOCAL) pname->service = NULL;
```
btovar: Tim, please add { } to this 'if'.

btovar:
It looks good to me.

PR #2135, opened 2019-08-29
standard configure options

btovar:
- Adds   `packaging/scripts/configure-from-image`   which standardizes  how cctools is built (i.e., how ./configure is called)  in travis and autobuild.
- Adds `packaging/scripts/generate-images` which is the script used to generate the docker and singularity images used in travis and autobuild.
- Moves all travis support files to `packaging/travis` (except for .travis.yml)
- Removes `configure.afs`.
- Updates debian to 9.9.
- Removes opensuse and ubuntu from travis, as maintaining them was more trouble than its worth (generating the images for them was painfully slow).
- WPI: add the script used to deploy to the Notre Dame campus.

[Read here](https://ccl-tech-notes.readthedocs.io/en/latest/software_eng/git-to-release) for further comments on configure-from-image.


btovar:
Closing as travis is building the wrong commit...

PR #2143, opened 2019-09-10
Fix compilation for Chirp SWIG bindings for Python 3.7

trshaffer:
From a [thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/cctools-nd/lwt9A_EZJK4) on the mailing list, there are compilation errors on Python 3.7 using conda. I grabbed the fix in 3fd3826554c0c4dbc3a1a2425175c541ff07baab and put it on the Chirp bindings, and things seem to work.

### DIFF
--- a/chirp/src/bindings/python3/Makefile
+++ b/chirp/src/bindings/python3/Makefile
@@ -4,7 +4,7 @@ include $(CCTOOLS_HOME)/rules.mk
 # Python always uses 'so' for its modules (even on Darwin)
 CCTOOLS_DYNAMIC_SUFFIX = so
 # SWIG produces code that causes a lot of warnings, so use -w to turn those off.
-LOCAL_CCFLAGS = -fPIC -w $(CCTOOLS_PYTHON3_CCFLAGS)
+LOCAL_CCFLAGS = -fPIC -w -DNDEBUG $(CCTOOLS_PYTHON3_CCFLAGS)
 LOCAL_LINKAGE = $(CCTOOLS_PYTHON3_LDFLAGS) $(CCTOOLS_GLOBUS_LDFLAGS) $(CCTOOLS_EXTERNAL_LINKAGE)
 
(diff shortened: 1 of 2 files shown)


dthain:
This makes sense to me, please keep the conversation going on the mailling list.
FYI @btovar

dthain:
@trshaffer can you summarize how rpath works on OSX, for our benefit?

trshaffer:
The issue here is more to do with the general linking conventions on OSX. Their version of RPATH works pretty much like on Linux (with slightly different syntax/rules). Conda just happens to use RPATH, which triggers this weird interaction.

On Linux, `ld` searches for libraries following `-L` specifications and adds the library names it finds to the resulting executable. Normally `DT_SONAME` for a library will be a logical name like `libc.so.6`. If you have strange library search paths, you'll need to use `$LD_LIBRARY_PATH`, `DT_RUNPATH`, etc. to tell the dynamic linker how to resolve the logical names.

On OSX, libraries have an install name, akin to `DT_SONAME`. This is normally set to the *absolute path* to the library. Thus a compiled executable ends up referring to absolute paths rather than short logical names. When using Conda, install names are set like `@rpath/libz.1.dylib`, and the library of course has its RPATH set. When we try to link against such a Conda-provided library, the linker copies over the install name as is, but doesn't do anything with the RPATH. Thus our compiled programs don't know the Conda prefix and fail to link.

As a workaround, I added a linker flag on OSX to set both the library search path and the RPATH. This works as long as install names are of the form `@rpath/whatever.dylib`, but any other use of RPATH would still be broken. Aside from installing CCTools via Conda and letting them patch up the headers, I'm not sure there's a general and correct way to handle this...

dthain:
@btovar

btovar:
I think this is ok for now.  With OSX, we only expect work_queue_worker to execute locally or inside a conda environment.

Issue #2142, opened 2019-09-10
Typo in makeflow manual


rykermcintyre:
In /cctools/doc/makeflow.html on line 282, "produceds" should be "produces"

PR #2144, opened 2019-09-10
Fix typo in makeflow manual on line 282

fschicke:
Corrected Typo in makeflow manual #2142: In /cctools/doc/makeflow.html on line 282, "produceds" should be "produces"


btovar:
Thanks!

PR #2152, opened 2019-09-18
expand newlines, etc. in batch-options for condor/wq-factory

btovar:
This allows to easily modify the condor submit file the factory uses.
E.g., as needed by mfrayer at Wisconsin:

work_queue_factory -T condor --batch-options="universe=docker\ndocker_image=debian" ...etc...


btovar:
Currently, I can achieve the same with something like:

```sh
work_queue_factory --batch-options="$(/bin/echo -e universe=docker\\ndocker_image=debian)"  ...
```

which is a little clunky.

PR #2156, opened 2019-10-02
Compile work_queue_worker statically with travis

btovar:
This pr modifies `packaging/scripts/configure-from-image` such that the `work_queue_worker` binary is compiled statically. This eases the deployment of workers across machines with the same architecture.

If this works as intended, we can follow a similar approach for compiling cctools with conda, such that workers do not depend on the conda environment's shared libraries.


dthain:
Is musl the critical piece needed to make this static?

btovar:
Yes. Otherwise the executables depend on glibc.

PR #2158, opened 2019-10-04
batch_job_amazon reuse machine images

epyifany:
Hi Prof. Thain,
I have created a list of occupied/idle running instances.
No more vms shutting down between makeflow jobs.
30s idle period to pick up new tasks before shutting down.
It handles shut down signal and terminates running vms.

Still have to debug some situations but I will keep working on it this weekend.

### DIFF
--- a/batch_job_amazon.c
+++ b/batch_job_amazon.c
@@ -1,0 +1,834 @@
+#include "batch_job_internal.h"
+#include "process.h"
+#include "batch_job.h"
+#include "stringtools.h"
+#include "debug.h"
+#include "macros.h"
+#include "nvpair_jx.h"
+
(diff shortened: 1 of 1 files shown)


dthain:
Something is strange about this pull request: it shows that every single line in the file is new.

dthain:
Oh, I see the problem.
It looks like you added a new file called batch_job_amazon.c at the root of the repository.

Don't do that: change the file in it's proper location (cctools/batch_job/src/batch_job_amazon.c), commit that, and make a pull request.  Then I can see what's actually changed and give you some feedback.

dthain:
Please close this pull request and make a new one..

PR #2162, opened 2019-10-09
Reuse instances: trouble shutting down

epyifany:
Hi Prof. Thain,

I have so far incorrectly used sleep(5) to wait to check if instances have been reused to shut down instances but it just makes the entire CPU sleep. 

Would it be a good idea to fork and wait? 
Do you mind taking a look at my code? (aws_terminate_idle_instance, line 256, sleep(5), line 266)

Thanks,
Yifan


dthain:
Ok, so you don't want to be calling sleep() in order to wait for things.

Keep in mind the overall structure here:
Makeflow is in charge of the main loop.

- When it wants to submit new jobs, it will call batch_job_amazon_submit.
- When it has nothing else to do and wants to wait for jobs to complete, it will call batch_job_amazon_wait.


Neither of those methods should be sleeping.  Instead, whenever a job completes, put its virtual machine on an "available" list.  And, whenever a new job is submitted, look for a compatible VM on the available list.  If there is one, use it.  If not, then create a new one.  Either way, dispatch the job to that VM.

Does that clarify?

batch_job/src/batch_job_amazon.c
```diff
@@ -1,9 +1,3 @@
- /*
```
dthain: Don't remove the copyright, please.

batch_job/src/batch_job_amazon.c
```diff
@@ -176,6 +171,68 @@ static struct jx * aws_describe_instance( struct aws_config *c, const char *inst
 	return j;
 }
 
+/*
```
dthain: If I am reading this correctly, what you are doing is asking AWS to put labels on each of the machines to describe their state.  I don't think this is the right way to go about it, since those queries and modifications can be quite expensive.

I think a much easier way to go about it would be to simply keep a local list of "available" VMs.  Whenever a job completes, put the VM on the list.  Whenever a new job is submitted, look for an available VM on the list.  If one is available/compatible, then use it.  Otherwise, create a new one.

PR #2172, opened 2019-10-17
Install instructions

btovar:
Updates install instructions.


dthain:
So, judging from these install instructions, the draft Spack module does not yet set PATH, PYTHONPATH, and LD_LIBRARY_PATH.  But it could with some more work.  Did I get that correct?

btovar:
Correct.

btovar:
> So, judging from these install instructions, the draft Spack module does not yet set PATH, PYTHONPATH, and LD_LIBRARY_PATH. But it could with some more work. Did I get that correct?

Further, what we need is  `spack load cctools` to do that for us. `spack load` uses a modulefile which we still need to write.  I update the instructions with `spack load` accordingly.

dthain:
Some suggestions for streamlining the instructions:

1 - This command seems scary to the un-initiated:
```
tar -C ~/cctools --strip-components=1 -xf cctools-*-source.tar.gz
```

While I get that you are trying to strip off the versioned directory string, I think that's likely to be confusing to the novice.  How about this more explicit approach?

```
tar xzf cctools-*-source.tar.gz
cd cctools-*-source
./configure
make
make install
```

2 - For the git install example, make it complete by adding the few lines of ./configure; make; make install

3 - Maybe drop the binary tarball section completely?

btovar:
> Some suggestions for streamlining the instructions:

Done.

btovar:
I think this is ready to be merged.
The only thing not working is spack, but it is marked appropriately.

dthain:
Ben, are we ready to update the warnings on Spack, and move ahead?

Issue #2179, opened 2019-10-25
Memory Leak - work_queue.c rmsummary


97xuzy:
I have discover 3 possible memory leak with `struct rmsummary` when testing my work_queue master with valgrind.
I have compiled with both v7.0.19 library and the library from compiling master branch, the problem exist in both.

In `work_queue.c`, function `add_task_report()`, line ~3266,
`tr->resources = rmsummary_copy(t->resources_allocated);`
a `struct rmsummary` is allocated via `rmsummary_copy()`, but it is not freed, it should be freed along with `free(tr)` on line ~3275, and after tr is freed, tr->resource will be lost.

In `work_queue.c`, function `compute_capacity()`, line ~3291
`capacity.resources = rmsummary_create(0);`
a `struct rmsummary` is allocated via `rmsummary_create()`, but it does not seem to be freed, and it does not seem to be passed to some other owner.

In `work_queue.c`, function `write_transaction_worker_resources()`, line ~6708
`struct rmsummary *s = rmsummary_create(-1);`
a `struct rmsummary` is allocated via `rmsummary_create()`, but it does not seem to be freed.

My wq master is basically just the barebone C example with some modification to run blast command.

Here is the valgrind heap summary
```
==27278== 
==27278== HEAP SUMMARY:
==27278==     in use at exit: 8,563 bytes in 169 blocks
==27278==   total heap usage: 543 allocs, 374 frees, 233,609 bytes allocated
==27278== 
==27278== 272 bytes in 1 blocks are definitely lost in loss record 157 of 169
==27278==    at 0x483780B: malloc (vg_replace_malloc.c:309)
==27278==    by 0x43CB01: rmsummary_create (rmsummary.c:910)
==27278==    by 0x40DAB2: compute_capacity (work_queue.c:3291)
==27278==    by 0x415C37: work_queue_get_stats (work_queue.c:6355)
==27278==    by 0x404342: log_queue_stats (work_queue.c:406)
==27278==    by 0x41217D: work_queue_create (work_queue.c:4948)
==27278==    by 0x403CB3: main (blast_wq_simple.c:88)
==27278== 
==27278== 272 bytes in 1 blocks are definitely lost in loss record 158 of 169
==27278==    at 0x483780B: malloc (vg_replace_malloc.c:309)
==27278==    by 0x43CB01: rmsummary_create (rmsummary.c:910)
==27278==    by 0x40DAB2: compute_capacity (work_queue.c:3291)
[...]

PR #2180, opened 2019-10-25
Correctly free the struct used for capacity

btovar:
Fixes for #2179 
When freeing the struct used for capacity, we we were not freeing the pointer to its resources.


btovar:
With this pr I get:
```text
==4065==    definitely lost: 0 bytes in 0 blocks
...
==4065== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

PR #2181, opened 2019-10-25
Add valgrind tests for  wq master, wq worker, and resource_monitor

btovar:


btovar:
Removed makeflow test for now.

PR #2186, opened 2019-10-28
batch_job_amaz uses list.h for idle instances with timestamps

epyifany:
Hi Prof. Thain and Ben,

I have used the existing linked list in dttools/src/list.h for consistency. I have also run and tested the list. 

I also included timestamps to shutdown idle instances that lingers around for more than "30 seconds". Should I work to make this a command-line option for Makeflow -T amazon? 

Thank you so much for going through this. 

Yifan


batch_job/src/batch_job_amazon.c
```diff
@@ -92,13 +108,113 @@ static const char * aws_instance_select( int cores, int memory, int disk )
...
 		}
 	}
 	return 0;
 }
 
+/*
+Lookup aws_instance_type given instance type name
+*/
+
+struct aws_instance_type * aws_instance_lookup(const char *instance_type)
```
dthain: Small thing: mark private functions as "static" throughout.

batch_job/src/batch_job_amazon.c
```diff
@@ -42,6 +45,19 @@ struct aws_config {
 	const char *keypair_name;
 };
 
+
```
dthain: Call this structure aws_instance throughout.  (You will see why in a bit.)

batch_job/src/batch_job_amazon.c
```diff
@@ -193,6 +309,44 @@ static int aws_terminate_instance( struct aws_config *c, const char *instance_id
...
+		debug(D_BATCH,"idle instance list empty\n");
+		return 0;
+	}
+	struct idle_instance* i = list_find(instance_list, (list_op_t) idle_instance_expire, timediff_p);
+	while (i != NULL){
+		aws_terminate_instance(c, i->instance_id);
+		list_remove(instance_list, i);
+		debug(D_BATCH,"erasing idle instance %s from list \n", i->instance_id);
+		debug(D_BATCH,"exiting terminate_idle_instance, current list count is %d\n", list_size(instance_list));
+		free(i);
```
dthain: For consistency, create a function called aws_instance_delete() and use it to dispose of an instance, instead of directly using free.  (Also, I think you may need to free members of the object.)

PR #2187, opened 2019-10-28
Do not link against libperl

btovar:
It is not needed for swig, and causes issues if libperl is not compiled
with '-fPIC', as with conda-forge.


dthain:
Aha, that's a much happier solution!

btovar:
Yes! libperl is only needed when compiling the module into a new perl executable.

PR #2191, opened 2019-10-29
Recycle Idle Instances in batch_job_amazon

epyifany:
Hi Prof. Thain

I have fixed several parts of my code based on the issues you have mentioned
1. Used different names for structures, 
2. added several missing static labels to private functions,
3. created a new function called aws_instance_delete() and free() malloc-ed recorded instances

To answer your question:
Does this require any change to the documentation?

I believe it does change documentation
Right now and I would still have to shut down the last several instances created at the end of the module manually. I will create a log file and write a clean-script. I think this should be the next feature to implement. And changes to documentation should follow as well. 

Thanks,
Yifan


batch_job/src/batch_job_amazon.c
```diff
@@ -92,13 +108,115 @@ static const char * aws_instance_select( int cores, int memory, int disk )
...
+*/
+
+static int compare_instance_type(struct aws_instance* instance, const char* type_2){
+	const char *type_1 = strdup(instance->instance_type);
+	if(strcmp(type_1,type_2) == 0){
+		return 1;
+	} else {
+		struct aws_instance_type *instance_1 = aws_instance_lookup(type_1);
+		struct aws_instance_type *instance_2 = aws_instance_lookup(type_2);
+		if (instance_1->cores<=instance_2->cores && instance_1->memory<=instance_2->memory){
```
btovar: This will segfault if aws_instance_lookup fails to find the type.

batch_job/src/batch_job_amazon.c
```diff
@@ -92,13 +108,115 @@ static const char * aws_instance_select( int cores, int memory, int disk )
...
+	return 0;
+}
+
+/*
+Compare if instance of type 1 does fit in instance of type 2.
+On fitting (cores and memory of type 1 instance <= type 2)
+return 1, otherwise return null
+*/
+
+static int compare_instance_type(struct aws_instance* instance, const char* type_2){
```
btovar: I suggest changing this to:
```C
static int instance_type_less_or_equal(struct aws_instance* instance, const char *type2) {
```

batch_job/src/batch_job_amazon.c
```diff
@@ -92,13 +108,115 @@ static const char * aws_instance_select( int cores, int memory, int disk )
...
+	const char *type_1 = strdup(instance->instance_type);
```
btovar: You don't need the strdup here. In fact it causes a memory leak as it is never freed.

batch_job/src/batch_job_amazon.c
```diff
@@ -92,13 +108,115 @@ static const char * aws_instance_select( int cores, int memory, int disk )
...
+Push an idle instance to the end of the list
+return 0 on error and 1 on success
+*/
+
+static int push_back_aws_instance(struct aws_instance* instance){
+
+	if (!instance_list) {
+		instance_list = list_create();
+	}
+	if (list_push_tail(instance_list, instance)){
```
btovar: No need to  check if  list_push_tail.  If it fails, we have worse problems and can't recover anyway.

batch_job/src/batch_job_amazon.c
epyifany: This is actually very true since I assumed the instance_types are already based on another table lookup. 

To prevent potential segfault, 
If say 

> if(!instance_1 || !instance_2) {return 0;} 

would this be too hard to follow?

Thanks Ben!

btovar: Yes, that works nicely.

dthain:
@epyifany please wrap this up and let us know when it's ready to merge.

epyifany:
Hi Prof. Thain,
 Sorry for the late comment and commit. I have made edits to Ben's comments. I test and committed the log functions to this branch. This branch should be ready to merge. 

Yifan

batch_job/src/batch_job_amazon.c
```diff
@@ -180,19 +296,59 @@ static struct jx * aws_describe_instance( struct aws_config *c, const char *inst
...
 
-static int aws_terminate_instance( struct aws_config *c, const char *instance_id )
+static int aws_terminate_instance( struct batch_queue * q, struct aws_config *c, const char *instance_id )
 {
 	char *str = string_format("aws ec2 terminate-instances --instance-ids %s --output json",instance_id);
 	struct jx *jresult = json_command(str);
 	if(jresult) {
 		jx_delete(jresult);
 		printf("deleted virtual machine instance %s\n",instance_id);
+		FILE *fp;
```
dthain: This is a bad pattern to start: Write a separate function called aws_log_message() or something that logs the desired message, so you don't keep repeating the get_option/fopen/fclose pattern

batch_job/src/batch_job_amazon.c
```diff
@@ -509,22 +665,37 @@ static batch_job_id_t batch_job_amazon_submit(struct batch_queue *q, const char
...
-	char *instance_id = aws_create_instance(aws_config,instance_type,ami);
+	/* Before create a new instance, fetch to see if an idle instance exists */
+	const char *instance_id = fetch_aws_instance(instance_type);
 	if(!instance_id) {
-		debug(D_BATCH,"aws_create_instance failed");
-		sleep(1);
-		return -1;
+  	/* Create a new instance and return its unique ID. */
+		printf("creating instance\n");
+  	instance_id = aws_create_instance(aws_config,instance_type,ami);
```
dthain: Indent

batch_job/src/batch_job_amazon.c
```diff
@@ -509,22 +665,37 @@ static batch_job_id_t batch_job_amazon_submit(struct batch_queue *q, const char
...
+    if(!instance_id) {
+      debug(D_BATCH,"aws_create_instance failed");
+      sleep(1);
+      return -1;
+    }
+		FILE *fp;
```
dthain: Repeated code here, pull it out into a function.

dthain:
Some little style things, please fix these up this week so we can merge it...

dthain:
This looks good to me. @btovar any remaining concerns?

btovar:
Yes. I marked a memory leak in the comments, and a couple of other comments that are unresolved.

dthain:
@epyifany please double check @btovar comments

epyifany:
Hi Prof. Thain and Ben,

I have pushed my updates resolving some of Ben's previous comments in addition to edits to the cleanup file. Despite several remaining bugs to be fixed, this module overall achieves the goal of recycling idle instances. 

I have run-tested this module update, using a workflow with the following command:

`$ makeflow -T amazon -J 20 --amazon-config <config-file> <workflow.mf>`

I have run tested a small BWA makeflow-example with 1000 sequences:
https://github.com/cooperative-computing-lab/makeflow-examples/tree/master/bwa

As we intend to test more than instance limit to test recycle, I recommend generating 40 sequences per split to generate a small workflow of 25 jobs(5 being recycled if the instances limit is 20). These 40-sequences-bwa jobs run smoothly on the default amazon t2.micro free tier instance. 
`$ ./make_bwa_workflow --ref ref.fastq --query query.fastq --num_seq 40 > bwa.mf`

I have also run-tested another self-constructed workflow of 100 Monte Carlo simulation jobs and the workflow work as intented.  

Please let me know what additional changes I should make and my work is ready to merge.

Issue #2195, opened 2019-10-30
work_queue_factory -T local -w not working as intented


zhxu73:
-w option seems to always spawn 1 less worker (-w 1-> 0 spawned, -w 2 -> 1 spawned), and I have double checked I have no other work_queue_worker running at the same time.

I have tested both on my laptop and on a cloud vm, with both master branch and 7.0.19, the problem persist

When using with option -w 1, meaning there need to be at least 1, however none is spawned
```
$ work_queue_factory localhost 9123 -T local -w 1
2019/10/30 08:17:28: |submitted: 0 |needed: 1 |waiting connection: 0 |requested: 0 
PROJECT            HOST                   PORT WAITING RUNNING COMPLETE WORKERS 
localhost          localhost              9123       0       0        0       1 

```

With -w 2, only 1 worker is spawned, the same with -w 3, -w 4, and so on, always 1 less worker.
```
$ work_queue_factory localhost 9123 -T local -w 2
2019/10/30 08:17:59: |submitted: 0 |needed: 2 |waiting connection: 0 |requested: 1 
PROJECT            HOST                   PORT WAITING RUNNING COMPLETE WORKERS 
localhost          localhost              9123       0       0        0       1 

work_queue_worker: creating workspace /tmp/worker-1000-9730
work_queue_worker: using 8 cores, 15921 MB memory, 7960 MB disk, 0 gpus
connected to master localhost:9123 via local address 127.0.0.1:39274

```

```
$ work_queue_factory localhost 9123 -T local -w 1 -dall
2019/10/30 08:28:09.25 work_queue_factory[10587] debug: work_queue_factory version 7.0.19 FINAL (released 2019-10-22 20:02:26 -0700)
2019/10/30 08:28:09.25 work_queue_factory[10587] debug: Built by johnxu@localhost.localdomain on 2019-10-22 20:02:26 -0700
2019/10/30 08:28:09.25 work_queue_factory[10587] debug: System: Linux localhost.localdomain 5.2.17-100.fc29.x86_64 #1 SMP Mon Sep 23 14:09:29 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
2019/10/30 08:28:09.25 work_queue_factory[10587] debug: Configuration:
2019/10/30 08:28:09.26 work_queue_factory[10587] batch: set feature `local_job_queue' to `yes'
2019/10/30 08:28:09.26 work_queue_factory[10587] batch: set feature `absolute_path' to `yes'
[...]

PR #2197, opened 2019-10-31
Don't count status connections as workers

btovar:
Fix for #2195

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -413,9 +413,9 @@ static void log_queue_stats(struct work_queue *q)
 
 	work_queue_get_stats(q, &s);
 
-	debug(D_WQ, "workers status -- total: %d, active: %d, available: %d.",
+	debug(D_WQ, "workers connections -- known: %d, connecting: %d, available: %d.",
 			s.workers_connected,
-			s.workers_connected - s.workers_init,
+			s.workers_init,
 			available_workers(q));
 
 	if(!q->logfile)
@@ -933,7 +933,7 @@ static void remove_worker(struct work_queue *q, struct work_queue_worker *w, wor
 	/* update the largest worker seen */
 	find_max_worker(q);
 
-	debug(D_WQ, "%d workers are connected in total now", hash_table_size(q->worker_table));
+	debug(D_WQ, "%d workers connected in total now", count_workers(q, WORKER_TYPE_WORKER | WORKER_TYPE_FOREMAN));
 }
 
 static int release_worker(struct work_queue *q, struct work_queue_worker *w)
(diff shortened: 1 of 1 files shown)


zhxu73:
I don't think it fixes the my problem, it still spawns 1 less worker.
The position I traced to is in `work_queue_factory.c`

work_queue_facotry.c, mainloop(), line ~832
a masters_list is created by do_direct_query()
```c
if(using_catalog) {
	masters_list = work_queue_catalog_query(catalog_host,-1,project_regex);
}
else {
	masters_list = do_direct_query(master_host,master_port);
}
```
work_queue_factory.c, do_direct_query(), line ~228
Creation process of the masters_list has something to do with the short "status" link that it connect to my WQ master.
```c
struct jx *jarray = jx_parse_link(l,stoptime);
```
line ~236
```c
struct jx *j = jarray->u.items->value;
j->type = JX_OBJECT;
	
struct list* master_list = list_create();
list_push_head(master_list, j);
```

work_queue_facotry.c, mainloop(), line ~856
The masters_list is then pass to count_workers_connected(), in which it returns 1 if my WQ master is running prior to the factory, and 0 otherwise (no other workers are running external to this).
So I persume that do_direct_query() definitely "record" the link to my WQ master, and count_workers_connected() mistakenly count that as a worker.
```c
int workers_connected = count_workers_connected(masters_list);
```

work_queue_facotry.c, mainloop() line ~892
since `workers_connected` is 1, thus cause workers_waiting_to_connect to be -1, which cause the factory to think there are some external worker.
```c
// if negative, this means workers external from this factory have
// connected.
int workers_waiting_to_connect = workers_submitted - workers_connected;

if(workers_waiting_to_connect < 0) {
	debug(D_WQ,"%d workers already connected from other sources", -workers_waiting_to_connect);
}
```

The -1 also reflect in the `-dall` logs as well
```bash
2019/10/31 14:50:19.55 work_queue_factory[9876] wq: applying minimum of 1 workers
2019/10/31 14:50:19.55 work_queue_factory[9876] wq: 1 workers already connected from other sources
```

zhxu73:
Ah, never mind, sorry, My recompiled WQ master isn't in the PATH, the old one is

zhxu73:
But I am still curious why is it a problem at `work_queue.c` rather than in `work_queue_factory.c`?
Is it because the status link actually exchange some info about both side, so that they what type of node the other one is?

btovar:
@zhxu73 When the factory decides how many workers to run, it takes into account the workers count at the master. The master was counting the connection of the factory itself as a worker connection.

This is why the problem was only observed when the factory directly asked the master, and not when the factory got its information from the catalog.

PR #2202, opened 2019-11-05
add documentation to resource_monitor.py

btovar:


dthain:
Looks great; be sure to add a short note to `resource_monitor/index.md` so it can be found from there, too...

PR #2211, opened 2019-11-25
Change pydoc search.

tkameyama:
pydoc is faiied when wrong PYTHONHOME environment variable is found.
```
$ PYTHONHOME=/tmp pydoc -w umbrella
ImportError: No module named site
```
This PR search pydoc in python install directory (expected to match PYTHONHOME ).

### DIFF
--- a/configure
+++ b/configure
@@ -1057,6 +1057,7 @@ then
 	fi
 fi
 
+PYDOC=""
 python=0    #to be set to the python path
 if [ $config_python_path != no ]
 then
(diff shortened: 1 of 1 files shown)


btovar:
Thanks for the pr!

However, if PYTHONHOME has an invalid value, none of the python scripts would work anyway. Since umbrella and its documentation are optional, I'm thinking of simply skipping them with a warning if the python environment is invalid.

tkameyama:
> 
> 
> Thanks for the pr!
> 
> However, if PYTHONHOME has an invalid value, none of the python scripts would work anyway. 

I am assuming the following situation:

- Multiple pythons (For example python2 and pyton3) are installed on build host.
- PYTHONHOME  environment variable is set for the python3.
- The user specify $PYTHONHOME/bin/python3 by configure option.
- configure detect pydoc for python2.

btovar:
@tkameyama, could you confirm whether pr #2217 does what you need?

We are in the process of changing how the configure script detect the python environment, to make sure, among other things, that the correct pydoc is found.

tkameyama:
> 
> 
> @tkameyama, could you confirm whether pr #2217 does what you need?

Yes.

> 
> We are in the process of changing how the configure script detect the python environment, to make sure, among other things, that the correct pydoc is found.

I hope it.

btovar:
This has been resolved by e6de50b9d8bff10a4e4159e743a4c9e72f5351fb and the latest ./configure.

Issue #2214, opened 2019-11-25
Add the len() function to JX


btovar:
Given a JSON array A, len(A) should return the number of elements in the array.  For any other type, len should signal an error.  

(This should be done in dttools/src/jx_function.{h,c}.)

dthain:
And just as important: update the documentation in jx-reference.md and jx-quick.md

camicarballo:
How should I go about testing this to make sure it does what I am expecting it to do?

btovar:
You can take a look at  dttools/test/TR_jx.sh.

Your input file would look something like:
```json
{
"count_1": len(["a", "b", "c"])
"count_2": len([])
}
count_1 == 3;
count_2 == 0;
```

with the expected file for comparison showing that both those expressions should be true.

camicarballo:
It looks like the function jx_array_length is incorrect. For an array of any length it consistently returns a length of 1.

btovar:
The iteration does there does seem suspect. Could you file an issue on it, and then change the iteration to use jx_iterate_array as shown in jx.h?   Thanks!

btovar:
Fixed by #2216

PR #2216, opened 2019-11-26
Add len() to JX

camicarballo:
Addresses #2214

### DIFF
--- a/doc/manuals/jx/jx.md
+++ b/doc/manuals/jx/jx.md
@@ -469,7 +469,7 @@ template("SM{PLATE}_{ID}.sam", {"PLATE": "10001", "ID": N/2 - 1})
 
 >       len([1,2,3]) -> Integer
 
-len() returns the length when passed in an array. This function is
+len() returns the length when passed in an array and errors when not an array. This function is
 based on Python's own len() function.
 
 ```python
--- a/doc/manuals/jx/jx.md
+++ b/doc/manuals/jx/jx.md
@@ -441,14 +441,15 @@ def format(spec, *args):
 
 #### template
 
->    template(A[,B]) -> String
->    where A = String and B = Object
+>       template(A[,B]) -> String
+>       where A = String and B = Object
 
 template() replaces format placeholders in the given string.
 This function is based on Python's sting formatting capabilities.
 Variable names enclosed in curly braces are looked up in the current context.
 Suppose that the ID variable is defined as 10.
-<pre><code>
+
+
 ```python
 template("file{ID}.txt")
 = "file10.txt"
@@ -463,6 +464,19 @@ Suppose that the variable N is defined as 48.
 template("SM{PLATE}_{ID}.sam", {"PLATE": "10001", "ID": N/2 - 1})
 = "SM10001_23.sam"
 ```
+
+#### len
+
+>       len([1,2,3]) -> Integer
+
+len() returns the length when passed in an array. This function is
+based on Python's own len() function.
+
+```python
+len([1,2,3])
+= 3
+```
+
 
 ### Comprehensions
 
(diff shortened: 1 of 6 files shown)


camicarballo:
@btovar Can you look over this?

btovar:
On it.

dttools/src/jx_function.c
```diff
@@ -591,3 +591,26 @@ struct jx *jx_function_template(struct jx *args, struct jx *ctx) {
...
 }
+
+struct jx *jx_function_len(struct jx *args){
+
+	assert(args);
+    assert(jx_istype(args, JX_ARRAY));
+	const char *funcname = "len";
+    void *i;
+    
+    if(args->type != JX_ARRAY){
```
btovar: This check is not needed, as you have the assert above.

dttools/src/jx_function.c
```diff
@@ -591,3 +591,26 @@ struct jx *jx_function_template(struct jx *args, struct jx *ctx) {
...
+        FAIL(funcname, args, "not an array");
+    }
+
+    struct jx* item = jx_iterate_array(args, &i);
```
btovar: jx_array_index(args, 0);  is clearer, and we can eliminate the i variable.

dttools/src/jx_function.c
```diff
@@ -591,3 +591,26 @@ struct jx *jx_function_template(struct jx *args, struct jx *ctx) {
...
+
+    if(args->type != JX_ARRAY){
```
btovar: I think you mean item instead of args here.

doc/manuals/jx/jx.md
```diff
@@ -464,6 +465,19 @@ template("SM{PLATE}_{ID}.sam", {"PLATE": "10001", "ID": N/2 - 1})
 = "SM10001_23.sam"
 ```
 
+#### len
+
+>       len([1,2,3]) -> Integer
+
+len() returns the length when passed in an array. This function is
```
btovar: Also specify that it is an error if the argument is not an array.

dttools/src/jx.h
```diff
@@ -121,6 +121,7 @@ typedef enum {
 	JX_BUILTIN_LISTDIR,
 	JX_BUILTIN_ESCAPE,
 	JX_BUILTIN_TEMPLATE,
+    JX_BUILTIN_LEN
```
btovar: spaces to tab

dttools/src/jx_eval.c
```diff
@@ -201,6 +201,7 @@ static struct jx *jx_eval_call(
 		case JX_BUILTIN_LISTDIR: return jx_function_listdir(args);
 		case JX_BUILTIN_ESCAPE: return jx_function_escape(args);
 		case JX_BUILTIN_TEMPLATE: return jx_function_template(args, ctx);
+        case JX_BUILTIN_LEN: return jx_function_len(args);
```
btovar: spaces to tabs

dttools/src/jx_function.c
```diff
@@ -591,3 +591,19 @@ struct jx *jx_function_template(struct jx *args, struct jx *ctx) {
 		FAIL(funcname, args, "at most two arguments are allowed");
 	}
 }
+
+struct jx *jx_function_len(struct jx *args){
+
+	assert(args);
+	assert(jx_istype(args, JX_ARRAY));
+
+	struct jx* item = jx_iterate_array(args, 0);
```
btovar: jx_array_index(args, 0)

dthain:
Note that the (much ignored) `cctools_indent` script in the root of the repository can be used to achieve the desired indenting effect...

btovar:
The problem with cctools_indent is that it may indent code not part of the pr. If cctools_indent is used, I would suggest adding the changes one by one with  git commit -p, rather than adding the whole file (such as with git add -u, git commit, or git commit -a).

camicarballo:
These three commits should fix everything you mentioned @btovar

Issue #2221, opened 2019-12-09
Inefficient parsing when using jx and --args


btovar:
The following takes less than a second to parse:

```json
# file: main_embedded.jx
{
    "define": {
        "DATA_FILE_LIST" : [ { "PATH" : "somepath." + n } for n in range(1500) ]
    },
    "rules": [
        {
            "command" : "echo hello >" + DATA["PATH"],
            "outputs" : [ DATA["PATH"] ]
        } for DATA in DATA_FILE_LIST
    ]
}
```

If instead we generate DATA_FILE_LIST to use in an args file as:

```json
#file: file_list_raw.jx
{
    "DATA_FILE_LIST": [ { "PATH" : "somepath." + n } for n in range(1500) ]
}

#file: main_external.jx
{
    "rules": [
        {
            "command" : "echo hello >" + DATA["PATH"],
            "outputs" : [ DATA["PATH"] ]
        } for DATA in DATA_FILE_LIST
    ]
}
```

$ jx2json file_list_raw.jx > file_list.jx
$ jx2json main_external --args file_list.jx

it takes about 40 s (which is much more than 1s when not using --args).

Any ideas @trshaffer, @dthain ?

dthain:
Nothing immediately jumps to mind, but I might guess that some internal `eval` is either too lazy or too eager, resulting in n^2 behavior for the double-for.

I would be inclined to jus run both examples through gprof to see where time is being spent.  A different number of calls to `jx_eval` would be a good hint.

btovar:
Many functions that evaluate jx do a jx_copy on the context.  The variables defined in --args end up as the context of the workflow.

For the examples above, using --args there are 197076363 calls to jx_copy. Without args, there are only 1613046. An increase of three orders of magnitude. 

Fixing this one will require some thought...

dthain:
Actually, I don't see that many instances of copying the context.

There is one in the top-level `jx_eval` that doesn't make sense to me.
And then there is one in `jx_eval_comprehension` that is more understandable.

It seems to me there is a substantial difference between `jx_eval(jx_copy( big list ))` and `jx_copy(jx_eval(list comprehension))`

Does that make sense?

btovar:
Yes, by commenting out the jx_copy/jx_delete in jx_eval parsing is done in 2.5s.

@trshaffer, do you see any problems with this?

btovar:
solved by #2224

PR #2224, opened 2019-12-10
do not copy context on eval

btovar:
Fix for #2221

### DIFF
--- a/dttools/src/jx.c
+++ b/dttools/src/jx.c
@@ -148,17 +148,6 @@ struct jx * jx_error( struct jx *err )
 	return j;
 }
 
-struct jx *jx_function(const char *name, jx_builtin_t op,
-	struct jx_item *params, struct jx *body) {
-	assert(name);
-	struct jx *j = jx_create(JX_FUNCTION);
-	j->u.func.name = strdup(name);
-	j->u.func.params = params;
-	j->u.func.body = body;
-	j->u.func.builtin = op;
-	return j;
-}
-
 struct jx * jx_arrayv( struct jx *value, ... )
 {
 	va_list args;
@@ -425,11 +414,6 @@ void jx_delete( struct jx *j )
 			jx_delete(j->u.oper.left);
 			jx_delete(j->u.oper.right);
 			break;
-		case JX_FUNCTION:
-			free(j->u.func.name);
-			jx_item_delete(j->u.func.params);
-			jx_delete(j->u.func.body);
-			break;
 		case JX_ERROR:
 			jx_delete(j->u.err);
 			break;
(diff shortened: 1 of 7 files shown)


trshaffer:
I think this might cause some issues with deltadb. The [calls to add builtins](https://github.com/cooperative-computing-lab/cctools/blob/master/dttools/src/jx_eval.c#L599) modify the context in place, which is fine for most of the uses of `jx_eval()` in the code. It looks like the added builtins would be observable during a checkpoint read in delatdb (and possibly when querying), which may cause other problems.

btovar:
What about a jx_eval_closed_context for deltadb that copies the context and then calls jx_eval?

trshaffer:
I would be more comfortable with a `jx_eval_inplace()` or something to that effect as a destructive version that's explicitly opt-in. For example, the command line tools like `jx2json` could switch to this version.

trshaffer:
Another issue with 2 eval flavors: we already have both `jx_eval()` and `jx_eval_with_defines()`, so it'd be awkward to do every permutation (e.g. `jx_eval_inplace_with_defines()`). It might be better to split these up, i.e. we have a function

    jx_ctx_from_defines(struct jx *j, struct jx *context)

that replaces `jx_eval_with_defines()` but returns the resulting context object rather than doing the eval itself. Then the caller can choose which of the two eval flavors to use.

I can push commits for that to this PR if you like.

btovar:
Tim, that sounds good! Thanks.

dthain:
@trshaffer I'm a little uncertain about how functions are currently handled.

Would there be a serious downside to treating them more generically, by parsing SYMBOL ( args ) as a generic function call, and then just performing a simple string comparison at call time?

trshaffer:
Not too serious. I was avoiding that as it's another weird case in `jx_eval_operator()` ([which is already sort of hairy](https://github.com/cooperative-computing-lab/cctools/blob/master/dttools/src/jx_eval.c#L338)).

But after looking at making changes in this PR, I'm thinking that lookup at call time might be the way to go.

dthain:
That makes sense to me, and it also has the effect that
one can parse and print arbitrary function calls without
having a hard-coded list of acceptable names at parse time.

dthain:
Thorough testing, as always...

trshaffer:
I ended up doing a fair amount of internal changes, to remove the `JX_FUNCTION` type (see 7c3b9ca77bd7801816be9e2fc5c08791bf7cc57b for details). Since we've been adding statically defined functions as needed, there's not a whole lot of reason to keep internal support for full lambda functions. With this simplification/cleanup, we can avoid the expensive copy in #2221.

trshaffer:
I was a bit concerned as there don't seem to be automated tests for DeltaDB stuff, but I don't see any issues from reading the code

dthain:
@btovar what do you think?

btovar:
Looks good to me.

Issue #2222, opened 2019-12-09
Fix executable path mismatch in work_queue_factory with Torque


annawoodard:
I'm trying to run the WQ [example](https://cctools.readthedocs.io/en/latest/work_queue/examples/work_queue_example.py) on a Torque system at University of Chicago. I've installed cctools following the conda install instructions. Submitting workers via 
```
torque_submit_workers cri16in001 8888 1
``` 

works as expected. Starting a factory via:

```
(wq) [cri16in001] ~/wq-hello >work_queue_factory -T torque cri16in001 8888
```
results in exit 127s for all submitted workers:
```
[cri16in001] ~/wq-hello >cat torque.status.19162063
start 1575923228
stop 127 1575923228
```

By default, the wrapper script is written inside of a `wq-pool-*` directory in the PWD (where I started the factory):
```
[cri16in001] ~/wq-hello >cat wq-pool-788688670/torque.wrapper
#!/bin/sh
#$ -S /bin/sh
[ -n "${PBS_JOBID}" ] && JOB_ID=`echo ${PBS_JOBID} | cut -d . -f 1`
cd /home/t.cri.awoodard/wq-hello
logfile=torque.status.${JOB_ID}
starttime=`date +%s`
cat > $logfile <<EOF
start $starttime
EOF

eval "$BATCH_JOB_COMMAND"

status=$?
stoptime=`date +%s`
cat >> $logfile <<EOF
stop $status $stoptime
EOF
```

Where `BATCH_JOB_COMMAND` is:
```
./work_queue_worker cri16in001 8888 -t 300 -C 'catalog.cse.nd.edu,backup-catalog.cse.nd.edu' -d all -o worker.log
```
The `work_queue_worker` executable also gets copied to that directory:
```
[cri16in001] ~/wq-hello >ls wq-pool-788688670/
torque.wrapper  work_queue_worker
```

As you can see, the wrapper script changes directory to the `/home/t.cri.awoodard/wq-hello`, but then tries to evaluate `./work_queue_worker`, which is in `/home/t.cri.awoodard/wq-hello/wq-pool-788688670`. Copying the `work_queue_worker` executable into `/home/t.cri.awoodard/wq-hello` fixes the problem. I think we want to modify  `BATCH_JOB_COMAND` to point to the `wq-pool-*` directory for the executable by default.

btovar:
[...]

PR #2225, opened 2019-12-10
use getcwd instead of PWD in cluster module

btovar:
Fix for #2222


btovar:
Thanks for letting us know!

Issue #2185, opened 2019-10-28
makeflow's DAG_FILE_TYPE_TEMP does not delete temp files


btovar:
Test TR_makeflow_040_sub_workflow_jx.sh is not deleting its arguments file `makeflow.jx.args.XXXXX` created [here](https://github.com/cooperative-computing-lab/cctools/blob/29c37b84c3b34a471907cfc079f97678914077e1/makeflow/src/makeflow.c#L216).

That file is registered with `makeflow_hook_add_input_file` as a temp file, but that does not seem to have any effect.

btovar:
One problem is that the args.XXXXX file is created while parsing, but the cleanup code is only called on node completion. For example, if makeflow is called with the -c option, the file is created but never cleaned as no node ever runs.

dthain:
Odd b/c the TEMP flag is added here:
https://github.com/cooperative-computing-lab/cctools/blob/master/makeflow/src/makeflow.c#L306

And then TEMP files are deleted on job completion here:
https://github.com/cooperative-computing-lab/cctools/blob/master/makeflow/src/makeflow.c#L909

btovar:
I think that's the problem. The cleanup is done only on job completion, but with -c or -k no job is ever completed.

That is why I added the manual fix to delete args.XXXXX files, which are created during parsing. A better solution would be to delete all temps regardless of job completion, but I did not want to do that just now because you were studying how slow is to stat files in the disk when ignoring the log.

dthain:
Ok, I see what you mean now -- the underlying architectural problem is that
clean on a sub-makeflow doesn't really "submit" and "complete" a job.  I
see a couple of ways forward on this:
1 - Make the clean operation really use the job execution interface in the
normal way.
2 - Duplicate some functionality from job execution into the sub-makeflow
clean operation.
3 - Pass arguments to sub-makeflows in a way that doesn't result in the
generation of extra files.  (but maybe there will still be

btovar:
Fixed in 38dd0a2cb7c727f6cf2d24b6d6b714a335a6f86a

PR #2226, opened 2019-12-10
Always remove temporary jx args files

btovar:
Fix for #2185. Ensure that the files created while parsing are always deleted.


dthain:
LGTM

dthain:
Ready to merge?

Issue #2161, opened 2019-10-09
Makeflow Gets Stuck if Submit Always Fails


epyifany:
Without specifying a remove job limit of 20(-J 20), makeflow -T amazon will be stuck in a loop of requesting for instances and amazon responding InstanceLimitExceeded.

This is a specific issue on amazon as other sources hold on to the jobs submitted.

It would be helpful to adjust part of makeflow_run in makeflow.c to wait after continuously failing to submit. 

YY

dthain:
More generally, the problem here is that Makeflow expects any failure in batch_job_submit to be a temporary one: if submit fails, then makeflow just keeps retrying.

This assumption is not correct in the case of Amazon EC2, when the user has run out of available instances: batch_job_submit() will just fail over and over again.

What needs to happen is that, in the case of submit failure, the makeflow main loop should just continue so that it eventually calls batch_job_wait(), allows a job to exit, and then batch_job_submit() will be able to succeed.

dthain:
Fixed.

PR #2234, opened 2019-12-19
wait for makeflow jobs even when remote submission fails.

btovar:
The use case is amazon batch jobs where we could not create new
submissions because some hard limit.

Solves #2161

### DIFF
--- a/makeflow/src/makeflow.c
+++ b/makeflow/src/makeflow.c
@@ -99,10 +99,17 @@ an example.
 /*
 Flags to control the basic behavior of the Makeflow main loop. 
 */
+enum job_submit_status {
+	JOB_SUBMISSION_HOOK_FAILURE = -1,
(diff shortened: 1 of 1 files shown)


btovar:
Got it.

Should we still try to submit local jobs? If makeflow_dispatch_ready_jobs breaks, then they won't run even if they could.  With the current pr they would get run after all the timeouts, which is also not desirable.

dthain:
Ah, good point!
Any suggestions on how to handle this cleanly?

btovar:
@dthain: cc8e7480496d63ab02701186542b522e4e23b3c4 makes it so that local jobs are still submitted even if the batch system times out.

epyifany:
@btovar I have just ran tested a makeflow example and the timeouts have helped. Jobs are successfully submitted. Thanks!

PR #2237, opened 2019-12-20
Warn when worker workspace on noexec filesystem

btovar:
Related to issue #2229.

I opted for simply warn rather than terminating the worker as command absolute paths would work when workspace is on a noexec filesystem.

@annawoodard, thanks again for debugging this issue for us!

### DIFF
--- a/dttools/src/host_disk_info.c
+++ b/dttools/src/host_disk_info.c
@@ -16,18 +16,6 @@ See the file COPYING for details.
 #include <sys/param.h>
 #include <sys/mount.h>
 
-#ifdef HAS_SYS_STATFS_H
-#include <sys/statfs.h>
-#endif
-
-#ifdef HAS_SYS_STATVFS_H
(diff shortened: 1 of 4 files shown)


dthain:
I'm thinking this should be a fatal error, rather than a warning.
If the user starts a bunch of workers via script or factory, will they even see the error?

(I think it's safe to drop the SUNOS ifdefs at this point as well!)

btovar:
We may not want a fatal error as the executable may not be under work queue control. It may break currently working applications where the executable comes from the singularity image, or even comes from a shared filesystem.  Also, killing the worker right away will just hit the batch system with resubmissions if using the factory.

Perhaps a better solution is to not kill the worker, but print a warning at the master if the task ran for less than 1/10 of a second and exited with error code 126.

In fact, we may want to print such warnings at the master if tasks run for a very short time with non-zero error codes.

PR #2240, opened 2019-12-21
WQ Master and Python API using JSON API for workqueue

camicarballo:


dthain:
This is on the right track!  I like how the end-user python client program looks a lot like a plain native Work Queue application, modulo the JSON aspects.  A few things small and large indicated throughout.

(Also, please add a brief description of what you are doing in the PR top-level description.)

work_queue/src/client.py
```diff
@@ -0,0 +1,35 @@
+import sys
```
dthain: Please add the standard CCTools copyright boilerplate (now 2020!) to the top of each file.

work_queue/src/client.py
```diff
@@ -0,0 +1,35 @@
...
+
+def main():
+
+    q = WorkQueueServer()
+
+    #connect to server
+    q.connect('127.0.0.1', 2345)
+  
+    #read in tasks and submit requests
+    f = open("tasks.txt", "r")
```
dthain: This requires that the task JSON objects be split line-by-line, which will eventually result in some trouble.
Could you make them as native data structures in this source file instead?

work_queue/src/json_server.py
```diff
@@ -0,0 +1,86 @@
...
+            "jsonrpc" : "2.0",
+            "method" : "submit",
+            "id" : self.id,
+            "params" : task
+        }
+
+        request = json.dumps(request)
+
+        request += "\n"
+        self.socket.send(request)
```
dthain: There is a lurking problem here: the TCP connection is a continuous stream, so there is no guarantee that when the sender writes N bytes, the receiver will read back exactly N bytes.  It probably works for small test cases on a single machine, but sooner or later, you will send() 100 bytes, and the recv() will only give back the first 50, and then chaos results.

What you need to do here is construct a small "envelope" that describes the length of the message, and then sends the contents of the message.  Then, when receiving, read back the fixed integer, and then read until you get exactly that number of bytes.  Then you know that you have a full message.

work_queue/src/json_server.py
```diff
@@ -0,0 +1,86 @@
...
+
+        response = self.socket.recv(1024)
```
dthain: Ibid.

work_queue/src/json_server.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def wait(self, timeout):
+        request = {
+            "jsonrpc" : "2.0",
+            "method" : "wait",
+            "id" : self.id,
+            "params" : timeout
+        }
+
+        request = json.dumps(request)
```
dthain: Factor out the common dump/send/recv pattern into a single shared function.

work_queue/src/json_server.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def remove(self, taskid):
+        request = {
+            "jsonrpc" : "2.0",
+            "method" : "remove",
+            "id" : self.id,
+            "params" : taskid
+        }
+
+        request = json.dumps(request)
```
dthain: Ibid.

work_queue/src/json_server.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def disconnect(self):
+        request = {
+            "jsonrpc" : "2.0",
+            "method" : "disconnect",
+            "id" : self.id,
+            "params" : None
+        }
+
+        request = json.dumps(request)
```
dthain: Ibid.

work_queue/src/tasks.txt
```diff
@@ -0,0 +1,5 @@
+{ "command_line" : "sed s/number/No./g file1.txt > file1.replaced" , "output_files" : [ { "local_name" : "file1.replaced" , "remote_name" : "file1.replaced" , "flags" : { "WORK_QUEUE_CACHE": true , "WORK_QUEUE_NOCACHE" : false , "WORK_QUEUE_WATCH" : false } } ] , "input_files" : [ { "remote_name": "sed" , "local_name" : "/usr/bin/sed" , "flags" : { "WORK_QUEUE_CACHE" : false , "WORK_QUEUE_NOCACHE" : true , "WORK_QUEUE_WATCH" : false } } , { "remote_name" : "file1.txt" , "local_name" : "file1.txt" , "flags" : { "WORK_QUEUE_CACHE" : false , "WORK_QUEUE_NOCACHE" : true , "WORK_QUEUE_WATCH" : false } } ] }
```
dthain: In the C interface, all of the symbolic constants had to have `WORK_QUEUE` appended to the front in order to ensure there would not be namespace conflicts.  That's a bit repetitive here, so let's have the JSON API just look for simple names like `"cache" : true` etc.

Also note that `WORK_QUEUE_CACHE` and `WORK_QUEUE_NOCACHE` are antonyms, so really only the first one is needed.

work_queue/src/work_queue_server.c
```diff
@@ -0,0 +1,197 @@
...
+char * workqueue = "{ \"name\" : \"server_wq\" , \"port\" : 1234 }";
+
+void reply(struct link *client, char *method, char* message, int id){
+
+    struct jx_pair *jsonrpc, *result, *idd;
+    struct jx *j;
+    char *response;
+    time_t t;
+
+    idd = jx_pair(jx_string("id"), jx_integer(id), NULL);
```
dthain: Get in the habit of combining the variable declaration with initial assignment.
It shortens things and makes it clearer what the scope of each declaration is.

work_queue/src/work_queue_server.c
```diff
@@ -0,0 +1,197 @@
...
+    t = time(NULL);
+    link_write(client, response, strlen(response), t + 10);
+
+    jx_delete(j);
+
+}
+
+int main(){
+
+    const char *key;
```
dthain: Ibid.

work_queue/src/work_queue_server.c
```diff
@@ -0,0 +1,197 @@
...
+    while(true){
+    
+        //clear work queue
+        work_queue_cancel_all_tasks(q);       
+
+        //wait for client to connect
+        struct link* client = link_serve(2345);
+    
+        if (!client){
+            printf("Could not serve on port 2345\n");
```
dthain: Don't hard-code the port into the func call and string: put it in a variable.

work_queue/src/work_queue_server.c
```diff
@@ -0,0 +1,197 @@
...
+            return 1;
+        }
+    
+        t = time(NULL);
+    
+        client = link_accept(client, t + 100);
+
```
dthain: Q: What happens if multiple clients attempt to connect?

work_queue/src/work_queue_server.c
```diff
@@ -0,0 +1,197 @@
...
+            error = NULL;
+            id = -1;
+    
+            //receive message
+            t = time(NULL);
+    
+            read = link_readline(client, message, BUFSIZ, t + 10);
+    
+            if (!read){
+                error = "Error reading from client";
```
dthain: Under some of these error circumstances, the server should probably drop the connection to the client.  Let's justify those clearly.

work_queue/src/work_queue_server.c
```diff
@@ -0,0 +1,197 @@
...
+        while (true){
+    
+            //reset
+            error = NULL;
+            id = -1;
+    
+            //receive message
+            t = time(NULL);
+    
+            read = link_readline(client, message, BUFSIZ, t + 10);
```
dthain: Code pattern for dealing with network timeouts:

- Define a global "timeout" parameter.
- Call `link_whatever(.... time(0)+timeout)`

work_queue/src/client.py
camicarballo: What exactly do you mean by "native data structures" in this case?

dthain: Something like this:

```
tasks = [ { task 1 ... }, { task 2 ... }, ... ]

for t in tasks
   submit t
```

camicarballo: There is an issue with making the tasks native data structures in the client file. What seems to happen is that putting the JSON description into python as opposed to just treating it as a string in a text file messes things up with the boolean values. Python wants them capitalized, but JSON wants them lower case and so one is going to end up throwing an error.

dthain: So, the user would have to write `"cache":True` in the python version.
Does `json.dumps(t)` return a string with the proper JSON captialization?

work_queue/src/work_queue_server.c
camicarballo: What seems to happen is that one client is able to connect and the other one waits until the server is available again. The server however, cannot serve on the same port again for some reason.

dthain: The first part makes sense to me: the server is a single thread, handling the first client, and doesn't get around to calling link_accept again.

And, it makes sense that a second process cannot listen on the same port simultaneously.  However, I would expect that, after stopping the server, you would be able to restart it on the same port.

camicarballo:
Tested with BWA as a real-world example.

dthain:
So, I think that this effort has shown us that we can effectively built an application based upon a JSON API using a persistent server.  I think what I would like to do here is, instead of working it into WQ, is to make it the first draft of a dataswarm master that makes use of the WQ master library and worker as the backend implementation...  PR forthcoming...

dthain:
@camicarballo I'm not seeing the example BWA application written in JSON style, can you add it, please?

camicarballo:
I have the BWA-JSON example in another branch. Should I create a separate PR for that? @dthain

dthain:
Can you just add it to this one?

camicarballo:
I can move some files around and add it to this branch/PR, yes

PR #2245, opened 2020-01-13
Remove unused os checks

btovar:
Removes compile options and #ifdef's for cygwin, sunos, aix, and bsd's that are not darwin.


dthain:
Wow, I didn't know AIX was still in there!

btovar:
Yes, it feels like destroying history!

PR #2250, opened 2020-01-28
Check generic facility during compile time

btovar:
This pr adds check_facility to check at configure time whether generic code will compile.

As an example of its use, we check whether struct statfs s; s.f_flags = 0; is valid code.

As found by Emily, some struct statfs do not have the field f_flags (e.g., the one when using the compilers included in conda).

This became an issue until we included pr #2237, which uses f_flags to check whether files can be executed from the worker's scratch directory.

### DIFF
--- a/configure
+++ b/configure
@@ -1309,9 +1309,25 @@ optional_include sys/statvfs.h HAS_SYS_STATVFS_H
 optional_include sys/vfs.h     HAS_SYS_VFS_H
 optional_include syslog.h      HAS_SYSLOG_H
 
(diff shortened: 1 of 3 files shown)


dthain:
Suggestion:

In the place where the worker checks the stat flags (which, apparently, are not portable),
have it instead just create a micro script and attempt to execute it.  That should be portable
and covers any other situation where execution might be prevented (extended access controls, selinux, etc)

btovar:
Sounds good!

PR #2253, opened 2020-02-18
Python interface for wq factory

btovar:
@trshaffer From your code I renamed it to work_queue.Factory, made a distinction between command line only and config file options, and made some variables and functions class attributes.

Let me know if it still works as you intended.

### DIFF
--- a/work_queue/src/bindings/python3/work_queue.binding.py
+++ b/work_queue/src/bindings/python3/work_queue.binding.py
@@ -13,6 +13,12 @@
 
 import copy
 import os
+import json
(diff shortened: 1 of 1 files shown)


btovar:
@trshaffer, Tim, does this pr looks good to you?

trshaffer:
@btovar it looks like the options aren't translating between dashes and underscores when going from Python identifiers to JSON option names?

btovar:
Let me see...

btovar:
@trshaffer  could you give it another try?

PR #2254, opened 2020-03-04
Python Packaging Cleanup

trshaffer:
This PR reworks the `python_package_*` tools. For `python_package_create` and `python_package_run` I mostly fixed the sharp corners and cleaned generally cleaned things up. I more or less rewrote `python_package_analyze`, though. In particular,
- Rather than just grepping the program text, I used the `ast` module to parse the code, which allows more control (e.g. only add imports from inside a function).
- Switched the spec format to use that of `conda env export`, which can be fed directly to `conda env create`. This captures both Pip and Conda packages.
- Rather than using the external dependency `stdlib_list`, I checked the path of each imported module against `{sys.prefix}/pythonX.Y/site-packages` to determine if Pip/Conda should know about it.
- In some cases, the imported name does not match the package name in Pip/Conda (e.g. `ndcctools` provides `work_queue`). I put in more complicated logic to list package contents and try to match things up if using the name doesn't work.
- In case of weird metapackage layouts, added a command line option for manually specifying which package provides a given import name.
- Prints a warning if a package can't be found. This could indicate a weird name/layout, or something installed outside the Conda environment.

It's been working for me under a variety of packaging setups, and can handle a lot more than the original implementation. Unfortunately, the performance is worse (since we have to list the full contents of all packages). We might make this extra (expensive) check optional so that users can just manually specify packages when the names don't match up.


btovar:
@trshaffer, is this pr ready to be merged?

python_packaging/python_package_run
```diff
@@ -1,45 +1,33 @@
-#!/bin/sh
+#!/bin/bash
+
+set -e
 
 usage() {
     echo "Usage: python_package_analyze [options] <environment-name> <python-command-string>"
```
btovar, line 6: This one should be: python_package_analyze

btovar: also <python-command-string> should be  <command> <arg> ...

PR #2261, opened 2020-03-13
wq_bwa: fixed off by two error when submitting tasks

camicarballo:
There was an error where two of the splits were not being created/submitted as tasks. A simple fix is proposed here.

### DIFF
--- a/apps/wq_bwa/wq_bwa
+++ b/apps/wq_bwa/wq_bwa
@@ -128,10 +128,12 @@ sub split_query {
 		$line_count++;
 	}
 
+    $num_outputs++;
+
 	close(infile);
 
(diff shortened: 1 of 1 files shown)


apps/wq_bwa/wq_bwa
```diff
@@ -128,10 +128,12 @@ sub split_query {
 		$line_count++;
 	}
 
+    $num_outputs++;
```
btovar: I would not add this one, and instead use $i <= $num_splits;  in line 144.

camicarballo: Getting rid of that makes all the print statements be off by one. Which I guess isn't such a big deal, but still. Do you still want me to change?

btovar: I see what you mean.

PR #2266, opened 2020-03-19
Return the number of physical CPUs from load_average_get_cpus()

### LINKED ISSUE #2262: Worker Resource Detection and Hyperthreads
Currently, the worker detects the number of "virtual CPUs" which corresponds to hyperthreads.  It should probably detect physical cores instead, or at least have an option to do so.

trshaffer:
Currently, we count *logical* cores when examining the host system. On systems that make use of hyper-threading (especially extreme cases like Xeon Phi), this can lead to accidental over-subscription issues.

This PR changes `load_average_get_cpus()` to return the number of *physical* cores on the machine. For Macs, this is a simple change. For Linux, it requires some digging into `/sys`. I've tested a number of CPU configurations:
- Single-socket, no HT (cclws20)
- Single socket, HT (cclws19)
- Multi-socket, no HT (crcfe02)
- Multi-socket, HT (earth)

From what I could find online, the Xeon Phi registers as a normal CPU with HT, but I don't think I have access to any real machines to test on.

Resolves #2262

### DIFF
--- a/dttools/src/load_average.c
+++ b/dttools/src/load_average.c
@@ -31,6 +31,10 @@ int load_average_get_cpus()
 #elif defined(CCTOOLS_OPSYS_LINUX)
 
 #include <stdio.h>
+#include <stdlib.h>
+
+#include "stringtools.h"
+#include "string_set.h"
 
 void load_average_get(double *avg)
 {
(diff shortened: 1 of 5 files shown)


dttools/src/load_average.c
```diff
@@ -45,20 +49,30 @@ void load_average_get(double *avg)
 
 int load_average_get_cpus()
 {
-	FILE *f;
+	struct string_set *cores;
+	cores = string_set_create(0, 0);
+
+	for (int i = 0; ; i++) {
+		char *p = string_format("/sys/devices/system/cpu/cpu%u/topology/thread_siblings", i);
```
btovar: Tim, could be use /proc/cpuinfo instead and count the number of unique 'core id' lines? This may be a smaller change, since we have been reading /proc/cpuinfo up to now.

trshaffer: Oops, I forgot about this PR!

The `core id` lines refer to cores within a physical CPU. On multisocket machines, the `core id`s are not unique across physical CPUs, so that would find the max number of cores in any one CPU.

To use `/proc/cpuinfo`, we would have to parse out multiple fields and count the unique `physical id`, `core id` pairs. That made me nervous since `/proc/cpuinfo` is made for human consumption, so I'm not certain that e.g. field ordering and delimiters are consistent everywhere. So instead, I did the same thing as `lscpu` from `util-linux`.

PR #2268, opened 2020-03-29
Support new HTCondor job log format containing year

GregThain:
As of HTCondor version 8.9.3, the default userlog format puts the year
into the job log, looking like this

005 (312.000.000) 2020-03-28 23:01:04

the old format looked like

005 (312.000.000) 03/28 23:01:04

Currently, makeflow hangs forever, not seeing completed jobs when HTCondor
uses the new format.

This patch allows cctools to parse both the old and the new format. Note that
the code used to assume the year was 2008, this been updated to 2020, in the
case where we don't know.

Note that you can configure HTCondor to emit the old format by setting
DEFAULT_USERLOG_FORMAT_OPTIONS = legacy

### DIFF
--- a/batch_job/src/batch_job_condor.c
+++ b/batch_job/src/batch_job_condor.c
@@ -238,8 +238,22 @@ static batch_job_id_t batch_job_condor_wait (struct batch_queue * q, struct batc
 			struct batch_job_info *info;
 			int logcode, exitcode;
 
-			if(sscanf(line, "%d (%" SCNbjid ".%d.%d) %d/%d %d:%d:%d", &type, &jobid, &proc, &subproc, &tm.tm_mon, &tm.tm_mday, &tm.tm_hour, &tm.tm_min, &tm.tm_sec) == 9) {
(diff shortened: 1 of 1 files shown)


btovar:
Thanks Greg!

dthain:
Thanks for the catch, Greg! I vaguely remember thinking (in 2008) that this is a hack and ought to be fixed up at some point...

And now that I look at it, I see two more bugs:
A - We should compute the current year instead of using 2020.
B - I'll bet we get the wrong values when rolling over to a new year...

@btovar I think A is worth the small effort to solve.  I don't think B is worth the trouble (given that HTCondor is evolving) but it's probably worth a comment in the code...

btovar:
On it. I'll merge this request as is, and then fix A.

GregThain:
CC Lab:

No worries, we apologize for breaking this in the first place. And if we 
were really on the ball, our log reader would be usable, and you won't 
have to write this code anyways.


Stay Safe and happy hacking,


-greg

PR #2270, opened 2020-03-31
HTCondor year fallback for old log format

btovar:
Folow up to #2268.

Computes current year in case HTCondor log does not provide (old log format). Before we simply used a fixed value.

### DIFF
--- a/batch_job/src/batch_job_condor.c
+++ b/batch_job/src/batch_job_condor.c
@@ -219,6 +219,18 @@ static batch_job_id_t batch_job_condor_wait (struct batch_queue * q, struct batc
 		}
 	}
 
+	time_t current;
+	struct tm tm;
+
+	/* Obtain current year, in case HTCondor log lines do not provide a year.
+	   Note that this fallback may give the incorrect year for jobs that run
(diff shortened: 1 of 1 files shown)


batch_job/src/batch_job_condor.c
```diff
@@ -219,6 +219,18 @@ static batch_job_id_t batch_job_condor_wait (struct batch_queue * q, struct batc
...
+	time_t current;
+	struct tm tm;
+
+	/* Obtain current year, in case HTCondor log lines do not provide a year.
+	   Note that this fallback may give the incorrect year for jobs that run
+	   when the year turns. However, we just need some value to give to a
+	   mktime below, and the current year is preferable than some fixed value.
+	   */
+	time(&current);
+	tm = *localtime(&current);
```
dthain: Just occurred to me to ask: is the condor log date in localtime or UTC?

btovar: The manual does not specify it, but a quick test shows that it is local time. Since some events only involve the schedd, I think it is most likely the local time at the machine running the schedd.

PR #2273, opened 2020-04-06
Pedantic warnings confuga

btovar:
Warnings found with gcc 7.3.0.

I simply silenced the strict aliasing one, as I am not sure what those are lines are doing, and we should be replacing json.h and json.c with jx anyway.


dthain:
In an ideal world, we would update the confuga code to use the jx module.
However, that would be a lot of internal work for little external benefit.
I split the difference by moving the old json module into the chirp code,
so that it is localized to that particular use case...

Issue #2289, opened 2020-04-30
work queue has inefficient input directory transfers


btovar:
Right now we give a name a cache name to each of the files, when only the top directory should get one. This also stops us from streaming the contents when copying making it inefficient. 

The transfer behavior should be similar to chirp's transfer of directories.

dthain:
After looking at it a little more, the problem is not so much the network representation per se.  A hierarchy still gets "streamed" inasmuch as the files and directories are all packed into the stream without response.  The problem is that each file is processed separately by the worker, with a call to `create_dir` to create the directory hierarchy for every single file.  So, if you have a deep directory tree with a colossal number of mkdirs!

PR #2290, opened 2020-04-30
WQ Recursive Put

dthain:
(Note: This hasn't been tested at all, I'm making a PR so I don't lose the code here.)

Address the performance problem of sending large directories via WQ.
The existing implementation of recursive put stinks b/c it sends every file
as a distinct transaction, forcing the worker to check for and create all
parent directories for every file.

This implementation provides a proper streaming put protocol, to go with
the streaming get protocol already in use.  Directories are sent via a stream
of data:

```
dir d
file x size mode
(data)
file y size mode
(data)
file z size mode
...
end
```

In this scheme, all file names must be simple names without paths.
Should minimize the metadata interactions that the worker has with the underlying filesystem.

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -2790,7 +2790,10 @@ static int send_symlink( struct work_queue *q, struct work_queue_worker *w, stru
 	int length = readlink(localname,target,sizeof(target));
 	if(length<0) return APP_FAILURE;
 
-	send_worker_msg(q,w,"symlink %s %d\n",remotename,length);
+	char remotename_encoded[WORK_QUEUE_LINE_MAX];
+	url_encode(remotename,remotename_encoded,sizeof(remotename_encoded));
+
+	send_worker_msg(q,w,"symlink %s %d\n",remotename_encoded,length);
 
 	link_write(w->link,target,length,time(0)+q->long_timeout);
 
@@ -2847,8 +2850,12 @@ static int send_file( struct work_queue *q, struct work_queue_worker *w, struct 
 		effective_stoptime = (length/q->bandwidth)*1000000 + timestamp_get();
 	}
 
+	/* filenames are url-encoded to avoid problems with spaces, etc */
+	char remotename_encoded[WORK_QUEUE_LINE_MAX];
+	url_encode(remotename,remotename_encoded,sizeof(remotename_encoded));
+
 	stoptime = time(0) + get_transfer_wait_time(q, w, t, length);
-	send_worker_msg(q,w, "put %s %"PRId64" 0%o %d\n",remotename, length, mode, flags);
+	send_worker_msg(q,w, "put %s %"PRId64" 0%o %d\n",remotename_encoded, length, mode, flags);
 	actual = link_stream_from_fd(w->link, fd, length, stoptime);
 	close(fd);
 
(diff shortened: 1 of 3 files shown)


dthain:
see #2289

dthain:
Outstanding problems:
- Handle filenames with spaces.
- Consolidate use of lstat on client side.

dthain:
New version transferring coffea as an unpacked directory:
2020/05/01 09:53:12.57 makeflow[32407] wq: cclws15.cse.nd.edu (127.0.0.1:34130) received 694.26 MB in **1.75s** (397.23s MB/s) average 488.53s MB/s

New version transferring coffea as a packed tarball:
2020/05/01 09:36:24.34 makeflow[30847] wq: cclws15.cse.nd.edu (127.0.0.1:34080) received 244.50 MB in **0.18s** (1383.24s MB/s) average 1383.24s MB/s

dthain:
Old version could not transfer the directory at all b/c it contains filenames with spaces.

dthain:
@btovar please have a look.  Give it a try with coffea code unpacked?

btovar:
For some reason I cannot make it work for input directories when cache=False. Investigating.

Relevant log line:
wq: unable to create cache/file-2-751f6c044aefeb6682bacd8069b9ee48-my-input-dir

btovar:
Here are the relevant lines from the worker's point of view:

```
2020/05/05 12:57:10.15 work_queue_worker[13157] wq: rx from master: unlink file-2-751f6c044aefeb6682bacd8069b9ee48-conda-coffea-wq-env-py3.6
2020/05/05 12:57:11.63 work_queue_worker[13157] wq: rx from master: dir file-4-751f6c044aefeb6682bacd8069b9ee48-conda-coffea-wq-env-py3.6
2020/05/05 12:57:11.63 work_queue_worker[13157] wq: unable to create cache/file-4-751f6c044aefeb6682bacd8069b9ee48-conda-coffea-wq-env-py3.6: File exists
```
Perhaps that unlink is doing nothing to a directory?

dthain:
Taking a look...

dthain:
Hmm, I'm not able to reproduce this, can you show more details?

btovar:
It seems to happen when a task does not fit the worker anymore. It is marked as foresaken and returned to the master. However, the master does not call delete_uncacheable_files in such cases.

In my tests I was not specifying disk for the task or the worker. The task is allocated the value of the free space of the disk of the worker at some of measurement. If that free spaces goes down, then the task is marked as not fitting the worker.

I think this is unrelated to this pr.

dthain:
Good catch!

btovar:
Some numbers when using this pr to transfer conda environments as tar, tar.gz, and whole directories. The new put is now in the order of sending the directory as tar with no compression.

Times in seconds.
commit: time to transfer
runtime: time to untar (if needed), unpack,and run hello python.
task1: first task running with that environment.
task2: second task, reusing cache, unpack, etc.

.tar
task1 commit 60.94   runtime 4.29  total 65.24
task2 commit 5.6e-05  runtime 0.47   total 0.47

.tar.gz
task1 commit 20.79 runtime 7.00   total 27.79
task2 commit 7.8e-05  runtime 0.55    total 0.55

dir
task1 commit 65.71   runtime 1.03    total 66.74
task2 commit 3.6e-05 runtime 0.68     total 0.68 

All tasks2 have practically the same numbers, which is expected as at that point all are executing the same steps.

tar.gz has the advantage of a quicker worker setup, but it relies on going outside of work queue's cache abstraction. The environment is expanded to a directory unknown to work queue.

The whole directory has the advantage that everything is under control of work queue. The transfer times are worse, and the debug log is full of file transfers (this times the number of workers).

The time gained by not compressing the tar file is much less than the time gained by sending the tar.gz. 

One alternative to have the best of both worlds is to declare an empty cached directory as the input for the task. The tasks can then unpack the conda environment to this directory. In this way, work queue will be aware of the directory. The tool in charge of unpacking would need to make sure that no two task try to do the unpacking at the same time, and not to unpack if a previous task has already set the environment. The current python_package_run ensures this. 

Note that this last point still breaks work queue's abstraction by having some coordination between tasks. It is similar to tasks in a worker sharing the same parrot cache.

dthain:
There are several issues here that I am trying to chip away at.  This fixes some, but not all.

I think these are fixed:
1 - The raw transfer of whole directories was unnecessarily inefficient, due to implied directory creation.  Also some bugs related to file names.
2 - Transferring and unpacking tar files results two copies of the data at the worker (the packaged and the unpacked); transferring a single raw directory reduces worker storage and makes it much easier to account for storage consumption. 

Now, I think these problems are still outstanding:
3 - The raw directory transfer protocol is the equivalent of tar (interleaving metadata and data) but not of gz, so the total data transferred is greater.  We don't have a compress-while-transferring method.
4 - The whole directory transfer is not accounted for.  Rather, we check for space and count up storage on each file.  Might be worthwhile to look at the directory transfer as one resource management option.

It might be constructive to look at the problem in a different way.  Namely, the data begins in form a, and the user wants it in form b.  What's the best way to deliver it?

If a=targz and b=dir, then the master should send the targz and the worker should unpack it on the fly.  Or if a=dir and b=dir, then perhaps the master and worker should work together to compress and unpack as they go.  We have run into this before, note the tar_stream utility used by bxgrid.

btovar:
One further issue here is that it is not simply how we transfer the data, but there is a non-standard operation after the transfer. It is a=targz or a=dir  and b=dir_with_postprocessing.  We need a way to attach that user-defined post processing step to the data. (A tar_stream_and_relocate, maybe?)

dthain:
Are there other examples of postprocessing that we know of?  The more specific the expectations, the better performance we can deliver.

For example, there is a big difference between:
stream -> command -> directory
and:
stream -> file/dir -> command -> file/dir

btovar:
For conda unpack we may actually be able to turn it into a stream. The conda-unpack of a particular environment has a list of files that needs to be modified, and how they need to be modified. So we can turn it upside down and pipe the files through some relocating filter.

dthain:
@btovar Any objections to this approach as an improvement to directory transfer?  A full conda environment solution is a bigger problem.

btovar:
Looks good to me.

PR #2293, opened 2020-05-06
python_package_run  time measurements

btovar:
- Adds debug log messages to measure the time each step takes (untaring, unpacking, command line, etc.). Activate with -d or --debug  options.

- Adds --help-env-creation which shows a recipe on how to create an environment that this script can consume. (Mainly ensure that conda and conda-pack are included.)

- Accept -- as a separator between options and command line to be executed.

- Accept a directory as a possible conda environment (as an alternative to a tar file).

This is the script used when testing pr #2290

### DIFF
--- a/python_environments/python_package_run
+++ b/python_environments/python_package_run
@@ -4,6 +4,15 @@
 # This software is distributed under the GNU General Public License.
 # See the file COPYING for details.
 
+# Executes a command line inside a conda environment.
+# The conda environment can be supplied via a tar file or a directory.
+#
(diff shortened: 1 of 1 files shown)


dthain:
LGTM is this meant to be a production change, or just a debugging crutch?

btovar:
Short term, it is a production change for debugging/evaluation purposes :). We want to know how long it takes to launch the environment vs just running the task, etc.  This wrapper seemed like the natural place to instrument that.
As for the long term for the whole tool, it is needed for coffea unless work queue supports conda environments natively.

PR #2296, opened 2020-05-13
WQ: Rename Unprotected Symbols

dthain:
A small thing, but it's bugging me:

Renamed unprotected symbols:
SUCCESS -> WQ_SUCCESS
WORKER_FAILURE -> WQ_WORKER_FAILURE
APP_FAILURE -> WQ_APP_FAILURE

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -1579,7 +1579,7 @@ static void fetch_output_from_worker(struct work_queue *q, struct work_queue_wor
 	}
 
 	/* print warnings if the task ran for a very short time (1s) and exited with common non-zero status */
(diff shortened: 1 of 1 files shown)


btovar:
Would you like to change WQ_ to WORK_QUEUE_ to match the rest of the enums?

dthain:
Well, I don't want to create confusion with the `WORK_QUEUE_RESULT_xxx` enumeration.

How about the convention that `WORK_QUEUE_x` indicates a public symbol, and `WQ_x` indicates an internal symbol?  That would entail some quick renaming via sed.

I am looking ahead to changing all relevant symbols `WORK_QUEUE` -> `DATA_SWARM` and `WQ` -> `DS`

btovar:
Yes, that makes sense. I'm almost tempted to go the other way around and make the public symbols the ones with the shorter names.

PR #2300, opened 2020-05-20
Adds ubuntu image to travis testing

btovar:
Major issues were that the dash shell does not expand {}, and that the swig version was old.


dthain:
RTM?

PR #2305, opened 2020-06-01
first pass at conda environments on wq

btovar:
Here is the prototype for conda environments in work queue. The major complication was assigning the correct failure to the task. By default, the worker disconnects if something goes wrong with an input file, but in this case that this not very useful.

The conda environment is untared and unpacked just after the sandbox for the task is setup. This ensures that the process is only tried once per environment. If that process fails, I create a fake pid to associate with the task, so that the rest of the worker machinery can deal with the task as usual. (It is a fake pid in the sense that the task never runs, but it is a real pid as it comes from a fork that immediately exits. This to ensure that we don't send a signal to the wrong process.)

When the task is run inside the environment, the directory for the environment is accessed inside the cache directory (i.e.: "source ../cache/.....env-dir-name-expanded; task command). This breaks abstraction a little bit, but it was the only safe choice in order not to create a name that the task will later use.

You can see an example of its use in `work_queue_conda_example`.  Some preliminary numbers for a worker running locally:

```
./work_queue_conda_example 
0.930142 task vanilla output: Python 2.7.5

9.346275 task with conda output: Python 3.8.2

9.355874 task with conda output: Python 3.8.2
```
It takes about one second for the worker to connect, and the task without an environment to run. Then about 8 seconds to send the environment and untar it/unpacked it. Once the environment is setup, the second task using the environment runs almost immediately.

If this looks ok, I can make the python/perl bindings.

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -3323,6 +3323,11 @@ static work_queue_result_code_t start_one_task(struct work_queue *q, struct work
 		}
 	}
 
+	if(t->conda_environment) {
+		/* t->conda_environment is a cached name, thus we do not need to url_encode it again. */
+		send_worker_msg(q,w, "conda_env %s\n", t->conda_environment);
+	}
+
 	// send_worker_msg returns the number of bytes sent, or a number less than
 	// zero to indicate errors. We are lazy here, we only check the last
 	// message we sent to the worker (other messages may have failed above).
@@ -4783,6 +4788,30 @@ void work_queue_task_specify_algorithm(struct work_queue_task *t, work_queue_sch
 {
 	t->worker_selection_algorithm = algorithm;
 }
+
+void work_queue_task_specify_conda_env(struct work_queue_task *t, const char *local_tarball) {
+	assert(t);
+	assert(local_tarball);
+
+	if(t->conda_environment) {
+		fatal("A conda environment for the task has been already specified.");
+	}
+
+	work_queue_task_specify_file(t, local_tarball, path_basename(local_tarball), WORK_QUEUE_INPUT, WORK_QUEUE_CACHE);
+
+	/* Search for the cached name given to the file. */
+	struct work_queue_file *tf;
+	list_first_item(t->input_files);
+	while((tf = (struct work_queue_file*)list_next_item(t->input_files))) {
+		if(!strcmp(local_tarball, tf->payload)) {
+			t->conda_environment = xxstrdup(tf->cached_name);
+			break;
+		}
+	}
+
+	assert(t->conda_environment);
+}
+
 
 void work_queue_task_specify_priority( struct work_queue_task *t, double priority )
 {
--- a/work_queue/src/work_queue_worker.c
+++ b/work_queue/src/work_queue_worker.c
@@ -549,7 +549,6 @@ static int setup_conda_environment(struct work_queue_process *p) {
 				"mkdir -p %s\n"
 				"/bin/tar -C %s -xf cache/%s\n"
 				"%s/bin/conda-unpack", e->expansion, e->expansion, p->task->conda_environment, e->expansion);
-		debug(D_WQ, cmd);
 		int status = system(cmd);
 		free(cmd);
 
--- a/work_queue/src/work_queue_worker.c
+++ b/work_queue/src/work_queue_worker.c
@@ -78,6 +78,12 @@ typedef enum {
 	CONTAINER_MODE_UMBRELLA
 } container_mode_t;
 
+struct environment {
+	char *tarball;     /* Conda environemnt as produced by conda-pack. */
+	char *expansion;   /* Directory in cache/. Only set when environemnt has been expanded.*/
+	int error;         /* Whether the expansion had an error. 1 for yes, 0 for no. */
+};
+
 #define DOCKER_WORK_DIR "/home/worker"
 
 // In single shot mode, immediately quit when disconnected.
@@ -204,6 +210,9 @@ static struct itable *procs_complete = NULL;
 
 //User specified features this worker provides.
 static struct hash_table *features = NULL;
+
+//Conda environment records. cache/cached_name of tarball -> struct environment
+static struct hash_table *environments = NULL;
 
 static int results_to_be_sent_msg = 0;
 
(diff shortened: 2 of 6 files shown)


work_queue/src/work_queue_worker.c
```diff
@@ -78,6 +78,12 @@ typedef enum {
 	CONTAINER_MODE_UMBRELLA
 } container_mode_t;
 
+struct environment {
```
dthain, line 81: Let's get more verbose about names.
struct `wq_conda_environment`: if it can really only be used for conda.
struct `wq_software_environment`: if it has potential use outside of conda.

work_queue/src/work_queue_worker.c
```diff
@@ -78,6 +78,12 @@ typedef enum {
...
+	char *tarball;     /* Conda environemnt as produced by conda-pack. */
+	char *expansion;   /* Directory in cache/. Only set when environemnt has been expanded.*/
+	int error;         /* Whether the expansion had an error. 1 for yes, 0 for no. */
```
dthain, line 84: Booleans always seem to grow beyond their original purpose.
Let's call this `status` and have an enumeration for the possible states.

work_queue/src/work_queue_worker.c
```diff
@@ -205,6 +211,9 @@ static struct itable *procs_complete = NULL;
 //User specified features this worker provides.
 static struct hash_table *features = NULL;
 
+//Conda environment records. cache/cached_name of tarball -> struct environment
+static struct hash_table *environments = NULL;
```
dthain, line 215: Likewise, `conda_environment_table` or `software_environment_table`

work_queue/src/work_queue_worker.c
```diff
@@ -521,6 +530,54 @@ int link_recursive( const char *source, const char *target )
...
+		free(cmd);
+
+		if(!WIFEXITED(status) || WEXITSTATUS(status)) {
+			debug(D_WQ, "failed to expand environment %s", p->task->conda_environment);
+			e->error = 1;
+		}
+	}
+
+	if(e->error) {
+		/* get a fake pid to insert to table */
```
dthain, line 562: Funny, but effective!
Maybe expand this comment to explain to the casual reader
that yes, this really is necessary, in order to avoid namespace conflicts.

work_queue/src/work_queue_worker.c
```diff
@@ -531,12 +588,26 @@ static int start_process( struct work_queue_process *p )
...
+	int fake_pid = setup_conda_environment(p);
+
+	if (fake_pid > 0) {
+		pid = fake_pid;
+		p->pid = fake_pid;
+		p->task_status |= WORK_QUEUE_RESULT_ENV_SETUP_ERROR;
+	}
+	else if (container_mode == CONTAINER_MODE_DOCKER) {
 		pid = work_queue_process_execute(p, container_mode, img_name);
-	else if (container_mode == CONTAINER_MODE_DOCKER_PRESERVE)
```
dthain, line 536: I think it's time to get rid of container modes...

work_queue/src/work_queue.c
```diff
@@ -4784,6 +4789,30 @@ void work_queue_task_specify_algorithm(struct work_queue_task *t, work_queue_sch
...
 
+void work_queue_task_specify_conda_env(struct work_queue_task *t, const char *local_tarball) {
+	assert(t);
+	assert(local_tarball);
+
+	if(t->conda_environment) {
+		fatal("A conda environment for the task has been already specified.");
+	}
+
+	work_queue_task_specify_file(t, local_tarball, path_basename(local_tarball), WORK_QUEUE_INPUT, WORK_QUEUE_CACHE);
```
dthain, line 4800: Would it make sense to require the user to explicitly call `specify_file` and then `conda_env` separately.
This will permit the tarball to arrive in different ways, using the current (and future) capabilities of `specify_file`.
This would also resolve the asymmetry in the worker below.

work_queue/src/work_queue_worker.c
```diff
@@ -952,6 +1024,16 @@ static int do_task( struct link *master, int taskid, time_t stoptime )
 				work_queue_task_specify_environment_variable(task,env,value);
 			}
 			free(env);
+		} else if(sscanf(line,"conda_env %s",filename)==1) {
```
dthain, line 1027: See above about separating `specify_file` and `conda_env`, that would eliminate the manual fixup here.

work_queue/src/work_queue_worker.c
btovar: Re: renaming. Yes, that's much better.

Re: the separation of specify_file. The protocol actions are separate. The tarball is still sent with infile. With regards to the API, this is another wrinkle. First I tried it separate, but it did not feel natural, as you would have to specify the file twice (once to send it, once to mark it as a the environment). Also, you would have to specify a remote name for the tarball, which would never be used in the task space.

To solve the above, I tried adding a WORK_QUEUE_TARBALL type, that let you specify an input file as a taball, with a remote name as a directory to expand to. This did not work so well, because not all tar files are expanded equally. In conda environments there is no directory containing all subdirectories, and you have to: mkdir mydir; tar -C mydir tarball.tgz. For almost any other tarball, you would expect: mkdir mydir; tar -C mydir --strip-components=1 tarball.tgz, as they usually have a directory that contains all. 

With that, the tarball infile message would need to contain the argument to strip-components, or we would have to create a message specially for tarballs.

As it is right now, there is the advantage that a user has no expectation on where the environment would be mounted, only that it will be activated. This is important, as it is unpacked for all tasks, and not for a particular remote name directory that a user may specify.

In summary, it was more complicated than it first appeared!

dthain: Ok, that makes sense to me.  Probably safer to have a "narrow" API that works in one way, rather than multiple touchpoints that can be misconfigured...

work_queue/src/work_queue.c
btovar: Let's try that.  Specifying the file twice did not seem natural to me at first, but the more I think about it, it matches better with the rest of the API.

work_queue/src/work_queue_worker.c
btovar: What's another process among friends? :)

PR #2306, opened 2020-06-01
Do not force debug messages to the console

btovar:
Instead of forcing the message to the terminal, messages are written to stderr.

Also, syslog and journal log options were removed.


dthain:
I like it -- maybe just double-check that nothing strange happens when parrot_run emits debug messages...

PR #2317, opened 2020-07-06
Work Queue JSON API (take 2)

dthain:
Once again, a PR to merge work by @camicarballo on the JSON WQ API.
Had to make a new PR to get around merge problems. 
Supersedes #2316 and #2240


dthain:
@camicarballo @btovar please take a look through and give me your thoughts on this version of the JSON API.

What I did here is to clean up the server process a bit by adding options and some cctools standard behavior,
and also cleaned up the BWA application so that everything is generated by a Makefile.

Please give me your thoughts on some naming issues:
- I moved the client library to work_queue/src/clients/python3/work_queue_server.py.  The idea is that the client library will be quite simple and can be implemented in a variety of languages.  However, the name is awkward, since it's not the server, but a library used to connect to the server.  What do we call this thing?
- work_queue_server / manager / helper / something else?

There are a couple of things that I think can be fixed/improved, but I would like to file them as separate issues and then let Cami work through them one-by-one:
- Message envelope: just send the message length as an integer on one line, and use readline to consume it.
- Application/server connection: a fixed port number will be unreliable, instead have the application open a random port number and as the "server" to connect to it.
- Configuration: add "config" and "getconfig" RPCs to read back things like the worker connection port, IP address, and so forth from the server to the application.  That will then allow the server to pick a random worker port, and the application to read it back.
- The server process should terminate itself whenever it determines that the parent process is no longer there.

dthain:
@camicarballo @btovar What do you think?

camicarballo:
looks good to me!

work_queue/src/clients/python3/work_queue_server.py
```diff
@@ -0,0 +1,123 @@
...
+            "params" : taskid
+        }
+
+        return self.send_recv(request)
+
+    def disconnect(self):
+        self.socket.close()
+        Popen.terminate(self.server)
+
+    def wq_empty(self):
```
btovar, line 108: simply  'empty'?   No other call has the wq_ prefix.

work_queue/src/clients/python3/work_queue_server.py
```diff
@@ -0,0 +1,123 @@
...
+            "method" : "remove",
+            "id" : self.id,
+            "params" : taskid
+        }
+
+        return self.send_recv(request)
+
+    def disconnect(self):
+        self.socket.close()
+        Popen.terminate(self.server)
```
btovar, line 106: A   Popen.wait()  is needed after  Popen.terminate()

work_queue/src/work_queue_json.c
```diff
@@ -44,7 +44,7 @@ static const char *work_queue_task_properties[] = { "tag", "command_line", "work
 	"time_execute_cmd_finish", "total_transfer_time", "cmd_execution_time",
 	"total_cmd_execution_time", "total_cmd_exhausted_execute_time",
 	"total_time_until_worker_failure", "total_bytes_received", "total_bytes_sent",
-	"total_bytes_transferred", "time_app_delay", NULL
+	"total_bytes_transferred", "time_app_delay", "cores", "memory", "disk", NULL
```
btovar, line 47: We may want to transform this array into a hash_table in the future.

work_queue/src/work_queue_json.h
```diff
@@ -34,6 +34,7 @@ Once returned, it is safe to re-submit the same take object via
 @param q A work queue object.
 @param str A JSON description of a task.
 
+<<<<<<< HEAD
```
btovar, line 37: It seems some conflict got merged...

work_queue/src/clients/python3/work_queue_server.py
dthain: done

work_queue/src/work_queue_json.c
dthain: Yes, validating incoming json will require some care in the future...

dthain:
Thanks for the careful readthrough!

Issue #2313, opened 2020-07-02
JX Project Behavior


dthain:
Likewise, either I'm misunderstanding the behavior of project or what it's expected to do:

Using jx_test:

```
{}
project( {"name":name, "port":port}, fetch("http://catalog.cse.nd.edu:9097/query.json"))
value:      error("on line 2, name: undefined symbol")
```

More simply:

```
{}
project( a, [ {"a":10}, {"a":20}, {"a": 30} ] )
value:      error("on line 2, a: undefined symbol")
```

I was expecting this to return:

```
[10,20,30]
```

What's the expected behavior here?

nkremerh:
Interesting. I will have to look into this.

nkremerh:
The simple example is failing because the context is empty when it is evaluated (https://github.com/cooperative-computing-lab/cctools/blob/master/dttools/src/jx_eval.c#L596). This top-level evaluation occurs _before_ project merges the context and each object being considered ({"a": 10}, ...).

If we bypass this by checking if context is empty in jx_eval (if so, skip lookup), we can essentially pass through to project's functionality of doing jx_merge(obj, context) which will then allow jx_eval to work as expected.

dthain:
Hmm, I don't think that's it.  `jx_lookup` does the right thing if `context` is null or the wrong type: it just returns `NULL` indicating no such item:
https://github.com/cooperative-computing-lab/cctools/blob/master/dttools/src/jx.c#L173

I think this is another instance of lazy versus eager evaluation, right?  In `project(a,[x,y,z])` you don't want `a` evaluated until `project` begins to run and computes `eval(a,merge(x,context))`.  Alternatively, `eval(a,eval(x,context))`

nkremerh:
We are saying the same thing.

dthain:
Ok, propose a solution?

nkremerh:
Either enforce lazy evaluation only for JX querying functions (select, project, fetch, schema) _or_ switch to all lazy evaluation for JX. The former I am willing to wrestle with, the latter is honestly far afield from my research goals and a guaranteed time sink.

Do you have an alternative?

dthain:
I don't think it's necessary to do it for all JX functions.  In fact, I'm not sure that you should even make a blanket change to all the query-related functions.  Just determine the minimal change needed...

nkremerh:
On it.

nkremerh:
Fixed.

PR #2322, opened 2020-07-07
Modifies project to perform lazy eval like select

nkremerh:
Copies the same enforced lazy eval of the query within select and adds it to project.

Addresses #2313


dthain:
As before, please add an appropriate test case(s) to ensure that the desired behavior is present.

nkremerh:
Done.

PR #2325, opened 2020-07-14
Adding wq_hypersweep to cctools/apps

tjuedema:
This is a first pass at writing a ResNet workload to be distributed using Work Queue. It builds a ResNet model, trains the model against a portion of the CIFAR-10 image dataset, and validates the accuracy of the constructed model against a reserved portion of the dataset. Multiple instances of the workload can be executed with varied hyperparameters and the results of these executions can be displayed as a heatmap to highlight optimal run configurations. 

It is still at a stage where it exists as a collection of scripts, runs sequentially, and requires reworking to run in a distributed context.


dthain:
Good, this is on the right track.
Converting `sweep.sh` into a WQ application should be pretty straightforward, and that's your next step.
Test it first by having a single worker locally with all of the pip packages available.
The trick will be carrying the necessary packages from the master to the worker.
For that, check out `cctools/python_packaging/python_package_{analyze|create|run}`

tjuedema:
Will do, thank you for the pointer to python_packaging

apps/wq_hypersweep/resnet.py
```diff
@@ -0,0 +1,139 @@
...
+
+#LOG_DIR = './log'
+#get_ipython().system_raw(
+#    'tensorboard --logdir {} --host 0.0.0.0 --port 6006 &'
+#    .format(LOG_DIR)
+#)
+
+
+# retrieve dataset
+(x_train, y_train), (x_test, y_test) = tf.keras.datasets.cifar10.load_data()
```
dthain, line 52: Is this where the download is actually happening?
If so, I think you will want to download it once at the beginning of the sweep and pass it in as a file to be transferred...

tjuedema: That is correct. I'll change this section to reflect that suggestion when I distribute the workload.

dthain:
@tjuedema Will it disrupt your workflow if I accept this PR as a milestone for a sequential implementation?

tjuedema:
@dthain It won't disrupt my workflow, feel free.

Issue #2321, opened 2020-07-07
WQ-JSON Worker Port Detection


dthain:
As written, the WQ-JSON client tells the server to use a fixed port number for workers.  Again, this won't work due to collisions between port numbers.

Native WQ applications handle this by telling the manager to listen on port zero, (meaning any port) and then read back the actual port via `work_queue_port`.

Fix this up in WQ-JSON by adding a new API operation `status` that reads back all the high level configuration of the server: port, project_name, tasks ready/running/etc, workers connected/etc and send that back in one JSON blob.

So then the standard WQ-JSON app will create the server listening on port zero, then immediately to a `status` to get the port number back, and continue that way.

dthain:
@camicarballo per our discussion today, you should be able to use the `link` object as it is.
`link_address_local` tells you the address and port of the local connection.
http://ccl.cse.nd.edu/software/manuals/api/html/link_8h.html#acc347b042e023a279eec60de1395e9ae

If you want to write a little helper function `link_local_port` as a wrapper around `link_address_local`, that would be fine.
But I don't think any major surgery to link is needed.

camicarballo:
Should all of the properties of the struct work_queue_stats be returned? Or just some _important_ ones?

camicarballo:
Also, the port number cannot be obtained from this work_queue_stats structure as it has been deprecated. So, for the purposes of port detection, this issue is addressed by #2332. In terms of adding a `stats` operation to the WQ-JSON API, I can create a separate issue specifically for that if needed.

dthain:
The JSON `status` command need not correspond exactly to one single WQ C API call. 
Use `work_queue_port` to get the port and put that in the status field.
There are a lot of fields there, I think it would be sufficient just to put the `task_*` and `worker_*` fields in there to start.

dthain:
It's a small thing, but let's call this JSON API operation `status` rather than `stats`.
The former implies the current state of the system, while the latter implies "statistics".

Issue #2320, opened 2020-07-07
WQ-JSON Server Port Selection


dthain:
As written, the WQ-JSON Python client works by *telling* the server what port it should listen on.  This won't work in general production, since there are many reasons why those ports might already be in use.  (The most likely is that another WQ is already running there.)

Instead, have the server listen on port zero (means any port is OK), observe the port actually obtained then print a little message to stdout.  The client (just reading back via popen) reads the port number and connects there.  Then the actual port number doesn't matter at all.

PR #2332, opened 2020-07-21
first pass at fixing worker port selection issue

camicarballo:
Address issue #2320 and the port detection aspect of issue #2321 

Does a simple read of stdout before connecting to get the port the server is listening on and connects there.

### DIFF
--- a/work_queue/src/work_queue_server.c
+++ b/work_queue/src/work_queue_server.c
@@ -280,8 +280,9 @@ int main(int argc, char *argv[])
 	}
 
     // what is the port chosen for the server?
-    char addr[LINK_ADDRESS_MAX];
-    int local = link_address_local(server_link, &addr, &server_port);
+    char *addr;
+    addr = (char *)malloc(LINK_ADDRESS_MAX);
+    int local = link_address_local(server_link, addr, &server_port);
     if(!local){
         printf("could not get local address: %s\n", strerror(errno));
         return 1;
--- a/work_queue/src/work_queue_server.c
+++ b/work_queue/src/work_queue_server.c
@@ -84,8 +84,6 @@ void mainloop(struct work_queue *queue, struct link *client)
 		ssize_t read = link_read(client, message, 1, time(NULL) + timeout);
 		while(message[0] != '{') {
 
-			//printf("server number read: %c\n", message[0]);
-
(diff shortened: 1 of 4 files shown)


work_queue/src/work_queue_server.c
```diff
@@ -84,8 +84,6 @@ void mainloop(struct work_queue *queue, struct link *client)
 		ssize_t read = link_read(client, message, 1, time(NULL) + timeout);
```
dthain, line 84: Use `link_readline` here to do what you want.

work_queue/src/work_queue_server.c
```diff
@@ -272,12 +270,26 @@ int main(int argc, char *argv[])
...
+
 	struct link *server_link = link_serve(server_port);
 	if(!server_link) {
 		printf("could not serve on port %d: %s\n", server_port,strerror(errno));
 		return 1;
 	}
 
+    // what is the port chosen for the server?
+    char addr[LINK_ADDRESS_MAX];
+    int local = link_address_local(server_link, &addr, &server_port);
```
dthain, line 284: The name of an array is the address of its first element.
So just pass `addr` not `&addr`

dthain:
@camicarballo also fix this up to use `link_readline` instead of `link_read` one byte at a time.

camicarballo:
It was originally a `link_readline` but changed to a `link_read` as you requested since we can't expect that any message will be followed by a newline.

dthain:
Sorry, I think we are both mixed up here.

After looking again, I see that you are reading digits until you get to an opening brace,
and then assuming that it is the start of the JSON body.  That's non ideal b/c (1) Reading one byte at a time with `link_read` is highly inefficient at scale and (2) We may want the body to contain something other than JSON in the future.

Please set up the message envelope like this:

1 - Send an integer as a string followed by a newline.
2 - Send the message body (in this case JSON) immediately following.

That way, a message send looks roughly like this:

```
sprintf(line,"%d\n",length);
link_write(link,line,strlen(line));
link_write(link,data,length);
```

And a message receive looks roughly like this:

```
link_readline(link,line,sizeof(line));
link_read(link,data,length);
```

dthain:
Looks good!

PR #2335, opened 2020-07-23
adds context switches to the resource_monitor

btovar:
The motivation is to discriminate slowdowns caused by machine load, IO, and memory swap. The implementation follows cpu_time, where at each time step we compute the change from accumulated values.


dthain:
Makes sense.  I wonder if it's possible to get cache misses in some form as well?

btovar:
It does not seem as easy. It seems that for cache misses the monitor would have to look at hardware events itself, rather than simply trying to read the info from /proc.

dthain:
RTM?

btovar:
What do you mean?

dthain:
Sorry: Is this Ready to Merge?

btovar:
I was reading it as 'read the manual' :). Yes, I will merge it now.

Issue #2327, opened 2020-07-15
compilation error ("conflicting types for 'gettimeofday'")


bcdarwin:
Compiling with various GCC versions from 4.9 to 7 and with glibc 2.31 on Linux (NixOS):

```
COMPILE parrot_namespace.o
LINK libparrot_helper.so
mkdir -p lib64
COMPILE lib64/libparrot_helper.so
parrot_helper.c:51:5: error: conflicting types for 'gettimeofday'
 int gettimeofday( struct timeval *tv, struct timezone *tz )
     ^~~~~~~~~~~~
In file included from parrot_helper.c:20:0:
/nix/store/lqn6r231ifgs2z66vvaav5zmrywlllzf-glibc-2.31-dev/include/sys/time.h:66:12: note: previous declaration of 'gettimeofday' was here
 extern int gettimeofday (struct timeval *__restrict __tv,
            ^~~~~~~~~~~~
make[2]: *** [../../rules.mk:29: libparrot_helper.so] Error 1
make[2]: *** Waiting for unfinished jobs....
parrot_helper.c:51:5: error: conflicting types for 'gettimeofday'
 int gettimeofday( struct timeval *tv, struct timezone *tz )
     ^~~~~~~~~~~~
In file included from parrot_helper.c:20:0:
/nix/store/lqn6r231ifgs2z66vvaav5zmrywlllzf-glibc-2.31-dev/include/sys/time.h:66:12: note: previous declaration of 'gettimeofday' was here
 extern int gettimeofday (struct timeval *__restrict __tv,
            ^~~~~~~~~~~~
make[2]: *** [Makefile:36: lib64/libparrot_helper.so] Error 1
ar rv libbatch_job.a batch_file.o batch_task.o batch_wrapper.o batch_job.o batch_job_amazon.o batch_job_amazon_batch.o batch_job_dryrun.o batch_job_chirp.o batch_job_cluster.o batch_job_blue_waters.o batch_job_condor.o batch_job_local.o batch_job_work_queue.o batch_job_lambda.o batch_job_mesos.o batch_job_k8s.o batch_job_mpi.o mesos_task.o
```

dthain:
@bcdarwin thanks for the bug report.  It looks like trouble is occurring in places where we are overriding existing system calls.  Always tricky to do this in a maximally portable way.

@trshaffer please set up a suitable Nix environment and see if you can reproduce.

bcdarwin:
To reproduce:

[...]

PR #2337, opened 2020-07-23
Fix for gettimeofday()

trshaffer:
Pulling this out of the other PR, as it's higher priority (hard error blocking compilation). See #2327 for discussion


dthain:
RTM?

PR #2339, opened 2020-07-23
Adds JX like regex matching function

nkremerh:
Useful for expanding querying functionality to include `LIKE`-style functionality from SQL.

### DIFF
--- a/dttools/src/jx_eval.c
+++ b/dttools/src/jx_eval.c
@@ -350,7 +350,7 @@ static struct jx * jx_eval_operator( struct jx_operator *o, struct jx *context )
 
 	if(o->type==JX_OP_CALL && jx_istype(o->left,JX_SYMBOL)) {
 		const char *name = o->left->u.symbol_name;
-		if(!strcmp("select",name) || !strcmp("project",name) || !strcmp("like",name)) {
+		if(!strcmp("select",name) || !strcmp("project",name)) {
 			struct jx *r = jx_array_shift(o->right);
 			r = jx_string(jx_print_string((r)));
 			jx_array_insert(o->right, r);
--- a/dttools/src/jx_eval.c
+++ b/dttools/src/jx_eval.c
@@ -226,6 +226,8 @@ static struct jx *jx_eval_call(struct jx *func, struct jx *args, struct jx *ctx)
 		return jx_function_project(args, ctx);
 	} else if (!strcmp(func->u.symbol_name, "schema")) {
 		return jx_function_schema(args, ctx);
+	} else if (!strcmp(func->u.symbol_name, "like")) {
+		return jx_function_like(args, ctx);
 	} else {
 		return jx_error(jx_format(
 			"on line %d, unknown function: %s",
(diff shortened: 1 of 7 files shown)


nkremerh:
@dthain @trshaffer A (hopefully) nice and simple addition to the JX querying functions.

dthain:
Looks good, I like it.  (ha)

Please add brief documentation to `doc/manuals/jx-quick.md` and `doc/manuals/jx.md` so others can find it and use it.  (Also add `select` and `project` while you are thinking about it...)

dttools/src/jx_eval.c
```diff
@@ -342,13 +344,13 @@ static struct jx * jx_eval_operator( struct jx_operator *o, struct jx *context )
...
 	/*
-	For function calls select and project, convert the first argument
+	For function calls select, project, and like, convert the first argument
 	to a string as a way of deferring evaluation of expressions.
 	*/
 
 	if(o->type==JX_OP_CALL && jx_istype(o->left,JX_SYMBOL)) {
 		const char *name = o->left->u.symbol_name;
-		if(!strcmp("select",name) || !strcmp("project",name)) {
+		if(!strcmp("select",name) || !strcmp("project",name) || !strcmp("like",name)) {
```
trshaffer, line 353: Is quoting the first argument necessary here? If I try to get the regex from a variable, it crashes

    expression: like(infile,"test")
    jx_function_like: jx_function.c:794[DEVELOPMENT]: Assertion 'jx_istype(val, JX_STRING)' failed.
    Aborted (core dumped)

nkremerh: The quoting is necessary for the lazy eval of keys. I pushed a fix for using a variable as the regex: eval the regex onto the context before the matching. Give it a try now.

dthain:
Er, I think you are fixing the problem backwards by doing `eval` on the quoted argument.
Why was it necessary to (implicitly) quote the argument in the first place?

nkremerh:
I put a demonstrable example in the JX tests. Copy the following query from the test example and remove the part Tim highlighted in `jx_eval.c`.

```
select(like(".es.*", y), [{"x": 0, "y": "test", "z": 1.0}]);
```

Now replace the part Tim highlighted, and run it again.

dthain:
Hmm, something about that doesn't quite add up.

It seems to me that `select` itself should delay the evaluation of the left argument
until the elements of the list in the right argument are ready to be used.

And that left argument could have *any* sort of function call, not just `like`.
So I don't think the problem is in `like` per se.
I think the problem is elsewhere.

Does that make sense?

trshaffer:
I didn't pull your latest commit, but if I remove the print/eval quoting on `like()` it works fine (passes your test)

nkremerh:
You are...correct. Say no more, changes have been pushed. Thanks for pushing back on my unnecessary redundancy.

dttools/src/jx_function.c
```diff
@@ -771,12 +771,41 @@ struct jx *jx_function_schema(struct jx *orig_args, struct jx *ctx) {
...
+}
+
+struct jx *jx_function_like(struct jx *orig_args, struct jx *ctx) {
+	assert(orig_args);
+	assert(jx_istype(ctx, JX_OBJECT));
+	const char *funcname = "like";
+	const char *err = NULL;
+
+	struct jx *args = jx_copy(orig_args);
+	struct jx *val = jx_eval(jx_array_shift(args), ctx);
```
trshaffer, line 790: You're gonna be so upset....Your latest push double-`eval()`s. The arguments already get evaluated in `jx_eval_operator()`, so there's no need to eval again here

nkremerh: I'll remove that and push the change along with the doc changes in a bit.

nkremerh:
@trshaffer I think we're good now.

PR #2342, opened 2020-07-23
Refactor JX functions

trshaffer:
This turned into a semi-major refactor of JX functions. The main architectural change is in the way functions are evaluated. The function call operator is now oblivious to the particulars of different functions. The operation from `jx_eval_operator()` is immediately dispatched to `jx_eval_call()` with no further eval/checks. The latter was already switching on function name and all that, so this PR keeps all the function-related activities there. I also changed the convention for calling the C implementation of functions. Argument lists are **always** owned by the callee. In practice, this means that most functions have a `jx_eval()` as part of their dispatch, and each function implementation is responsible for deleting the arg list. The notable exceptions are @nkremerh's new functions, which instead get a `jx_copy()` and do the eval themselves. This avoids the need to do string quoting and re-parsing.

I also made some implementation changes to the function code. I replaced the error handling macro with a proper function, and fixed a number of memory leaks and a few use-after-frees. I think the functions are more readable/straightforward now.

@nkremerh since you know your new functions better than me, you should check things over to make sure I didn't break your stuff.

### DIFF
--- a/dttools/src/jx_eval.c
+++ b/dttools/src/jx_eval.c
@@ -450,6 +450,7 @@ static struct jx_item *jx_eval_comprehension(struct jx *body, struct jx_comprehe
 	struct jx *list = jx_eval(comp->elements, context);
 	if (jx_istype(list, JX_ERROR)) return jx_item(list, NULL);
 	if (!jx_istype(list, JX_ARRAY)) {
+		jx_delete(list);
 		return jx_item(jx_error(jx_format(
 			"on line %d: list comprehension takes an array",
 			comp->line
--- a/dttools/src/jx_eval.c
+++ b/dttools/src/jx_eval.c
@@ -186,48 +186,43 @@ static struct jx *jx_eval_array(struct jx_operator *op, struct jx *left, struct 
 static struct jx *jx_eval_call(struct jx *func, struct jx *args, struct jx *ctx) {
 	assert(func);
 	assert(args);
-	assert(args->type == JX_ARRAY);
-
-	if (!jx_istype(func, JX_SYMBOL)) {
-		jx_error(jx_format(
-			"on line %d, unknown function: %s",
-			func->line,
-			func->u.symbol_name
-		));
-	}
+	assert(jx_istype(args, JX_ARRAY));
+	assert(jx_istype(func, JX_SYMBOL)); //TODO
 
 	if (!strcmp(func->u.symbol_name, "range")) {
-		return jx_function_range(args);
+		return jx_function_range(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "format")) {
-		return jx_function_format(args);
+		return jx_function_format(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "join")) {
-		return jx_function_join(args);
+		return jx_function_join(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "ceil")) {
-		return jx_function_ceil(args);	
+		return jx_function_ceil(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "floor")) {
-		return jx_function_floor(args);
+		return jx_function_floor(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "basename")) {
-		return jx_function_basename(args);
+		return jx_function_basename(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "dirname")) {
-		return jx_function_dirname(args);
+		return jx_function_dirname(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "listdir")) {
-		return jx_function_listdir(args);
+		return jx_function_listdir(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "escape")) {
-		return jx_function_escape(args);
+		return jx_function_escape(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "template")) {
-		return jx_function_template(args, ctx);
+		return jx_function_template(jx_eval(args, ctx), ctx);
 	} else if (!strcmp(func->u.symbol_name, "len")) {
-		return jx_function_len(args);
+		return jx_function_len(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "fetch")) {
-		return jx_function_fetch(args);
+		return jx_function_fetch(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "select")) {
-		return jx_function_select(args, ctx);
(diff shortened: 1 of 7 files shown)


dttools/src/jx_eval.c
```diff
@@ -186,48 +186,43 @@ static struct jx *jx_eval_array(struct jx_operator *op, struct jx *left, struct
...
-
-	if (!jx_istype(func, JX_SYMBOL)) {
-		jx_error(jx_format(
-			"on line %d, unknown function: %s",
-			func->line,
-			func->u.symbol_name
-		));
-	}
+	assert(jx_istype(args, JX_ARRAY));
+	assert(jx_istype(func, JX_SYMBOL)); //TODO
```
dthain, line 190: That assert seems correct to me, why TODO?

dttools/src/jx_eval.c
```diff
@@ -186,48 +186,43 @@ static struct jx *jx_eval_array(struct jx_operator *op, struct jx *left, struct
...
 	} else if (!strcmp(func->u.symbol_name, "len")) {
-		return jx_function_len(args);
+		return jx_function_len(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "fetch")) {
-		return jx_function_fetch(args);
+		return jx_function_fetch(jx_eval(args, ctx));
 	} else if (!strcmp(func->u.symbol_name, "select")) {
-		return jx_function_select(args, ctx);
+		// NB: deferred eval
+		return jx_function_select(jx_copy(args), ctx);
```
dthain, line 218: `jx_eval` vs `jx_copy` is important here but easy to overlook, this is a good place to separate these cases to the end and explain with a big comment.

dttools/src/jx_function.c
```diff
@@ -25,20 +26,27 @@ See the file COPYING for details.
...
+static struct jx *make_error(const char *funcname, struct jx *args, const char *fmt, ...) {
+	assert(funcname);
+	assert(args);
+	assert(fmt);
+
+	va_list ap;
+	buffer_t buf;
+
+	buffer_init(&buf);
+	buffer_printf(&buf, "function %s on line %d: ", funcname, args->line);
```
dthain, line 41: +1 for eliminating macros.

dttools/src/jx_eval.c
trshaffer: Whoops, forgot that one! I wanted to check if the parser only accepts symbols there (so an assert is correct) or if it's possible to get other types (so an error message would be correct)

trshaffer: Aha!

    expression: "bad_call"()
    jx_eval_call: jx_eval.c:190[DEVELOPMENT]: Assertion 'jx_istype(func, JX_SYMBOL)' failed.
    Aborted (core dumped)

dthain: Good catch.  I would be satisfied if this failed at parsing rather than execution.  Unless you can think of a reason for having expressions that evaluate to functions?

trshaffer: Not really. We currently don't eval function names, so you can't do higher-order things

dttools/src/jx_function.c
```diff
@@ -551,262 +550,323 @@ static struct jx *expand_template(struct jx *template, struct jx *ctx, struct jx
...
+
+	result = jx_array(0);
 
 	struct jx *item;
 	for(void *i = NULL; (item = jx_iterate_array(objlist, &i));) {
-		struct jx *j = jx_eval(val, jx_merge(ctx, item, NULL));
-		assert(jx_istype(j, JX_BOOLEAN));
-		if(!j) {
-			err = "error evaluating select expression";
+		if (!jx_istype(item, JX_OBJECT)) {
```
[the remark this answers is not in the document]
trshaffer: Forgot those functions *do* eval the second arg. Should be fixed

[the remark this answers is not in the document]
nkremerh: One step closer:

```
Enter context expression (or {} for an empty context):
{};
Now enter expressions:
select(like(".*work-queue.*", type), [fetch("/disk/d11/tlq/0483083f-dc09-4426-a0d8-3374b02e7da3.json")]);
expression: select(like(".*work-queue.*",type),[fetch("/disk/d11/tlq/0483083f-dc09-4426-a0d8-3374b02e7da3.json")])
value:   
```

We should expect this to return `[ ]`.

```
select("type" == "test", 
[fetch("/disk/d11/tlq/0483083f-dc09-4426-a0d8-3374b02e7da3.json")]);
expression: select("type"=="test",[fetch("/disk/d11/tlq/0483083f-dc09-4426-a0d8-3374b02e7da3.json")])
Segmentation fault
```

Trying to quote a key (`"type"`) used to return a symbol lookup error I believe.

trshaffer:
@nkremerh you should run this PR through its paces (now includes `keys()` and friends). Since it's blocking changes to JX functions, I'd like to merge this as soon as it's ready

dttools/src/jx_function.c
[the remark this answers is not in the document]
trshaffer: Hmmm.... using `"type"=="test"` should give an empty result (since the two strings are never equal), but definitely not a segfault. Mind posting a backtrace?

nkremerh:
Same kind of query keeps tripping up:

```
Enter context expression (or {} for an empty context):
{};
Now enter expressions:
select(like("work-queue-master", type), [fetch("/disk/d11/tlq/0483083f-dc09-4426-a0d8-3374b02e7da3.json")]);
expression: select(like("work-queue-master",type),[fetch("/disk/d11/tlq/0483083f-dc09-4426-a0d8-3374b02e7da3.json")])
SELECT ENTERED
OBJLIST: []
value:  
```

`fetch` seems to be returning nothing, so the returned value is NULL rather than a list. Thoughts?

trshaffer:
Hmmm....if it's just not printing anything, it's probably a segfault or `NULL` being returned. Do you have that JSON file so I can run it through GDB?

trshaffer:
Ah, found it. There was a use-after-free in `select()` if an error happened during eval. I can get your example to run, and Valgrind doesn't see any leaks when running `jx.input` or your example.

nkremerh:
This ugly query works, so consider me happy:

```
project(type, [schema(project(work_queue_master, select(like("work-queue-master",work_queue_master["type"]),[fetch("/disk/d11/tlq/0483083f-dc09-4426-a0d8-3374b02e7da3.json")])))]);
expression: project(type,[schema(project(work_queue_master,select(like("work-queue-master",work_queue_master["type"]),[fetch("/disk/d11/tlq/0483083f-dc09-4426-a0d8-3374b02e7da3.json")])))])
value:      ["string"]
```

dthain:
Nice!

trshaffer:
OK, I tracked down the memory issue (leak fix duplicated in #2343, rebase turned it into a double free). This looks good to me, @nkremerh?

nkremerh:
Looks good to me.

Issue #2341, opened 2020-07-23
jx_parse_stream Breaks on Sufficiently Large JSON Document


nkremerh:
`jx_parse_stream` crashes after a malloc on sufficiently large documents. As an example, [this Makeflow log](https://github.com/cooperative-computing-lab/tlq/blob/master/example-dataset/server-data/7a608781-c422-43a8-aba0-d691548ffff9.json) parsed to JSON is 34MB.

Downloading this document locally and fetching this in `jx_test` using `fetch("/path/to/doc")` yields a segfault.

@dthain @trshaffer will probably be most interested in this.

dthain:
Posting Nate's comment from Slack:
```
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-119.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /afs/crc.nd.edu/user/n/nkremerh/cctools-source/dttools/src/jx_test...done.
(gdb) run
Starting program: /afs/crc.nd.edu/user/n/nkremerh/cctools-source/dttools/src/./jx_test
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Enter context expression (or {} for an empty context):
{};
Now enter expressions:
fetch("t2.json"); //This is a locally renamed copy of the parsed log
expression: fetch("t2.json")
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff78730b1 in _int_malloc () from /lib64/libc.so.6
Missing separate debuginfos, use: debuginfo-install glibc-2.17-307.el7.1.x86_64
(gdb) bt
#0  0x00007ffff78730b1 in _int_malloc () from /lib64/libc.so.6
#1  0x00007ffff7877184 in calloc () from /lib64/libc.so.6
#2  0x0000000000412c66 in xxcalloc (nmemb=1, size=32) at xxmalloc.c:51
#3  0x0000000000402b6d in jx_create (type=JX_STRING) at jx.c:48
[...]

PR #2343, opened 2020-07-24
JX Iterative Parsing

dthain:
Updated list and object parsing to be iterative instead of recursive.
This should reduce stack depth and address crashes in parsing large datasets per #2341
Still has a bug I haven't sorted out yet...


dthain:
@trshaffer please look over and make sure I didn't miss something...

Issue #2347, opened 2020-07-25
JX Cannot Index Immediately After Function Call


nkremerh:
The following expressions _should_ work but currently fail at the parsing stage:

```
select( like( "example", type ), [ fetch("test.json")["subcomponent1"] ] );
"jx parse error: line 2: list of items missing a comma or closing delimiter"
```

```
range(3)[0]
"jx parse error: line 3: unexpected token: 0"
```

Uncovered by @trshaffer and myself while testing for #2342.

dthain:
@trshaffer are you doing this one, or should I take a look?

dthain:
However:
```
(range(3))[0];
expression: range(3)[0]
value:      0

range(3)[0];
expression: range(3)[0]
value:      0
```

PR #2348, opened 2020-07-27
JX Prefix and Postfix Operators

dthain:
Fixed up parsing to allow for stacked prefix and postfix operators, for example:
```
not not true;
range(1,10)[5];
[1,[2,3,4],5][1][2];
```


dthain:
Closes #2347

Issue #2328, opened 2020-07-15
SLURM MPI Dispatch


dthain:
Makeflow does not currently dispatch MPI jobs to SLURM correctly.
The underlying issue is that SLURM expects a somewhat different invocation for MPI vs non-MPI jobs.

For example, a 4-core multithreaded job should be submitted like this:
```
sbatch -N 1 -n 1 -c 4 job.sh
```

While a 4-process MPI job (each process using one core should be submitted like this:
```
sbatch -N 1 -n 4 -c 1 job.sh
```

Further, `job.sh` must include a call to `srun --mpi=pmi2` in order to set up the MPI environment properly, in the latter case.

Solving this requires two things:
1 - Some sort of syntax in Makeflow to convey that a particular job employs MPI and the desired configuration of the job.
2 - Appropriate support in `batch_job_cluster` to convert that information into the appropriate `sbatch` commands for SLURM.

May need to consider whether similar support is needed for other systems.

btovar:
We have .MAKEFLOW_CORES to set the number of cores. One thing we could do is to add .MAKEFLOW PROCESSES. The example above would look like:

.MAKEFLOW CORES 1
.MAKEFLOW PROCESSES 4

I am not sure about the `srun --mpi=pmi2`, as it is modifying the command line for a specific batch system. For slurm we could have it so that  `srun ...` etc. is automatically added if  PROCESSES > 1.  

We would also need to add an option that disables any of these and only honors BATCH_OPTIONS. We have that for sge, but only for sge.

dthain:
I think that's on the right track.  Are the resources then divided among the processes, or is each process assumed to consume the stated cores, memory, and disk?

Perhaps we should be explicit about an application being MPI?

Agreed about the resource disabling.  Isn't that what `--safe-mode` does?

btovar:
I think we need to assume that each process consumes the stated resources.  For example, the resource monitor would have to check for --max-concurrent-processes=4 --cores=4.   Otherwise, we would get --cores=1, which probably is not what we want.

dthain:
That makes sense.

How about we make `PROCESSES` explicitly `MPI_PROCS` so that it's clear that we are enabling MPI functionality?
Is there any other supported programming model where we are expecting to dispatch P x ( CORES, MEM, DISK ) in parallel?

btovar:
There is the local MPI mode that Kyle added, and most likely some sge installations will have something similar.

dthain:
[...]

PR #2352, opened 2020-07-30
Correctly set number of MPI processes for slurm

btovar:
In response to issue #2328


btovar:
1. I have not tried it with slurm yet. I believe that stampede has slurm and that may be easy to try.
2. It is worth a try!

dthain:
This looks good to me.  RTM?

PR #2354, opened 2020-08-04
Adds distributed version of wq_hypersweep (PR off of Pr/2325)

tjuedema:
My pup chewed through the charger to my Macbook so I forked the original PR on my Windows machine to create a new dev environment. Probably not the best work around, but it seemed the most agreeable with the Github Desktop app.

Changes in this commit:
- Moves resnet code into a main loop in resnet.py
- Adds function load_cifar_data to resnet.py. This downloads/prepares the dataset just as tf.keras.datasets.cifar10.load_data would, but caches the dataset in the local work directory instead of in ~/.keras, allowing the dataset to be distributed to workers without modifying the permissions of the master.
- Adds output file option to resnet.py
- Cleans out commented Tensorflow logging functions from resnet.py
- Adds test.py, the Work Queue master program for this application.  At the moment, this is hardcoded to generate 400 tasks that execute resnet.py while varying its number of residual blocks and dropout rate. Currently uses blacklisting and task resubmission to resolve failing workers.
- Adds script.sh, a pilot script for use by test.py. Contains the commands to unpack and activate the conda environment for resnet.py, as well as execute it.
- Updates README.md to reflect latest build/execute instructions

Still needed:
- Add result consolidation and handling to test.py
- Adapt plot.py to automatically generate/name/label plots given a correct input
- Rename component scripts to something more meaningful
- Consolidate component scripts into one parameterized run script
- Adjust test.py to allow user to choose parameters and ranges to run sweeps on from the command line


dthain:
Ok to close #2325 ?

apps/wq_hypersweep/README.md
```diff
@@ -0,0 +1,78 @@
...
+python_package_anaylze resnet.py env.yaml
+python_package_create env.yaml env.tar.gz
+```
+<ul>The same task can be accomplished using conda-pack directly:</ul>
+
+```bash
+conda create -n res_env python=3.7 numpy=1.18.1 tensorflow=1.15.0  -o env.tar.gz
+conda pack -n res_env -o res_env.tar.gz
+```
+
```
dthain, line 46: Might be helpful here to list each of the executables and their purposes.

apps/wq_hypersweep/test.py
```diff
@@ -0,0 +1,64 @@
...
+def compose_task(i,j):
+	id = (i-1)*20+j
+	d_rate = i*0.05
+	r_blok = j
+	outfile = "results%d.csv" % id
+	command = "./script.sh results%d.csv %f %d" % (id,d_rate,r_blok)
+
+	t = Task(command)
+
+	t.specify_file("env.tar.gz", "env.tar.gz", WORK_QUEUE_INPUT, cache=False)
```
dthain, line 14: I think all these inputs should be cacheable since they don't change throughout the run, yes?

apps/wq_hypersweep/test.py
```diff
@@ -0,0 +1,64 @@
...
+				if t.hostname not in whitelist:
+					whitelist.append(t.hostname)
+			if t.return_status != 0:
+				print("stdout:\n{}".format(t.output))
+				print("Blacklisting host: %s" % t.hostname)
+				q.blacklist(t.hostname)
+				blacklist.append(t.hostname)
+				q.submit(t)
+				print("Resubmitted task (id# %s): %s" % (t.id, t.command))
+
```
dthain, line 56: Yep, this is the right idea for the first pass. For the next step the idea is to run a range of jobs, find the minimum, and then run again at a finer grain around the optimal.  Repeat N times so that we end up with multiple waves of jobs.

tjuedema:
Latest commit addresses comments on last code review. Now working on search refinement, improving linkage between the computation and plotting portions of the code, renaming the scripts, and consolidating the function of some of scripts directly into the program.

dthain:
@tjuedema I think it's helpful to proceed in bite size pieces.
If this version is working well, then let's merge the PR.
And you can make a new PR for the next round of cleanup, better search, etc.
Are you OK to merge this?

tjuedema:
I'm okay with this being merged. I don't have write access to cctools though.

PR #2355, opened 2020-08-04
WIP Application API for DataSwarm

camicarballo:


dataswarm/manager/dataswarm_file.h
```diff
@@ -0,0 +1,10 @@
+#ifndef DATASWARM_FILE_H
```
dthain, line 1: For consistency with other modules, this should be `dataswarm_file_declare`, `dataswarm_file_commit` etc

dataswarm/manager/dataswarm_file.c
```diff
@@ -0,0 +1,51 @@
...
+            project = value->u.integer_value;
+        } else if(!strcmp(key, "metadata")){
+            metadata = value->u.string_value;
+        }
+
+    }
+
+    // assign a UUID to the file
+    // TODO: create vv function
+    int uuid = dataswarm_assign_uuid();
```
dthain, line 38: For future generality, a uuid needs to be a string, since it may have a form like this:
https://en.wikipedia.org/wiki/Universally_unique_identifier

Also see `dttools/src/uuid.h` for our local implementation.

dataswarm/manager/dataswarm_manager.c
```diff
@@ -119,7 +120,21 @@ void handle_client_message( struct dataswarm_manager *m, struct dataswarm_client
...
 	} else if(!strcmp(action,"file_submit")) {
-		/* */
+        // read in json message
+        const char *params = jx_lookup_string(msg, "params");
+	   
+        // convert params argument to jx
+        struct jx *file = jx_parse_string(params);
+        
+        // declare_file()
+        int uuid = dataswarm_declare_file(file);
```
dthain, line 130: string uuid

dataswarm/manager/dataswarm_manager.c
```diff
@@ -119,7 +120,21 @@ void handle_client_message( struct dataswarm_manager *m, struct dataswarm_client
...
+
+        /* create jx of uuid response */
+        struct jx *response;
```
dthain, line 133: Need to create a JX object and add something to it.  (e.g. the response and the UUID.)

dataswarm/apps/dataswarm_test
```diff
@@ -32,9 +32,15 @@ def main():
     ds.connect('127.0.0.1', 1234)
 
     # Here, we are declaring tasks in WQ style.
-    # Instead, we should declare and upload the shared input ile.
+    # Instead, we should declare and upload the shared input file.
     # Then, declare each output file and task.
 
+    #TODO: declare input file
+    input_file = { "type": "input", "url": "/usr/share/dict/words", "size": "4.8M" "metadata": {}, "project": None }
+    ds.file_submit(input_file)
```
dthain, line 40: Make sure you get the response back here and do something with it.

dthain:
@camicarballo keep going, add a few more functions and lets see how it fits together.

dataswarm/manager/dataswarm_file.c
camicarballo: should it be of type string or of type cctools_uuid_t?

dthain: Generate a UUID, but pass it around as a plain string with no special interpretation.  The worker just knows that the manage r refers to blobs by name, and the naming scheme is up to the manager.  likewise, the application shouldn't make any limiting assumptions about the file name.

PR #2357, opened 2020-08-07
Adds TLQ URL retrieval to Makeflow, WQ Master, and WQ Worker

nkremerh:
This allows for Work Queue masters and workers to retrieve their TLQ URL from the catalog server and to advertise their URLs to each other.

Makeflow and workers are passed a `--tlq=<host:port>` flag to provide the phone home address TLQ uses to tell a user it has their logs. This is used to check that we are getting the right URL from the catalog. Makeflow passes this phone home address to the master (along with the path to the debug log) via the batch queue interface.

This functionality is turned on only when both the makeflow (by extension the master) and at least one worker have been given the `--tlq` option.


nkremerh:
Closing to rework on a more scalable approach.

PR #2358, opened 2020-08-11
Non-blocking IO module

trshaffer:
This PR adds a module for sending/receiving messages over sockets. Internally, it's all done using non-blocking IO. This approach allows for multiplexed IO, i.e. sending/receiving on multiple connections in parallel. Rather than taking a threaded approach, this module does as much IO as possible (until using up socket buffers), then does other work. Sends are queued, making the send buffer unlimited (up to system memory). Receives and socket accepts must be handled immediately by the application, however. (Note that keeping a received message will cause `_wait()` to return immediately without doing any IO, so it's important to handle events promptly.)

It ended up being quite a bit of code, since there's a little event loop in there that needs to be able to handle interruption and resumption at any time. By design, this module hides the gory details (which makes the implementation somewhat complex). There's also a moderate amount of magic involved in getting everything to work, though users of the module shouldn't need to know about that. 

There are still some items to do:
- **File transfer** Transferring files and/or directories will be an important addition for Data Swarm usage. The details of how to transfer/encode are TBD.
- **Serial/blocking implementation for comparison** I think this will just require small changes, however this mode is also more difficult to write tests for as it's vulnerable to deadlocks. This might be easier to evaluate once we have clients and servers going.
- **Error handling** Using non-blocking and async operations means that errors can show up in unexpected places, possibly distant from their sources. I tried to capture errors as much as possible, but I'm certain that the error handling isn't finished.
- **API** Not sure if the API is the most ergonomic. It might be more convenient to provide direct access to the buffer storage, or something else.
- **Extra features** In-stream compression would be the first thing that comes to mind.
- **`epoll()`** I designed the API such that we could switch to `epoll()` down the road without requiring changes elsewhere. If we're operating at scale (>1000 simultaneous connections) and `ppoll()` becomes a performance problem, we might want to look at this.

This PR implements enough to do JSON messages and in-memory buffers for either single connections or a pool of workers, which should be sufficient to get started with Data Swarm. I wrote a couple tests, and things seem to be working. Also runs clean under Valgrind. I'm sure there are still changes necessary, but this should be enough to get discussion going.


dttools/src/mq.c
```diff
@@ -0,0 +1,716 @@
...
+#include "xxmalloc.h"
+#include "debug.h"
+#include "jx_print.h"
+#include "jx_parse.h"
+
+
+#define HDR_SIZE (sizeof(struct mq_msg) - offsetof(struct mq_msg, magic))
+#define HDR_MAGIC "MQmsg"
+
+enum mq_msg_type {
```
dthain, line 35: Is it necessary to make mq aware of the payload type?
I'm wondering if it is sufficient to just move opaque buffers back and forth, and let the caller interpret it.

dttools/src/mq.c
```diff
@@ -0,0 +1,716 @@
...
+	out |= (uint64_t) d[4]<<24;
+	out |= (uint64_t) d[3]<<32;
+	out |= (uint64_t) d[2]<<40;
+	out |= (uint64_t) d[1]<<48;
+	out |= (uint64_t) d[0]<<56;
+	return out;
+}
+#endif
+
+static int ppoll_compat(struct pollfd fds[], nfds_t nfds, int stoptime) {
```
dthain, line 135: Does this compat implementation have a race condition, compared to native `ppoll`?
If so, we can't avoid it, but it should probably be documented.
Perhaps move to `dttools/src/ppoll_compat.c`?

configure.tools.sh
```diff
@@ -224,7 +224,7 @@ require_path ()
 
 require_function ()
 {
-	if check_function $1 $2
+	if check_function "$1" "$2" "-D_GNU_SOURCE"
```
btovar, line 227: It might be better to add  this define _GNU_SOURCE to the small test program that check_function writes. Note that we add _GNU_SOURCE in ccflags_append_define anyway.

dttools/src/buffer.c
```diff
@@ -66,12 +66,14 @@ void buffer_free(buffer_t * b)
...
 }
 
-/* make room for at least n chars */
-static int grow(buffer_t * b, size_t n)
+int buffer_grow(buffer_t * b, size_t n)
 {
 	size_t used = inuse(b);
 	size_t newlen = sizeof(b->initial); /* current buf is always at least as big as b->initial */
 
+	/* Keep using the initial buffer if possible */
```
btovar, line 74: Doesn't realloc do this anyway?

dttools/src/mq.c
```diff
@@ -0,0 +1,716 @@
...
+		case MQ_MSG_BUFFER:
+		case MQ_MSG_JSON:
+			break;
+		default:
+			return -1;
+	}
+	return 0;
+}
+
+static int unpack_msg(struct mq_msg *msg) {
```
btovar, line 264: As above, we may want to leave any jx out of this library, and add the jx conversion in DS proper.

dttools/test/TR_mq_poll.sh
```diff
@@ -0,0 +1,116 @@
+#!/bin/sh
+
+. ../../dttools/test/test_runner_common.sh
+
+exe="mq_poll.test"
+
+prepare()
+{
+	${CC} -I../src/ -g $CCTOOLS_TEST_CCFLAGS -o "$exe" -x c - -x none ../src/libdttools.a -lm <<EOF
+#include <assert.h>
```
btovar, line 10: I suggest making this example and the one below proper .c files. In that way it will be compiled with the flags as any other file. Otherwise, if we update ./configure, then we may also need to update this TR*.sh file.

dttools/src/link.c
```diff
@@ -437,21 +438,28 @@ struct link *link_serve_address(const char *addr, int port)
 struct link *link_accept(struct link *master, time_t stoptime)
 {
 	struct link *link = 0;
+	int fd = -1;
 
 	if(master->type == LINK_TYPE_FILE) {
 		return NULL;
 	}
 
+	do {
```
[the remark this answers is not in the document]
trshaffer: If I'm reading it correctly, the socket is set to nonblocking mode ([here](https://github.com/cooperative-computing-lab/cctools/blob/6c252547900f74aca35a1560c2e332c00f6d2a5c/dttools/src/link.c#L426)), so it *should* be safe to call `accept()` once to check for errors before sleeping?

dttools/src/mq.c
trshaffer: It does, I can move that to its own module

dttools/src/buffer.c
trshaffer: If we're still using the initial buffer in the struct, there would be an unconditional `malloc()` [here](https://github.com/cooperative-computing-lab/cctools/blob/6c252547900f74aca35a1560c2e332c00f6d2a5c/dttools/src/buffer.c#L91). It looks like the other functions check the available space before calling `grow()`, so to make this a friendly public function I checked that here too.

trshaffer:
From the comments, I'm thinking about an API change. First, remove JX parsing from the module and leave that up to the caller. Second, allow the sender and receiver to select the disposition of messages. Rather than `mq_send()`/`mq_recv()` and the `mq_wrap_*`/`mq_unwrap_*` functions, we would have
- `void mq_send_buffer(buffer_t buf)` (takes ownership of the passed-in buffer)
- `buffer_t mq_recv_buffer()`

Then we could add other dispositions, e.g.
- `int mq_send_file(const char *path)` (or maybe use fd?)
- `int mq_recv_file(const char *path)`
- `int mq_send_popen(const char *command)` (or maybe `argv` instead)
- `int mq_recv_popen(const char *command)`

The main question is how to tell the module where to put things *before* the message is received. That was the purpose of the separate `mq_message` type. Not sure if we should keep that approach, or do things in a more async way, i.e. to route the next message to a file, call `mq_recv_file()` immediately after handling the RPC header message (**before** waiting again). This could return some sort of `EALREADY`, then after waiting and receiving a message call again and return success. Thoughts?

trshaffer:
I made those requested changes and reworked the API to allow directly passing buffers (and in the future streaming to/from files and such) @btovar @dthain

dthain:
I think it's looking pretty good.

Now, considering your comments about how to deal with incoming data.  This is a bit of a problem, because the library doesn't know the intended destination until the caller of `mq_recv_whatever` indicates the type.  I suppose that you could just put some threshhold of message less than X are buffered in memory, and over X are sent to a file, but I think that's going to burn us sooner or later.

To me, that's an argument to *only* have buffering on sends, and not on receives.  (We only need one side buffered in order to avoid deadlocks anyway.)  That way, you don't start doing any work of reception until it becomes clear what the intent is. 

Or, put the two together, and you could be aggressive about reading messages below some size X, and delay on larger ones until the caller tells you what to do.  (Although, remember that you already get O(64KB) of buffering for free in the kernel via the TCP window.  And you could increase that via `link_window_set`)

What do you think?

trshaffer:
This latest round of changes does something similar to that. If the caller doesn't do anything special, messages just go to in-memory buffers. If you want to route to a file instead, you would tell the library that *before* receiving any data by calling `mq_store_file()` (or other functions if we have different ways to route messages). It's necessary to make that call either before waiting for the first time or immediately after receiving a message and before waiting again. This approach only works because we don't buffer received messages. So for Data Swarm, it might go like this:

1. Call `mq_recv()` with no special options, sends the JSON RPC message to an in-memory buffer.
2. Get a File Transfer message.
3. Call `mq_store_file(path)` **before** waiting (in the same loop iteration) to indicate that the next message should go to `path`.
4. Call `mq_wait()`/`mq_poll_wait()`, which does IO in the background.
5. Call `mq_recv()`, which indicates that the file is ready.

trshaffer:
Note that I didn't implement different storage stuff yest, just made room for it

trshaffer:
I think this is everything I wanted for the first draft. @btovar @dthain if this looks OK (especially the API to control storage [here](https://github.com/cooperative-computing-lab/cctools/pull/2358/files#diff-b2f1ac35e05916d85d87cda7210bf01fR327) that would work similarly for files), I think this is ready to merge.

PR #2359, opened 2020-08-12
python_package_run  do not perform unpack step every time

btovar:
When the environment directory is reused, this reduces the overhead of setting the environment by about 2 seconds.


dthain:
Looks good, merge if you are satisfied.

PR #2360, opened 2020-08-12
Adds TLQ URL retrieval to Makeflow, WQ Master, and WQ Worker

nkremerh:
This allows for makeflows, masters, and workers to retrieve their TLQ URL from their local TLQ log server and to advertise their URLs to each other.

Makeflow and workers are passed a --tlq=<port> flag. Makeflow passes this port to the master (along with the path to the debug log) via the batch queue interface. Makeflow, WQ master, and WQ workers each look up their TLQ URL from their local log server via TCP connection.

This functionality is turned on only when the makeflow (by extension the master) or a worker have been given the --tlq option.

Rework of #2357

### DIFF
--- a/dttools/src/address.c
+++ b/dttools/src/address.c
@@ -157,27 +157,4 @@ int address_parse_hostport( const char *hostport, char *host, int *port, int def
 	}
 }
 
-char *address_get_tlq_url( int port, const char *log_path ) {
-	struct sockaddr_in serv_addr;
-	char buffer[256];
-	strcpy(buffer, log_path);
-	serv_addr.sin_family = AF_INET;
-	serv_addr.sin_addr.s_addr = inet_addr("127.0.0.1");
-	serv_addr.sin_port = htons(port);
-	debug(D_NOTICE, "set server socket information");
-
-	int sockfd = socket(AF_INET, SOCK_STREAM, 0);
-	if(sockfd < 0) debug(D_NOTICE, "error opening local INET socket: %d - %s", errno, strerror(errno));
-	if(connect(sockfd, (struct sockaddr *) &serv_addr, sizeof(serv_addr)) < 0) debug(D_NOTICE, "error connecting to local INET socket: %d - %s", errno, strerror(errno));
-	int serv_write = write(sockfd, buffer, strlen(buffer));
-	if(serv_write < 0) debug(D_NOTICE, "error writing to local INET socket: %d - %s", errno, strerror(errno));
-	bzero(buffer, 256);
-	int serv_read = read(sockfd, buffer, 255);
-	if(serv_read < 0) debug(D_NOTICE, "error reading from local INET socket: %d - %s", errno, strerror(errno));
-	close(sockfd);
-	char *url = xxstrdup(buffer);
-	if(!url) url = "NONE";
-	return xxstrdup(buffer);
-}
-
 /* vim: set noexpandtab tabstop=4: */
--- a/dttools/src/address.c
+++ b/dttools/src/address.c
@@ -1,10 +1,14 @@
 #include "address.h"
 #include "debug.h"
+#include "xxmalloc.h"
 #include <assert.h>
+#include <errno.h>
 #include <string.h>
 #include <arpa/inet.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/socket.h>
+#include <unistd.h>
 
 int address_check_mode( struct addrinfo *info ) {
 	assert(info);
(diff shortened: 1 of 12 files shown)


dttools/src/address.c
```diff
@@ -153,4 +157,27 @@ int address_parse_hostport( const char *hostport, char *host, int *port, int def
 	}
 }
 
+char *address_get_tlq_url( int port, const char *log_path ) {
```
dthain, line 160: Some cleanup needed on this one:

1 - Instead of using raw socket code here, just use the `link` abstraction.  So, this bit of code becomes the much simpler `link_connect / link_write / link_read `.  There are a lot of technical hiccups that `link` solves for you, like ipv6, port ranges, timeouts, all sorts of things.

2 - Move this to a separate module called `tlq_config` or something like that.  The `address` module is just supposed to handle textual parsing of addresses, and is used by `link` and `datagram`, and  we want to avoid having a circular dependency.

work_queue/src/work_queue.c
```diff
@@ -562,6 +567,31 @@ work_queue_msg_code_t process_name(struct work_queue *q, struct work_queue_worke
...
+		else {
+			debug(D_TLQ, "error setting master TLQ URL - setting it to NONE");
+			q->tlq_url = "NONE";
+		}
+	}
+	else if(q->tlq_port && !q->debug_path) debug(D_TLQ, "cannot get master TLQ URL: no debug log path set");
+
+	//send master TLQ URL if there is one. otherwise send NONE
+	debug(D_TLQ, "sending master TLQ URL to worker (%s)", w->addrport);
+	send_worker_msg(q, w, "%s\n", q->tlq_url ? q->tlq_url : "NONE");
```
dthain, line 591: - Perhaps I'm missing something, but don't you want to send something like `"tlq %s"` rather than just the url alone?
- Suggest that you just not send a message at all if TLQ is not configured locally.

work_queue/src/work_queue_worker.c
```diff
@@ -213,6 +217,11 @@ static int total_tasks_executed = 0;
 static const char *project_regex = 0;
 static int released_by_master = 0;
 
+char *tlq_url;
```
dthain, line 220: static?

work_queue/src/work_queue_worker.c
```diff
@@ -442,6 +451,30 @@ static void report_worker_ready( struct link *master )
 	send_master_message(master,"workqueue %d %s %s %s %d.%d.%d\n",WORK_QUEUE_PROTOCOL_VERSION,hostname,os_name,arch_name,CCTOOLS_VERSION_MAJOR,CCTOOLS_VERSION_MINOR,CCTOOLS_VERSION_MICRO);
 	send_master_message(master, "info worker-id %s\n", worker_id);
 	send_features(master);
+
```
dthain, line 454: Break this out into a separate function `send_tlq_config` or the like, to keep an even high level structure of `report_worker_ready`

work_queue/src/work_queue_worker.c
```diff
@@ -442,6 +451,30 @@ static void report_worker_ready( struct link *master )
...
+		debug(D_TLQ, "looking up worker TLQ URL");
+		tlq_url = address_get_tlq_url(tlq_port, debug_path);
+		if(tlq_url) debug(D_TLQ, "set worker TLQ URL: %s", tlq_url);
+		else {
+			debug(D_TLQ, "error setting worker TLQ URL - setting it to NONE");
+			tlq_url = "NONE";
+		}
+	}
+
+	if(tlq_url) {
```
dthain, line 466: I would avoid blocking for a response here -- I don't think there is any other case where a blocking RPC is initiated by the worker, and strange things will happen if you have a version message.

Instead, move this logic down into `handle_master` so you can simple handle a `tlq` message whenever it happens to arrive.

nkremerh:
Ready for another pass, @dthain.

dttools/src/tlq_config.c
```diff
@@ -15,7 +15,10 @@ char *tlq_config_url( int port, const char *log_path ) {
 	struct link *server;
```
dthain, line 15: Pro-tip -- wherever possible, pass in `stoptime` as a parameter to your function, then just use that `stoptime` for each of the underlying `link` operations.  That way, timeouts get set in one place (in the worker) rather than having constants scattered throughout the code.

dttools/src/tlq_config.c
```diff
@@ -15,7 +15,10 @@ char *tlq_config_url( int port, const char *log_path ) {
...
 	strcpy(buffer, log_path);
 	server = link_connect("127.0.0.1", port, time(0) + 10);
-	if(!server) debug(D_NOTICE, "error opening local INET socket: %d - %s", errno, strerror(errno));
+	if(!server) {
+		debug(D_NOTICE, "error opening local INET socket: %d - %s", errno, strerror(errno));
+		return NULL;
+	}
 	
 	ssize_t serv_write = link_write(server, buffer, sizeof(buffer), time(0) + 10);
 	if(serv_write < 0) debug(D_NOTICE, "error writing to local INET socket: %d - %s", errno, strerror(errno));
```
dthain, line 24: Don't forget to close the link!

work_queue/src/work_queue.c
```diff
@@ -654,6 +688,8 @@ static work_queue_msg_code_t recv_worker_msg(struct work_queue *q, struct work_q
 		result = process_name(q, w, line);
 	} else if (string_prefix_is(line, "info")) {
 		result = process_info(q, w, line);
+	} else if (string_prefix_is(line, "tlq")) {
```
dthain, line 691: Much cleaner.

work_queue/src/work_queue.c
```diff
@@ -562,6 +566,36 @@ work_queue_msg_code_t process_name(struct work_queue *q, struct work_queue_worke
...
+		q->tlq_url = "NONE";
+	}
+
+	char worker_url[WORK_QUEUE_LINE_MAX];
+	int n = sscanf(line, "tlq %s", worker_url);
+	if(n != 1) debug(D_TLQ, "empty TLQ URL received from worker (%s)", w->addrport);
+	else debug(D_TLQ, "received worker (%s) TLQ URL %s", w->addrport, worker_url);
+
+	//send master TLQ URL if there is one. otherwise send NONE
+	if(q->tlq_url) {
```
dthain, line 592: I think there is some inconsistency here: in some places, you use `tlq_url="NONE"` to indicate that TLQ is disabled, and in others, it's `tlq=NULL`  For simplicity, I would suggest you stick with the latter, because then you can simply protect all the TLQ code with `if(tlq_url) { ... }`.

nkremerh:
These should be addressed now.

Issue #2346, opened 2020-07-24
Not all Python code is Python3 compatible


bcdarwin:
e.g., in the 7.1.6 release:

```
$ 2to3 makeflow_gatk 
RefactoringTool: Skipping optional fixer: buffer
RefactoringTool: Skipping optional fixer: idioms
RefactoringTool: Skipping optional fixer: set_literal
RefactoringTool: Skipping optional fixer: ws_comma
RefactoringTool: Refactored ./result/bin/makeflow_gatk
--- ./result/bin/makeflow_gatk	(original)
+++ ./result/bin/makeflow_gatk	(refactored)
@@ -18,7 +18,7 @@
         while rargs:
             try:
                 optparse.OptionParser._process_args(self,largs,rargs,values)
-            except (optparse.BadOptionError,optparse.AmbiguousOptionError), e:
+            except (optparse.BadOptionError,optparse.AmbiguousOptionError) as e:
                 largs.append(e.opt_str)
 
 def parse_req(parser):
RefactoringTool: Files that need to be modified:
RefactoringTool: ./result/bin/makeflow_gatk
```

and likely other places

dthain:
Hello @bcdarwin the `apps` directory has some examples of varying vintage that don't receive the same attention as our core code.

Is GATK something you are specifically interested in running at scale?

You may find the Makeflow Examples to be more helpful:
https://github.com/cooperative-computing-lab/makeflow-examples

bcdarwin:
I'm not specifically interested in this example, but building `cctools` does cause the `makeflow_gatk` binary to be installed.  If it's just an unmaintained example, perhaps that shouldn't be the case?

btovar:
Fixed in #2362

PR #2362, opened 2020-08-13
Update Apps to Python3

dthain:
- python 2->3 changes recommended by 2to3 (mostly print and except)
- /usr/bin/env python -> /usr/bin/env python3


dthain:
@btovar Is this consistent with how we are handling python versioning?  I seem to recall we abandoned the `/usr/bin/env cctools_python VERSION` approach.

dthain:
#2346

btovar:
Working on some minor fixes with extra parentheses.

PR #2374, opened 2020-08-28
dataswarm: client: add operations to client

camicarballo:
Add the following operations to the dataswarm client: 

task-submit, task-remove, task-retrieve
file-submit, file-commit, file-delete, file-copy
service-submit, service-delete
project-create, project-delete
status
wait


dthain:
This is the eright idea, however, I think when sending the parameters for each method, you want each parameter named independently, like this:

```
"params" {
  "timeout" : 10 
}
```

Rather than this:
```
"params" : 10
```

Make sense?

PR #2407, opened 2020-09-16
Stream file descriptors via MQ

trshaffer:
This PR adds support for sending/receiving of file descriptors via MQ. These could be regular files, pipes, `stdin`/`stdout`, etc. In the context of DataSwarm, this is useful for sending/receiving files and directories (e.g. spawn `tar` and hook the stdio stream to MQ). The library is designed to select over all the different file descriptors and handle both the TCP connections and send/recv fds asynchronously. I tried out a number of configurations and checked that everything is being sent properly, no busy waiting, etc. I included tests for most of the cases I can think of and everything seems to be working, so I think this is ready to try out in DataSwarm.

This PR took a while to get right, since there are a number of tricky bits.
- Support unknown-length streams (`tar` will output ??? bytes)
- Streaming fd on one side to buffer on the other side or vice versa
- Switching between socket and/or fd(s) with each message
- Dealing with hangups/disconnects

### DIFF
--- a/dttools/src/mq.h
+++ b/dttools/src/mq.h
@@ -55,6 +55,7 @@ See the file COPYING for details.
  *	struct mq *mq = mq_connect("127.0.0.1", 1234);
  *	mq_send(mq, msg);
  *	while (true) {
+ *		mq_store_buffer(mq, &buf);
  *		switch (mq_wait(mq, time() + 30)) {
  *			case 0:
  *			// interrupted by timeout or signal
@@ -62,7 +63,7 @@ See the file COPYING for details.
  *			// error or closed socket, mq should be deleted
  *			default:
  *			// got some messages!
- *			mq_recv(mq, &buf);
+ *			mq_recv(mq, NULL);
  *			// process the result
  *			// make sure something breaks out of the loop!
  *		}
@@ -120,7 +121,6 @@ See the file COPYING for details.
 
 typedef enum {
 	MQ_MSG_NONE = 0,
-	MQ_MSG_NEWBUFFER,
 	MQ_MSG_BUFFER,
 	MQ_MSG_FD,
 } mq_msg_t;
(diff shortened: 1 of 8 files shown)


dttools/src/mq.h
```diff
@@ -303,6 +315,22 @@ void *mq_poll_error(struct mq_poll *p);
...
+ * after passing it in. This function will read from fd until reaching the
+ * end of the file. This function will not seek fd, so it's possible to send
+ * slices from within files.
+ * @param mq The message queue.
+ * @param fd The file descriptor to read.
+ * @returns 0 on success. Note that this only indicates that the message was
+ *  successfully queued. It gives no indication about delivery.
+ * @returns -1 on failure.
+ */
+int mq_send_fd(struct mq *mq, int fd);
```
dthain, line 332: I think you want to pass in the size of the buffer as an argument here -- it's going to be a common case that the size of the file is known in advance (or it's coming from a messages of known length from the client)

dthain (commented):
Tim, this is looking good, except I think it's important to have the option of sending a known length instead of reading to end of file.

dthain:
So, if I am reading this correctly, you have separated the provisioning of storage (`mq_store_fd` or `mq_store_buffer`) from the actual blocking for a message (`mq_recv`).  So, that allows the message loop to stream a file to disk, and let you know when the whole thing is done.  The only complication here is that the recipient will need to keep track of the buffer/fd it provided in some auxiliary structure.  Did I get that right?

trshaffer:
OK, I think I cleared that up. Linux and OSX seem to act differently with buffered pipe data after a hang up.

trshaffer:
> The only complication here is that the recipient will need to keep track of the buffer/fd it provided in some auxiliary structure. Did I get that right?

Correct, though the return code from `mq_recv()` can "remind" the loop what it was doing. After parsing the messages received, the main loop can decide if the next thing should go to fd (just got a put) or to buffer (most messages). It just needs to make that decision before calling `mq_poll_wait()` again.

PR #2408, opened 2020-09-16
blob async rpcs

btovar:
Got the sequence  create->commit->delete for blobs working.

If this looks like a good idea, I can start working on put and copy, also tasks.

### DIFF
--- a/dataswarm/manager/dataswarm_blob_rep.h
+++ b/dataswarm/manager/dataswarm_blob_rep.h
@@ -4,52 +4,54 @@
 #include "dataswarm_rpc.h"  /* needed for dataswarm_result_t */
 
 typedef enum {
-    DS_BLOB_ACTION_NEW = 0,
-    DS_BLOB_ACTION_CREATE,
-    DS_BLOB_ACTION_PUT, DS_BLOB_ACTION_COPY,
-    DS_BLOB_ACTION_REQ_GET,
-    DS_BLOB_ACTION_GET,
-    DS_BLOB_ACTION_COMMIT,
-    DS_BLOB_ACTION_DELETE,
-} dataswarm_blob_action_t;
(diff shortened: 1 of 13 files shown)


dthain:
Per our discussion today, let's retreat to a simpler model of asynchrony, such that each object may have up to one operation outstanding.  You can simpllify the queue of requests to be a single requested_state and request_id.  Then, when responses come back, either advance or retreat the state.

btovar:
Now a blob goes through these stages:
https://github.com/cooperative-computing-lab/cctools/pull/2408/files#diff-02f8cae6f502e68fe908909684fbff85R16

blob-get is the trickiest, as it has to be done in two steps.

dataswarm/manager/dataswarm_blob_rep.h
```diff
@@ -0,0 +1,55 @@
+#ifndef DATASWARM_BLOB_REP_H
+#define DATASWARM_BLOB_REP_H
+
+#include "dataswarm_rpc.h"  /* needed for dataswarm_result_t */
+
+typedef enum {
```
dthain, line 6: I like this approach much better, the code seems much more straightforward to me.
Although shouldn't these be `DS_BLOB_STATE_X rather than `ACTION_X`?

dthain (approved):
This looks good!

dataswarm/manager/dataswarm_blob_rep.h
btovar: Yes, I wasn't too happy with the names. If we go for state, I would prefer:

DS_BLOB_STATE_CREATED, COMMITTED, etc.

dthain: Sure

PR #2410, opened 2020-09-22
Add File and Task Hash Tables

camicarballo:
Added the hash tables to the manager and a file struct to DataSwarm_file.h

Still need to hook up receivers to use these tables.


camicarballo:
@dthain @btovar I created a symlink to the DataSwarm_task.h/c files in the manager directory and I get a make error with a few of the #include statements. Should I unlink these two file and just create new ones in this directory or should I add the other required files here as well?

dthain:
I think (temporarily) you should add all the necessary files as symlinks to get this PR working.
And then our next step will be to move the shared objects into a common directory.

PR #2417, opened 2020-09-23
DS common files

btovar:
Moves code from task. blob and messages to directories task, blob, and comm.
Renames any non-public file from dataswarm_ to ds_, DATASWARM_ to DS_.

Only public facing files, such as dataswarm_manager, dataswarm_worker and datswarm_client_ops kept the full name.


dthain:
Hmm, I'm not convinced this is the right slicing.

For example, dataswarm_blob.c and dataswarm_task.c are definitely shared objects b/c they do nothing except translate between C and JX representations.

However, dataswarm_blob_table.c and dataswarm_task_table.c are definitely coupled to the worker because they implement the organization and action on blobs/tasks that are stored at the worker.

Can we make it work with just manager, worker, and common directories?

btovar:
Sounds good!

dthain:
This looks like the right idea to me, except for a few tidbits:

- Rename all files dataswarm_x to ds_x for consistency?
- ds_process is peculiar to the worker, I think.
- helpers -> ds_validate

Issue #2418, opened 2020-09-24
makeflow hangs on timeout when specifying directly -t  in BATCH_OPTIONS


btovar:
The issue is that on timeout, slurm is not updating the status file of the job. E.g:

```sh
$ cat slurm.status.6502438
start 1600959872
```
but:

```sh
squeue -u $USER
JOBID   PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
```
and:

```sh
$ sacct
6502438       makeflow0     normal tg-phy130+        272    TIMEOUT      0:0 
```

bcdarwin:
I have this issue as well with `makeflow -Tslurm` using `cctools 7.1.6` and a rather old `slurm 17.02.7'.

dthain:
@btovar recall that the status file is not created by SLURM per se, but by the wrapper script around the job.  So, if SLURM decides on its own to cancel the whole job, then the wrapper script doesn't get a chance to write the completion record.

btovar:
Ah yes, I was treating it as condor log.

btovar:
Fixed with #2419

PR #2419, opened 2020-09-25
Cluster hearbeats

btovar:
For issue #2418. Tested on stampede2.


batch_job/src/batch_job_cluster.c
```diff
@@ -316,6 +317,7 @@ static batch_job_id_t batch_job_cluster_wait (struct batch_queue * q, struct bat
 					}
 				}
 				fclose(file);
+				info->log_pos = ftell(file);
```
dthain, line 320: Dangling pointer to file

dthain:
Let's discuss with the reporter(s) about their expectations to make sure it's actually solving the problem.

PR #2444, opened 2020-10-07
DS unify blob enums

btovar:
Split ds_blob_table_delete into ds_blob_table_deleting and ds_blob_table_delete.   

*_deleting is called when the rpc blob-delete is received.
*_delete is called in a new function ds_blob_table_advance finds that there is something to do with a blob.

The final delete state is sent to the manager in an asynchronous update, as we do with tasks.

### DIFF
--- a/dataswarm/common/ds_blob.c
+++ b/dataswarm/common/ds_blob.c
@@ -94,3 +94,18 @@ int ds_blob_to_file( struct ds_blob *b, const char *filename )
 	return 1;
 }
 
+char *ds_blob_state_string( ds_blob_state_t state )
+{
+	switch(state) {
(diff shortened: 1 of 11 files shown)


dthain:
No - we want to keep the deleting state.

Keep in mind that deletion at both the manager and the worker may be slow or delayed.

When the client calls "file_delete" it will immediately mark the file as deleting, and return quickly.  Then, the scheduler will work through the blobs of the file deleting them, and then eventually mark the file as deleted when done.  

A similar story applies when the manager talks to the worker: first it marks things as deleting, then transitions to deleted and removes the record when the deletion is complete.

Same story with tasks.

PR #2456, opened 2020-10-19
JX json schema test

btovar:
This is what JX validation would look like if we use jsonschema (https://json-schema.org/)

### DIFF
--- a/doc/manuals/jx/jx-schema.json
+++ b/doc/manuals/jx/jx-schema.json
@@ -1,0 +1,139 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "$id": "https://cctools.readthedocs.io/en/latest/jx/jx",
+
+    "definitions": {
+        "environment": {
+            "type": "object",
+            "patternProperties": {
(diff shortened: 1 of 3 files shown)


dthain:
A couple thoughts:

1 - Going forward, we should be careful to distinguish between JX proper (just JX plus expressions, no schema) and JX used to express workflows, which does have a schema.  Maybe we need a different name for the latter.

2 - You could also look at this from the perspective of just the parser structure.  Right now, the JX -> makeflow translation happens like this, and ignores unexpected properties:

```
foreach ( expected key k ) { value = jx_lookup(object,k); }
```

When really what we want is this:

```
foreach( key k,v in object ) { if( k is expected ) use k,v; else error;
```

btovar:
In general I think we should change to  if (k is expected...) etc., not only for makeflow, but for DS, etc.  I also think that it would be very useful to have explicit schemas mirroring the implementation, for documentation, sanity checks, etc.

dthain:
Yes, I like that idea.

PR #2475, opened 2020-11-17
Dataswarm DB Initial

dthain:
A first pass at a simple persistence layer for the manager.
On restart, the manager loads all objects from disk.
Manipulate objects in hash tables as before.
When you complete a transaction on a task/file, then call ds_db_commit_task/file to force it to disk.
Not hooked up to anything yet.

### DIFF
--- a/dataswarm/common/ds_blob.c
+++ b/dataswarm/common/ds_blob.c
@@ -94,7 +94,7 @@ int ds_blob_to_file( struct ds_blob *b, const char *filename )
 	return 1;
 }
 
-char *ds_blob_state_string( ds_blob_state_t state )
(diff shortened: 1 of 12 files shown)


dthain:
I'm still pondering the persistence model and how to make it fit with the data queries needed for the basic scheduler.

Do we store nested data that corresponds to the expected access patterns?
One table of tasks, which refer to files by uuid.
One table of files, which contain blobs as sub-documents.
One table of workers.  (Do we need workers to be a persistent data structure?)

Or do we make flat tables as if we had a normalized database?
One table each for tasks, files, blobs, workers,...
This would be nice and easy to store, but requires that we construct some complex queries to answer questions like:
- Where are all the blobs for file f?
- What blobs does worker w contain?

PR #2476, opened 2020-11-19
Makeflow Status File

tphung3:
Changes are in 
makeflow.c: added calls to new function makeflow_file_summary
makeflow_catalog_reporter.c: added function is here
makeflow_catalog_reporter.h: added function header in here


dthain:
Good, you are on the right track.  Expect that every PR you may will result in comments and refinements until it is ready to commit.  Some things to fix up:

1 - The automated build tests are failing -- you can click through the reports above to see the build and test logs on each platform.  Note that you can run the tests yourself on your workstation by running `make` and then `make test` at the top level.  (It's not immediately obvious to me what's the source the failure.)

2 - Remove the excess printfs and commented-out code.  That's fine when you are testing, of course, but clean up when committing.

3  - In the output, add a little colored bar representing the total workflow progress.

trshaffer:
Looks like some of failures are related to Makeflow gc tests. Those ones are expecting specific files to be cleaned up mid-workflow, so they fail if the current directory listing changes. If your new feature is behind a command line flag, it should resolve that

dthain:
Right, the behavior that would be consistent with the rest of makeflow is to only write the status file if the option `--status-file <filename>` is given.

tphung3:
Should I fix the status file name to status.html and make the status update interval configurable - something like `--status-file <update interval in seconds>` - or should I fix the interval to 60s and make the status file name configurable as you suggested above?

dthain:
Suggest that you do both separately:

```
--status-file <filename>
--status-file-interval <time>
```

makeflow/src/makeflow.c
```diff
@@ -2119,6 +2152,14 @@ int main(int argc, char *argv[])
 			case LONG_OPT_TLQ:
 				tlq_port = atoi(optarg);
 				break;
+			case LONG_OPT_FILE_STATUS:
+				file_status_on = 1;
+				file_status_name = optarg;	
+				break;
+			case LONG_OPT_FILE_STATUS_INTERVAL:
+				file_status_on = 1;
+				file_status_interval = atoi(optarg);
```
dthain, line 2161: You can use the function `string_parse_metric` in `dttools/src/stringtools.h` which parses various formats of hours, minutes, seconds, etc.

makeflow/src/makeflow.c
```diff
@@ -1109,7 +1138,7 @@ static void makeflow_run( struct dag *d )
 
 	/* Always make final report to catalog when workflow ends. */
```
dthain, line 1139: Make sure to emit a final status file when the workflow is done, so that the status file reflects all jobs complete.

dthain (commented):
On the right track, it passes all of the tests now.

dthain:
Careful!  I think you just pushed reverts that removed all of your work!
That's probably because this PR is based off of your master branch.

tphung3:
I got a conflict in my origin master when trying to pull changes from ccl master, so I decided to make a branch in my fork - the makeflow_status_file branch. My apologies I should have notified you first.

PR #2484, opened 2020-12-03
WIP: Catalog Remote Query

dthain:
Add new URL to catalog server: `/query/XXX` where XXX is a base-64 encoded JX query expression.
Will be needed for efficiency as we increase number of items reported to catalog.


dthain:
@btovar, your thoughts on the following?

Catalog server exposes new URLs:
`/query/XXX` - Gives back JSON matching JX expression XXX.
`/history/AAA/BBB/XXX` - Gives back DeltaDB event stream from time AAA to time BBB matching JX expression XXX.

Client side tools:
`catalog_query` will now send the query string over to the catalog for execution there.  If that fails, it falls back to the old method of fetching all data via `/query.json` and filtering at the client.
`deltadb_query` can now fetch a data stream from the catalog server  and apply all of the deltadb query tools to that stream.

To make this happen:
- Factored out deltadb queries into a standalone module.
- Moved catalog_server over into deltadb directory to satisfy dependencies.
- Hooked up catalog server to deltadb queries directly.  (Previously, it implemented write-only deltadb via jx_database.  That module is now renamed deltadb.)

Placement:
The catalog clients and tools are still in `dttools` while the `catalog_server` is now in `deltadb`.  Does this make sense, or would it just be simpler to roll everything in `deltadb` back into `dttools` ?

dthain:
This is running on the backup catalog server and at least initially appears to be good.

Example query:

```
/deltadb_query --catalog backup-catalog.cse.nd.edu --from 2020-12-01 --every 1h --filter '(type=="wq_master")' --output 'COUNT(name)' --output 'SUM(tasks_running)'
```

However, the output is polluted by some periodic outputs from `curl`.

PR #2489, opened 2020-12-08
Work Queue Failure Only Flag

dthain:
Files marked at WORK_QUEUE_FAILURE_ONLY will only be returned if the task does not succeed.  This is handy for tasks that generate large log files at runtime.  Don't bother bringing them back if not needed.


btovar:
Do you want me to add the python bindings?  Like for cache we have cache = True|False, which is much nicer in python.

dthain:
Please!

dthain:
Ok, ready to merge?

Issue #2479, opened 2020-11-30
Makeflow: Add LSF Support


dthain:
Add support for the LSF batch system to Makeflow, and work with Larry Singh to make sure that it works for his particular LSF system and workflow.

At first glance, this looks like a straightforward update of `batch_job_cluster.c`: add LSF to the batch type enumeration and add cases for running `bsub` and appropriate commands to execute each makeflow job.  Consult the LSF user documentation for the key ideas:
https://www.ibm.com/support/knowledgecenter/SSWRJV_10.1.0/lsf_admin_foundations/working_lsf.html

dthain:
Adding @larryns to the ticket.

dthain:
Documentation for the facility:
https://hpcwiki.pmacs.upenn.edu/wiki/index.php/HPC:User_Guide

dthain:
@larryns can you share the command lines that you use to get your jobs running, along with the output that comes back, showing the job id that was submitted?

larryns:
Sure, there are two ways I usually do it:

1. bsub -J <jobname> -M <memory in Mb> -n <number of cores> -o <output file> -e <error file> <shell script>

e.g. bsub -J runmeJob -M 42000 -n 12 -o runme.o -e runme.e runme.sh
will request 42Gb, 12 cores and run the script runme.sh

2. Alternatively you can put the requirements in the shell script, e.g.

#!/bin/bash
#BSUB -J runmeJob
#BSUB -e runme.e
#BSUB -o runme.o
#BSUB -M 42000
#BSUB -n 12

... rest of shell script follows...
----------------------------

Submit with: 
> bsub < runme.sh

For both submissions, you'll get a response like:

Job <67550203> is submitted to default queue <normal>.

I use both methods; I guess it depends on which is more convenient for you to code.

-Larry.

dthain:
Is that the literal output, or did you add the angle brackets around the job number?

```
Job <67550203> is submitted to default queue <normal>.
```

larryns:
Literal, straight copy and paste. I didn't add the angle brackets.

dthain:
PR #2481

dthain:
@larryns we have a prototype here for you to try.  There is always some unexpected oddity that comes up with a new system.  Please try this out and let us know how it goes for you:

You will have to install from source:
```
git clone https://github.com/cooperative-computing-lab/cctools
cd cctools
./configure --prefix $HOME/cctools-test
make install
[...]

PR #2490, opened 2020-12-10
Batch Job LSF Fixup

dthain:
Put dot-slash in front of the batch job wrapper, solves a submission problem on LSF.


dthain:
#2479

PR #2503, opened 2021-01-05
allow jobs to run on sge with sge_submit_makeflow

btovar:
Previously, submitted makeflow would run rules using work queue workers. With these changes, rules may run as sge jobs.


btovar:
Incorrect approach. Nodes usually cannot submit sge jobs themselves.

PR #2505, opened 2021-01-08
adds WQ tasks resources display

tphung3:
WQ resources for tasks are now displayed when using 
`work_queue_status -R`

### DIFF
--- a/dttools/src/jx_table.c
+++ b/dttools/src/jx_table.c
@@ -112,7 +112,13 @@ void jx_table_print( struct jx_table *t, struct jx *j, FILE * f, int columns_max
 		}
 		else if(t->mode == JX_TABLE_MODE_MEGABYTES) {
 			line = malloc(10);
-			string_metric(jx_lookup_integer(j,t->name), -1, line);
+			string_metric(jx_lookup_integer(j,t->name), 0, line);
+			for (int i = 0; line[i] != '\0'; i++){
+				if (line[i] == '.'){
+					line[i] = '\0';
+					break;
+				}				
+			}
 		} else {
 			int found;
 			struct jx *v = jx_lookup_guard(j,t->name,&found);
--- a/dttools/src/jx_table.c
+++ b/dttools/src/jx_table.c
@@ -109,6 +109,10 @@ void jx_table_print( struct jx_table *t, struct jx *j, FILE * f, int columns_max
 			line = malloc(10);
 			string_metric(jx_lookup_integer(j,t->name), -1, line);
 			strcat(line, "B");
+		}
+		else if(t->mode == JX_TABLE_MODE_MEGABYTES) {
+			line = malloc(10);
+			string_metric(jx_lookup_integer(j,t->name), -1, line);
 		} else {
 			int found;
 			struct jx *v = jx_lookup_guard(j,t->name,&found);
--- a/dttools/src/jx_table.h
+++ b/dttools/src/jx_table.h
@@ -7,6 +7,7 @@
 typedef enum {
 	JX_TABLE_MODE_PLAIN,
 	JX_TABLE_MODE_METRIC,
+	JX_TABLE_MODE_MEGABYTES,
 	JX_TABLE_MODE_URL
 } jx_table_mode_t;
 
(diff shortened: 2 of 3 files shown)


dthain:
Looks good so far.  Two refinements:

- Look at `jx_table.h`.  You can select the display mode `METRIC` for the memory and disk columns, and then they should be much more readable.

- Arrange the columns so that you have task cores, total cores, task memory, total memory, etc.  That way, it's easy for the user to compare the used/available fraction.
-

tphung3:
I'm having a weird parse error in the picture.
Deleting all changes to `work_queue_status.c` doesn't solve this issue for some reason.
The METRIC display is in KB and MB so it's a bit counter-intuitive, as the memory and disk are assumed to be measured in MB. Should I go to the jx_table.c and change them to just K and M?
image

dthain:
Don't worry about the parse error for the moment, I think that's a separate issue to be addressed.

Hmm, I agree that it is confusing to mix units in this display.

If you want to force all the columns to be MB, don't change the behavior of `JX_TABLE_MODE_METRIC`, because other bits of code depend on that behavior.  Instead, add a new option to `jx_table` like `JX_TABLE_MODE_MEGABYTES` that implements the desired behavior.

Also, it would help to right-justify all the numeric fields so that everything lines up nicely.

tphung3:
WQ resources display looks better now!
image

btovar:
Looks good!

I wonder if instead of  CORES, TASK_CORES it would be better to have:
TOTAL CORES      USED CORES  etc.

tphung3:
On the line `btovar-cms-production` the MEMORY is 249.9K but the TASKS_MEMORY IS 376.2K. So is MEMORY the workers' currently memory usage and is TASKS_MEMORY the tasks' declared memory usage? If it is so then TOTAL MEMORY and USED MEMORY and so on like what @btovar you suggested don't really fit in I think. Maybe WORKERS_MEMORY and TASKS_MEMORY and so on might be better?..

btovar:
Sorry, it should be   TOTAL CORES    NEEDED CORES,  as task_total_cores, etc.  gives the count of memory needed. So you are correct, the TASK_MEMORY is total declared memory needed.

Memory values are computed in MB, so that K is incorrect.

tphung3:
That sounds right. I will update the column headers to TOTAL and NEEDED. 
The K and M were meant to indicate thousands and millions (in megabytes), but it is still confusing I guess. Maybe I can add something like `TOTAL_MEMORY (in MB)` in the column header and make columns a bit wider?

btovar:
Yes, I think something like TOTAL MEMORY (MB) should work well!

dthain:
Wait, now I am confused.  :)

I thought total_cores was the total number of cores reported by connected workers, while total_task_cores was the total number of cores allocated by running tasks.  So, total_task_cores should never exceed total_cores.  Am I mistaken?

btovar:
total_task_cores  is the sum of all the cores the task have declared.  If you want the number that are currently used you need: `cores_inuse`. Similarly for `memory_inuse` and `disk_inuse`.

I would prefer *_inuse to be reported by this -R option, it seems more informative.

tphung3:
Final revision I hope? :)
image

dthain:
That's looking nice, except the units for memory and disk are unclear.  The header says (MB), but the values seem quite small to me.  Are those workers really only report KB of disk?

tphung3:
K and M were meant for thousands and millions, but I'll change them to something else more explicit, maybe like e3 or e6?

dthain:
That's not what I meant --  K and M have a perfectly clear meaning.  I'm questioning whether those values are correct: I would expect that workers would be reporting memory and disk available in the GB range.

btovar:
I think in this case K means x1000, so for example it is 62.7 MB x K = 62.7 GB. I think using K like that is the source of confusion.

Memory and disk are reported in MB. For the display it seems practical to report it in GB (e.g., TOTAL_MEMORY (GB)).

dttools/src/jx_table.c
```diff
@@ -109,6 +109,10 @@ void jx_table_print( struct jx_table *t, struct jx *j, FILE * f, int columns_max
 			line = malloc(10);
 			string_metric(jx_lookup_integer(j,t->name), -1, line);
 			strcat(line, "B");
+		}
+		else if(t->mode == JX_TABLE_MODE_MEGABYTES) {
+			line = malloc(10);
+			string_metric(jx_lookup_integer(j,t->name), -1, line);
```
dthain, line 115: Oh, I see the problem.
This case isn't doing anything different than the `JX_TABLE_MODE_METRIC` case.
Take a look at how string_metric works:
https://github.com/cooperative-computing-lab/cctools/blob/master/dttools/src/stringtools.c#L238

If -1, it chooses the metric prefix automatically.
If >=0 then you can choose the power of 10.

So just set this to always show megabytes.

tphung3:
I think string_metric is meant to convert values in bytes to other units, so a raw report like this might be good ( I use 0 for the magnitude power)? I double-checked and the values are consistent with `work_queue_status -R`.
image

Issue #2497, opened 2020-12-18
WQ: Fix work_queue_hungry


dthain:
`work_queue_hungry` is designed to simplify the question of whether one should produce more tasks for the queue.
It's used like this:

```
while(work_queue_hungry(q)) {
     task = create another task
     work_queue_submit(task);
}
```

The current implementation dates from way back when we assumed that one task would run on each worker.  So, it assumes that the number of idle jobs in the queue should be roughly equal to the number of workers.  It returns true if there are fewer than this number of jobs in the queue.

But this doesn't work in a multi-core world.  Tasks and workers are described by (cores,memory,disk) and multiple tasks can run on each worker. So, fix up `work_queue_hungry` so that it takes this into account.

Basic idea:
- Sum up the total (cores,memory,disk) available in the workers.
- Sum up the total (cores,memory,disk) of the waiting tasks.
- Decide which resource is the limiting factor.
- Then, if the sum of tasks is less than the sum of workers, return true.

Perhaps I haven't thought this fully through, and there is a better solution.

dthain:
@tphung3 This would be a good next step issue to work on.
It also applies to hypersweep, since using `work_queue_hungry` would allow you to pipeline jobs.

tphung3:
How can I close this issue?

dthain:
Ah, I had to adjust the permissions, can you see the "close" button now?

tphung3:
Great!

PR #2508, opened 2021-01-21
Check actual resources consumption in work_queue_hungry

tphung3:
The check for actual resources consumption can be improved to avoid the problem of tasks being submitted in a loop while resources are fragmented? For example, a more comprehensive check that matches specific available workers to appropriate tasks. What do you think?

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -6184,17 +6184,55 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
 
 int work_queue_hungry(struct work_queue *q)
 {
-	int ready = task_state_count(q, NULL, WORK_QUEUE_TASK_READY);
-	if(q->stats->tasks_dispatched < 100)
-		return MAX(100 - ready, 0);
+	//int ready = task_state_count(q, NULL, WORK_QUEUE_TASK_READY);
+	//if(q->stats->tasks_dispatched < 100)
+	//	return MAX(100 - ready, 0);
 
-	//BUG: fix this so that it actually looks at the number of cores available.
+	struct work_queue_task *t;
+	
+	//get total available resources consumption (cores, memory, disk) of all workers of this manager
+	//available = total (all) - committed (actual in use)
+	int64_t workers_total_avail_cores;
+	int64_t workers_total_avail_memory;
+	int64_t workers_total_avail_disk;
 
+	workers_total_avail_cores = q->stats->total_cores - q->stats->committed_cores;
+	workers_total_avail_memory = q->stats->total_memory - q->stats->committed_memory;
+	workers_total_avail_disk = q->stats->total_disk - q->stats->committed_disk;
+
+	//get total required resources (cores, memory, disk) of the waiting tasks
+	int64_t ready_tasks_total_cores;
+	int64_t ready_tasks_total_memory;
+	int64_t ready_tasks_total_disk;
+
+	ready_tasks_total_cores = 0;
+	ready_tasks_total_memory = 0;
+	ready_tasks_total_disk = 0;
+
+	list_first_item(q->ready_list);
+	while ((t = list_next_item(q->ready_list))){
+		ready_tasks_total_cores += t->resources_allocated->cores;
+		ready_tasks_total_memory += t->resources_allocated->memory;
+		ready_tasks_total_disk += t->resources_allocated->disk;
+	}
+	
+	//Check limiting factors (expand features later?)
+	if (ready_tasks_total_cores > workers_total_avail_cores){
+		return 0;
+	}
+ 	if (ready_tasks_total_memory > workers_total_avail_memory){
+		return 0;
+	}
+	if (ready_tasks_total_disk > workers_total_avail_disk){
+		return 0;
+	}
+
+	return 1;	//all good
 	//i = 1.1 * number of current workers
 	//i-ready = # of tasks to queue to re-reach the status quo.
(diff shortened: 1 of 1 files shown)


dthain:
Yes, this looks like the right idea.
Make sure to test this with a simple application to be sure all the corner cases are handled.
For example, what happens if there are no workers connected yet?

btovar:
One thing that worries me is the initial state of hungry and accepting new workers. Say my application does something like:

```sh
while queue not empty:
    if queue is hungry:
       submit some tasks
    wq wait
```

At the start, there are no tasks and no workers connected, so queue hungry is false. Since no task is ever submitted, wq wait returns immediately and no new workers connect. With this, hungry is always false.

dthain:
@tphung3 please push your latest changes, so we can wrap this one up.

dthain:
Ah, I think this may be too complicated -- you don't want to block inside this call, because nothing will happen while you are sleeping.

Better to just return true in any unusual case, which will encourage the caller to compose a job and call `work_queue_wait`

dthain:
Let me be more specific:

- Count up the number and resources of jobs waiting, as you are doing now.

- Return true if the sum of those resources is less than the sum of the resources of the workers.  (We want enough tasks to feed all available workers.)

- Return true if the number of waiting jobs is less than, say, 10.  (We want a minimum number of tasks available regardless of workers.)

Does that make sense?

tphung3:
Yes that makes sense. I agree with point 1 and 3 but point 2 can be improved. Assume that we have a lot of small-sized tasks that are ready and workers can only take **some** of them, but the total required resources are higher than the amount of resources workers have, so in this case we actually are not submitting anything, as `work_queue_hungry` always return false. So instead of summing up the required resources, maybe we can check the ready tasks individually?

dthain:
I don't quite follow.  Can you give an example?

tphung3:
Let's say all workers have 10 cores and 1GB of memory and disk available. We also have 100 tasks, and each requires 1 core and 100 MB of memory and disk. If we sum up all required resources of tasks, then WQ hungry will return false since all 100 tasks amount to 100 cores and 10GB of memory and disk. But if we check each task one at a time, then WQ hungry will return true as the workers can take in a small number of tasks (10 in this case).

tphung3:
How likely do you think it can actually happen?

dthain:
Ah, I see what you mean.

So, from the big picture, we have to balance several things:

- We want to have a sufficient backlog of idle jobs in the queue, so that WQ can dispatch new tasks quickly when old ones complete.  But, we don't want to materialize an unnecessary number of idle jobs, otherwise the master's time/memory will be occupied managing the queue.
- The time/complexity of work_queue should be low, because it is going to get called about once for every submission.  We don't want to re-implement the whole scheduling algorithm here, which takes into account data locality, network performance, special capabilities, etc.  If we have to simplify, then work_queue_hungrey should err on the side of true, rather than false.

You are right that simply counting up the resources is going to result in some inaccuracy, because the tasks may not pack into the workers perfectly.

In your example, the tasks don't nicely pack into the workers, because the tasks don't consume much memory.  What will happen is that WQ will assign tasks to workers until the critical resource (cores) is exhausted.  So, you will get 10 tasks per worker, which is the desired outcome.

Thus, we want work_queue_hungry to keep returning true until the critical resource (in this case, cores) is exhausted.

tphung3:
The field `q->stats->tasks_waiting` seems to fail at tracking the number of ready tasks so I used 'q->stats->submitted - q->stats->dispatched' instead.

work_queue/src/work_queue.c
```diff
@@ -6184,17 +6184,73 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
...
-	if(q->stats->tasks_dispatched < 100)
-		return MAX(100 - ready, 0);
+	//check if queue is initialized
+	//return false if not
+	if (q == NULL){
+		return 0;
+	}
 
-	//BUG: fix this so that it actually looks at the number of cores available.
+	//check if queue's statistics are initialized
```
dthain, line 6193: Ah, I see the problem here.  `q->stats` is used internally to track some stats incrementally, but doesn't track those that are expensive to compute.  This is done in `work_queue_get_stats` which takes some things from `q->stats` and then adds the computed items.

So, what you want to do is this:
```
work_queue_stats s;
work_queue_get_stats(q,&s);
```

Then everything in `s` (like `tasks_waiting` will be up to date and can be used normally.

work_queue/src/work_queue.c
```diff
@@ -6184,17 +6184,73 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
...
+		if (q->stats->tasks_submitted - q->stats->tasks_dispatched < 10){
+			return 4;
+		}
+		return 0;
+	}
+	
+	//if number of ready tasks is less than 10, return true for more tasks in queue 
+	//ready = submitted - dispatched
+	//10 is chosen to be the default number of ready tasks in queue to keep queue efficient
+	if (q->stats->tasks_submitted - q->stats->tasks_dispatched < 10){
```
dthain, line 6213: As above, use `s.tasks_waiting`

dthain (changes requested):
Almost there: see comment about `q->stats`

dthain:
Yes, looks good!
One more thing, add gpus as a fourth resource.
(I just fixed this up in a different PR.)

dthain (approved):

dthain:
@tphung3 go ahead and close #2497!

Issue #2494, opened 2020-12-15
WQ: Should the flag WORK_QUEUE_SUCCESS be added when adding files to a task?


btovar:
Similar to #2488 (WORK_QUEUE_FAILURE_ONLY), should we add a flag to retrieve a file only when the task succeeds?

dthain:
I think we talked about several cases yesterday:
`WORK_QUEUE_FAILURE_ONLY` - Only return this file if the task fails.  (already exists)
`WORK_QUEUE_SUCCESS_ONLY` - Only return this file if the task succeeds.
`WORK_QUEUE_OPTIONAL` (is there better name?) Do not consider the job a failure if this file was not created.

Is there another case we should be concerned about?

Related question: should it be possible to express these constraints in Makeflow?

PR #2509, opened 2021-01-21
WQ: Added SUCCESS_ONLY to parallel FAILURE_ONLY

dthain:


dthain:
Addresses #2494.  I did not add OPTIONAL b/c WQ already treats a missing output as a non-fatal error for a task, precisely b/c one needs to be able to see partial outputs on failure.  (Although note that Makeflow is more persnickety.)

PR #2511, opened 2021-01-27
WQ GPU Factory/Capacity Fixups

dthain:
The WQ factory and the capacity computations did not include GPUs as a resource.  This PR fixes that.


dthain:
@btovar you did a nice comparison of CI platforms, can you link that document here?

batch_job/src/work_queue_factory.c
```diff
@@ -199,6 +208,10 @@ int manager_workers_needed_by_resource(struct jx *j) {
 		needed = MAX(needed, DIV_INT_ROUND_UP(tasks_total_disk, disk));
 	}
 
+	if(gpus > 0 && tasks_total_gpus > 0) {
```
btovar, line 211: gpus should also be added to configuration file (around line 704)

btovar (changes requested):

batch_job/src/work_queue_factory.c
dthain: Indeed, thanks!

PR #2515, opened 2021-01-31
Show resources when using work_queue_status with -T and -W

tphung3:
I'm not sure how to handle the -1 GPU...
image

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -6184,17 +6184,55 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
 
 int work_queue_hungry(struct work_queue *q)
 {
-	int ready = task_state_count(q, NULL, WORK_QUEUE_TASK_READY);
-	if(q->stats->tasks_dispatched < 100)
-		return MAX(100 - ready, 0);
+	//int ready = task_state_count(q, NULL, WORK_QUEUE_TASK_READY);
+	//if(q->stats->tasks_dispatched < 100)
+	//	return MAX(100 - ready, 0);
 
-	//BUG: fix this so that it actually looks at the number of cores available.
+	struct work_queue_task *t;
+	
+	//get total available resources consumption (cores, memory, disk) of all workers of this manager
+	//available = total (all) - committed (actual in use)
+	int64_t workers_total_avail_cores;
+	int64_t workers_total_avail_memory;
+	int64_t workers_total_avail_disk;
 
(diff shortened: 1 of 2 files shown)


btovar:
The -1 means that it was not specified. In general we use -1 to mean use all available at the worker. I think we would want to show -1 as * (star) for all the resources in that table.

dthain:
@btvoar this makes me wonder if we have the defaults set right.

If the user does not specify any resources for a task, then we want it to use the whole worker, that's correct.

But, if they user specifies, say, one core, and doesn't specify anything else, do we will still want the other resources set to "consume all"?  I'm inclined to think they should be set to zero.

btovar:
I think both ways are reasonable. The current implementation is conservative, maybe overly so, but ensures that we can fulfill the promises made to a task.

Simply changing the default value to 0 breaks the current implementation, though. If we set unspecified resources to 0, there are three cases that worry me:

1) Task A only explicitly specifies to use all of some resource in a machine, (say memory). Task B does the same for some other resource, (say cores). WQ could then send these two tasks to the same worker, but then we will be breaking the promise to each of these tasks (any task uses some memory or some core).

2) If we don't specified say cores and disk, and say we specify 1MB for memory. If the worker has 1GB, then we will have 1000 tasks sent to that worker, which sounds bad.

3) When the worker runs out of resources, all tasks will be canceled and we would not know if a particular task was the culprit. With the current implementation the resource monitor can check the resources per task. Setting a default to 0, we would only know that the whole worker ran out of resources, but nothing more.

For points 1) and 2) one thing we could do is to use the whole worker when no resource is specified, but when some resource is specified then:

memory if unspecified defaults to 0.
disk if unspecified defaults to 0.
gpu if unspecified defaults to 0.
cores if unspecified defaults to 1.

This would naturally limit the number of tasks on a worker. 

For point 3), we could retry tasks using a whole worker.

dthain:
Ok, let's discuss at the meeting today and not take over this PR from Thanh.

dthain:
@tphung3 This looks good, except that it contains commits from your previous PR.  Hence the "unable to merge" because some other things have changed in the meantime.

Please submit a new PR that contains only commit ea0ba1a.  Suggest that you go back to your master branch, pull origin master to get up to date, then create a new branch for this feature.

Also, please adjust the column widths so that things fit nicely into 80-columns by default.  Otherwise the display will be quite difficult to read for someone with a default window.  (Negative values will stretch to fill if the display is wide.)

Issue #2514, opened 2021-01-29
WQ: Assign Specific GPUs


dthain:
Currently, the WQ framework allocates GPUs by just counting the number in use.

For example, if you submit three jobs: J1 requires one GPU, J2 requires one GPU, and J3 requires two GPUs, then WQ will correctly dispatch three jobs to share a four GPU machine.

However, this won't work well because it relies upon the user to set up the job command line to only use the indicated number of GPUs.  They often don't do this, and many applications will jsut use all available GPUs on the machine, resulting in trouble.

To fix this, we need the worker to try a little harder:
- Keep track of each GPU individually.
- Assign running tasks to specific GPUs.
- Inform tasks of their assignment by setting the environment variable CUDA_VISIBLE_DEVICES.

So, J1 would get CUDA_VISIBLE_DEVICES=0, J2 would get 1, and J3 would get 2,3

PR #2517, opened 2021-02-01
WQ: Assign Tasks to Specific GPUs

dthain:
The worker now keeps track of which specific GPUs each task is assigned to and then sets CUDA_VISIBLE_DEVICES to match that assignment.  That will  now allow us to run multiple concurrent GPU task such that each one stays in its lane.

There is still an oddity in how the manager assigns the total number of gpus to an execution request, but I'll bring that up in a separate issue.


dthain:
#2514

PR #2524, opened 2021-02-03
Add `specify_gpus` to WQ Python bindings

trshaffer:
Need this to add support for GPU specifications in Parsl WQ executor.

@btovar can you do a Conda release with this?


btovar:
Tim, could you add it to the perl bindings too?

trshaffer:
Looks like the perl bindings already have it?
https://github.com/cooperative-computing-lab/cctools/blob/50884b858a6803611a9ee81d80c467f0baaf5a08/work_queue/src/bindings/perl/Work_Queue/Task.pm#L247

PR #2525, opened 2021-02-04
WQ: Deny GPU access if not declared.

dthain:
Proposing this, but slightly concerned it could backfire.   If the manager tells the worker that gpus=0, then the worker will set CUDA_VISIBLE_DEVICES=-1 to prevent the task from using any GPUs and messing others up.

Note that the question what what the **user** tells the **manager** about resources is a little more complicated.  Currently, if the user tells the manager nothing about resources, then the manager will assign all resources (including all gpus) to the task, and this clause does not come into play.


dthain:
@btovar what do you think?

btovar:
I think we should only set it if it is not set in the env before.  I can see the use case where I launch a worker as:

CUDA_VISIBLE_DEVICES=... work_queue_worker ....

dthain:
Good point, I like it.

dthain:
Decided to abandon this one, so as to not break existing codes.

PR #2532, opened 2021-02-10
Simplify category code

btovar:
Focus on cores, memory, disk, and gpus only.
Turn macros into functions.
Remove resource_monitor_histogram tool in favor of plotting histograms with other tools (like python+matplotlib).


dthain:
I am supportive of a general cleanup here.  Perhaps we want to retain wall clock time as a top-level resource, since that is mentioned as a makeflow resource?  (Not sure if it is a WQ resource.)

btovar:
wall_time would still be a top level resource. This is only whether we want tasks to find automatic limits for wall_time. I don't think it makes much sense for a couple of reasons:

- The automatic labeling works under the assumption that we want to run as many tasks as possible concurrently maximizing throughput or minimizing waste. Time fields do not fit this idea very well, as time stacks differently than the other resources (e.g., if all other resources stack vertically, time stacks horizontally in a x-y plot). Automatic labeling of time still makes sense with something similar to fast_abort, but the code in category.c does not fit the use case.
- The computations assume independence of resources, i.e., we do not look at memory for computing an efficient value for cores. With wall_time this is not respected. E.g., the cores value may be computed thinking that most tasks would run for 2 hours, but the computed optimal wall_time may come as 1 hour (this value, because of the previous point is probably just noise). Since tasks are killed at 1 hour, our value for cores is not useful.

dthain:
Ok, makes sense to me.

PR #2535, opened 2021-02-15
WQ: Add simple API for fetching manager data via HTTP.

dthain:
Hitting the manager host and port with a web browser will give a simple data API,
and then allow you to grab task/worker/resources data via JSON, in the same
format as work_queue_status.


dthain:
@btovar need your sanity check on this one.

btovar:
wq_factory and status don't quite work. Both take a long time to respond to direct connections  to the master, and then show as DISCONNECTION FAILURE in the transaction log, rathen than DISCONNECTION STATUS_WORKER.  I'm debugging that right now.

dthain:
Good catch!

btovar:
I think it is ready to merge.

PR #2545, opened 2021-03-03
Proposal for default resources

btovar:
In summary:
- default not specified gpus are 0.
- In fixed mode (the default) cores, memory and disk not specified get a value proportional to the resources specified, rounded, and taking the max if more than one of cores, memory, disk or **gpus** are specified.
- Not specifying anything gets you the whole worker.
- Specifying everything to 0 uses the whole worker.

Nothing is implemented, this is an update to the wq manual with the proposed changes.

### DIFF
--- a/doc/man/md/work_queue_status.md
+++ b/doc/man/md/work_queue_status.md
@@ -129,7 +129,7 @@ The **-A** option shows a summary of the resources observed per task
 category.
 
 ```
-$ work_queue_status -T cclws16.cse.nd.edu 9001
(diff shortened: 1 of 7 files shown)


dthain:
Yes, that seems about right to me.

dthain:
Sorry, I just merged #2541 after it was sitting out there for a while, you may need to update..

dthain:
I like this description of categories, but that's a more complex use case.
Make sure to start with the simple case of specify_cores/memory/disk for a single task.

PR #2549, opened 2021-03-05
Fix task.specify_buffer

tphung3:
`task.specify_buffer` calls method `_determine_file_flags(flags, cache)` but this method requires an additional argument `failure_only`. The fix is that I simply supply `None` to this argument as `specify_buffer` doesn't have `failure_only` argument. SHADHO runs with this fix too.


btovar:
Thanks Thanh, it looks good!

PR #2559, opened 2021-03-11
add access-control header to worker response to allow js/browser access

jmrundle:


work_queue/src/work_queue.c
```diff
@@ -2663,6 +2663,7 @@ static work_queue_msg_code_t process_http_request( struct work_queue *q, struct
 	} else {
 	        // Other requests get raw JSON data.
 		send_worker_msg(q,w,"Content-type: text/plain\n\n");
+		send_worker_msg(q,w,"Access-Control-Allow-Origin: *\n\n");
```
dthain, line 2666: Careful: you only want two newlines after the final header message.  (That indicates the end of headers and the start of data.)

dthain (changes requested):

work_queue/src/work_queue.c
jmrundle: Ah good catch, should be good now

dthain (approved):

PR #2572, opened 2021-03-23
Update WQ Factory Python docs

### LINKED ISSUE #2570: WQ: Python Factory Docs
@trshaffer please add docs for the WQ Factory interface, starting from this page:
http://ccl.cse.nd.edu/software/manuals/api/html/namespaceWorkQueuePython.html

trshaffer:
Resolves #2570 

I also adjusted the interface so it can be used directly rather than via context manager to make things easier in Jupyter notebooks. I tried to note in the docs that this is not the recommended way of doing things outside of Binder.

### DIFF
--- a/work_queue/src/bindings/python3/work_queue.binding.py
+++ b/work_queue/src/bindings/python3/work_queue.binding.py
@@ -10,6 +10,7 @@
 #
 # - @ref work_queue::WorkQueue
 # - @ref work_queue::Task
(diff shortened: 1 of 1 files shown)


dthain:
LGTM

dthain:
Are we ok to merge on this one?  @wiggles35 would like to use this in the Binder example.

btovar:
Looks good to me.

PR #2573, opened 2021-03-24
wq: add API call to translate task result from integer to string

btovar:
This like a perror but for work_queue_result_task_t. For debugging and user interaction it is more helpful to see for example a 'INPUT_MISS" rather than the integer 1.


dthain:
LGTM

Issue #2542, opened 2021-02-24
Doc: Improve WQ Example Application


dthain:
The current example WQ program (gzip some files) isn't an ideal example to show to new users.  It's simple, which is good, but doesn't represent a typical use of WQ.  And it has strange ergonomics in that the user must provide some extra files to compress. 

Let's think of a better example application.  It should have the following properties:

- High level program should be simple and straightforward (generate tasks, submit, and wait)
- Should involve a modest amount of I/O so that we can clearly demonstrate how to transfer/cache files.
- Application should have modest runtime (e.g. 5-10s) so that it's easy to show some parallel speedup on a small number of tasks.
- Should be self-contained so that the user doesn't have to download/install/configure anything additional.

Suggestions?

dthain:
Some ideas:

- Task: An N-body simulation of the solar system + a comet at a particular location.  App: parameter sweep of initial conditions.    Pros: Easy to make this require an arbitrary amount of computation. Cons: Request some effort to understand/visualize the output.
- Task: Analyze one text in Project Gutenberg for word frequency, etc.  App: Run in parallel across all Pros: Easy to understand inputs/outputs.  Cons: Doesn't involve a huge amount of computation time.

btovar:
We can do a brute-force algorithm for some known number, such as pi by probability:

https://mste.illinois.edu/activity/buffon/

Each task does the same thing, and then we collect everything in a final task.

btovar:
Closing as wq is now in maintenance mode.

PR #2574, opened 2021-03-24
WQ: Brownian Motion Example

btovar:
Simulate a toy diffusion example.

Each task simulates a toy Brownian motion: Flip a coin, if 0 move left,
if 1 move right. Do that for a certain number of steps.

At the end, we process the results and compute the entropy per time
step. We should observe entropy increasing, e.g.:

Task 98 has returned.
Task 99 has returned.
Task 100 has returned.
All task have finished.
Processing results...
Computing entropy:
time-step       entropy
0               0.000
10              0.798

This example has the advantage that tasks are very simple and can be
implemented in pure bash, and that we can make a clear distinction on wq
code and app code.

If it looks ok, I can clean it up, document it properly, and copy it to C and perl.

To solve #2542


btovar:
I agree with you with regards to bash expectations. The current implementation is a compromise to ensure the tasks can run almost anywhere without worrying if python/perl/etc are available.  We only need random and modulo, so we just need a common tool that has those that is not bash.

As for the second point, you can increase the number of steps, or I can put a sleep at each iteration to artificially increase the time required. Without sleep, the bash implementation runs 32000 steps in about 13 seconds. So, enough to start to see some benefits of parallelism, but nothing impressive.

dthain:
Let's put this in the work-queue-examples repo.

PR #2588, opened 2021-04-20
jx_repl tool for JX language exploration

jmrundle:
Not yet ready for merge -- still a working prototype.  I'm open to any ideas/changes to make this a more useful tool.  The code is also a little messy, so any tips on how to improve it are also welcome.

### DIFF
--- a/dttools/src/jx_repl.c
+++ b/dttools/src/jx_repl.c
@@ -19,8 +19,6 @@ The program exits once EOF is reached or after the user enters the `quit` comman
 #include "jx_pretty_print.h"
 
 #include <stdio.h>
-#include <readline/readline.h>
-#include <readline/history.h>
 
 
 #define CATALOG_URL "http://catalog.cse.nd.edu:9097/query.json" 
@@ -135,14 +133,14 @@ int main(int argc, char *argv[]) {
 
         char in[14];
         char out[14];
-        char prompt[20];
 
         sprintf(in, "in_%d", i);
         sprintf(out, "out_%d", i);
-        sprintf(prompt, "%s  : ", in);
 
-        char *line = readline(prompt);
-        add_history(line);
+        printf("%s  : ", in);
+        char *line;
+        size_t len = 0;
+        getline(&line, &len, stdin);
 
         if (!line) {
             // EOF
--- a/dttools/src/jx_repl.c
+++ b/dttools/src/jx_repl.c
@@ -7,8 +7,8 @@ See the file COPYING for details.
 /*
 This is a program to explore the jx library.
 It reads in a JX expression, evaluates it, prints out the result, then saves it to the context.
-Thus, expressions can reference previous expressions with the `Out` array
-The program exits once EOF is reached or after `quit` command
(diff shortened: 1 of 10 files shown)


dttools/src/jx_repl.c
```diff
@@ -0,0 +1,228 @@
...
+#include <stdio.h>
+#include <string.h>
+
+#ifdef HAS_LIBREADLINE
+#include "readline/readline.h"
+#include "readline/history.h"
+#endif
+
+#define CATALOG_URL "http://catalog.cse.nd.edu:9097/query.json" 
+#define MAX_LINE 200
```
dthain, line 30: Make this at least 4KB, JX expressions can get complex quickly.

dttools/src/jx_repl.c
```diff
@@ -0,0 +1,228 @@
...
+    "  basename( path )\n"
+    "  dirname( path )\n"
+    "  escape( string )\n"
+    "  len( array )\n"
+    "  fetch( URL/path )\n"
+    "  select( boolean, array )\n"
+    "  project( expression, array )\n"
+    "  schema( object )\n"
+    "  like( string, object )\n"
+    "\n";
```
dthain, line 67: This is a nice help text, but we should find a way to keep it synchronized with the list of available functions in jx_exal / jx_function.  Maybe we just need some comments in both places to remember?

dttools/src/jx_sub.c
```diff
@@ -0,0 +1,235 @@
...
+static struct jx * jx_sub_call( struct jx *func, struct jx *args, struct jx *ctx) {
+    assert(func);
+    assert(args);
+    assert(jx_istype(args, JX_ARRAY));
+    assert(jx_istype(func, JX_SYMBOL));
+
+    struct jx *left  = jx_copy(func);
+    struct jx *right = NULL;
+
+    if (!strcmp(func->u.symbol_name, "range")    ||
```
dthain, line 22: I think we need to factor this out into a function called something like `jx_is_function_name` that is shared with the corresponding code in `jx_eval` and `jx_function` so that we don't get trapped by changes to the set of functions.  (Which can and do happen)

dttools/src/jx_sub.c
```diff
@@ -0,0 +1,235 @@
...
+        !strcmp(func->u.symbol_name, "fetch")    ||
+        !strcmp(func->u.symbol_name, "schema")   ||
+        !strcmp(func->u.symbol_name, "like")     ||
+        !strcmp(func->u.symbol_name, "keys")     ||
+        !strcmp(func->u.symbol_name, "values")   ||
+        !strcmp(func->u.symbol_name, "items")) {
+        
+        right = jx_sub(args, ctx);
+
+    } else if (!strcmp(func->u.symbol_name, "select") ||
```
dthain, line 42: Right, there are two categories of functions: eager and lazy evaluation.  I guess we need **two** functions to match names...

dthain (changes requested):
I like it, this is looking good and already a useful tool.  Please add a descriptive page to the user manual, and you are all set.

dthain (approved):

PR #2593, opened 2021-05-05
WQ: Add Minimum Time as a Resource

tphung3:
**Modifications:**
Line 3722 to 3726 in `work_queue.c` checks whether a worker has enough time for a task.
Line 22 in `work_queue_resources.h` adds time to the `work_queue_resources` structure.
Line 383 to 390 in `work_queue_worker.c`updates the manager on the time a worker has left.

**Diagnosis:**
When code is compiled and run with the example WQ program in the [docs](https://cctools.readthedocs.io/en/latest/work_queue/) (this program needs to add a call to `work_queue_task_specify_running_time` for the issue to appear), it seems that `work_queue_worker` is stuck in a loop starting at line 2962 in `work_queue_worker.c`. The main reason for this is that line 2979 in `work_queue_worker.c` returns the variable `result ` with a value of 0. A deeper dive into function `serve_manager_by_hostport_list` doesn't give much information though.

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -244,7 +244,7 @@ struct work_queue_worker {
 	timestamp_t start_time;
 	timestamp_t last_msg_recv_time;
 	timestamp_t last_update_msg_time;
-	long long int end_time;
+	int64_t end_time;
 };
 
 struct work_queue_task_report {
@@ -636,7 +636,7 @@ work_queue_msg_code_t process_info(struct work_queue *q, struct work_queue_worke
 		w->workerid = xxstrdup(value);
 		write_transaction_worker(q, w, 0, 0);
 	} else if(string_prefix_is(field, "worker-end-time")) {
-		w->end_time = atoll(value);	
+		w->end_time = MAX(0, atoll(value));
 	}
 
 	//Note we always mark info messages as processed, as they are optional.
@@ -1035,6 +1035,7 @@ static void add_worker(struct work_queue *q)
 	w->current_tasks_boxes = itable_create(0);
 	w->finished_tasks = 0;
 	w->start_time = timestamp_get();
+	w->end_time = -1;
 
 	w->last_update_msg_time = w->start_time;
 
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -3719,6 +3719,11 @@ static int check_hand_against_task(struct work_queue *q, struct work_queue_worke
 	if(w->resources->gpus.inuse + limits->gpus > overcommitted_resource_total(q, w->resources->gpus.total, 0)) {
 		ok = 0;
 	}
+		
+	//if the wall time for the worker is specified and there's not enough time for the task, then not ok
+	if(w->resources->time_left != -1 && w->resources->time_left < t->resources_requested->wall_time){
+		ok = 0;
+	}
 
 	rmsummary_delete(limits);
 
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -4591,20 +4591,20 @@ void work_queue_task_specify_running_time( struct work_queue_task *t, int64_t us
 	}
 }
 
-void work_queue_task_specify_running_time_max( struct work_queue_task *t, int64_t useconds )
+void work_queue_task_specify_running_time_max( struct work_queue_task *t, int64_t seconds )
 {
-	work_queue_task_specify_running_time(t, useconds);
+	work_queue_task_specify_running_time(t, 1000000*seconds); 	//convert to useconds for backward compat
 }
 
-void work_queue_task_specify_running_time_min( struct work_queue_task *t, int64_t useconds )
+void work_queue_task_specify_running_time_min( struct work_queue_task *t, int64_t seconds )
 {
-	if(useconds < 1)
+	if(seconds < 1)
 	{
 		t->min_running_time = -1;
 	}
 	else
 	{
-		t->min_running_time = useconds;
+		t->min_running_time = 1000000*seconds;					//convert to useconds for backward compat
 	}	
 }
 
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -244,6 +244,7 @@ struct work_queue_worker {
 	timestamp_t start_time;
 	timestamp_t last_msg_recv_time;
 	timestamp_t last_update_msg_time;
+	long long int end_time;
 };
 
 struct work_queue_task_report {
@@ -634,6 +635,8 @@ work_queue_msg_code_t process_info(struct work_queue *q, struct work_queue_worke
 		free(w->workerid);
 		w->workerid = xxstrdup(value);
 		write_transaction_worker(q, w, 0, 0);
+	} else if(string_prefix_is(field, "worker-end-time")) {
+		w->end_time = atoll(value);	
 	}
 
 	//Note we always mark info messages as processed, as they are optional.
@@ -3721,7 +3724,17 @@ static int check_hand_against_task(struct work_queue *q, struct work_queue_worke
 	}
 		
 	//if the wall time for the worker is specified and there's not enough time for the task, then not ok
-	if(w->resources->time_left != -1 && w->resources->time_left < t->resources_requested->wall_time){
+	if(w->end_time){
+		if(t->resources_requested->wall_time > 0){
+			if(w->end_time - time(0) > t->resources_requested->wall_time) {
+				ok = 1;
+			}
+			else{
+				ok = 0;
+			}
+		}
+	} 
+	else if(w->end_time < t->resources_requested->end){
 		ok = 0;
 	}
 
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -3731,10 +3731,10 @@ static int check_hand_against_task(struct work_queue *q, struct work_queue_worke
 
 	//if wall time for worker is specified and there's not enough time for task, then not ok
 	if(w->end_time > 0){
-		if(t->resources_requested->wall_time > 0 && w->end_time < timestamp_get() + t->resources_requested->wall_time){
+		if(t->resources_requested->wall_time > 0 && ((uint64_t) w->end_time) < timestamp_get() + t->resources_requested->wall_time){
 			ok = 0;
 		}
-		if(t->resources_requested->end > 0 && w->end_time < t->resources_requested->end) {
+		if(t->resources_requested->end > 0 && ((uint64_t) w->end_time) < t->resources_requested->end) {
 			ok = 0;
 		}
 	}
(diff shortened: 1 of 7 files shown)


dthain:
Well, this looks like the right direction.
And the fact that it passes all the automated tests means that the basic application works, as-is.

dthain:
Oh, I see.  At what point does the worker send "time remaining" back to the manager?

work_queue/src/work_queue.c
```diff
@@ -3719,6 +3719,11 @@ static int check_hand_against_task(struct work_queue *q, struct work_queue_worke
 	if(w->resources->gpus.inuse + limits->gpus > overcommitted_resource_total(q, w->resources->gpus.total, 0)) {
 		ok = 0;
 	}
+		
+	//if the wall time for the worker is specified and there's not enough time for the task, then not ok
+	if(w->resources->time_left != -1 && w->resources->time_left < t->resources_requested->wall_time){
```
btovar, line 3724: Here we should also check for t->resources_requested->end (if > 0).  end is an absolute time, so something like:
t->resources->time_left > t->resources_requested->end - timestamp(). or something like that.

Also, not all tasks define -> wall_time, so you have to check for wall_time > 0
I would also change time_left != -1 to > 0 for consistency.

work_queue/src/work_queue_resources.h
```diff
@@ -19,6 +19,7 @@ struct work_queue_resource {
 
 struct work_queue_resources {
 	int64_t tag;                       // Identifies the resource snapshot.
+	time_t time_left;			   	   // Marks the time left for work queue workers
```
btovar, line 22: I think it would be easier to have a end_time in the work_queue_worker.c, rather than a resource field here. time_left behaves differently than the other resources anyway.
With this, in wq worker we compute end_time as time() + time specified from the command line. (in seconds?)

work_queue/src/work_queue_worker.c
```diff
@@ -380,6 +380,14 @@ static void send_resource_update(struct link *manager)
 		total_resources->disk.total    = MAX(0, local_resources->disk.total);
 		total_resources->disk.largest  = MAX(0, local_resources->disk.largest);
 		total_resources->disk.smallest = MAX(0, local_resources->disk.smallest);
+		
+		//if workers are set to expire at some time, send the amount of time left to manager
+		if(manual_wall_time_option != 0) {
+			total_resources->time_left = worker_start_time + manual_wall_time_option - time(0);
```
btovar, line 386: If following above, this would be:
end_time = time(0) + manual_wall_time_option, which is simpler.

Also, make the check manual_wall_time_option > 0, otherwise negative times would terminate the worker right away.

btovar (changes requested):

btovar:
In wq worker, in the function report_worker_ready you want something like:

    send_manager_message(manager, "info worker-end-time %d\n", end_time);

which should be processed accordingly in work_queue.c process_info(...)
The struct work_queue_worker in work_queue.c also needs an end_time field.

work_queue/src/work_queue_worker.c
tphung3: `end_time` is always updated since `time(0)` is always updated I think. Maybe `end_time = worker_start_time + manual_wall_time_option` is better?

tphung3:
Here's the initial update based on above suggestions (Thanks!). I haven't run tests though, just want to put the updates here first. Also in line 3727 in work_queue.c it should be `if(w->end_time > 0){`, I forgot about that.

work_queue/src/work_queue.c
```diff
@@ -3719,6 +3722,21 @@ static int check_hand_against_task(struct work_queue *q, struct work_queue_worke
 	if(w->resources->gpus.inuse + limits->gpus > overcommitted_resource_total(q, w->resources->gpus.total, 0)) {
 		ok = 0;
 	}
+		
+	//if the wall time for the worker is specified and there's not enough time for the task, then not ok
+	if(w->end_time){
```
btovar, line 3727: if(w->end_time > 0) ...

work_queue/src/work_queue.c
```diff
@@ -3719,6 +3722,21 @@ static int check_hand_against_task(struct work_queue *q, struct work_queue_worke
...
+		if(t->resources_requested->wall_time > 0){
+			if(w->end_time - time(0) > t->resources_requested->wall_time) {
```
btovar, line 3729: we can simplify with:
if(w->end_time < time(0) + t->resources_requested->wall_time) {
ok = 0;
}

btovar (changes requested):

tphung3:
**Potential bug:**
In function [enforce_processes_max_running_time](https://github.com/tphung3/cctools/blob/1917fc4899df96f50d83b8e8aa2d11670ed3c35b/work_queue/src/work_queue_worker.c#L1580) in `work_queue_worker.c`,
on line [1592](https://github.com/tphung3/cctools/blob/1917fc4899df96f50d83b8e8aa2d11670ed3c35b/work_queue/src/work_queue_worker.c#L1592),
shouldn't the comparator sign be `>` instead of `<` since `now` must be higher than the other side to trigger a stop?

btovar:
> **Potential bug:**
> In function [enforce_processes_max_running_time](https://github.com/tphung3/cctools/blob/1917fc4899df96f50d83b8e8aa2d11670ed3c35b/work_queue/src/work_queue_worker.c#L1580) in `work_queue_worker.c`,
> on line [1592](https://github.com/tphung3/cctools/blob/1917fc4899df96f50d83b8e8aa2d11670ed3c35b/work_queue/src/work_queue_worker.c#L1592),
> shouldn't the comparator sign be `>` instead of `<` since `now` must be higher than the other side to trigger a stop?

It looks that way. Good catch.

work_queue/src/work_queue.c
```diff
@@ -3719,6 +3722,18 @@ static int check_hand_against_task(struct work_queue *q, struct work_queue_worke
...
+		
+	//if wall time for worker is specified and there's not enough time for task, then not ok
+	if(w->end_time > 0){
+		if(t->resources_requested->wall_time > 0){
+			if(w->end_time < time(0) + t->resources_requested->wall_time) {
+				ok = 0;
+			}
+		}
+	} 
+	else if(w->end_time < t->resources_requested->end){
```
btovar, line 3734: This one won't work if both are -1. I think you want something like:
```C
if(w->end_time > 0) {
   if(t->resources->requested->wall_time > 0  && w->end_time < time(0) + t->resources_requested->wall_time) {
      ok = 0;
  }
   if(t->resources->requested->end_time > 0  && w->end_time < t->resources_requested->endl_time) {
    ok = 0;
 }
}
```

work_queue/src/work_queue_worker.c
```diff
@@ -113,6 +113,9 @@ static int init_backoff_interval = 1;
 // Maximum value for backoff interval (in seconds) when worker fails to connect to a manager.
 static int max_backoff_interval = 60;
 
+// Absolute end time for worker, worker is killed after this point.
+static time_t end_time = 0;
```
btovar, line 117: Let's set this one to -1, as we usually use -1 to mean unset.

work_queue/src/work_queue_worker.c
```diff
@@ -380,6 +383,14 @@ static void send_resource_update(struct link *manager)
 		total_resources->disk.total    = MAX(0, local_resources->disk.total);
 		total_resources->disk.largest  = MAX(0, local_resources->disk.largest);
 		total_resources->disk.smallest = MAX(0, local_resources->disk.smallest);
+		
+		//if workers are set to expire in some time, send the expiration time to manager
+		if(manual_wall_time_option > 0) {
+			end_time = worker_start_time + manual_wall_time_option;
+		}
+		else {
```
btovar, line 391: No need for the else case here. (This eliminates confusion between the value set here and the default in line 117).

btovar (changes requested):

tphung3:
**Issue:** The behavior of [check_hand_against_task](https://github.com/tphung3/cctools/blob/4b36c88dcdfdec562f79a96622a50cd417728055/work_queue/src/work_queue.c#L3685) in [work_queue.c](https://github.com/tphung3/cctools/blob/1917fc4899df96f50d83b8e8aa2d11670ed3c35b/work_queue/src/work_queue.c) is to let tasks transferred to workers after checking several conditions. However, if workers aren't able to send messages about their infos in time, tasks will be transferred as all conditions will fail due to the lack of workers' information. For example, all conditions like `w->resources->*` or `w->end_time` will be evaluated to 0 since those fields of object `w` aren't initialized as such information is yet to be sent from workers. Thus, with this behavior, we can't stop a task being transferred to workers even if workers don't have enough time for that task.

**Potential solution:** Instead of setting `ok` to 1 as default in line [3708](https://github.com/tphung3/cctools/blob/4b36c88dcdfdec562f79a96622a50cd417728055/work_queue/src/work_queue.c#L3708), we can set `ok` to 0 then conditions following this will flip `ok` to 1.

Comments?

dthain:
What is the desired outcome if the task declares a time limit, but the worker does not?

dthain:
So the worker would kill a task for running over a time limit, even if the worker has no time limit?

btovar:
Yes, what @tphung3 said, only that the worker does not disconnect from the manager. If the task goes above the wall time, it is terminated with result `WORK_QUEUE_RESULT_TASK_MAX_RUN_TIME`. This is previous behavior not modified by the current pr.

btovar:
@tphung3: Thanh, just added some small commits to your pr. Mainly I moved everything to usecs and added some tests for python.

dthain:
Following yesterday's discussion with @btovar, we need to break out two different use cases in the API:

`work_queue_task_specify_time_min` should give the minimum time needed to execute this task.  This will prevent the task from being assigned to a worker that has run out of allocated time.

`work_queue_task_specify_time_max` should give the maximum allowed execution time for this task.  If it runs over, then the worker should kill the task.

Then, for backwards compatibility, `work_queue_task_specify_time` should be an alias for `time_max`

tphung3:
Everything should work now. I still need to write wrappers for python API though (t.specify_running_time_min and max are not yet available.)

dthain:
Yes, that is the right idea.  Python bindings are next , and also some mention in the WQ manual under the resources section.  Then, let's get it out to Parsl+DES.

tphung3:
Bindings are done. Where can I find the code that generates WQ manual?

dthain:
https://github.com/cooperative-computing-lab/cctools/tree/master/doc/manuals

dthain:
This is looking good overall, except one thing:
The end user has to give the time in microseconds, which seems unnecessary.
How about we let them do it in seconds?

btovar:
That sounds good. We can keep the previous specify_running_time in usecs for backwards compatibility.

dthain:
@btovar if this is looking good to you, please expedite a minor release, and then let's get it in the hands of benc for use with DES/LSST.

PR #2605, opened 2021-06-09
WQ: API to Summarize Workers

David-Simonetti-ND:
This adds in functionality for creating a wsummary struct and then displaying the information in the command line (sort able by resource type)


dthain:
@David-Simonetti-ND looks like there are some build failures above -- if you click through, you can see exactly what the problem is in the logs.

dthain:
Share an example output?

David-Simonetti-ND:
image
Here's what I have so far.
I think there are a few edge cases where a worker or two gets duplicated, but I will iron out those bugs tomorrow

David-Simonetti-ND:
image
I think we need to rethink the bucket sizing for memory. I calculated the memory lost by the logarithmic rounding scale and memory has been hovering around 50-60% lost, with disk around 8-12% lost.

David-Simonetti-ND:
image
Adding 8 subcategories made a major improvement. Memory less is less than 10% now and disk loss is under 5%.

dthain:
Looks better!

work_queue/src/work_queue.c
```diff
@@ -7413,4 +7412,295 @@ int work_queue_specify_min_taskid(struct work_queue *q, int minid) {
 	return q->next_taskid;
 }
 
+//the functions below are used by qsort in order to sort the work_queue_wsummary data
+int comparecount(const void *a, const void *b)
```
dthain, line 7416: Make this a `static` function, so that it is private to the module.

work_queue/src/work_queue.c
```diff
@@ -7413,4 +7412,295 @@ int work_queue_specify_min_taskid(struct work_queue *q, int minid) {
...
+{
+	return ( ((struct work_queue_bucket *)b)->count - ((struct work_queue_bucket *)a)->count );
+}
+int comparecores(const void *a, const void *b)
+{
+	return ( ((((struct work_queue_bucket *)b)->cores - ((struct work_queue_bucket *)a)->cores) == 0) ?
```
dthain, line 7422: Stylistic simplification:
Let `struct work_queue_bucket *x = a, *y=b` then you can use `x` and `y` instead of all the casts below.

work_queue/src/work_queue.c
```diff
@@ -7413,4 +7412,295 @@ int work_queue_specify_min_taskid(struct work_queue *q, int minid) {
...
+{
+	return ( ((struct work_queue_bucket *)b)->count - ((struct work_queue_bucket *)a)->count );
+}
+int comparecores(const void *a, const void *b)
+{
+	return ( ((((struct work_queue_bucket *)b)->cores - ((struct work_queue_bucket *)a)->cores) == 0) ? 
+			 ((struct work_queue_bucket *)b)->count - ((struct work_queue_bucket *)a)->count : 
+			 ((struct work_queue_bucket *)b)->cores - ((struct work_queue_bucket *)a)->cores);
+}
+int comparememory(const void *a, const void *b)
```
dthain, line 7426: Same for the following three functions.

work_queue/src/work_queue.c
```diff
@@ -7413,4 +7412,295 @@ int work_queue_specify_min_taskid(struct work_queue *q, int minid) {
...
+}
+int comparegpus(const void *a, const void *b)
+{
+	return ( ((((struct work_queue_bucket *)b)->gpus - ((struct work_queue_bucket *)a)->gpus) == 0) ? 
+			 ((struct work_queue_bucket *)b)->count - ((struct work_queue_bucket *)a)->count : 
+			 ((struct work_queue_bucket *)b)->gpus - ((struct work_queue_bucket *)a)->gpus);
+}
+
+// function used by other functions
+void sort_work_queue_worker_summary(struct work_queue_wsummary *data, wsummary_sort_category sortby)
```
dthain, line 7446: static

work_queue/src/work_queue.c
```diff
@@ -7413,4 +7412,295 @@ int work_queue_specify_min_taskid(struct work_queue *q, int minid) {
...
+	{
+		worker_data->count[i] = temp.count[i];
+		worker_data->cores[i] = temp.cores[i];
+		worker_data->memory[i] = temp.memory[i];
+		worker_data->disk[i] = temp.disk[i];
+		worker_data->gpus[i] = temp.gpus[i];
+		total_worker_count += temp.count[i];
+	}
+	worker_data->length = temp.length;
+	printf("Total memory losses: %.0lfmb, Average memory loss per worker: %.2lfmb\nTotal memory available: %.0lfmb, %% of memory lost %.2lf\n",
```
dthain, line 7625: `printf` is fine while troubleshooting, but we don't want that in the final version of the library.  Could replace it with a `debug` if you want to keep it.

work_queue/src/work_queue.c
```diff
@@ -7413,4 +7412,295 @@ int work_queue_specify_min_taskid(struct work_queue *q, int minid) {
...
+				(worker_data->count[i] == 1) ? "is" : "are",
+				worker_data->count[i], (worker_data->count[i] == 1) ? "worker" : "workers",
+				worker_data->cores[i], (worker_data->cores[i] == 1) ? "core" : "cores",
+				worker_data->memory[i], worker_data->disk[i],
+				worker_data->gpus[i], (worker_data->gpus[i] == 1) ? "gpu" : "gpus");
+	}
+}
+
+// creates the buckets of workers and then sorts workers into those buckets
+void add_worker_to_wsummary(struct work_queue_wsummary *worker_data, struct work_queue_worker *w)
```
dthain, line 7656: static

work_queue/src/work_queue.c
```diff
@@ -7413,4 +7412,295 @@ int work_queue_specify_min_taskid(struct work_queue *q, int minid) {
...
+	printf("Total memory losses: %.0lfmb, Average memory loss per worker: %.2lfmb\nTotal memory available: %.0lfmb, %% of memory lost %.2lf\n",
+	memory_losses, memory_losses / total_worker_count, total_memory, 100 * (memory_losses) / total_memory);
+	
+	printf("Total disk losses: %.0lfmb, Average disk loss per worker: %.2lfmb\nTotal disk space available: %.0lfmb, %% of disk space lost %.2lf\n\n", 
+	disk_losses, disk_losses / total_worker_count, total_disk, 100 * memory_losses / total_disk);
+}
+
+// sorts and displays data summary of worker resources
+// call this function with "gpu" "count" or "disk" in the sortby argument to sort the output by those aspects
+void display_work_queue_worker_summary(struct work_queue_wsummary *worker_data, char *sortby)
```
dthain, line 7634: This function doesn't need to be in the library, since the user will have their own method for displaying/using the data.
It would be a good thing to have in a sample application and/or test case, though.

work_queue/src/work_queue.h
```diff
@@ -346,6 +346,38 @@ struct work_queue_stats {
 	int workers_blacklisted;         /**< @deprecated Use workers_blocked instead. */
 };
 
+/** Statistics describing the workers in a work queue. Workers are grouped by index. For example, count[0] contains the number of workers of type 0, cores[0] contains the number of cores each of those workers has, memory[0] displays the memory of those workers, etc**/
+
+struct work_queue_wsummary {
```
dthain, line 351: Hmm, the fixed array sizes are going to be a problem at some point.

I would suggest something more like this.  Let the structure describe one bucket:

```
struct work_queue_wsummary {  int count; int cores; int memory; ... };
```

Then, let the user indicate the length of array they are willing to deal with:

```
work_queue_summary( struct work_queue *q, struct work_queue_wsummary [], int length ); 
```

So then it would get invoked like this:

```
struct work_queue_wsummary summary[64];
work_queue_summary(queue,summary,64);
```

And that way, the library will work with any arbitrary length.

dthain (changes requested):

work_queue/src/work_queue.c
```diff
@@ -7416,4 +7415,259 @@ int work_queue_specify_min_taskid(struct work_queue *q, int minid) {
 	return q->next_taskid;
 }
 
+//the functions below are used by qsort in order to sort the work_queue_wsummary data
+static int comparecount(const void *a, const void *b)
```
dthain, line 7419: Looks much better.

dthain, line 7419: Looks much better!

work_queue/src/work_queue.h
```diff
@@ -346,6 +346,28 @@ struct work_queue_stats {
...
+    int count;    /**< Number of workers of a certain type */
+    int cores;   /**< Number of cores of the workers */
+    int memory;  /**< Amount of memory of the workers */
+    int disk;    /**< Disk available to the workers */
+    int gpus;    /**< GPUs available to the workers */
+};
+
+typedef enum {
```
dthain, line 359: Since this is only used internally, it can be moved to `work_queue.c` just above the relevant function.

work_queue/src/work_queue.h
```diff
@@ -346,6 +346,28 @@ struct work_queue_stats {
...
+	NONE = 0,
+	COUNT = 1,
+	CORES,
+	MEMORY,
+	DISK,
+	GPUS
+} wsummary_sort_category;
+
+static const int MAX_POWER_OF_TWO = 25; // used for snapping memory and disk values to logarithmic scale
```
dthain, line 368: Likewise.

work_queue/src/work_queue.c
```diff
@@ -7416,4 +7415,259 @@ int work_queue_specify_min_taskid(struct work_queue *q, int minid) {
...
+		data[i].count = list_of_buckets[i].count;
+		data[i].cores = list_of_buckets[i].cores;
+		data[i].memory = list_of_buckets[i].memory;	
+		data[i].gpus = list_of_buckets[i].gpus;
+		data[i].disk = list_of_buckets[i].disk;	
+	}
+	free(list_of_buckets);
+}
+
+void convert_wsummary_to_log_scale(struct work_queue_wsummary worker_data[], int *current_length)
```
dthain, line 7492: Since this is a public part of the interface, call it `work_queue_wsummary_compact`

dthain (changes requested):
Looking good!  Last few things noted below.

Also: Remove jx_repl and work_queue_example binaries from the PR.

(Perhaps they need to be added to .gitignore.)

PR #2606, opened 2021-06-09
Custom class ads for condor_submit_workers

btovar:
Adds --class-ad to specify custom class ads when submitting workers to condor.

NDCMS use case:
```sh
condor_submit_workers ... --class-ad "+IsExpressJob" ...
```


dthain:
Do we also need a corresponding feature for the `work_queue_factory`?

(While I don't want to remove the helper scripts, I do want to encourage the factory as the best supported way of submitting workers consistently.)

btovar:
We already have it in work queue factory (--batch-options does the same thing for condor)

The use case here did not fit the factory well, as these express jobs look more like one-shots. (E.g., there is only a single node that cares about the ad, and the jobs only run for a few minutes.)

dthain:
As long as there is a way to do it, that's fine.

PR #2610, opened 2021-06-14
WQ: Add API for Python Function Evaluation

BarrySlyDelgado:


work_queue/src/bindings/python3/PyTask/PyTask.py
```diff
@@ -0,0 +1,107 @@
+import work_queue as wq
```
dthain, line 1: Please add the CCL copyright boilerplate.

work_queue/src/bindings/python3/PyTask/PyTask.py
```diff
@@ -0,0 +1,107 @@
...
+		return self._task
+	
+	def python_function_command(self, func, *args):
+			
+		func_file = os.path.join(self._tmpdir, 'function_{}.p'.format(self._id))
+		args_file = os.path.join(self._tmpdir, 'args_{}.p'.format(self._id))
+		out_file = os.path.join(self._tmpdir, 'out_{}.p'.format(self._id))
+		package_file = os.path.join(self._tmpdir, 'dep.yml')
+		tar_file = os.path.join(self._tmpdir, 'venv.tar.gz')
+		pp_run = '/afs/crc.nd.edu/user/b/bslydelg/cctools/python_environments/src/python_package_run'
```
dthain, line 31: This hardcoded path won't work for anyone except you.  :)
How about `shutil.which("python_package_run")` or similar.

work_queue/src/bindings/python3/PyTask/PyTask.py
```diff
@@ -0,0 +1,107 @@
...
+		else:
+			print('no reusly for task {}, error code {} {}'.format(p_task._task.tag, p_task._task.result, P_task._task.result_str))
+			func_result = NoResult()
+		
+		return func_result
+
+def divide(dividend, divisor):
+	return dividend/divisor
+
+def main():
```
dthain, line 91: This is a nice example code, but shouldn't be part of the library itself.
Move to a test case in the WQ directory, so that it gets run automatically by `make test`

work_queue/src/bindings/python3/PyTask/PyTask.py
```diff
@@ -0,0 +1,107 @@
+import work_queue as wq
+import dill
+import os
+import sys
+import tempfile
+import uuid
+import traceback
+
+
+class PyTask(object):
```
dthain, line 10: Should be in the work_queue namespace, I think.

work_queue/src/bindings/python3/PyTask/README.md
```diff
@@ -0,0 +1,32 @@
+# PyTask
```
dthain, line 1: This is a good introductory text, but please integrate it into the WQ manual (see docs/manuals/workqueue) so that others will be able to find it easily.

work_queue/src/bindings/python3/PyTask/py_wrapper.py
```diff
@@ -0,0 +1,24 @@
+if __name__=='__main__':
```
dthain, line 1: Ok, this is the right file, but you don't want it to live in the source tree.  What the PyTask module should do is create a file with these contents in the current working directory, at runtime.

work_queue/src/bindings/python3/PyTask/PyTask.py
dthain, line 10: Also, I think you want the PyTask object to extend the Task object, so that it supports all the usual functions to set resources, add additional files, etc...

dthain (changes requested):
Good, you are on the right track.  Some additional refinements below...

dthain (changes requested):
Good, almost there.

I missed this this first time around:   Just move the PyTask (PythonTask) code to `work_queue/src/bindings/python3/work_queue.bindings.py` and make it part of the `work_queue` namespace.  So, then you get it for free like this:

```
import work_queue as wq
pt = wq.PythonTask(...)
```
Also, notice that the other code is using pydoc to generate developer documentation.  You will want to do the same thing...

btovar:
That one is on me. I recommended to @BarrySlyDelgado to not incorporate it to the work_queue bindings until we had it running as we wanted, as it is much easier to test outside the bindings.

work_queue/src/bindings/python3/PyTask/PyTask.py
```diff
@@ -0,0 +1,125 @@
...
+			
+		
+		with open(self._func_file, 'wb') as wf:
+			dill.dump(func, wf)
+		
+		with open(self._args_file, 'wb') as wf:
+			dill.dump([*args], wf)
+		
+		#command = "python {wrapper} {function} {args} {out}".format(
+		command = "./{pprun} -e {tar} python {wrapper} {function} {args} {out}".format(
```
btovar, line 69: Here lets do a:
```python
'./{pprun} -e {tar} --unpack-to "$WORK_QUEUE_SANDBOX"/{unpack}-env  python {wrapper} {function} {args} {out}".format(
...
unpack=os.path.basename(self._tar_file)
...
```

work_queue/src/bindings/python3/PyTask/PyTask.py
```diff
@@ -0,0 +1,125 @@
...
+	def python_result(task):		
+		if task.result == wq.WORK_QUEUE_RESULT_SUCCESS:
+			try:
+				with open(os.path.join(task._tmpdir, 'out_{}.p'.format(task._id)), 'rb') as f:
+					func_result = dill.load(f)
+			except Exception as e:
+				func_result = e
+		else:
+			print('no reuslt for task {}, error code {} {}'.format(task.tag, task.result, task.result_str))
+			func_result = 'no result'
```
btovar, line 122: For this one, let's create a class:
```python
PyTaskNoResult(Exception):
   pass
```

and return that instead. We can even add task.result and result_str as fields of the class.

work_queue/src/bindings/python3/PyTask/PyTask.py
```diff
@@ -0,0 +1,125 @@
...
+					exec_out = e
+
+				with open(out, 'wb') as f:
+					dill.dump(exec_out, f)
+
+				print(exec_out)'''))
+			 
+				
+	@staticmethod
+	def python_result(task):
```
btovar, line 113: I think we want  python_result as a method of PyTask (or the rename PythonTask), to make it more pythonic:

```python
t = q.wait(5)
...
t.python_result()
```

btovar (changes requested):

dthain:
> That one is on me. I recommended to @BarrySlyDelgado to not incorporate it to the work_queue bindings until we had it running as we wanted, as it is much easier to test outside the bindings.

My mistake then, carry on with Ben's instruction.

work_queue/src/bindings/python3/PyTask/PyTask.py
BarrySlyDelgado: what does this do?

btovar: Instead of unpacking the environment for each task, it unpacks it once to a known location so that further tasks can reuse it. It saves about 20s per task (Other than the very first task executing in a worker.)

PR #2613, opened 2021-06-21
Pytask to PythonTask bindings

btovar:
@BarrySlyDelgado could you confirm that this still works for you?

I made some minor changes, but I think you can follow the example at:

work_queue/src/bindings/python3/PythonTask_example.py


BarrySlyDelgado:
@btovar Is there something I have to do so that `import work_queue` includes the PythonTask  class?

BarrySlyDelgado:
Nevermind figured it out

work_queue/src/bindings/python3/work_queue.binding.py
```diff
@@ -821,6 +839,133 @@ def resources_allocated(self):
...
+        self.specify_output_file(self._out_file, cache=False)
+
+
+    ##
+    # creates the wrapper script which will execute the function. pickles output.
+    def _create_wrapper(self):
+        with open(self._wrapper, 'w') as f:
+            f.write(textwrap.dedent('''\
+                    import sys
+                import dill
```
BarrySlyDelgado, line 946: This indent of `import sys` caused an error I believe. After unindented it, it ran and completed successfully.

BarrySlyDelgado (changes requested):

PR #2623, opened 2021-07-08
use debug D_NOTICE to print wq password warning.

btovar:
Forcing it with fprintf to stderr may cause noise in applications that
do not expect output in stderr. (E.g., progress bars in coffea.)


dthain:
Well, I think this message is meant to be disruptive, since it's alerting the user to a high level problem...

btovar:
Got it. Let me see if I can make the progress bars to print after those warnings.

PR #2626, opened 2021-07-20
PONCHO Tools for Python Packaging

BarrySlyDelgado:


dthain:
Ok, I'm going to try it out and get back to you with some comments..

dthain:
Ok, I was able to successfully make it work.  Nice!

A couple of little usability things to fix:

1 - If I simply run `poncho_package_analyze test.py` then it just sits there with no output.  It took me a while to realize that it was blocking reading from stdin.  I think you want to detect that and emit some kind of error.

2 - poncho_package_create ends with the message "To activate this environment... / To deactivate this environment" which is a little confusing b/c it only created that temporarily.  Suggest that you add some output between and after the phases to explain that poncho_package_create is processing the conda package, processing the pip packages, etc. and then end with a message explaining how to use the created package.

dthain:
Then, to make sure that it is built and installed along with the rest of the cctools:
- Add poncho to the list of packages in `configure`
- Add `poncho/Makefile` so that you have a proper install rule.

If you have it all working, then you ought to be able to `make install` from the top level directory and see that `poncho_package_*` gets installed to `$HOME/cctools/bin`

dthain:
Next - The documentation you have written is great, but it isn't visible from the source directory.  You will need to move things and reformat to get them into their proper places.

- The general overview of Poncho in `README.md` and the spec file documentation `spec_doc.md` should be merged and placed into `doc/manuals/poncho/index.md`.  That will make it part of our online readthedocs manual.  You should also add a table of contents entry to `mkdocs.yml`

- The man pages currently concatenated in `README.md` should be broken into separate files and moved into `doc/man/m4/poncho_package_{analyze|create|run}.m4`.  If you look at the other man page examples, you will see that this just uses a simple macro language, and this is automatically converted into man, html, and md formats for use in various places.

dthain:
And finally -

- Modify `batch_job/src/work_queue_factory.c` and just replace `python_package` with `poncho_package` throughout.  That will make it easy to start workers with poncho packages.

dthain:
Noting some points from our discussion today:

Quick things:
- Set things up so that it isn't required to tar/untar an environment.  This will make it easy to transfer/share/reuse packages without the cost of packing/unpacking.
- Make sure that it is possible to nest calls to poncho_package_run, then you can compose environments from the worker to individual tasks.
- Update `work_queue_factory --package` so that it uses `poncho_package_run` instead of `python_package_run`.  Then you can use that to start workers with a consistent environment.
- Update the Python wq.Factory interface so that you can `specify_environment` for the factory, causing that to use the `--package` feature in the factory.   Then you will be able to easily compare worker-environments and task-environments.
-

dthain:
Looks like `poncho` needs a `clean` rule:
https://github.com/cooperative-computing-lab/cctools/runs/3146066945#step:6:296

dthain:
Something that @btovar clarified at our team meeting yesterday:

Using a package isn't as simple as just un-tarring it: the unpack process also has to modify some items within the package to account for the path of the locally installation.

So, while it is possible to move around a package as a directory without untarrting it, the package won't be immutable.  It still has to be specialized for the local machine.

btovar:
The steps are:
1. untar to `/x/y/z`
2. update paths of files (e.g. #! and ld paths) to `/x/y/z` with `/x/y/z/bin/conda-unpack`
3. activate the environment with `source /x/y/z/bin/activate`

`conda-unpack` is specialized for each environment. It contains the list of files to modify.

poncho/src/Makefile
```diff
@@ -0,0 +1,20 @@
+include ../../config.mk
+include ../../rules.mk
+
+PROGRAMS = poncho_package_analyze poncho_package_create poncho_package_run
+
+all:
+
+clean:
```
dthain, line 8: The `clean` rule should remove derived programs in the build directory, not in the install directory.
This rule should be a no-op.

batch_job/src/work_queue_factory.c
```diff
@@ -1056,9 +1056,9 @@ void add_wrapper_input( const char *filename )
 	if(!wrapper_input) {
 		wrapper_input = strdup(filename);
 	} else {
-		char *tmp = string_format("%s,%s",wrapper_input,filename);
-		free(wrapper_command);
-		wrapper_command = tmp;
+		char *tmp = string_format("%s %s",wrapper_input,filename);
```
dthain, line 1059: Definitely my mistake, thanks for finding it!

work_queue/src/bindings/python3/work_queue.binding.py
```diff
@@ -1965,6 +1971,9 @@ def _write_config(self):
         with open(self._config_file, 'w') as f:
             json.dump(opts_subset, f, indent=4)
 
+    def _specify_environment(self, env):
```
dthain, line 1974: Suggest that you add `factory.specify_package()` and `task.specify_package()` as aliases for `specify_environment`, since all of the other tooling describes the tarballs as packages rather than environments.

dthain (changes requested):
Looking good, just a few little items remaining.

dthain (approved):

dthain:
Looks great!

dthain:
Add PONCHO packaging tools.

PR #2631, opened 2021-07-21
WQ: Add convert/starch example.

dthain:


dthain:
Note that we can get ImageMagick via Conda, which makes the example more reliable, I think:
https://anaconda.org/conda-forge/imagemagick

PR #2633, opened 2021-07-21
python_package_run: lock on unpack dir rather than custom file

btovar:
Take advantage that ${UNPACK_TO} is created by mkdir, which is atomic most of the time.


dthain:
@BarrySlyDelgado make sure you pick up this change for PONCHO as well.

PR #2673, opened 2021-09-03
always use release/ with test and deploy github action

btovar:


dthain:
@btovar proceed or close this one?

PR #2679, opened 2021-09-06
evaluate values in factory config file as JX expressions

jmrundle:


batch_job/src/work_queue_factory.conf.example
```diff
@@ -4,8 +4,9 @@
 	"min-workers": 1,
 	"cores":     8,
 	"disk":      500,
+	"memory": cores * 4000,
 	"timeout": 300,
-	"tasks-per-worker": 8.0,
+	"tasks-per-worker": 8,
```
btovar, line 9: actually this one becomes:

```
"tasks-per-worker": cores
```

which is very nice.

batch_job/src/work_queue_factory.c
```diff
@@ -718,7 +722,7 @@ int read_config_file(const char *config_file) {
 
 	assign_new_value(new_factory_timeout_option, factory_timeout, factory-timeout, int, JX_INTEGER, integer_value)
 
-	assign_new_value(new_tasks_per_worker, tasks_per_worker, tasks-per-worker, double, JX_INTEGER, integer_value)
+	assign_new_value(new_tasks_per_worker, tasks_per_worker, tasks-per-worker, int, JX_INTEGER | JX_DOUBLE, integer_value)
```
btovar, line 725: I think we can safely convert all numeric JX_DOUBLE to ints after a ceil() around line 676. We will need another if to check that it is a double and do the corresponding ceil.

btovar (approved):

PR #2688, opened 2021-09-15
wq: wait do one of each

btovar:
Changes wq wait internal so that connect new workers and submit tasks get executed at least once per call. This has some consequences:

- allows to connect workers when the queue is empty, targeting workflows that submit tasks according to the resources available.
- work_queue_status connecting directly to the manager is ensured to succeed (unless there are very large tasks transfers). Before, a status call would only be processed when the manager did not have anything to do, which was seldom the case when running hundreds of tasks and workers.
- a task is submitted as soon as one is retrieved. This means that the number of tasks running on workers does not decrease as long as there are tasks waiting. However, in steady state, this means that more tasks may have results computed waiting at the workers. In stable environments this is more efficient, as results can be retrieved while new tasks start to run, but in environments with high eviction some results may be lost and they will need to be recomputed.

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -6283,7 +6283,7 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
    - go to S
 */
 {
-	int events = 0;
+	int events_total = 0;
 
 	// account for time we spend outside work_queue_wait
 	if(q->time_last_wait > 0) {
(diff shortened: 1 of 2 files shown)


tphung3 (approved):

dthain:
Hmm, I am not (yet) convinced.  Just looking at the high level description at the top, it seems that the former approach was essentially priority-based: we prefer to retrieve all tasks, then send tasks, then connect workers, etc.  The idea behind that was to make sure that we are not holding on to more workers than we are able to satisfy.

The new approach looks like round-robin rather than prioritized.  So, in one round, we will retrieve a task, send a task, connect a worker, etc.  The downside to this is that it will accept new workers, even when the master is "too busy" to deal with them effectively.  

Now, your point about the status requests is a good one: it would make sense to prioritize status connections.  However, we can't tell the difference at this point, since both workers and status requests arrive at the same port.  (Perhaps they should be split into different ports?)

btovar:
Closing as it has not the desired effect.

PR #2694, opened 2021-09-17
Small fixes for over-commitment to workers

btovar:
Changes the name of `asynchrony-multiplier` to `resource-submit-multiplier` tune parameter.  This factor (default to 1) is applied to cores, memory and gpus when deciding if a task should be sent to a worker. The idea is to allow for tasks to wait at workers.

This pr also fixes an old bug where a task was sent to a worker because of the apparent resources given `asynchrony-multiplier`, but the task would not be able to run because of the actual worker resources.

It also eliminates asynchrony-modifier, as it only had an effect for the old unlabeled tasks before work queue managed resources.

Lastly, it makes work_queue_hungry take into account the over-commitment.


tphung3:
I somehow can't view the commit code when clicking on the commit (github opens 404 error page.)

PR #2702, opened 2021-09-24
Add keep-wrapper-stdout option.

dthain:


dthain:
Some ways this could be better -
- Also redirect stderr for systems that handle them separately.
- Let makeflow choose the output file name.  (Although that has some dangers)
- Clean up the filename when the job is complete.

PR #2708, opened 2021-10-04
Bwisema3/python task abstractions v1.0

bwiseman77:
NOT READY FOR MERGE

Got map, par, and tree-reduce working, but not sure if done correctly. I am still struggling on how to get from my own python into work_queue.py (via work_queue.binding.py) Possibly I have multiple work_queue.py's? I can see my test method (just prints something) in work_queue.py after doing a make install to see if I can update work_queue.py but is not working (does not see method). Also finding some problems with my origin code when moving it, such as I always use the same port, but should probably let the user choose that? I was able to make 3 functions to user Python Task, but I am stuck when trying to place in work_queue.

Update: got test print method to work, but only when in the cctools/bin/python3.9/site-packages dir. Tried with functions, but was not able to create a port ( i am guessing my lack of python class knowledge is showing, but my guess is I can't create a port in that same .py file? Im not sure why it is not happy with that )


work_queue/src/bindings/python3/work_queue.binding.py
```diff
@@ -1719,6 +1722,27 @@ def wait(self, timeout=WORK_QUEUE_WAITFORTASK):
         return None
 
 
+def qmap(f, a):
```
btovar, line 1725: @bwiseman77 I think we want these functions to be a method of WorkQueue, so that we can call them like:

```python
import work_queue as wq
q = wq.WorkQueue(port=9123)

fn = lambda x: x*x
results = q.map(fn, [1,2,3,4])
```

Further, in the case of `map`, we need to return the results in the same order as the arguments were given to us. I suggest keeping a dictionary of  `p_task.id -> i`, where `i` is the index of the argument in something like:

```
for (i, arg) in enumerate(args):
      p_task = PythonTask(f, arg)
      ...
```

work_queue/src/bindings/python3/brett.py
```diff
@@ -0,0 +1,92 @@
...
+
+def wqtreereduce(f, a):
+    q = wq.WorkQueue(9123)
+
+    tmp = a
+    size = len(a)
+
+    while size != 1:
+        results = []
+        for i in range(len(tmp)//2):
```
btovar, line 50: This will loose results if `len(tmp)` is odd.

work_queue/src/bindings/python3/brett.py
```diff
@@ -0,0 +1,92 @@
...
+                print("Task {} failed and did not generate a result.".format(t.id))
+            else:
+                results.insert(0, x)
+
+    return results
+
+def wqallpairs(f, a, b):
+    q = wq.WorkQueue(9123)
+    results = []
+    for i in a:
```
btovar, line 26: We can use here:

```python
import itertools

for i, (x,y) in enumerate(itertools.product(a, b)):
    p_task = wq.PythonTask(f, x, y)
   ... keep a dictionary from p_task to i so that we know the order of submission.
  ... and when inserting to results, we do it in a predictable (i.e., itertools.product) order.
```

As with `map`, all these should be methods of WorkQueue, thus no need to create a queue here and worry about the port number.

btovar (changes requested):

btovar:
For testing, I would `cd work_queue/src/bindings/python3` and create an example file with:

```python
import work_queue as wq

q = wq.WorkQueue(port=9123)

results = q.map(lambda x: x*x, [1,2,3])
... etc ...
```
Then after modifying `work_queue.binding.py`, do a `make` and try your example. Otherwise it may get confusing which `work_queue.py` you are using.

work_queue/src/bindings/python3/brett.py
bwiseman77: I had a question about that - would that odd one just carry down till the end? I was not sure what to do with that

work_queue/src/bindings/python3/work_queue.binding.py
bwiseman77: Ok i was thinking i needed to keep track and assumed I was getting lucky but wasnt sure the best way of keeping track.

work_queue/src/bindings/python3/brett.py
bwiseman77: Ok that makes a lot of sense.

bwiseman77:
Thank you for your very useful feedback ben! Ill get started on fixing these things up

work_queue/src/bindings/python3/brett.py
btovar: The odd one, if any, will always be the last in `tmp`, so we can add it as the last of `results`. Eventually it will be processed in some iteration.

bwiseman77:
@btovar This will be showing my lack of python skills, but since we want it to be q.map (vs wq.map), that means its in the WorkQueue class right? and so would that mean that I need to use self at all to reference all the information about q(such as the submits and waits)?

btovar:
Yes, you got it. A `q.wait(...) will be a `self.wait(...)`, etc.

bwiseman77:
Ok makes sense Thanks!

work_queue/src/bindings/python3/brett.py
bwiseman77: @btovar What do you mean here by "in a predictable (i.e., itertools.product) order"?

bwiseman77:
I believe I have made those new changes. Please let me know if there is anything else. I am going to try to look into other ways of dealing with the exceptions but not add them yet

bwiseman77:
@btovar ^ (sorry forgot in comment)

doc/manuals/work_queue/examples/work_queue_abstractions.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
+
+# Example of how to use work_queue abstractions
+import work_queue as wq
+
+def main():
+    # Set up queue
+    q = wq.WorkQueue(port=9223)
```
btovar, line 8: port=9123   as that's the default everywhere else.

btovar (changes requested):

doc/manuals/work_queue/examples/work_queue_abstractions.py
bwiseman77: Oh shoot Sorry I meant to change that but forgot. Thanks

btovar:
Merged in 2116f783f870daa7dd0bfdc571ffaa427e625157

PR #2710, opened 2021-10-04
WQ: Detect tasks that do not match any worker.

Cpreciad:
Implemented check_task_compatibility()

- am I checking all of the correct things within task_is_compatible()?
- any pointers on how to keep track of time for a given task?

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -4165,10 +4165,7 @@ static void reap_task_from_worker(struct work_queue *q, struct work_queue_worker
 static int send_one_task( struct work_queue *q )
 {
 	struct work_queue_task *t;
-
 	struct work_queue_worker *w;
-	
-	
 
 	timestamp_t now = timestamp_get();
 
@@ -4176,18 +4173,14 @@ static int send_one_task( struct work_queue *q )
 	list_first_item(q->ready_list);
 	while( (t = list_next_item(q->ready_list))) {
 
-
 		// Skip task if min requested start time not met.
 		if(t->resources_requested->start > now) continue;
-
 
 		// Find the best worker for the task at the head of the list
 		w = find_best_worker(q,t);
 
 		// If there is no suitable worker, consider the next task.
-		if(!w){
-			continue;
-		}
+		if(!w) continue;
 		// Otherwise, remove it from the ready list and start it:
 		commit_task_to_worker(q,w,t);
 
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -6371,7 +6371,10 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
 		compute_manager_load(q, 0);
 
 		if(q->wait_for_workers <= hash_table_size(q->worker_table)) {
-			q->wait_for_workers = 0;
+			if(q->wait_for_workers > 0) {
+				debug(D_WQ, "Target number of workers reached (%d).", q->wait_for_workers);
+				q->wait_for_workers = 0;
+			}
 			// tasks waiting to be dispatched?
 			BEGIN_ACCUM_TIME(q, time_send);
 			result = send_one_task(q);
@@ -6425,13 +6428,17 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
 			}
 		}
 
-		// return if queue is empty.
-		BEGIN_ACCUM_TIME(q, time_internal);
-		int done = !task_state_any(q, WORK_QUEUE_TASK_RUNNING) && !task_state_any(q, WORK_QUEUE_TASK_READY) && !task_state_any(q, WORK_QUEUE_TASK_WAITING_RETRIEVAL) && !(foreman_uplink);
-		END_ACCUM_TIME(q, time_internal);
+		// return if queue is empty and something interesting already happened
+		// in this wait.
+		if(events > 0) {
+			BEGIN_ACCUM_TIME(q, time_internal);
+			int done = !task_state_any(q, WORK_QUEUE_TASK_RUNNING) && !task_state_any(q, WORK_QUEUE_TASK_READY) && !task_state_any(q, WORK_QUEUE_TASK_WAITING_RETRIEVAL) && !(foreman_uplink);
+			END_ACCUM_TIME(q, time_internal);
 
-		if(done)
-			break;
+			if(done) {
+				break;
+			}
+		}
 
 		/* if we got here, no events were triggered. we set the busy_waiting
 		 * flag so that link_poll waits for some time the next time around. */
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -174,8 +174,7 @@ struct work_queue {
 
 	struct list *task_reports;	      /* list of last N work_queue_task_reports. */
 
-	double asynchrony_multiplier;     /* Times the resource value, but disk */
-	int    asynchrony_modifier;       /* Plus this many cores or unlabeled tasks */
+	double resource_submit_multiplier; /* Times the resource value, but disk */
 
 	int minimum_transfer_timeout;
 	int foreman_transfer_timeout;
@@ -200,7 +199,7 @@ struct work_queue {
 	int keepalive_timeout;
 	timestamp_t link_poll_end;	//tracks when we poll link; used to timeout unacknowledged keepalive checks
 
-    char *manager_preferred_connection; 
+    char *manager_preferred_connection;
 
 	int monitor_mode;
 	FILE *monitor_file;
@@ -357,16 +356,11 @@ void work_queue_disable_monitoring(struct work_queue *q);
 /********** work_queue internal functions *************/
 /******************************************************/
 
-static int64_t overcommitted_resource_total(struct work_queue *q, int64_t total, int cores_flag) {
+static int64_t overcommitted_resource_total(struct work_queue *q, int64_t total) {
 	int64_t r = 0;
 	if(total != 0)
 	{
-		r = ceil(total * q->asynchrony_multiplier);
-
-		if(cores_flag)
-		{
-			r += q->asynchrony_modifier;
-		}
+		r = ceil(total * q->resource_submit_multiplier);
 	}
 
 	return r;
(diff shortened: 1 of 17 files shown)


work_queue/src/work_queue.c
```diff
@@ -4111,23 +4161,35 @@ static void reap_task_from_worker(struct work_queue *q, struct work_queue_worker
...
 
+
 		// Find the best worker for the task at the head of the list
 		w = find_best_worker(q,t);
 
 		// If there is no suitable worker, consider the next task.
-		if(!w) continue;
-
+		if(!w){
+			timestamp_t temp_time = timestamp_get();
```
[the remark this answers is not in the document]
btovar: @Cpreciad I don't think we should increase the events here. In fact, this code is a fail-safe for when no events can happen for some tasks.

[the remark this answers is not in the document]
Cpreciad: 👍

Cpreciad:
personal note to take a look at what David sent to me on slack

work_queue/src/work_queue.c
```diff
@@ -3965,6 +3969,56 @@ static struct work_queue_worker *find_worker_by_time(struct work_queue *q, struc
 	}
 }
 
+
+// checks workers against tasks and returns if a task has the  
+// possibility of fitting a worker
+static int check_task_fit_in_worker(struct work_queue *q, struct work_queue_task *t, struct work_queue_worker *w)
```
btovar, line 3975: @Cpreciad  We want to compare against all the possible declared resources for the task. Things get a little bit more complicated when resources are declared in a category of tasks rather than directly with  `t->resources_requested`.

What this means is that you have to do (just after the `int ok = 1;`):

```C
    struct rmsummary *l = task_worker_box_size(q, w, t);
```

And then anywhere you use `t->resources_requested` use simply `l`.

Finally, just before the `return ok;`, you need to add:

```C
rmsummary_delete(l);
```

btovar (changes requested):

work_queue/src/work_queue.c
Cpreciad: made the change, and ensured that an unfit task was printed to console

Cpreciad (commented):
Made the requested change to use rmsummary struct to compare if a task can fit to a worker

work_queue/src/work_queue.c
```diff
@@ -84,6 +84,10 @@ See the file COPYING for details.
 
 #define MAX_NEW_WORKERS 10
 
+// Seconds for how often the user can be notified about when a task cannot fit to any workers
+// Used in check_task_fit_in_connected_workers()
+#define THRESHOLD 60000000
```
btovar, line 89: let's make this one:
```
static timestamp_t  interval_tasks_fit_check  180000000;  //3 minutes in usecs
```

work_queue/src/work_queue.c
```diff
@@ -3965,6 +3969,57 @@ static struct work_queue_worker *find_worker_by_time(struct work_queue *q, struc
...
+	//TODO ensure that these resource_requested stats are the correct ones from rmsummary
+	if(w->resources->cores.total < l->cores ) {
+		ok = 0;
+	}
+
+	if(w->resources->memory.total < l->memory ) {
+		ok = 0;
+	}
+
+	if(w->resources->disk.total < l->disk ) { /* No overcommit disk */
```
btovar, line 3994: Remove comment.

work_queue/src/work_queue.c
```diff
@@ -3965,6 +3969,57 @@ static struct work_queue_worker *find_worker_by_time(struct work_queue *q, struc
...
+{
+	char *key;
+	struct work_queue_worker *w;
+	hash_table_firstkey(q->worker_table);
+	
+	
+	while(hash_table_nextkey(q->worker_table, &key, (void**)&w))
+	{
+		if (check_task_fit_in_worker(q, t, w)){
+			//printf("It's Fine. task %d will eventually match a worker\n", (int)t->taskid);
```
btovar, line 4014: Make this a comment. `// Task could run on a currently connected worker`

work_queue/src/work_queue.c
```diff
@@ -4137,6 +4191,26 @@ static int send_one_task( struct work_queue *q )
...
+static void check_all_tasks(struct work_queue *q)
+{
+	struct work_queue_task *t;
+
+	list_first_item(q->ready_list);
+	
+
+	while( (t = list_next_item(q->ready_list))){
+		// check each task against the queue of connected workers
+		if (!check_task_fit_in_connected_workers(q, t)){
```
btovar, line 4203: This could be a very long debug output. We can just print a handful of them and warn than there are others, or we can put them in a list and summarize them somehow.

For the summary, one way I was thinking is to get the maximum per resource of cores, memory, disk, and gpus of all the workers, and the max of cores, memory, disk, and gpus of the tasks that don't fit. Then, per resource, we check how many tasks don't fit in the maximum and print something like:
```
43 waiting tasks need more than 2 cores (the maximum available). The largest task needs 4 cores.
10 waiting tasks need more than 1024 MB of memory (the maximum available). The largest task needs 2000 MB.
```
It may be that all tasks do fit some resource. E.g., 0 gpus needed, 0 gpus offered by workers. If no waiting task goes above the maximum available, then we print nothing.

What do you think?

btovar (changes requested):

work_queue/src/work_queue.c
Cpreciad: Where should I place this variable? should this be a global variable?

btovar: Yes, in the same place where wq_option_blocklist_slow_workers_timeout is. (Around line 125.)

work_queue/src/work_queue.c
Cpreciad: Yeah I really like this actually I thing the first portion would be quite easy to implement, as I can just count up the number of tasks that. I think I can use a sort of bit mask to determine which resources a task check fails. 

I think I would reasonably be able to scan the largest task as well.

Cpreciad: using work _queue_get_stats to get the maximum stats for the workers

Cpreciad: Please take a look at how I tried to get the max stats for the tasks

work_queue/src/work_queue.c
```diff
@@ -4137,6 +4192,64 @@ static int send_one_task( struct work_queue *q )
...
+		if (bit_set & (1<<3))
+			unfit_gpu++;	
+	}
+	struct work_queue_stats s;
+	work_queue_get_stats(q, &s);	
+	
+	if(unfit_core)
+		printf("%d waiting tasks need more than %d cores(max available). The largest task needs %d cores\n",unfit_core,(int)s.max_cores, max_task_stats[0]);
+	if(unfit_mem)
+		printf("%d waiting tasks need more than %d MB of memory(max available). The largest task needs %d MB of memroy\n",unfit_mem,(int)s.max_memory, max_task_stats[1]);
```
btovar, line 4246: s/memroy/memory/

work_queue/src/work_queue.c
```diff
@@ -3965,6 +3968,59 @@ static struct work_queue_worker *find_worker_by_time(struct work_queue *q, struc
...
+	{
+		int new_set = check_task_fit_in_worker(q, t, w);
+		
+		if (bit_set == 0){
+			// Task could run on a currently connected worker, immediately return	
+			return 1;
+		}
+
+		// Inherit the unfit criteria for this task
+		bit_set = bit_set | new_set;
```
btovar, line 4016: Clever... but it may give confusing information to the user.   Say

my task is: 2 cores, 10 GB of memory
worker A: 1 core, 1 GB of memory
worker B: 1 core, 20 GB of memory

The message for memory will read:
1 waiting tasks need more than 20000 MB of memory(max available). The largest task needs 10000 MB of memroy

work_queue/src/work_queue.c
```diff
@@ -4137,6 +4192,64 @@ static int send_one_task( struct work_queue *q )
...
+		if (bit_set & (1<<2))
+			unfit_disk++;	
+		if (bit_set & (1<<3))
+			unfit_gpu++;	
+	}
+	struct work_queue_stats s;
+	work_queue_get_stats(q, &s);	
+	
+	if(unfit_core)
+		printf("%d waiting tasks need more than %d cores(max available). The largest task needs %d cores\n",unfit_core,(int)s.max_cores, max_task_stats[0]);
```
btovar, line 4244: change all of these to:

waiting task(s) need(s) more than

btovar (changes requested):

work_queue/src/work_queue.c
```diff
@@ -3441,7 +3448,9 @@ static struct rmsummary *task_worker_box_size(struct work_queue *q, struct work_
 			}
 
 			if(limits->memory < 0) {
+				printf("b: %lf, %lf\n",limits->cores, limits->memory);
```
btovar, line 3451: Debugging code...

work_queue/src/work_queue.c
```diff
@@ -3441,7 +3448,9 @@ static struct rmsummary *task_worker_box_size(struct work_queue *q, struct work_
...
 				limits->memory = MAX(1, floor(w->resources->memory.largest * max_proportion));
+				printf("a: %lf, %lf\n",limits->cores, limits->memory);
```
btovar, line 3453: Same.

work_queue/src/work_queue.c
```diff
@@ -3965,6 +3972,57 @@ static struct work_queue_worker *find_worker_by_time(struct work_queue *q, struc
...
+	while(hash_table_nextkey(q->worker_table, &key, (void**)&w))
+	{
+		int new_set = check_task_fit_in_worker(q, t, w);
+		if (new_set == 0){
+			// Task could run on a currently connected worker, immediately return	
+			return 0;
+		}
+
+		// Inherit the unfit criteria for this task
+		bit_set = bit_set | new_set;
```
btovar, line 4019: I still think this causes the smallest worker to cause spurious messages.

btovar (changes requested):

work_queue/src/work_queue.c
Cpreciad: I have to put a lot more thought into how I would be able to solve this problem. for now, I will only increment on when a task does not fit to a resource, regardless if it does fit somewhere else. clearly using a bit set does not solve the nuances of potential workers that can fit, so i'll have to go back to the drawing board

btovar:
See #2771

PR #2715, opened 2021-10-12
wq: adds by_apparent_ip mode

btovar:
ignores the list of interfaces and directly tries to connect to the
manager address the catalog observes.

This is the original work queue behaviour from years ago, and it was
already present as a fallback from when a manager did not report a list
of interfaces.


btovar:
@dthain, does this one look to you?

PR #2716, opened 2021-10-19
replace python_environments with poncho

btovar:
Supersedes  #2659


dthain:
Have Coffea and Parsl been updated to use poncho instead?

PR #2725, opened 2021-11-01
Add recurse option to serialization

BarrySlyDelgado:
The default recurse option using dill is false. Thus, previously, if a user used PythonTask for serialization, any sub functions would not be included in the. This would fix that.


btovar:
@BarrySlyDelgado Is this one ready to merge?

BarrySlyDelgado:
@btovar It should be.

Issue #2691, opened 2021-09-17
wq: delayed-wait tune parameter


btovar:
Add the tune parameter "delayed-wait" to `work_queue_tune` and `struct work_queue` `delayed_wait`, which is a boolean (int).

If 0 (the default), then work queue works as it currently does. As soon as a task changes state to `WORK_QUEUE_TASK_DONE`, we break out of the wait loop.

If not 0, then we do not break out of the loop, but do so only if the result of `send_one_task` and `receive_one_task` are both zero.  This allows to send and received as much as tasks as possible given the timeout allowed while respecting the manager capacity. (I suggest adding an `int task_event = 0` to the inner `while`, and do a `task_event |= result` after each *_one_task call.)

We have to be careful not to change more than one task to `WORK_QUEUE_TASK_DONE` per wait call. Thus, the call to `t = task_state_any(q, WORK_QUEUE_TASK_RETRIEVED)` should not be done if `t` is already not NULL.

dthain:
#2746

PR #2726, opened 2021-11-01
Added delayed_wait parameter to alter work_queue loop

David-Simonetti-ND:
Pull request for #2691
Added a delayed_wait paramter to work_queue_tune that, if enabled, alters the function of the main work_queue_wait loop. The loop now only breaks if we fail to send or receive a task, and not only if one task changed to WORK_QUEUE_TASK_DONE.

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -6317,32 +6317,31 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
 	while( (stoptime == 0) || (time(0) < stoptime) ) {
 
 		BEGIN_ACCUM_TIME(q, time_internal);
-		// keep track of whether a task has been sent and recieved
+		// keep track of whether a task has been sent or recieved
 		int task_event = 0;
 		// task completed?
 		if (t == NULL)
 		{
 			t = task_state_any(q, WORK_QUEUE_TASK_RETRIEVED);
-		}
-		if(t) {
-			change_task_state(q, t, WORK_QUEUE_TASK_DONE);
+			if(t) {
+				change_task_state(q, t, WORK_QUEUE_TASK_DONE);
 
-			if( t->result != WORK_QUEUE_RESULT_SUCCESS )
-			{
-				q->stats->tasks_failed++;
-			}
+				if( t->result != WORK_QUEUE_RESULT_SUCCESS )
+				{
+					q->stats->tasks_failed++;
+				}
 
-			// return completed task (t) to the user. We do not return right
-			// away, and instead break out of the loop to correctly update the
-			// queue time statistics.
-			events++;
-			END_ACCUM_TIME(q, time_internal);
-			if (!q->delayed_wait)
-			{
-				break;
+				// return completed task (t) to the user. We do not return right
+				// away, and instead break out of the loop to correctly update the
+				// queue time statistics.
+				events++;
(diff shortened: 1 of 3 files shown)


work_queue/src/work_queue.c
```diff
@@ -6315,9 +6317,13 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
...
-
+		// keep track of whether a task has been sent and recieved
+		int task_event = 0;
 		// task completed?
-		t = task_state_any(q, WORK_QUEUE_TASK_RETRIEVED);
+		if (t == NULL)
+		{
+			t = task_state_any(q, WORK_QUEUE_TASK_RETRIEVED);
+		}
 		if(t) {
```
btovar, line 6327: @David-Simonetti-ND Let's move this `if(t)` inside the `if(t == NULL)`. Otherwise the `change_task_state` and `task_failed++` may be executed several times for the same task.

work_queue/src/work_queue.c
```diff
@@ -6331,7 +6337,10 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
 			// queue time statistics.
 			events++;
 			END_ACCUM_TIME(q, time_internal);
-			break;
+			if (!q->delayed_wait)
```
btovar, line 6340: Then move this if outside the  `if(t==NULL)` and change it to:  `if(t && !q->delated_wait)`

work_queue/src/work_queue.c
```diff
@@ -6770,6 +6784,9 @@ int work_queue_tune(struct work_queue *q, const char *name, double value)
 	} else if(!strcmp(name, "wait-for-workers")) {
 		q->wait_for_workers = MAX(0, (int)value);
 
+	} else if(!strcmp(name, "delayed-wait")){
+		q->delayed_wait = MAX(0, (int)value);
```
btovar, line 6788: Add documentation to `work_queue.h` in `work_queue_tune` and `bindings/python3/work_queue.binding.py`

work_queue/src/work_queue.c
```diff
@@ -6315,9 +6317,13 @@ struct work_queue_task *work_queue_wait_internal(struct work_queue *q, int timeo
 	while( (stoptime == 0) || (time(0) < stoptime) ) {
 
 		BEGIN_ACCUM_TIME(q, time_internal);
-
+		// keep track of whether a task has been sent and recieved
```
btovar, line 6320: sent __or__ received

btovar (changes requested):

btovar:
@David-Simonetti-ND It looks good. Have you been able to test it with the setup you had over the summer? (I.e. with the synthetic application where you control how much time is spent outside wq wait?)

David-Simonetti-ND:
Performance Different based on Task Time
Performance Different based on Max Tasks
Performance Different based on Concurrent Tasks
Performance Different based on Postprocess Time

btovar:
These look great!

btovar:
Seems like I need to update the docker images to include libssl-dev.

dthain:
The remaining failure here is due to the intermittent WQ allocations test, which will be fixed elsewhere.
@btovar is this otherwise ready to merge?

btovar:
Almost. We are renaming it "wait_retrieve_many" (vs wait_retrieve_one).

PR #2728, opened 2021-11-02
adds wait_for_tag to python bindings

btovar:
q.wait_for_tag(tag) will block until a task with the corresponding tag returns or until the queue is empty. 
Tasks without the desired tasks are kept in a set, and returned when calling wait() of wait_for_tag() as if they have just completed.

E.g.:

```python
t_a = Task('echo hello')
t_a.specify_tag('hello')
q.submit(t_a)

t_b = Task('echo world')
t_b.specify_tag('world')
q.submit(t_b)

t = q.wait_for_tag('world')
print(t.output)    # prints 'world'
t = q.wait()
print(t.output)    # prints 'hello' (assuming no other task in queue)
```

### DIFF
--- a/work_queue/src/bindings/python3/work_queue.binding.py
+++ b/work_queue/src/bindings/python3/work_queue.binding.py
@@ -1045,6 +1045,10 @@ class WorkQueue(object):
         self._stats_hierarchy = None
         self._task_table = {}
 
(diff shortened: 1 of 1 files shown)


dthain:
I think this properly belongs natively in the C library, not in Python.  (And it's a good thing for Brett to work on.)

PR #2729, opened 2021-11-02
Simplify poncho_package_create

btovar:
Moved env cache and base/full packages back to topeftenv.py. In this
way, poncho_package_create can just focus on creating environments. 
This means that the output argument is required again.

I still don't think the reading the spec from stdin works correctly.


BarrySlyDelgado (approved):
Everything looks good.

Issue #2635, opened 2021-07-22
WQ: Add TLS Support to Manager-Worker Connection


dthain:
Add (optional) support for TLS security in WQ.  This requires several levels of changes:
- Update `configure` and friends to identify libgnutls.
- Update conda and spack build recipes to require libgnutls.
- Modify `link.c` to perform TLS send/recv in place of plain read/write.
- Modify the top level tools to accept options for enabling TLS and loading identity types.

bbockelm:
Dumb question:  why `libgnutls` instead of OpenSSL?  As a consumer, I'm a bit ambivalent between the two -- but as someone who might be able to help debug in the future, I know my way around the OpenSSL code base quite well!

btovar:
We already link to openssl, so my guess is that `libgnutls` above is a typo that means "the chosen tls library". :)

dthain:
No strong opinion about the SSL library.

@bbockelm To be consistent with how you are using Coffea+Dask in coffea-casa, can you tell us a little more about your authentication setup?  e.g. are you using a PSK between manager and workers, or something else?

bbockelm:
Dask does a pretty traditional mTLS setup.  On first login, we generate a CA cert, user cert, and host cert and configure Dask to find the three credentials as needed (client gets user cert, scheduler + worker gets host cert).

bbockelm:
@btovar -- Excited to hear about the TLS support?  In the current implementation, are you using SNI to indicate the desired hostname to the server?

This is the key trick we use in Coffea-Casa to run 100 users behind the same IP address.

btovar:
@bbockelm We are! The workers use the `SSL_set_tlsext_host_name(ssl, hostname)`, so that information should be available before any handshake happens.

PR #2733, opened 2021-11-09
wq: openssl support

btovar:
(Issue #2635)

@dthain: If it looks ok, I can update the manual and man pages.

Still need to add the options to makeflow when using Work Queue, but I think we can do that in another pr.

I tried to make it so that current code does not need any change. Only a handful of functions from link.c had to be wrapped with openssl code.

For the catalog, the managers report "ssl: true when appropriate.  A worker can then initiate a TLS connection. When using HOST PORT, a worker needs the --ssl flag.  The --ssl flag is always needed for `work_queue_status HOST PORT` when the manager to query is using ssl.


btovar:
Seems like I need to update the docker images to include libssl-dev

dttools/src/link.c
```diff
@@ -346,13 +363,73 @@ void sockaddr_set_port( struct sockaddr_storage *addr, int port )
...
         }
-}               
+}
 
 struct link *link_serve(int port)
 {
 	return link_serve_address(0, port);
 }
 
+int _ssl_errors_cb(const char *str, size_t len, void *use_warn) {
```
dthain, line 373: static functions?

dttools/src/link.c
```diff
@@ -560,13 +715,70 @@ struct link *link_connect(const char *addr, int port, time_t stoptime)
 	return 0;
 }
 
+ssize_t _read(struct link *link, char *data, size_t count) {
```
dthain, line 718: Name of _read likely conflicts with libc symbols.

dttools/src/link.c
```diff
@@ -560,13 +715,70 @@ struct link *link_connect(const char *addr, int port, time_t stoptime)
...
+
+		return result;
+	}
+#endif
+
+	//if no ssl support, or no link->ssl, fallback to read syscall
+	return read(link->fd, data, count);
+}
+
+ssize_t _write(struct link *link, const char *data, size_t count) {
```
dthain, line 745: Ibid

dthain (changes requested):
Yes, I think this is really well done, with the changes being quite compact.
A few naming items in the middle, but the approach looks good overall.

btovar:
Build fails because TLS_client_method() in more recent versions of openssl. Adding a config check to select the correct openssl call...

btovar:
@dthain, this is ready to merge from my side. You may want to take a look at the note in the wq manual. 
https://github.com/cooperative-computing-lab/cctools/pull/2733/commits/33db544c64b3f1231f5c754a926074eeba408353

doc/manuals/work_queue/index.md
```diff
@@ -1222,6 +1222,56 @@ API](http://ccl.cse.nd.edu/software/manuals/api/html/work__queue_8h.html).
 
 ### Security
```
dthain, line 1224: At the beginning of the section, place the note that, by default, no authentication is done,
and so you can do SSL, password, or both.

doc/manuals/work_queue/index.md
```diff
@@ -1222,6 +1222,56 @@ API](http://ccl.cse.nd.edu/software/manuals/api/html/work__queue_8h.html).
...
 
+#### TLS (SSL) support
+
+Work Queue can encrypt the communication between manager and workers using TLS.
+For this, you need to specify the key and certificate (in PEM format) of your
+server when creating the queue.
+
+If you do not have a key and certificate at hand, but you want the
+communications to be encrypted, you can create them using a command like:
+
```
dthain, line 1233: Explain a little bit more here that this is self-signed, with the usual proviso...

doc/manuals/work_queue/index.md
```diff
@@ -1222,6 +1222,56 @@ API](http://ccl.cse.nd.edu/software/manuals/api/html/work__queue_8h.html).
...
+work_queue_worker (... other args ...) --ssl HOST PORT
+work_queue_factory (... other args ...) --ssl HOST PORT
+work_queue_status --ssl HOST PORT
+condor_submit_workers -E'--ssl' HOST PORT
+```
+
+
+#### Password Files
+
 By default, Work Queue does **not** perform any authentication, so any workers
```
dthain, line 1275: Move this paragraph above...

doc/manuals/work_queue/index.md
dthain: Also a bit that SSL provides manager-to-worker authentication, while password gives you worker-to-manager auth.

dthain: Also, that reminds me that the warning message emitted by the manager should now mention using either password or ssl.

doc/manuals/work_queue/index.md
```diff
@@ -1222,6 +1222,56 @@ API](http://ccl.cse.nd.edu/software/manuals/api/html/work__queue_8h.html).
 
 ### Security
 
+#### TLS (SSL) support
```
dthain, line 1225: Call this SSL consistently, since that's how we name the options and code...

Issue #2734, opened 2021-11-10
makeflow_monitor: tasks_total is defined both as an attribute and a method decorated with @property


btovar:
At least with python 3.7.5 (as reported by Willem Marais) this causes an exception when we try to set `self.tasks_total = 0`, as it is interpreted as the @property, not the attribute.

btovar:
Willem confirmed it now works for him.

dthain:
Great!

PR #2735, opened 2021-11-10
makeflow_monitor fixes

btovar:
Fix for #2734 

Also other fixes for newer python versions.


dthain:
Can you add a suitable smoke test to make sure makeflow_monitor works correctly with a simple makeflow?

btovar:
Will do.

Issue #2731, opened 2021-11-02
WQ Allocations Test


dthain:
This test seems to periodically fail on Mac OS.

When running on Linux, it has the following problem:
```
./TR_work_queue_allocations.sh: line 37: ./wq_alloc_test.py: Permission denied
```

After fixing that, I get this, indicating that the allocated number of gpus is not equal to the number of requested gpus:

```
(base) bash-4.2$ ./TR_work_queue_allocations.sh run
/afs/crc.nd.edu/user/d/dthain/work/cctools/work_queue/test
Writing port 1024 to file wq.port
/afs/crc.nd.edu/user/d/dthain/miniconda3/lib/python3.8/site-packages/work_queue.py
Waiting for manager to be ready.
Master is ready on port 1024 
Running worker.
work_queue_worker: creating workspace /tmp/worker-102744-30607
work_queue_worker: using 4 cores, 200 MB memory, 200 MB disk, 8 gpus
connected to manager localhost:1024 via local address 127.0.0.1:56970
all_specified:
expected: {'cores': 1, 'memory': 2, 'disk': 3, 'gpus': 4}
allocated: {'cores': 1, 'memory': 2, 'disk': 3, 'gpus': 0}
Traceback (most recent call last):
  File "./wq_alloc_test.py", line 53, in <module>
    check_task('all_specified', wq.WORK_QUEUE_ALLOCATION_MODE_FIXED, max = r, min = {}, expected = r)
  File "./wq_alloc_test.py", line 34, in check_task
    assert(t.resources_allocated.gpus == expected['gpus'])
AssertionError
disconnected from manager localhost:1024
work_queue_worker: deleting workspace /tmp/worker-102744-30607
Worker completed.
```

btovar:
Taking this one over to speed the release of 7.3.6

PR #2739, opened 2021-11-11
add debug switch to TR_work_queue_allocations.sh

btovar:
The test succeeds running locally on cclmac02. Adding `set -x` as we are not getting any output when the test fails with the github action.


btovar:
To investigate #2731.

btovar:
And of course now it succeeds! bah.

PR #2740, opened 2021-11-11
WQ Trash Directory

dthain:
First pass at deleting items in a common trash directory.


work_queue/src/work_queue_worker.c
```diff
@@ -2092,28 +2090,43 @@ workspace_prepare is called every time we connect to a new manager,
 static int workspace_prepare()
 {
 	debug(D_WQ,"preparing workspace %s",workspace);
+
 	char *cachedir = string_format("%s/cache",workspace);
 	int result = create_dir(cachedir,0777);
 	free(cachedir);
 
 	char *tmp_name = string_format("%s/cache/tmp", workspace);
```
dthain, line 2098: @btovar Do you have any idea why the temp directory was set up inside the cache directory?  Would it make sense to put it at the same level in the hierarchy?

btovar: I don't. It puzzled me too when I saw it recently.

dthain:
@btovar can you test this branch and see if it eliminates the poncho_package_run errors?

btovar:
Will do.

dttools/src/trash.c
```diff
@@ -0,0 +1,62 @@
...
+
+#include <sys/stat.h>
+#include <string.h>
+#include <errno.h>
+
+static char *trashdir = 0;
+
+void trash_setup( const char *dir )
+{
+	trashdir = strdup(dir);
```
btovar, line 16: Rather than keeping a global trashdir, I think `trash_file` and `trash_empty` should have as an argument their `trash_dir`. I can see us in the future trying to set trash directories per task and being confused that only the last trash directory created works.

In any case, the setup should do nothing if the trashdir has been setup once.

Issue #2671, opened 2021-09-02
Move the script cctools_gpu_autodetect to gpu_info.c


btovar:
The script is simple and does not need its own file. Further, we wouldn't have to worry to send it together with the worker.

dthain:
@Cpreciad another good starter one.  This one is simple but surprisingly important: just invoke `nvidia-smi` directly instead of through `cctools_gpu_autodetect`

dthain:
Just looked at gpu_info and observed that gpu_name_get() does approximately the right thing, but gpu_info_get is far too complex.

PR #2741, opened 2021-11-12
WQ: Move Script into Program

Cpreciad:
This is for issue #2671


dthain:
Yes, this is on the right track.  I wonder if you can make it even smaller.
For example, can you use the `--query` option to `nvidia_smi` in a similar way as `gpu_name_get`?

dthain:
Whoops, I think this is mixing up two pull requests.  It has some code form the no-available-worker feature.

Cpreciad:
made new pull request with clean branch

PR #2744, opened 2021-11-15
WQ: Abstractions with Chunk Size

bwiseman77:
first attempt at chunk_sized map
not sure if it is right to use normal python map?
I used it because I was not sure how else to do each "chunk" with only one tast
not sure if using maht.ceil is ok
not sure if how I check for end of array is ok
I believe it is working, and is faster than last version


dthain:
Yes, that's the right idea, I like how you are using a native map within the PythonTask.
A couple things to check/report on:
- Fix up the merge conflicts.  Touch base with @btovar if you need some help there.
- Double check that you have all the boundary conditions right: for example, what happens when the number isn't evenly divisible by the chunksize, etc.
- Measure the performance in terms of functions/second for large arrays with a chunksize of 1000.
- Extend the idea to `wq.pairs` and `wq.treeReduce`
-

PR #2745, opened 2021-11-15
WQ: Abstractions with Chunk Size - Map

bwiseman77:
Conflicts:
	work_queue/src/bindings/python3/work_queue.binding.py

### DIFF
--- a/work_queue/src/bindings/python3/work_queue.binding.py
+++ b/work_queue/src/bindings/python3/work_queue.binding.py
@@ -1748,7 +1748,7 @@ class WorkQueue(object):
     # @param seq        The sequence that will call the function
     # @param chunk_size The number of elements to process at once
 
-    def Nmap(self, fn, array, chunk_size=1):
+    def map(self, fn, array, chunk_size=1):
         size = math.ceil(len(array)/chunk_size)
         results = [None] * size
         tasks = {}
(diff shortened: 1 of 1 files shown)


bwiseman77:
Not sure What I should do with other PR, so I will just leave both open for now? git is confusing and I dont wanna mess anything else up

bwiseman77:
todo: 
draw/think about tree reductions
time map
fix map output

dthain:
@bwiseman77 I think a constructive thing to do here is to replace the old `wq.map` with your new map implementation (just call it `wq.map`), and complete and commit that as a single PR and call it done.

Then, you can start thinking about how to improve `wq.treereduce` and then make that a new PR for consideration.  Likewise for `wq.allpairs`.

bwiseman77:
Ok, I just replaced map on this PR, is that ok? Should I close the other PR?

dthain:
Yes, close the other one if you don't need it any more.

work_queue/src/bindings/python3/work_queue.binding.py
```diff
@@ -1745,19 +1746,30 @@ def wait(self, timeout=WORK_QUEUE_WAITFORTASK):
     # @param self       Reference to the current work queue object.
     # @param fn         The function that will be called on each element
     # @param seq        The sequence that will call the function
-    def map(self, fn, seq):
-        results = [None] * len(seq)
+    # @param chunk_size The number of elements to process at once
+
+    def Nmap(self, fn, array, chunk_size=1):
```
dthain, line 1751: Call this one `map` since it's a drop in replacement for the old one.

bwiseman77: Op yes sorry I meant to but forgot to delete the N.

dthain:
Looks good!

Issue #2737, opened 2021-11-11
wq: add wait_for_tag


btovar:
Add a `work_queue_wait_for_tag(q, tag)` call when only done tasks with the desired tasks are returned. It should work in conjunction with `work_queue_wait`, so that done tasks with the incorrect tag are not lost.

The use case is #2708, where only tasks of the corresponding calls are included in the results.

dthain:
Done?

PR #2749, opened 2021-11-16
wq: work_queue_wait_for_tag

btovar:
For #2737.

Decided to modify `work_queue_wait_internal` directly, as that allows us to use the RETRIEVED state for tasks with the incorrect tag. Otherwise we would need to add another state, or maintain a list of done tasks.

Asking for the NULL (None) tasks means to wait for any task (e.g. q.wait_for_tag(None, time) == q.wait(time))

### DIFF
--- a/work_queue/src/work_queue.h
+++ b/work_queue/src/work_queue.h
@@ -786,6 +786,16 @@ If the task could not, then the <tt>result</tt> field will be non-zero and the
 */
 struct work_queue_task *work_queue_wait(struct work_queue *q, int timeout);
 
+
+/** Wait for a task with a given task to complete.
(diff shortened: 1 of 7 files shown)


work_queue/src/work_queue.h
```diff
@@ -786,6 +786,16 @@ If the task could not, then the <tt>result</tt> field will be non-zero and the
 */
 struct work_queue_task *work_queue_wait(struct work_queue *q, int timeout);
 
+
+/** Wait for a task with a given task to complete.
```
dthain, line 790: "given tag"

dthain (changes requested):
Looks good!  One little typo.

PR #2755, opened 2021-11-22
WQ Abstractions: Tree Reduce

bwiseman77:
work in progress, but I believe it is close to done, as it works, and bigger chunks seem to be faster. Need to do an actual timed test with large data. 

For breaking up into tasks, associative operations work no matter how you break up the data, but operations like divison will change based on the size. The way I tried to fix this was by only breaking up the data in powers of 2, this way breaking up the tasks is the same as doing it as one big task. However, I am not sure if this is the best way, and there would be a gap between the requested chunk_size and the used one

### DIFF
--- a/work_queue/src/bindings/python3/work_queue.binding.py
+++ b/work_queue/src/bindings/python3/work_queue.binding.py
@@ -1796,15 +1796,49 @@ class WorkQueue(object):
     # @param self       Reference to the current work queue object.
     # @param fn         The function that will be called on each element
     # @param seq        The seq that will be reduced
+    # @param chunk_size The number of elements per Task (for tree reduc, must be greater than 1)
 
-    def tree_reduce(self, fn, seq):
-        size = len(seq)
+    def tree_reduce(self, fn, seq, chunk_size=2):
+        def reduce(fn, seq):
+            ''' Serializable tree reduce function '''    
+            while len(seq) > 1:
+                results = []
+                for i in range(len(seq)//2):
+                    results.append(fn(seq[(i*2)], seq[(i*2)+1]))
+
+                if len(seq) % 2 == 1:
+                    results.append(seq[-1])
+                
+                seq = results
+    
+            return seq[0]
+
+        def find_pow(n):
+            ''' Finds the largest power of 2 smaller than n '''
+            bit = 1
+            while(bit <= n):
+                bit = bit << 1
+
+            return bit >> 1
+
+        # parallel function
         tasks = {}
-
+        n = find_pow(chunk_size)
+        
         while len(seq) > 1:
-            results = [None] * (len(seq)//2)
-            for i in range(len(seq)//2):
-                p_task = PythonTask(fn, seq[(i*2)], seq[(i*2)+1])
+
+            size = math.ceil(len(seq)/n)
+            results = [None] * size
+        
+            for i in range(size):
+                start = i*n
+                end = start + n
+
+                if end > len(seq):
+                    p_task = PythonTask(reduce, fn, seq[start:])
+                else:
+                    p_task = PythonTask(reduce, fn, seq[start:end])
+
                 self.submit(p_task)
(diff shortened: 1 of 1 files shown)


bwiseman77:
even normal associative:

        24
      /      \
   2         12
 /    \      /   \
1    2    3    4

even parallel associative(ck_sz=2): 
('|' represents seperated tasks)

        24
      /      \
   2         12
 /    \   |   /   \
1    2  |  3    4

odd normal associative:

             720
           /          \
         24          30
      /       \            \
   2          12       30 
 /    \      /    \      /     \ 
1    2    3    4    5     6

odd parallel associative(ck_sz=3):
 
             720
           /          \
       6              120
     /   \      |      /     \
   2     3    |   20    6
 /    \     \   |   /   \      \ 
1    2    3  |  4    5     6


even normal non-associative:

        1
      /      \
   2         2
 /    \      /   \
8    4    2    1

even parallel non-associative(ck_sz=2):

         1
      /      \
   2         2
 /    \   |   /   \
8    4  |  2    1

odd normal non-associative:

                1/2
             /          \
             1           2
         /       \           \
      2           2           2 
   /    \        /    \       /     \ 
32    16    8    4    2     1

odd parallel non-associative (ck_sz=3):
 
                 1/8
           /                \
         1/4               2
        /   \       |      /     \
     2      8     |    2      1
   /    \       \   |   /   \      \ 
32    16    8  |  4    2     1

odd parallel non-associative (but go to largest pow of 2 < chunk_size -> so ck_sz = 3, n = 2):
                1/2
             /          \
             1           2
         /       \     |      \
      2           2  |         2 
   /    \    |   /    \    |   /     \ 
32    16 |   8    4 |   2     1

bwiseman77:
^need to learn how to comment

dthain:
This is coming along nicely.  As we discussed today, please extend the work (on this branch) to allow for n-ary reduction functions, like `max`.  For example, if I do `wq.tree_reduce(max,arr,100)` where `arr` has 1000 values, then I should get 10 tasks that compute `max` on 100 items, and then one final task that takes the max of the combined results.

bwiseman77:
is the 100 the "chunk_size" that I have already, or would that be a separate variable? I am trying to think of a way of determining how to know how many arguments a function takes, and I feel that should be different from the "chunk_size". For example, if u pass that into a binary function, I will have a task do a tree reduces on 100 items, but in that function, it breaks up those 100 elements to be passed into the function. But for something like max, that would be 100 for the chunk, and then every element should get passed into max, but I do not know of a way of determining how many of those 100 elements should be passed into a given function at a time.

dthain:
I think the way to proceed here is to assume that the reduction function is of the form `value = function( iterable )` and it will combine as many items as you give it into one value.

bwiseman77:
I think I get that part, but say for a binary function will look like this:
value = function((x, y)) or value = function( (*2 element iterable*) )
but I'm giving the task 100 elements to do that to, so id have to break up the larger 100 elements into groups of 2,
but for the max, the chunk_size and the iterable is the same size, so I am "breaking" it up into 1 group of 100
but I am not sure how to tell how to break up the "chunk" into the correct iterable?

bwiseman77:
I could see how I could do it given the number of arguments the function wants, but that seems undesirable

dthain:
If the task has 100 elements and the function takes an iterable, then just invoke `function( [100 items] )` and return the result.

bwiseman77:
How would I tell the difference between that (all 100 at once) and one that needs to be done multiple times(2 per func call but done 50 times in this case)?

bwiseman77:
I guess how can I tell if I need to break up the 100 elements or just pass the whole thing

bwiseman77:
Or maybe that is the wrong question

dthain:
What I mean is that we **require** the user to pass in a function that takes in an iterable as an argument.

dthain:
Looks good please add docs in next PR.

PR #2758, opened 2021-11-22
WQ: Simplify GPU Detection

Cpreciad:
Issue #2671 is now in this branch. cleaned up the branch


dttools/src/gpu_info.c
```diff
@@ -23,36 +23,20 @@ See the file COPYING for details.
...
+	if(access(nvidia_cmd, X_OK) != 0){	
+			return 0;
 	}
+	
+	FILE *pipe = popen(nvidia_cmd, "r");
+	if (!pipe)
+		return 0;
+	
+	int gpus;
+	fscanf(pipe, "%d", &gpus);
```
dthain, line 36: Make sure to check for the case that fscanf fails, so you can handle systems where nvidia-smi fails and the error message doesn't match "%d"

PR #2763, opened 2021-11-23
Catalog: Add HTTPS/SSL query capability.  (WIP)

dthain:
First pass, as yet untested.


dthain:
Doesn't quite work yet b/c the existing code breaks the `link` abstraction by getting the underlying fd with `link_fd` and then creating a stdio stream, which breaks the SSL abstraction.  What we need to do here is create a `link_printf` that sits on top of `link_write`.

dthain:
Catalog now serves up results via ssl correctly.

work_queue/src/work_queue.c
```diff
@@ -5677,6 +5686,10 @@ void work_queue_delete(struct work_queue *q)
 			free(q->manager_preferred_connection);
 
 		free(q->poll_table);
+
+		if(q->ssl_cert) free(q->ssl_cert);
```
btovar, line 5690: No need of `if`.  `free(NULL)` does nothing.

dttools/src/link.c
```diff
@@ -532,30 +532,17 @@ struct link *link_serve_address(const char *addr, int port)
...
-	return 0;
-}
-
-
-int link_ssl_wrap_accept(struct link *parent, struct link *link) {
-#ifdef HAS_OPENSSL
-	if(parent->ctx) {
 		debug(D_TCP, "setting up ssl state for %s port %d", link->raddr, link->rport);
 
-		link->ctx = _create_ssl_context(/* is client */ 1);
```
btovar, line 551: Drop the argument from `_create_ssl_context`, and use `TLS_method()` or `SSLv23_method()` as appropriate.

btovar (changes requested):

dthain:
Catalog is working now, WQ fixed up slightly to deal with changes to link.
Remaining piece here is that deltadb needs the capability to write to a link as well as a stream.
Also need to look at the performance impact of link_printf versus printf.
The former does a single link_write for each link_printf, whereas the latter is buffering blocks.
We may need to add an output buffer for link_printf.

btovar (changes requested):

btovar:
The changes to `link_accept` also fix a current bug with ssl in osx and conda where SSL_accept fails with a SSL_SYSCALL_ERROR.  Thus, I would like to add this pr to 7.4.0.

dthain:
I think this branch needs more cooking -- I'm leery of updating the catalog server too aggressively.  Are you willing to backport the ssl changes?  I think it's just a matter of link.c and work_queue.c

btovar:
Yes, I was thinking of backporting those changes.

re nonblocking: It is a possibility, but I did try changing the SSL_accept wrapper to also retry on want_read/write, but the error is fatal for the connection.

dthain:
This is stuck at the moment b/c the OSX build does not detect the presence of openssl in configure.
This is because OSX no longer puts dynamic libraries in /usr/lib, so configure cannot find them.
Yep, really:

_New in macOS Big Sur 11.0.1, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)_

https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes

So, we need to figure out how to adjust configure appropriately without turning the whole thing inside out...

btovar:
Great.

We'll need to support both methods (path and dlopen), as we need the path method for conda.

dthain:
Ah, are you telling me that when we install on osx via conda, the libraries are found in the usual way?

btovar:
Yes, openssl comes from conda.

btovar:
Actually, looking around, dlopen may not help us much as the openssl headers may not be there. People seem to recommend installing openssl from brew, etc.

dthain:
Worked around this by writing the SSL test to depend up on CCTOOLS_OPENSSL_AVAILABLE.
I think we are ok as long as Conda is the preferred installation method.

PR #2771, opened 2021-12-08
work_queue: print a warning for tasks that don't fit workers

btovar:
Every three minutes we check for tasks that don't fit connected workers.
The warning includes the minimum size of worker needed to run these
tasks.

(Supersedes #2710: fixes merge history and renames some functions.)
(@Cpreciad could you check it with your examples?)

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -4250,7 +4250,7 @@ static int send_one_task( struct work_queue *q )
 static void print_large_tasks_warning(struct work_queue *q)
 {
 	timestamp_t current_time = timestamp_get();
-	if(current_time - q->time_last_large_tasks_check > interval_check_for_large_tasks) {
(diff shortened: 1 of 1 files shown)


work_queue/src/work_queue.c
```diff
@@ -4173,6 +4246,77 @@ static int send_one_task( struct work_queue *q )
 	return 0;
 }
 
+
+static void print_large_tasks_warning(struct work_queue *q)
+{
+	timestamp_t current_time = timestamp_get();
+	if(current_time - q->time_last_large_tasks_check > interval_check_for_large_tasks) {
```
Cpreciad, line 4253: should be < instead of >

Cpreciad (changes requested):

work_queue/src/work_queue.c
btovar: Indeed, good catch.

Cpreciad (approved):

PR #2774, opened 2021-12-10
WQ PythonTask Abstractions Docs

bwiseman77:


dthain:
Yes, that's the right idea, but please flesh out the examples to be a little more complete.
Show us what's in the sequences and what comes out as results..

dthain:
@bwiseman77 this one was left over from the end of last semester, please make the examples a little more detailed, for the user to follow.

bwiseman77:
yes, working on that now. Got env all figured out. Is there a way to see how the changes will look? to me it is still magic how it goes from that file to the man pages.

dthain:
The process is a bit complicated.  Once committed, the website readthedocs pulls everything from our repo and turns the MD into a manual via the tool mkdocs.

It is possible to run mkdocs locally if you conda install it, and then run `mkdocs serve` at the root of our repo.
https://www.mkdocs.org/getting-started/

But since hte formatting here isn't super complicated, I think it's sufficient just to draft things and commit when ready.

doc/manuals/work_queue/examples/wq_python_abstractions.py
```diff
@@ -0,0 +1,37 @@
+#import work_queue
+import work_queue as wq
+
+# Example funtion for the Map Abstraction
+def Map(x):
```
dthain, line 5: This is the right idea, but the naming is confusing b/c you have both `map` and `Map` and only the former does mapping.  :)

Suggest you call these three  `addfunc` `mulfunc` and `maxfunc` or something similar that describes what each function actually does.

dthain (changes requested):

doc/manuals/work_queue/examples/wq_python_abstractions.py
bwiseman77: That makes sense. I was going for "map function is used for map abstraction", But I agree that is definitely less confusing :smile:

dthain:
Looks good!

dthain (approved):

PR #2776, opened 2021-12-13
WQ: Fix backwards compatibility for Python 2.7

dthain:
Fixed some minor idioms in work_queue.binding.py that were valid Python 3 but not valid Python 2.7, resulting in test regressions.

### DIFF
--- a/work_queue/src/bindings/python3/work_queue.binding.py
+++ b/work_queue/src/bindings/python3/work_queue.binding.py
@@ -881,7 +881,9 @@ class Task(object):
 try:
     import dill
     pythontask_available = True
-except ModuleNotFoundError:
+except Exception:
+    # Note that the intended exception here is ModuleNotFoundError.
+    # However, that type does not exist in Python 2
     pythontask_available = False
 
 class PythonTask(Task):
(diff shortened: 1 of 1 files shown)


RyanGutenkunst:
@dthain The behavior of PythonTask has changed recently (breaking our code), and my guess is that this commit did it.

As a quick example, the manager.py code here: https://cctools.readthedocs.io/en/latest/work_queue/ now fails with a TypeError:
  File "/Users/rgutenk/Desktop/temp.py", line 18, in <module>
    task = wq.PythonTask(my_sum, value, value)
  File "/Users/rgutenk/cctools/lib/python3.9/site-packages/work_queue.py", line 1838, in __init__
    self._serialize_python_function(func, *args)
  File "/Users/rgutenk/cctools/lib/python3.9/site-packages/work_queue.py", line 1901, in _serialize_python_function
    dill.dump(list(*args), wf, recurse=True)
TypeError: list expected at most 1 argument, got 2

Was this change in behavior intendend?

dthain:
@RyanGutenkunst moving this discussion over to issue #2792 so that it doesn't get lost

Issue #2326, opened 2020-07-15
Compilation error with GCC 8 and above


bcdarwin:
Compiling from source with no special flags using GCC 8.x and glibc 2.31 (not triggered with GCC 7.x and glibc 2.31):

```
In file included from <...>-glibc-2.31-dev/include/fcntl.h:329,
                 from libforce_halt_enospc.c:13:
libforce_halt_enospc.c: In function 'open':
<...>-glibc-2.31-dev/include/bits/fcntl2.h:41:1: error: inlining failed in call to always_inline 'open.localalias.0': redefined extern inline functions are not considered for inlining
 open (const char *__path, int __oflag, ...)
 ^~~~
libforce_halt_enospc.c:47:12: note: called from here
   err_fd = open(filename, O_RDWR | O_CREAT);
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
COMPILE jx2json.o
make[2]: *** [../../rules.mk:6: libforce_halt_enospc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory '/build/cctools-7.1.6-source/dttools/src'
make[1]: *** [Makefile:10: src] Error 2
make[1]: Leaving directory '/build/cctools-7.1.6-source/dttools'
make: *** [Makefile:23: dttools] Error 2
```

dthain:
@trshaffer same as #2327, thanks.

bcdarwin:
To reproduce, use the same steps as detailed in #2327, but in pkgs/top-level/all-packages.nix in the Nixpkgs repo, replace `cctools = callPackage ./pkgs/applications/science/misc/cctools { stdenv = gcc7Stdenv; };` with a `gccXStdenv` of your choice higher than 7 (or simply use `{ }` to get the current one, currently gcc 9).

btovar:
Compilation is now confirmed to work with gcc 11.2.0.

bcdarwin:
This inlining error still occurs for me with GCC 8 and above when compiling 7.4.1 or trunk (currently b1829604).

btovar:
@bcdarwin thanks again for your report, and I'm sorry I closed this issue prematurely. Our compilation tests with conda (now the preferred delivery method) did not show any issues with the most recent compilers. I've just submitted a pr that seems to fix most issues with nix. The only thing missing is the recently added openssl support.

PR #2777, opened 2021-12-13
Nix fixes

btovar:
@dthain: The most important is the change from sigsetmask to sigprocmask. sigsetmask has been deprecated and it causes an error with pedantic compile options.

(For #2326. Note that the compiler versions are ok with conda, but not with nix. I assume nix is setting more pedantic compilation options. )

### DIFF
--- a/configure
+++ b/configure
@@ -95,7 +95,7 @@ ccompiler=${CC:-gcc}
 cxxcompiler=${CXX:-g++}
 linker="${ccompiler}"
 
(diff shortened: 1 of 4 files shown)


btovar:
Also, openssl does not work yet with nix. nix separates where it puts `include/` and `lib/` directories for openssl. Our `configure` scripts assumes that include and lib have the same root for each package.

dthain (approved):

Issue #2784, opened 2022-01-13
wq: low space on AFS seems to confuse transfer of output files


btovar:
Seen with topEFT. When there is low free space in the AFS volume to which the manager writes output files, the transfer of output files seems to occur correctly: the manager checks that there is enough space, and then reports that all expected bytes were written. However, the output files do not contain any bytes.  This was observed in a volume filled at 98%.

I tried to reproduce the error using a loopback device on a local disk, but in that case wq correctly detects that there is not enough space (which leads to the bug #2783). Thus, if it is indeed the case that the problem is related to low disk free space, then it seems related to some AFS quirk. 

Perhaps we need to sync after every transfer and do an fstat to check that the file is the expected size. Syncs are not very cheap...

dthain:
The AFS quirk here is that the file is not transmitted from the client to the server until close().
And so a file write looks successful all the way until you attempt to close(), and then the close() fails with an error.
Turns out that we have to check the error message from close!

btovar:
Oh yes, we do not check the close exit status. Let me do a pr on that. For now I will make it a worker error to be consistent with other transfer failures, but we may want to change that later.

btovar:
From the close man page:

>   Not  checking  the  return  value of close() is a common but nevertheless serious programming error.  It is quite possible
       that errors on a previous write(2) operation are first reported at the final close().  Not checking the return value  when
       closing the file may lead to silent loss of data.  This can especially be observed with NFS and with disk quota.

That's us!

Issue #2783, opened 2022-01-13
wq: low disk space keeps manager running forever


btovar:
Before the manager transfers back an output file from a worker, it first checks that there is enough local disk space. If there is not, the transfer fails, the task fails, and it is reported as a worker error, which means that the worker is disconnected. If using the work queue factory, then this means that workers are replenished, and tasks are retried, but no progress is ever made. If not using the factory, all workers are eventually disconnected, and again no progress is made.

Some alternatives:
- Report the error as an APP_ERROR. The manager then terminates and the workflow fails.
- Reset the transfer attempt, so that results are kept in the worker. Print a warning that disk space is needed. Periodically retry the transfer again.

dthain:
Or an even simpler placeholder:
- Emit a big warning message stating (disk is full) -- and block in that state until the user fixes it...

btovar:
One of the issues about blocking the state is that it is not clear when to unblock. Should we unblock as soon as there is space for the transfer that just failed?

On top of that, like for AFS, the filesystem does seem to have space, but if fails until the close, so it is not exactly clear how much free space is really needed. Because of these, I'm partial to simply fail with an APP_FAILURE.

dthain:
Right: `df /afs` (statfs) doesn't produce useful information about space.  `fs quota` does though...

btovar:
That's was also a confusing aspect. For the topEFT workflow  fs quota gave a 98% utilization, with about 5GB free. Files we were trying to transfer were as small as 40MB, but they all failed. They were definitely some larger than that, but it seemed that those last GB were not available to us. 

But say that `fs quota` does give a number we can use. Do you propose to call `fs quota` if we detect we are running on afs? Maybe we can just call it when we detect that the `close` call fail.

btovar:
Studying the code more closely, the error reported is APP_FAILURE, but the task simply fails with OUTPUT_MISSING. As discussed in our meeting, I'll add a MANAGER_FAILURE that will terminate the workflow immediately. We may want to think of a more refined error handling in the future.

dthain:
I think this issue has been addressed as far is it can be, yes?

btovar:
Correct.

PR #2786, opened 2022-01-14
wq: check close for errors at manager

btovar:
An error on close may report an error on a previous write (e.g. on AFS).
We observed this with AFS volumes close to their quota, where output
files seemed to be transferred correctly to the manager, but failed to
update in the shared file system.

For #2784.  Preserves current behavior as if the `write` calls failed (see #2783).


dthain:
This is definitely an improvement in that we are now catching an error that was not caught before.
I'm not sure if I agree that failure-to-close implies an application-level failure, though.

Maybe it would be instructive to ask some of our users "What do you expect us to do in this case?"  :)

PR #2790, opened 2022-01-20
DeltaDB: Various Query Fixups

dthain:


dthain:
OSX failure of WQ test here is very odd.  Transient issue?

btovar:
I think it is transient.  The error comes from  `t=q.wait(5)` timing out and returning None because no worker connected.

PR #2793, opened 2022-01-24
Adds kwargs to PythonTask

### LINKED ISSUE #2792: WQ: PythonTask Compatibility Regression
(moving discussion over from closed PR)

@RyanGutenkunst reports:

The behavior of PythonTask has changed recently (breaking our code), and my guess is that [this commit](https://github.com/cooperative-computing-lab/cctools/pull/2776#issuecomment-1018921846) did it.

As a quick example, the manager.py code here: https://cctools.readthedocs.io/en/latest/work_queue/ now fails with a
```
TypeError: Traceback (most recent call last):
File "/Users/rgutenk/Desktop/temp.py", line 18, in
task = wq.PythonTask(my_sum, value, value)
File "/Users/rgutenk/cctools/lib/python3.9/site-packages/work_queue.py", line 1838, in init
self._serialize_python_function(func, *args)
File "/Users/rgutenk/cctools/lib/python3.9/site-packages/work_queue.py", line 1901, in _serialize_python_function
dill.dump(list(*args), wf, recurse=True)
TypeError: list expected at most 1 argument, got 2
```
Was this change in behavior intendend?

btovar:
Also fixes #2792 by serializing arg list as is, rather than expanding and creating a new list.

### DIFF
--- a/work_queue/src/bindings/python3/work_queue.binding.py
+++ b/work_queue/src/bindings/python3/work_queue.binding.py
@@ -893,7 +893,8 @@ class PythonTask(Task):
     # @param self 	Reference to the current python task object
     # @param func	python function to be executed by task
     # @param args	arguments used in function to be executed by task
-    def __init__(self, func, *args):
+    # @param kwargs	keyword arguments used in function to be executed by task
+    def __init__(self, func, *args, **kwargs):
         self._id = str(uuid.uuid4())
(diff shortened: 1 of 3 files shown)


dthain (approved):
Looks like a reasonable approach. +1 for adding the test case.

btovar:
For the test to work we need dill in the docker images. I'm thinking how to do it easily, maybe from pip?

PR #2802, opened 2022-01-31
No strcpy to same buffer

btovar:
Updated compiler used in conda does not like that source and destination of strcpy are the same buffer. `deltadb_process_stream` modified so that the copy is not necessary, it must be imperceptibly faster too.


btovar:
The commit does not hide the copy with an alias, but avoids the copy all together.  Rather than copying characters towards index 0, the line pointer is moved to signal a new start of the string.  The original line is kept in the `whole_line` buffer.

dthain:
My mistake, I didn't read carefully enough!

dthain (approved):

PR #2812, opened 2022-02-06
WQ Factory: Fix shared scratch space

benjaminlyons:


benjaminlyons:
I'm not entirely sure how to test this fix on the crc machines without screwing up my environment, but I think this is a fix for the issue i had last week

benjaminlyons:
It seems to have not broken anything I guess thats a good sign?

btovar (approved):

dthain:
@David-Simonetti-ND this is the "orthogonal" fix to what we discussed last week.  We still need your contribution to ensure that the Factory class waits for the work_queue_factory to actually exit after being killed...

PR #2818, opened 2022-02-11
add specify_path to pair() in wq bindings

bwiseman77:
Env are all nice i think, but dill module still not working? maybe i have it set up wrong, but I am not sure why its not working. It connects to the condor jobs but return error because of no dill. if I just use specify_env in my test file, it works, it is just with in work_queue.binding,py, but not sure why


btovar:
Did you `make clean all install` again?

bwiseman77:
yes, still same error

`python test.py`
`warning: using plain-text when communicating with workers.`
`warning: use encryption with a key and cert when creating the manager.`
`Traceback (most recent call last):`
  `File "pytask_wrapper.py", line 2, in <module>`
 `   import dill`
`ImportError: No module named dill`

btovar:
In your `test.py` find the call to `WorkQueue(...)` and add: `debug_log=":stdout"` as an argument.  When you run it again, please look for the command line that has `pytask_wrapper`. I want to know if it is being executed by itself, or with `./poncho_package_run ...`.

bwiseman77:
I added a file with the log output, I see pytask wrapper sometimes with poncho_package_run

btovar:
Note:
```
python pytask_wrapper.py function_1fea624f-8e13-462e-8340-3fb196bdaccd.p args_1fea624f-8e13-462e-8340-3fb196bdaccd.p out_1fea624f-8e13-462e-8340-3fb196bdaccd.p
```
So I don't think the environment is being set at all, probably the incorrect version of `work_queue.py` is being used.

bwiseman77:
why might that be? and how can i make sure i am using the right version?

btovar:
@David-Simonetti-ND I don't think so. I need to cleanup the pr as it includes a log file that should be there. The error is not related to the pr.

Issue #2819, opened 2022-02-11
work_queue_factory does not properly wait for children to finish before it returns


David-Simonetti-ND:
work_queue_factory does not properly wait for all children to return before itself quits. An example of the problem is given below
<img width="1434" alt="Screen Shot 2022-02-04 at 10 34 03 AM" src="https://user-images.githubusercontent.com/31227025/153625094-aa5d36c1-846a-4daa-83d2-8ada39c9e456.png">
In this picture, the process tree for before and after sending SIGTERM to the work_queue_factory process is pictured. As shown, the factory process and calling python process are able to successfully exit before both poncho_package_run and its child work_queue_worker are able to return.
The reason for this issue is illustrated below.
<img width="447" alt="Screen Shot 2022-02-11 at 9 59 43 AM" src="https://user-images.githubusercontent.com/31227025/153625504-19b85a4b-b299-4d8a-a7fb-60e054ba2793.png">
When cleaning up, work_queue_factory calls remove_all_workers
<img width="597" alt="Screen Shot 2022-02-11 at 9 59 23 AM" src="https://user-images.githubusercontent.com/31227025/153625599-c451249a-c8bd-4c07-91ed-b73be60473a7.png">
Which calls batch_job_remove
<img width="767" alt="Screen Shot 2022-02-11 at 9 59 16 AM" src="https://user-images.githubusercontent.com/31227025/153625637-13370e09-7922-458f-b208-3cdd51bda856.png">
Which calls (at least in the case shown above) batch_job_local_remove, and this function sends SIGTERM to the child process it wishes to wait for.
<img width="642" alt="Screen Shot 2022-02-11 at 9 59 06 AM" src="https://user-images.githubusercontent.com/31227025/153625718-ae7ceb7f-4e42-446c-9055-6ea41a8280e3.png">
Which calls process_waitpid()
<img width="464" alt="Screen Shot 2022-02-11 at 9 58 58 AM" src="https://user-images.githubusercontent.com/31227025/153625752-56e9e546-59c6-465d-8e4f-1af246f8dccc.png">
Which repeatedly calls process_work() to check for a change in the process state (looking for when the process changes from running to zombie presumably)
<img width="464" alt="Screen Shot 2022-02-11 at 9 58 58 AM" src="https://user-images.githubusercontent.com/31227025/153625972-a7a1d119-30c4-4649-9e4b-e6f78910261a.png">
[...]

PR #2822, opened 2022-02-17
batch_job_local_remove properly waits for child to finish and poncho_package_run properly cleans up and waits for children

David-Simonetti-ND:
Fix for #2819
Problem was that process_waitpid() had a timeout set to 0, so it would never wait for the given subprocess to finish before returning. It now sends SIGTERM to the process every 5 seconds and only returns when the child process is finished


David-Simonetti-ND:
Also added a patch so that when poncho_package_run calls its cleanup function, it sends SIGTERM to its child (i.e. a work_queue_worker) and properly waits for it to exit before finishing itself

David-Simonetti-ND:
Screen Shot 2022-02-16 at 11 50 38 PM
batch_job_local_remove now loops forever sending SIGTERM to the child process we want to finish. It waits up to 5 seconds checking if the process finished, and only returns if the process actually returns, otherwise it sends SIGTERM again and waits up to five more seconds in a loop.
Screen Shot 2022-02-16 at 11 53 32 PM
Pictured above is with the code in the pull request. As shown, each process properly waits for all children to finished before returning, and no parentless processes are left.

batch_job/src/batch_job_local.c
```diff
@@ -110,20 +110,23 @@ static batch_job_id_t batch_job_local_wait (struct batch_queue * q, struct batch
 
 static int batch_job_local_remove (struct batch_queue *q, batch_job_id_t jobid)
 {
-	if(kill(jobid, SIGTERM) == 0) {
+	while(kill(jobid, SIGTERM) == 0) {
```
btovar, line 113: We may want to try only a handful of times, and then send a SIGKILL. If the batchjob ignores the SIGTERM, then we wait forever.

poncho/src/poncho_package_run
```diff
@@ -152,6 +152,12 @@ function cleanup {
     s=$((total_ms/1000))
 
     logmsg total runtime: $(printf "%d.%0.3d seconds" "${s}" "${ms}")
+    cpids=$(pgrep -P "$$")
```
btovar, line 155: I don't think pgrep is a standard utility.  Can we get the same information from `ps`?

poncho/src/poncho_package_run
```diff
@@ -152,6 +152,12 @@ function cleanup {
...
+    for cpid in $cpids;
+    do
+        kill -SIGTERM $cpid
+        wait $cpid
```
btovar, line 159: We want to send the kill to all  cpids and then `wait` without arguments for all of them.
In the current implementation, if one process is ignoring the SIGTERM, then we do not sent the signal to the rest of the processes. If at that point poncho_package_run gets another signal (e.g. from C-c), it will quit without sending the SIGTERM to all the processes.

btovar (changes requested):
Nice ideas!

poncho/src/poncho_package_run
btovar: Nevermind, in linux `ps` and `pgrep` come from the same package (procps).

batch_job/src/batch_job_local.c
btovar: I wonder if it would be too difficult to add a function that gets us the pid's of all the children of jobid, like `pgrep` does, and repeat here what you do for poncho, that is, send a kill to all the children too.

poncho/src/poncho_package_run
dthain: I hesitate to ask.... but what if those children forked other processes?

btovar: Just checked, pgrep is not recursive, bummer.

dthain: Let's not make the perfect the enemy of the good here, but at least drop a comment indicating that the solution isn't complete..

batch_job/src/batch_job_local.c
David-Simonetti-ND: I added a functionality so that it attempts up to 12 times (60 seconds) for a clean termination with SIGTERM, and then tries another 12 times/60 seconds sending SIGKILL, returning 0 if this is still not successful.
Doesn't work_queue_factory already track all of the child processes/jobs and when it calls remove_all_workers it calls batch_job_local_remove on each of the children, so wouldn't have a function to do the same thing be repetitive?

poncho/src/poncho_package_run
David-Simonetti-ND: How responsible should a process be for cleaning up its children/grandchildren/great-grandchildren/etc? For example, if we have
work_queue_factory -> poncho_package_run -> work_queue_worker -> cctools_monitor
Should processes be responsible for all their progeny? If so the factory would send termination signals to poncho, the worker, and the resource monitor? Poncho will then also send termination signals to the worker and the monitor, and then the worker will also send a termination signal to the monitor.  It seems a bit wasteful to me.
Or can we rely on each process cleaning up its immediate child. For example the factory sends SIGTERM to poncho, which sends SIGTERM to the worker, which sends SIGTERM to the monitor, monitor returns, worker returns, poncho returns, and finally the factory returns.
If we are looking for a recursive solution like the first one listed here, pstree along with some parsing might be of great assistance.
Screen Shot 2022-02-17 at 9 22 34 PM

poncho/src/poncho_package_run
```diff
@@ -152,6 +152,20 @@ function cleanup {
...
+    for cpid in $cpids; # loop through and send termination signals to all children
+    do
+        kill -SIGTERM $cpid
+    done
+    # WARNING!!!!! this is unfinished. Any child process that blocks/ignores SIGTERM will cause this program to wait infinitely for them to finish
+    # the proper solution would be to set a timer that sends an interrupt signal to this process after a certain wait (say 60 seconds)
+    # after that time, if any child processes remain running, send SIGKILL to them because that can not be blocked and attempt to wait again for them to return
+    # unfortunately, bash has not native implementation of alarm(), and the only way for a process to send a signal to itself in the future is to create another background process to handle that
+    # but creating another process to send a signal to oneself has its own host of issues...
+    for cpid in $cpids; # wait for all child processes to finish
```
btovar, line 165: No need for a loop here. I think we can do:
```sh
wait $cpids
```

btovar (changes requested):

poncho/src/poncho_package_run
```diff
@@ -152,6 +152,17 @@ function cleanup {
...
+    for cpid in $cpids; # loop through and send termination signals to all children
+    do
+        kill -SIGTERM $cpid
+    done
+    # WARNING!!!!! this is unfinished. Any child process that blocks/ignores SIGTERM will cause this program to wait infinitely for them to finish
+    # the proper solution would be to set a timer that sends an interrupt signal to this process after a certain wait (say 60 seconds)
+    # after that time, if any child processes remain running, send SIGKILL to them because that can not be blocked and attempt to wait again for them to return
+    # unfortunately, bash has not native implementation of alarm(), and the only way for a process to send a signal to itself in the future is to create another background process to handle that
+    # but creating another process to send a signal to oneself has its own host of issues...
+    wait $cpids
```
btovar, line 165: One thing to consider is that there is no way to wait for grandchildren processes, so the use of pgrep together with wait may not work as we want. We can only wait on the child process that we spawn when running the command line.

batch_job/src/batch_job_local.c
```diff
@@ -110,20 +110,37 @@ static batch_job_id_t batch_job_local_wait (struct batch_queue * q, struct batch
 
 static int batch_job_local_remove (struct batch_queue *q, batch_job_id_t jobid)
```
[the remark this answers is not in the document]
btovar: Actually, SIGHUP may be more appropriate.

dthain:
@btovar are you satisfied that this is an improvement, or is there more to do?

btovar:
I think it needs to be updated to use `process_kill_waitpid`.

David-Simonetti-ND:
I changed batch_job_local_remove to use process_kill_waitpid so this should be good to merge.

PR #2824, opened 2022-02-18
initial serverless prototype

nicklocascio:
This is the beginning of what we have been discussing for moving towards a serverless feature in work queue. Right now, when an "@FUNCTION" call is made in a manager, the worker will send a message to the network_function.py server and wait for a response from the server. The current configuration sends a pseudo-event JSON string to network_function.py where a simple function is invoked and then a response is sent back to the worker.


work_queue/src/work_queue_process.c
```diff
@@ -256,6 +260,60 @@ pid_t work_queue_process_execute(struct work_queue_process *p )
...
+			char * result_json = "{\"result\": \"success\"}";
+			write(p->output_fd, result_json, strlen(result_json));
+			close(p->output_fd);
+			exit(0);
+			*/
+			
+			// Should this be split into a separate function?
+			// i.e. void work_queue_process_network_function(struct work_queue_process *p)
+			// build necessary networking components
+			struct hostent *hp;
```
dthain, line 275: This is the right idea, but please use the modules in `dttools` for dealing with the network,
which already deal with a variety of cross platform issues, ipv6, and so forth:

`link` provides an abstractions for network connections, and handles 
https://github.com/cooperative-computing-lab/cctools/blob/master/dttools/src/link.h

`domain_name` performs domain name lookups:
https://github.com/cooperative-computing-lab/cctools/blob/master/dttools/src/domain_name.h

work_queue/src/work_queue_process.c
```diff
@@ -256,6 +260,60 @@ pid_t work_queue_process_execute(struct work_queue_process *p )
...
+			struct sockaddr_in sin;
+			char * host;
+			char * port;
+			int s, len;
+			socklen_t addr_len;
+			char buf[BUFSIZ];
+
+			host = "crcfe02.crc.nd.edu";
```
dthain, line 283: Connect to localhost

work_queue/src/work_queue_process.c
nicklocascio: I have started to change the code to use `link` but I am having some trouble with `link_connect` throwing a connection refused error. I'm running `network_function.py` on the same machine as the worker and the port is hardcoded right now, so I pushed up the changes to see if you see something that looks wrong.

dthain: Aha, your server is listening on a udp port, but the link module does tcp.
Use the `datagram` module instead if you want to do udp instead.

nicklocascio: Fixed!

nicklocascio:
I am currently working on trying to read data from an input file to send to the network function. I'm not sure if I am using the wrong method to open and read the input files or what exactly is going on, but I figured that I would push what I am trying to see if you have any feedback. 

I know that the file is being sent from the manager to the worker successfully from the debug logs, so I just need to figure out how to read it

work_queue/src/work_queue_process.c
```diff
@@ -267,6 +267,37 @@ pid_t work_queue_process_execute(struct work_queue_process *p )
...
+			struct list *infiles = p->task->input_files; 
+			char *file = (char *)list_pop_head(infiles);
+			printf("file: %d", list_size(p->task->input_files));
+			*/
+
+			// this is returning a data from f->payload, but says that there is no such file when I try to open it
+			struct work_queue_file *f;
+			list_first_item(p->task->input_files);
+			while((f = list_next_item(p->task->input_files))) {
+				char *file_name = f->payload;
```
dthain, line 281: I think you want t->remote_name here, instead of t->payload.  (See work_queue_internal.h)

nicklocascio: That worked. The latest commit is fully working reading data from input file and sending to network function. It didn't look like there was anything in `dttools`for this so I used `fread` but if I missed something I can change that

nicklocascio: I also tested with different formats of JSON on the manger end and modifying the handler in the network function accordingly and that worked, so the JSON schema doesn't matter at all in terms of the worker, which is what we're looking for

work_queue/src/work_queue_process.c
```diff
@@ -256,6 +259,81 @@ pid_t work_queue_process_execute(struct work_queue_process *p )
...
+				}
+				
+				// get the file size for reading
+				fseek(fp, 0L, SEEK_END);
+				numbytes = ftell(fp);
+				fseek(fp, 0L, SEEK_SET);
+				
+				// allocate memory and read
+				file_buf = (char *)calloc(numbytes, sizeof(char));
+				fread(file_buf, sizeof(char), numbytes, fp);
```
dthain, line 287: Use `full_fread` from `dttools/full_io` which will ensure that you get all the requested bytes.

work_queue/src/work_queue_process.c
```diff
@@ -256,6 +259,81 @@ pid_t work_queue_process_execute(struct work_queue_process *p )
...
+
+			// get response from network function
+			bzero(buf, sizeof(buf));
+			int bytes_recv = datagram_recv(dgram, buf, sizeof(buf), addr, &port, timeout);
+			if(bytes_recv < 0) {
+				fatal("error receving from network function: %s", strerror(errno));
+			}
+
+			// write response to manager output file
+			write(p->output_fd, buf, strlen(buf));
```
dthain, line 332: Likewise use `full_write` here.

work_queue/src/work_queue_process.c
```diff
@@ -256,6 +259,81 @@ pid_t work_queue_process_execute(struct work_queue_process *p )
 		if(result == -1)
 			fatal("could not dup pipe to stderr: %s", strerror(errno));
 
+		if(strcmp(p->task->command_line, "@FUNCTION") == 0) {	
+			// read the contents from the input file
```
dthain, line 263: You are doing all the right things here, but factor this out into a bunch of small functions, each one doing one little thing, something like this:

```
input = load_input_file( task )
output = invoke_coprocess_function( task, input )
write_output_file( output )
```

etc...

work_queue/src/network_function.py
```diff
@@ -0,0 +1,48 @@
...
+
+def main():
+	s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+	
+	try:
+		s.bind(('localhost', 45107))
+	except Exception as e:
+		s.close()
+		print(e)
+		exit(-1)
```
btovar, line 21: Let's do exit(1) here. Negative numbers may be associated with signals.

nicklocascio:
Updated and fully working! Adding a null char to the buffer did not work but `calloc` as opposed to `malloc` did the trick, I think because `calloc` will zero out the entire string first

work_queue/src/work_queue_process.c
```diff
@@ -194,6 +198,68 @@ static void specify_resources_vars(struct work_queue_process *p) {
...
+
+	int bytes_read = full_fread(fp, buf, fsize);
+	if(bytes_read < 0) {
+		fatal("error reading file: %s", strerror(errno));
+	}
+
+	return buf;	
+}
+
+char * invoke_coprocess_function(char *input) {
```
btovar, line 221: static char *invoke_...

work_queue/src/work_queue_process.c
```diff
@@ -256,6 +322,19 @@ pid_t work_queue_process_execute(struct work_queue_process *p )
 		if(result == -1)
 			fatal("could not dup pipe to stderr: %s", strerror(errno));
 
+		if(strcmp(p->task->command_line, "@FUNCTION") == 0) {
```
btovar, line 325: This makes me a little nervous. I think having special command lines is ok for testing, but current users would expect to "opt-in", as it is possible that somebody does have a "@FUNCTION" executable.

I suggest adding a  --python-function argument to the work queue worker which will keep the command_line to check against, e.g.  work_queue_worker `--python-function @my_func`.   In the future we will then execute some_module.my_func.

nicklocascio:
The two most recent commits include the minor fixes that you noted @btovar as well as a new python-function argument in work_queue_worker

btovar (approved):

Issue #2800, opened 2022-01-28
Catalog: 


dthain:
@jmrundle, collecting together a variety of things we have discussed.

- [ ] Front page of catalog should summarize number of records of each type.
- [ ] Drill down into records should show relevant details for that type.
- [ ] Should be possible to view current state, snapshot at a given Unix time, or grab history stream from time X to Y
- [ ] When ready, should use deltadb multiple tables to segregate types.  (do this last)
- [ ]

dthain:
Another element that would be nice:
- [ ] Allow a DELETE verb to remove a record from the catalog (if from the same host/ip addr).  This makes it possible for a cleanly shutdown service to remove itself when desired.
- [ ]

dthain:
Thank you @jmrundle for the next work on the catalog and deltadb to get the historical view exposed and working.

Next piece for another student to work on is an improved landing page that summarizes records by type.

PR #2826, opened 2022-02-22
DeltaDB: historical snapshots on catalog webpage

jmrundle:
https://github.com/cooperative-computing-lab/cctools/issues/2800#issue-1117809035

### DIFF
--- a/deltadb/src/catalog_export.c
+++ b/deltadb/src/catalog_export.c
@@ -11,7 +11,6 @@ See the file COPYING for details.
 
 #include <stdlib.h>
 #include <string.h>
-#include <stdio.h>
 
 static char * unquoted_string( struct jx *j )
 {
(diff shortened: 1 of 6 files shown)


deltadb/src/catalog_server.c
```diff
@@ -475,23 +477,71 @@ static void handle_query( struct link *ql, time_t st )
...
+		// continue on
+	} else if (matches == 1) {
+		strncpy(path, "/", sizeof("/"));
+	} else {
+		strcpy(path, full_path);
+	}
+
+	// reset catalog to timestamp
+	if (timestamp > 0) {
+		// deltadb_delete(table);
```
jmrundle, line 532: NOTE:  unsure sure if not deleting `table` before overwriting it causes a leak, since the variable is inherited during process fork.  Either way, it might be better to avoid deleting since it impacts performance, and the leak will be recovered shortly after on process exit.

deltadb/src/catalog_export.c
```diff
@@ -160,3 +159,39 @@ void catalog_export_html_footer( struct link *l, struct jx_table *h, time_t stop
...
+void catalog_export_html_datetime_picker( struct link *l, time_t stoptime, time_t current) {
+	struct tm *t1 = gmtime(&current);
+	struct tm *t2 = gmtime(&current);
+	struct tm *t3 = gmtime(&current);
+
+	int year = t1->tm_year + 1900;
+	int month = t1->tm_mon + 1;
+	int day = t1->tm_mday;
+
+	t2->tm_mday--;
```
dthain, line 172: What happens if tm_mday goes negative here?

deltadb/src/catalog_export.c
```diff
@@ -160,3 +159,39 @@ void catalog_export_html_footer( struct link *l, struct jx_table *h, time_t stop
...
+	t2->tm_hour = t2->tm_min = t2->tm_sec = 0;
+	time_t yesterday = mktime(t2);
+
+	t3->tm_mday += 2;
```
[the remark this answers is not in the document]
dthain: Not sure if midnight belongs to the prior day or the current day.
Perhaps using 12:01 will result in more clarity?

deltadb/src/deltadb.c
```diff
@@ -444,13 +443,17 @@ static int log_recover( struct deltadb *db, time_t snapshot )
 	return 1;
 }
 
-struct deltadb * deltadb_create( const char *logdir )
+struct deltadb * deltadb_create( const char *logdir, time_t timestamp )
```
dthain, line 446: Let's make the API more specific here, so as to avoid misuse:

`deltadb_create(logdir)` - creates a new instance at the current time, and permits appending.
`deltadb_create_snapshot(logdir,time)` - creates a new instance at a prior time, and inhibits appending.

deltadb/src/deltadb.c
```diff
@@ -461,7 +464,7 @@ struct deltadb * deltadb_create( const char *logdir )
 
 	if(logdir) {
 		db->logdir = strdup(logdir);
-		log_recover(db,time(0));
+		log_recover(db,timestamp);
 	}
 
 	return db;
```
dthain, line 470: After this: `update` and `remove` should be a no-op if the object was created as a snapshot.

dthain (changes requested):

dthain:
@jmrundle this is very close, please make a few quick changes and we can merge.  Looks like you have a minor compile warning on one platform as well.

deltadb/src/catalog_export.c
jmrundle: mktime handles that for us :)

https://www.cplusplus.com/reference/ctime/mktime/

> The values of the members tm_wday and tm_yday of timeptr are ignored, and the values of the other members are interpreted even if out of their valid ranges (see [struct tm](https://www.cplusplus.com/tm)). For example, tm_mday may contain values above 31, which are interpreted accordingly as the days that follow the last day of the selected month.
> 
> A call to this function automatically adjusts the values of the members of timeptr if they are off-range or -in the case of tm_wday and tm_yday- if they have values that do not match the date described by the other members.

deltadb/src/catalog_export.c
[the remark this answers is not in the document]
jmrundle: Ah just realized it was because I had decremented the value a few lines earlier to get yesterdays value, so I needed to increment twice to reach tomorrows value.

deltadb/src/deltadb.c
```diff
@@ -322,7 +322,11 @@ static int log_replay( struct deltadb *db, const char *filename, time_t snapshot
 	FILE *file = fopen(filename,"r");
 	if(!file) return 0;
 
-	while(fgets(line,sizeof(line),file)) {
+	while(fgets(whole_line,sizeof(whole_line),file)) {
```
jmrundle, line 325: Comes from [deltadb_stream](https://github.com/cooperative-computing-lab/cctools/blob/master/deltadb/src/deltadb_stream.c#L40)

Issue #2825, opened 2022-02-21
WQ: Coprocess Option


dthain:
@David-Simonetti-ND please work on this one, it will provide a clean setup and teardown step for #2824 and perhaps other facilities we may invent.

Add to `work_queue_worker` an option called `--coprocess` that will start an arbitrary process when the worker starts up.  (And also kill it when the worker shuts down.)

Then, that will allow @nicklocascio to couple a network function to a worker like so:
```
work_queue_worker --coprocess network_function.py
```

Then, add a similar option to `work_queue_factory` that just passes the same string through to the worker:
```
work_queue_factory --coprocess network_function.py . . .
```

Since we already have the option `--python-package` to deploy the worker in a given poncho environment, putting them together like this:
```
work_queue_factory --python-package package.tar.gz --coprocess network_function.py . . .
```

Will get us workers that look like this:
```
poncho_package_run package.tar.gz work_queue_worker --coprocess network_function.py
```

dthain:
Let's make it a feature that is separate from tasks for now.

dthain:
Sure, you could put it in `dttools/src/process.c` and call it `process_really_kill` or something to that end,
and then call it from both modules.

dthain:
Complete.

PR #2827, opened 2022-02-24
Wqcoprocess

David-Simonetti-ND:
Initial PR for issue #2825

### DIFF
--- a/dttools/src/process.c
+++ b/dttools/src/process.c
@@ -98,6 +98,66 @@ struct process_info *process_waitpid( pid_t pid, int timeout)
 	return 0;
 }
 
+int process_kill_waitpid(pid_t pid, int timeout, int max_attempts)
+{
+	int flags = 0;
+        struct sigaction new_action, old_action;
+
+	if(timeout == 0) {
+                flags = WNOHANG; // if the timeout is zero, we want waitpid to return immediately
+        } else { // set up the signal handler for alarm to get us out of wait
+                flags = 0;
+                new_action.sa_handler = alarm_handler;
+                sigemptyset(&new_action.sa_mask);
+                new_action.sa_flags = 0;
+                sigaction(SIGALRM, &new_action, &old_action);
+        }
+	int num_attempts = 0; // keep track of how many time we attempted to signal process pid
+	int current_signal = SIGTERM; // start by sending SIGTERM, move on to SIGKILL
+	int status; 
+	int ret_val = 1; // return value from this function: 0 means process exited cleanly, 1 means there was an error
+	while (kill(pid, current_signal) == 0) // send signal to process
+	{
+		if (timeout != 0) {alarm(timeout);} // and set an alarm for timeout seconds
+		num_attempts++;
+		if (num_attempts == max_attempts) // change to sending SIGKILL after max_attempts tries
+		{
+			current_signal = SIGKILL;
+		}
+		else if (num_attempts >= max_attempts * 2) // and give up after 2 * max_attempt tries
+		{
+			break;
+		}
+        	pid_t return_pid = waitpid(pid, &status, flags); // we will wait for the process to return for timeout seconds before the alarm signal is sent and we break out of wait
+		if (return_pid < 0) // handle errors
+		{
+			if (errno == EINTR) // this is the case that the alarm woke us up so we try again to signal
+			{
+				continue;
+			}
+			else // if there is any other error, it is fatal and we should just exit
+			{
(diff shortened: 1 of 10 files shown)


David-Simonetti-ND:
I am unhappy with how the end_coprocess code is right now because it is very much a copy paste of the code work_queue_factory uses. would there be some way to unify these functions?

btovar:
@David-Simonetti-ND, yes, that's a good idea.  Could you add such a function to `dttools/src/process.*` with a name such as `process_kill_waitpid(pid, timeout)` ?  It should work in the same way as `process_waitpid`.

dttools/src/process.c
```diff
@@ -98,6 +98,66 @@ struct process_info *process_waitpid( pid_t pid, int timeout)
...
+                sigemptyset(&new_action.sa_mask);
+                new_action.sa_flags = 0;
+                sigaction(SIGALRM, &new_action, &old_action);
+        }
+	int num_attempts = 0; // keep track of how many time we attempted to signal process pid
+	int current_signal = SIGTERM; // start by sending SIGTERM, move on to SIGKILL
+	int status; 
+	int ret_val = 1; // return value from this function: 0 means process exited cleanly, 1 means there was an error
+	while (kill(pid, current_signal) == 0) // send signal to process
+	{
```
btovar, line 120: to match the style of the other functions, move this brace to previous line

dttools/src/process.c
```diff
@@ -98,6 +98,66 @@ struct process_info *process_waitpid( pid_t pid, int timeout)
...
+		if (timeout != 0) {alarm(timeout);} // and set an alarm for timeout seconds
```
btovar, line 121: expand to three lines ( if {..., alarm,   } )

dttools/src/process.c
```diff
@@ -98,6 +98,66 @@ struct process_info *process_waitpid( pid_t pid, int timeout)
...
+		num_attempts++;
+		if (num_attempts == max_attempts) // change to sending SIGKILL after max_attempts tries
```
btovar, line 123: move brace to previous line, same for all the rest of the function

btovar: Also, `if(num_attempts >= max_attempts)`

dttools/src/process.c
```diff
@@ -98,6 +98,66 @@ struct process_info *process_waitpid( pid_t pid, int timeout)
 	return 0;
 }
 
+int process_kill_waitpid(pid_t pid, int timeout, int max_attempts)
+{
+	int flags = 0;
+        struct sigaction new_action, old_action;
```
btovar, line 104: Is the file function indented uniformly? The rest of the file uses only tabs.

btovar (changes requested):

dttools/src/process.c
```diff
@@ -98,6 +98,66 @@ struct process_info *process_waitpid( pid_t pid, int timeout)
...
+			if (errno == EINTR) // this is the case that the alarm woke us up so we try again to signal
+			{
+				continue;
+			}
+			else // if there is any other error, it is fatal and we should just exit
+			{
+				break;
+			}
+		}
+		else if (return_pid == 0) // if the return value is 0, it means we set the timeout to zero and nothing changed, so we try to signal and wait for the process again
```
btovar, line 143: Doesn't this trigger a busy-wait? I.e.. If the timeout is zero, and waitpid is no hang, then if the processes has not changed, this loop will be executed over and over again with no pause in between.

btovar: I think it should be ok to always require `process_kill_waitpid` to have a positive timeout and use waitpid without the no hang flag. Also, I'm not sure about the re-sending of signals until max_attempts. Shouldn't we just send the signal once and then do the sigkill?  What worries me is that a processes may modify its cleanup when receiving the same signal in succession (e.g. first signal triggers cleanup, second or more terminates immediately).

btovar (changes requested):

David-Simonetti-ND:
I reworked process_kill_waitpid to use process_waitpid because there may be cases where a given child has already return and is in the completed process list. It now only sends SIGTERM once, waits for timeout seconds, and then sends SIGKILL and attempts to wait for another timeout seconds. At the end it just gives up.
I also added functionality in the work_queue_worker for checking if the coprocess specified in the command line has stopped running and restarting it if this is the case.

btovar:
Looks good. Ready to merge?

btovar (approved):

PR #2828, opened 2022-03-25
Pipes added to coprocess and limit to coprocess restarts

David-Simonetti-ND:
I added a pipe connected to the stdin and stdout of the coprocess because 
1: the coprocess should be cut off from reading in from the stdin of the worker or outputting to the stdout of the worker
2: it might prove helpful to have a method for the worker to communicate with the coprocess
Also added a check that the coprocess is currently running and a max limit of 10 restarts to the coprocess so that the worker would not continuously restart it.

### DIFF
--- a/work_queue/src/work_queue_worker.c
+++ b/work_queue/src/work_queue_worker.c
@@ -2504,7 +2504,7 @@ struct list *parse_manager_addresses(const char *specs, int default_port) {
 void start_coprocess() {
 
 	if (pipe(coprocess_in) || pipe(coprocess_out)) { // create pipes to communicate with the coprocess
-		debug(D_WQ, "couldn't create coprocess pipes: %s\n", strerror(errno));
+		fatal("couldn't create coprocess pipes: %s\n", strerror(errno));
 		return;
 	}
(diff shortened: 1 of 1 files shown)


work_queue/src/work_queue_worker.c
```diff
@@ -2498,27 +2501,76 @@ struct list *parse_manager_addresses(const char *specs, int default_port) {
 	return(managers);
 }
 
-void start_coprocess()
-{
+void start_coprocess() {
+
+	if (pipe(coprocess_in) || pipe(coprocess_out)) { // create pipes to communicate with the coprocess
+		debug(D_WQ, "couldn't create coprocess pipes: %s\n", strerror(errno));
```
btovar, line 2507: I think we want this one to be fatal(), rather than debug().

btovar (changes requested):

work_queue/src/work_queue_worker.c
David-Simonetti-ND: Change is pushed.

btovar (approved):

PR #2829, opened 2022-03-28
Tcp serverless refactor

nicklocascio:
@btovar Here is the complete refactor to TCP. Everything is working correctly with the new coprocess option! There are some stray commits in there from when I pulled David's branch from his forked copy, but I made sure that I merged in the main master branch so everything is up to date. The last commit in the list is all of my changes

### DIFF
--- a/work_queue/src/work_queue_process.c
+++ b/work_queue/src/work_queue_process.c
@@ -234,14 +234,14 @@ static char * invoke_coprocess_function(char *input) {
 	timestamp_t curr_time = timestamp_get();
 	time_t stoptime = curr_time + timeout;
 
-	bool connected = false;
+	int connected = 0;
 	struct link *link;
 	int tries = 0;
 	// retry connection for ~30 seconds
 	while(!connected && tries < 30) {
 		link = link_connect(addr, port, stoptime);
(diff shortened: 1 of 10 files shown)


work_queue/src/work_queue_process.c
```diff
@@ -229,28 +231,58 @@ static char * invoke_coprocess_function(char *input) {
 		fatal("could not lookup address of localhost");
 	}
 
-	struct datagram *dgram = datagram_create(DATAGRAM_PORT_ANY);
-	if(!dgram) {
-		fatal("could not create datagram: %s", strerror(errno));
+	timestamp_t curr_time = timestamp_get();
+	time_t stoptime = curr_time + timeout;
+
+	bool connected = false;
```
btovar, line 237: Let's change this one to  `int connected = 0`, and use `1` for true, to make it uniform with the rest of the code.

btovar (changes requested):
Looks good! Just a minor change, but it should be ready to merge.

work_queue/src/work_queue_process.c
nicklocascio: Done!

Issue #2770, opened 2021-12-08
poncho_package_create:  local pip --editable ignored if package has a version in the spec.


btovar:
Example: with an spec that includes

```json
"pip": [
        "coffea"
        ]
```
if coffea is locally installed (e.g. --editable), then the local version is added to the final package, otherwise the latest stable version from pip is installed. However, with:

```json
"pip": [
        "coffea==0.7.11"
        ]
```

the local version is not found, and the stated version is installed. Since --editable packages takes precedence over any explicit stated version (e.g., because we are developing for such package), the expected behavior is for the --editable package to be included. I think the problem is that we are taking the whole "coffea==0.7.11" as the name of the package, rather than just "coffea".

(Related to this, the message that warns about editable packages not present in the spec is missing a `.format(pip_name)`)

PR #2835, opened 2022-04-07
poncho_package_create: ignore version when looking at locally pip packages

btovar:
For #2770


BarrySlyDelgado (approved):
This looks good.

PR #2840, opened 2022-04-08
Bugfix worker segfault

nicklocascio:
@btovar We just discovered that flushing the info from network function 3 times would result in some info making to the worker but sometimes info would be left behind, thus resulting in the segfault that we were seeing in the worker. This quick fix to flush all the info at once appears to have solved the problem, we tested it with a lot of workers

### DIFF
--- a/work_queue/src/network_function.py
+++ b/work_queue/src/network_function.py
@@ -13,18 +13,23 @@ def function_handler(event):
 	return int(event["a"]) + int(event["b"])
 
 def main():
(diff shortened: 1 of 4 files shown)


btovar:
Sounds good. Could you rebase this pr to the latest master?

nicklocascio:
I've actually never done that before so I'm not sure the steps do so. Could you walk me through it or is it possible for you to do it?

PR #2854, opened 2022-05-12
move coprocess to its own .h.c

btovar:
E.g.
```
t = Task("my_sum")
t->specify_coprocess("my_coprocess")
```

will call the function "my_sum" in the coprocess with the "my_coprocess" name. Tasks are only sent to workers with matching coprocesses.

Run `work_queue/src/coprocess_example.py` as an example. (The file works both as the coprocess and the wq manager.)

### DIFF
--- a/work_queue/src/Makefile
+++ b/work_queue/src/Makefile
@@ -13,6 +13,7 @@ SOURCES_LIBRARY = \
 
 SOURCES_WORKER = \
 	work_queue_process.o \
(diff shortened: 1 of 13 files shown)


David-Simonetti-ND:
Looks good to me! Is the coprocess_example purely a placeholder for now, because I can not seem to find that module.

btovar:
@David-Simonetti-ND No, my mistake. Added now.

Issue #2863, opened 2022-05-24
Support specifying path to gnuplot in makeflow_graph_log


sghuang19:
`makeflow_graph_log` and `work_queue_graph_log` both plots the log with `gnuplot`. `makeflow` only checks `gnuplot` in `PATH`, while `work_queue` only accepts the path to `gnuplot` with `-c` option.

Use `gnuplot` in `PATH` by default, use others if specified could be better.

sghuang19:
#2864

PR #2864, opened 2022-05-24
Add options in makeflow_graph_log to support specifying paths for gnuplot and convert

sghuang19:
```
makeflow_graph_log [options] <makeflowlog> <imagefile>
```

Use `-g` to specify path to `gnuplot`, `-c` to specify path to `convert`. `PATH` is used by default.

### DIFF
--- a/makeflow/src/makeflow_graph_log
+++ b/makeflow/src/makeflow_graph_log
@@ -29,11 +29,11 @@ do
 		;;
 	-g)
 		shift
-		gnuplot=$1
+		gnuplot="$1"
 		;;
 	-c)
 		shift
-		convert=$1
+		convert="$1"
 		;;
 	*)
 		break
@@ -50,7 +50,7 @@ fi
 
 infile=$1
 outfile=$2
-tempfile=$outfile.tmp
+tempfile="${outfile}.tmp"
 output_format=${outfile##*.}
 
 # check for the necessary programs in the path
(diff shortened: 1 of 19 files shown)


sghuang19:
#2863

makeflow/src/makeflow_graph_log
```diff
@@ -47,7 +61,7 @@ do
    then
 	  true
    else
-	  echo "$0: could not find $prog in your PATH"
+	  echo "$0: could not find $prog in the PATH variable or in the specified path"
```
btovar, line 64: Let's change this whole if to:
```sh
if ! which $prog >& /dev/null
then
   echo ...
  exit 1
fi
```

makeflow/src/makeflow_graph_log
```diff
@@ -10,17 +10,31 @@
...
 while [ $# -gt 0 ]
 do
 	case $1 in
 	-h |--help )
 		show_help
 		exit 1
 		;;
+	-g)
+		shift
+		gnuplot=$1
```
btovar, line 32: "$1"

makeflow/src/makeflow_graph_log
```diff
@@ -10,17 +10,31 @@
...
+		;;
+	-c)
+		shift
+		convert=$1
```
btovar, line 36: "$1"

btovar (changes requested):

btovar (changes requested):

PR #2865, opened 2022-05-24
Addition of remotemap, remotepair, remotetree_reduce

David-Simonetti-ND:
Cleaned up version of Brett's PR #2831 adding the remote versions of map, pair, and tree reduce.


work_queue/src/bindings/python3/work_queue.binding.py
```diff
@@ -1917,7 +1917,136 @@ def tree_reduce(self, fn, seq, chunk_size=2):
...
+
+    # map, but remote
+    def Rmap(self, fn, array, name, chunk_size=1):
+        size = math.ceil(len(array)/chunk_size)
+        results = [None] * size
+        tasks = {}
+
+        for i in range(size):
+            start = i*chunk_size
+            end = start + chunk_size
```
btovar, line 1929: Let's change this one to:
```python
end = min(len(array), start+chunk_size)
```
Then you don't need the `if` below. (use the else branch)

work_queue/src/bindings/python3/work_queue.binding.py
```diff
@@ -1917,7 +1917,136 @@ def tree_reduce(self, fn, seq, chunk_size=2):
             seq = results
 
         return seq[0]
-           
+
+    # map, but remote
+    def Rmap(self, fn, array, name, chunk_size=1):
```
btovar, line 1922: Let's rename all of these to `remote_` (e.g. `remote_map`), as an initial capital letter is meant for classes, no functions.

btovar (changes requested):

work_queue/src/bindings/python3/work_queue.binding.py
```diff
@@ -1918,22 +1918,28 @@ def tree_reduce(self, fn, seq, chunk_size=2):
...
+    ##
+    # Maps a function to elements in a sequence using work_queue remote task 
+    #
+    # Similar to regular map function in python, but creates a task to execute each function on a worker running a coprocess
+    #
+    # @param self       Reference to the current work queue object.
+    # @param fn         The function that will be called on each element. This function exists in coprocess.
+    # @param seq        The sequence that will call the function
+    # @param coprocess  The name of the coprocess that contains the function fn.
+    # @param name       This defines the key in the event json that wraps the data sent to the coprocess.
```
btovar, line 1930: For these functions we are in charge of the coprocess, right? So, `fn` is a user function that does not know about "event".  This means that we don't need the "name" argument, and we can do something like:

```python
    def remote_map(self, fn, array):
                ...
                 event = json.dumps({"args" : array[start:]})
                 def fn_wrap(event, response):
                       try:
                              response["Result"] = fn(event["args"])
                              response["StatusCode"] = 200
                      except Exception as e:
                              response["StatusCode"] = 500
                     p_task = RemoteTask(fn_wrap, event, name_of_wq_standard_coprocess)           
```

Since we need to do this wrapping for all the functions, we may want to create a decorator with it. This is a good tutorial on decorators: https://www.datacamp.com/tutorial/decorators-python

David-Simonetti-ND: I don't quite understand which functions should be decorated. Is the idea that remote_map, remote_pair, and remote_tree_reduce will all be decorated by one fn_wrap decorator?

David-Simonetti-ND: fn is just a string in all the python abstraction functions, so would the decorator have to live in the network function side?

btovar: The user functions `fn` are the ones being decorated. E.g.,

```python
def wrap_fn_for_remote(fn):
    def wrapped_fn(event, response):
        try:
            response["Result"] = fn(event["args"])
            response["StatusCode"] = 200
        except Exception as e:
            response["StatusCode"] = 500
    return wrappend_fn

def remote_map(self, fn, array):
    ...
    event = json.dumps({"args" : array[start:]})
    p_task = RemoteTask(wrap_fn_for_remote(fn), event, name_of_wq_standard_coprocess)
    ...
```

btovar: Ah, you are right. I forgot `fn` is just a string. If you are ok with, I can merge as it is right now, but I would like to later talk about how are we passing arguments. This event dict feels a little clunky, as I don't like that we have to pass the "name" argument to the `remote_` functions.

PR #2871, opened 2022-05-26
Fix for dall bug

David-Simonetti-ND:
I think this is an elegant solution for the dall bug. It allows for the coprocess output to be accessed in the exact same way as a regular task, but also allows wq debug output to still be output to the screen for the forked work_queue_worker process that communicates with the coprocess.


btovar:
Nice, ready to merge?

PR #2875, opened 2022-05-27
remote_execute decorator and switch to explicit fork()

David-Simonetti-ND:
Reworks the decorator to also wrap up the output of the user function into the response dict, and allows the user to specify the remote function arguments when creating the RemoteTask. Also switched from the multiprocess library to explicitly forking a process to run the function.

### DIFF
--- a/work_queue/src/network_function.py
+++ b/work_queue/src/network_function.py
@@ -48,32 +48,47 @@ def main():
                 input_spec = conn.recv(size).decode('utf-8').split()
                 function_name = input_spec[0]
                 event_size = int(input_spec[1])
+            try:
+                if event_size:
+                    response = ""
+                    # receive the event itself
+                    event = conn.recv(event_size)
+                    event = json.loads(event)
+                    print('Network function: recieved event: {}'.format(event), file=sys.stderr)
+                    # create a forked process for function handler
+                    p = os.fork()
+                    if p == 0:
+                        response = getattr(wq_worker_coprocess, function_name)(event)
+                        os.write(write, json.dumps(response).encode("utf-8"))
+                        os._exit(-1)
+                    elif p < 0:
+                        print('Network function: unable to fork', file=sys.stderr)
+                        response = { 
+                            "Result": "unable to fork",
+                            "StatusCode": 500 
+                        }
+                    else:
+                        chunk = os.read(read, 65536).decode("utf-8")
+                        all_chunks = [chunk]
+                        while (len(chunk) >= 65536):
+                            chunk = os.read(read, 65536).decode("utf-8")
+                            all_chunks.append(chunk)
+                        response = "".join(all_chunks).encode("utf-8")
+                        os.waitid(os.P_PID, p, os.WEXITED)
+                    
+                    response_size = len(response)
 
-            if event_size:
-                # receive the event itself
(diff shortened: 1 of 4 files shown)


work_queue/src/network_function.py
```diff
@@ -50,18 +52,17 @@ def main():
...
                 # create a forked process for function handler
-                manager = multiprocessing.Manager()
-                response = manager.dict()
-                p = multiprocessing.Process(target=getattr(wq_worker_coprocess, function_name), args=(event, response))
-                p.start()
-                p.join()
-
-                response = json.dumps(dict(response))
+                p = os.fork()
+                if p == 0:
```
btovar, line 59: If the fork fails, the os.read below may wait forever.  Let's do the full `if p == 0  elif p < 0 else`.

work_queue/src/network_function.py
```diff
@@ -50,18 +52,17 @@ def main():
...
+                    response = getattr(wq_worker_coprocess, function_name)(event)
+                    os.write(write, json.dumps(response).encode("utf-8"))
+                    os._exit(-1)
+                response = os.read(read, 9999999999)
```
btovar, line 63: Here let's put the os.read in a loop that reads 64K  (65536) at a time, as we do in link.c, until os.read returns the empty string.  Put each read result in a list an concatenate at the end.

work_queue/src/network_function.py
```diff
@@ -50,18 +52,17 @@ def main():
             if event_size:
```
btovar, line 52: Put this `if event_size` inside a  `try ... except ...` block, so that the coprocessor does not die if something goes wrong reading the response, or forking, etc.

btovar (changes requested):

work_queue/src/network_function.py
David-Simonetti-ND: because we are reading from a pipe and not a file, if we try to read when there is nothing left in the pipe then python blocks instead of returning the empty string. As a result, I read chunks from the pipe until we got a chunk that was not 64k (indicating it has the last of the data)

David-Simonetti-ND:
It appears that the max length json that work_queue_coprocess could handle receiving was 1024 bytes (was giving me no output because the process was segfaulting), so I changed it to dynamically allocate the size of the output buffer depending on the length sent

David-Simonetti-ND:
I added a feature where the user can specify the execution method they want for the remote task. they must specify the method argument when creating a remote task to either thread, fork, or direct depending on the execution method they desire. Defaults to fork if the user specifies nothing. 
For example,
t = wq.RemoteTask("my_sleep", "my_coprocess_example", t=sleep_time, method="fork")

David-Simonetti-ND:
Added two example functions to one. tensorflow_import just calls a quick tensorflow function, and the idea behind this is to see how much time RemoteTask saves on imports compared to PythonTask. Tests of 10 tasks show RemoteTask taking 1 second to complete, but PythonTask took 30 seconds because of having to reimport tensorflow 10 times. 
The other example function tensorflow_train trains a basic tensorflow model. The RemoteTask verison runs in roughly 120 seconds, while the PythonTask version runs in roughly 160 seconds.

btovar (approved):

PR #2879, opened 2022-05-31
wq: add input and output lines in the transaction log

sghuang19:
I'm not sure how to test it, but the codes are done.

### DIFF
--- a/work_queue/src/work_queue.c
+++ b/work_queue/src/work_queue.c
@@ -1405,7 +1405,7 @@ static work_queue_result_code_t get_output_file( struct work_queue *q, struct wo
 
 		debug(D_WQ, "%s (%s) sent %.2lf MB in %.02lfs (%.02lfs MB/s) average %.02lfs MB/s", w->hostname, w->addrport, total_bytes / 1000000.0, sum_time / 1000000.0, (double) total_bytes / sum_time, (double) w->total_bytes_transferred / w->total_transfer_time);
 
-        write_transaction_transfer(q, w, t, total_bytes, sum_time, f->remote_name, 0);
+        write_transaction_transfer(q, w, t, total_bytes, sum_time, f->remote_name, WORK_QUEUE_OUTPUT);
 	}
 
 	// If the transfer was successful, make a record of it in the cache.
@@ -3390,7 +3390,7 @@ static work_queue_result_code_t send_input_file(struct work_queue *q, struct wor
 		q->stats->bytes_sent += total_bytes;
 
 		// Write to the transaction log.
-		write_transaction_transfer(q, w, t, total_bytes, elapsed_time, f->remote_name, 1);
+		write_transaction_transfer(q, w, t, total_bytes, elapsed_time, f->remote_name, WORK_QUEUE_INPUT);
 
 		// Avoid division by zero below.
 		if(elapsed_time==0) elapsed_time = 1;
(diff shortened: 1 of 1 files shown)


sghuang19:
#2879

work_queue/src/work_queue.c
```diff
@@ -1403,6 +1404,8 @@ static work_queue_result_code_t get_output_file( struct work_queue *q, struct wo
 		w->total_transfer_time += sum_time;
 
 		debug(D_WQ, "%s (%s) sent %.2lf MB in %.02lfs (%.02lfs MB/s) average %.02lfs MB/s", w->hostname, w->addrport, total_bytes / 1000000.0, sum_time / 1000000.0, (double) total_bytes / sum_time, (double) w->total_bytes_transferred / w->total_transfer_time);
+
+        write_transaction_transfer(q, w, t, total_bytes, sum_time, f->remote_name, 0);
```
btovar, line 1408: Instead of 0, let's use:  `WORK_QUEUE_OUTPUT`

work_queue/src/work_queue.c
```diff
@@ -3386,6 +3389,9 @@ static work_queue_result_code_t send_input_file(struct work_queue *q, struct wor
 
 		q->stats->bytes_sent += total_bytes;
 
+		// Write to the transaction log.
+		write_transaction_transfer(q, w, t, total_bytes, elapsed_time, f->remote_name, 1);
```
btovar, line 3393: As above, change 1 to `WORK_QUEUE_INPUT`

work_queue/src/work_queue.c
```diff
@@ -7543,6 +7549,25 @@ static void write_transaction_worker_resources(struct work_queue *q, struct work
 }
 
 
+static void write_transaction_transfer(struct work_queue *q, struct work_queue_worker *w, struct work_queue_task *t, size_t sizeinmb, int walltime, char* filename, int type){
+	struct buffer B;
+	buffer_init(&B);
+	buffer_printf(&B, "TRANSFER ");
+	if (type)
```
btovar, line 7556: Use braces for both the `if` and `else`, even it it is only one statement.
Also, now use
```
if(type == WORK_QUEUE_INPUT) {  
    ...
```

work_queue/src/work_queue.c
```diff
@@ -7543,6 +7549,25 @@ static void write_transaction_worker_resources(struct work_queue *q, struct work
 }
 
 
+static void write_transaction_transfer(struct work_queue *q, struct work_queue_worker *w, struct work_queue_task *t, size_t sizeinmb, int walltime, char* filename, int type){
```
btovar, line 7552: Note that size and walltime come in bytes, and microseconds, respectively.  I'd suggest changing the parameters to:
```C
..., size_t size_in_bytes, int time_in_usecs, work_queue_file_type_t type) {
```

and convert to MB and seconds as needed. Use "%f" when printing the numbers below.

btovar (changes requested):

work_queue/src/work_queue.c
btovar: For memory, divide by `((double) MEGABYTE)`.  For time, divide by `((double) USECOND)`, and use "%f" in buffer_print for both.

btovar (approved):

PR #2880, opened 2022-06-02
activate env before conda-unpack

btovar:
Needed because conda-unpack looks for a 'python' in PATH, but rh8 has
none.


BarrySlyDelgado (approved):

BarrySlyDelgado:
Everything looks good to me.

PR #2881, opened 2022-06-03
Small fixes for running workers with factory

David-Simonetti-ND:
This fixes some incorrect handling of systemcall return values.

### DIFF
--- a/work_queue/src/work_queue_coprocess.c
+++ b/work_queue/src/work_queue_coprocess.c
@@ -34,7 +34,12 @@ int work_queue_coprocess_write(char *buffer, int len, int timeout)
 	int poll_result = poll(&read_poll, 1, timeout);
 	if (poll_result < 0)
 	{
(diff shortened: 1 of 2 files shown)


work_queue/src/work_queue_coprocess.c
```diff
@@ -98,6 +118,10 @@ char *work_queue_coprocess_setup(int *coprocess_port)
 		if (curr_bytes_read < 0) {
 			fatal("Unable to get information from coprocess\n");
 		}
+		if (curr_bytes_read == 0)
```
btovar, line 121: Could `4096 - buffer_offset` (e.g. the len argument for w_q_coprocess_{read,write}) ever be zero? If so, I would make it so that `curr_bytes_read` returns -1 when a retry is needed, and > -1 on failure.  Otherwise if 0 is the correct number of bytes read, this may fail

PR #2888, opened 2022-06-07
poncho_package_serverize to create network functions

David-Simonetti-ND:
This pull request contains a variety of things
1) poncho_package_serverize depends on poncho_package_create and poncho_package_analyze in order to run. The solution to this is to create a poncho python package that contains all the poncho functions. This will also allow greater flexibility later in creating remote functions (for example the manager script can import poncho_package_serverize and create the network function before submitting tasks). The Makefile creates a folder called poncho in the $(CCTOOLS_PYTHON3_PATH) folder and places all the poncho programs in there along with a setup.py.
2) poncho_package_analyze has a create_spec function that returns the conda packages that a file/function depend on 
3) poncho_package_serverize itself was changed to be an executable script which accepts a source input file, a list of functions to put in the network function, an output file path for the network function and an optional argument to specify the path where an environment tarball will be created that has all the dependencies required to run the network function. This allows the creation of an environment to be done once (since it is so slow) and small changes to the function the user wishes to run (without changing imports) means only the network function has to be remade.
For example, the first time they create the network function the user would run
`poncho_package_serverize --source INPUT_FILE --function FUNCTION_A --function FUNCTION_B --out OUTPUT_FILE --pack OUTPUT_ENV`
But after they have created the environment, if they don't make changes to any imports and only change FUNCTION_A/B, then they can run
`poncho_package_serverize --source INPUT_FILE --function FUNCTION_A --function FUNCTION_B --out OUTPUT_FILE
`In order to only regenerate the network function.
The creation of the network function works by generating an ast of the python code for the users function, looking for all imports and any of the functions the user specified in the input. It then will paste the imports and the functions it found into an output network function.


David-Simonetti-ND:
Also, for right now the network code is just kept in a string inside of poncho_package_serverize, but I am wondering if there is a better place to keep it.

btovar (approved):

btovar:
Anything that is copied into the prefix installation dir (from ./configure --prefix dir) is copied to the final package. If you need to install to site-packages, look at how we do it in `work_queue/src/bindings/python3/Makefile`.  Since most of the code is not python, we do not use `setup.py` to install cctools.

David-Simonetti-ND:
Since the code for the network function lives in poncho_package_serverize now, I figured it would be messy to have two locations with the identical code so I removed the one in the work_queue directory

dthain:
@David-Simonetti-ND @btovar what's the status on this one?

David-Simonetti-ND:
My last commit includes the changes @btovar requested but I never pinged him to review it again. Should be close to complete.

btovar (approved):

PR #2889, opened 2022-06-08
Fix transfer log

sghuang19:
- Add log synopsis for transfer log
- Replace worker_id with worker_addrport for transfer log


btovar:
It seems that the only the header of the log was modified. You also want to delete line 7577.

btovar:
Ready to merge?

sghuang19:
> Ready to merge?

Yep

Issue #1099, opened 2016-01-15
Ability to drain a Work Queue Factory


klannon:
I'm not sure if this is even a sensible request, but it would be really nice if you could "drain" a work queue factory--in other words, tell a factory that it should reduce the number of connected workers, but only by removing workers after they complete their current tasks.

dthain:
That is a great idea!

klannon:
You can probably even guess why it occurred to me.  :)

dthain:
@charleszheng44 please add a new capability to `work_queue_factory` so that if `drain:true` is set in the configuration file, then workers are allowed to exit, but are not actively removed.  Put that together with `max-workers:0` and you get a system that will shut itself down slowly without killing jobs...

charleszheng44:
Did you mean that in order to slowly decreased the number of workers, we keep decrease the value of max-workers to 0?

btovar:
Look at work_queue_shut_down_workers(q, n);
It releases at most n workers, but it does not release a worker if it is running a task.

Call work_queue_shut_down_workers in(q, n); in work_queue_factory when drain is true, and max-workers is smaller than the current active workers.

dthain:
Er, Ben, I think you are confused: `work_queue_shutdown_workers` is at the master program, not in the factory.

btovar:
Ah, indeed. 

The behavior that you describe is already there, without the need of drain: true. Simply set max_workers and min_workers to 0, workers are not regenerated, and the idle out as the master sends less work.

However, if you do not like the waiting for the idling out, then I do not see how we can do this from work_queue_factory alone.

klannon:
This is not quite what I had in mind.  Currently, I have let's say 2k 
workers in a factory.  I'm told by the university that I'm breaking the 
internet and I need to scale back.  So, I'd like to go to 1k workers 
with the least amount of wasted work, but as quickly as possible.  
Currently, it seems that if I set max workers to 1k, the factory doesn't 
launch any new jobs, but the workers only seem to be removed by 
eviction.  To have them terminate faster, but still not lose any work, 
it would be nice if the factory could tell them, "please disconnect 
after you finish your next task."  (And of course, only tell this to 
enough workers that I go from 2k running to 1k running.  Note, for me, 
it would be fine if it got close to 1k but not quite--i.e. when it's 
removing workers, if it accidentally remove a few too many, I wouldn't 
be upset.)  So, the master keeps sending work, but we shrink the pool of 
[...]

PR #2912, opened 2022-06-30
wq: work_queue_factory limits max workers

sghuang19:
As discussed #1099, add support for "draining" a work_queue_factory.

- [x]  Work Queue Factory includes `FACTORY_NAME` to indicate where the workers come from.
    - [x]  A command line argument for `work_queue_factory` to specify `factory_name`.
    - [x]  A command line argument `--from-factory <factory_name>` for `work_queue_worker`.
    - [x]  In `read_config_file`
    - [x]  Work Queue Factory launches the workers with `--from-factory=..` argument.
- [x]  Factory reports this `workers_max` to the catalog server
- [x]  Manager reads the info and take actions
    - [x]  Manager reads `workers_max` from the catalog server.
    - [x]  Manager maintains a dictionary of factory structs.
    - [x]  Manager sends shutdown signal to excessive workers in the factory that are not running any tasks.
    - [x]  Do not dispatch tasks to workers until currently connected workers is less then `workers_max`
    - [x]  Whenever the last task on the worker is returned, shut down that worker.


sghuang19:
```mermaid
graph LR
a[Manager reads catalog] --"factory:max_workers"-->
b[Maintaining a<br>dict of factories]
c[Iterate through workers] --"factory:cur_workers"--> b
```

- [ ] Every time we update the catalog, trigger the dict update
- [ ] In updating this dict, if the limit is exceeded, immediately shutdown the idle workers.
- [ ] In dispatching the tasks, refer to the dict and don't give tasks to worker in draining factories.
- [ ] In returning of the final task, refer to the dict for whether shutting down the worker.

If we adopt this strategy, the "accuracy" (i.e. shutting down more workers than expected) of draining would be determined by how frequently we update the catalog.

Besides, this dictionary of factories might be also used for other future usages, so it could be good to invent it extensible. For example, use JX structure?

@btovar It's getting tricky, what do you think?

sghuang19:
#2919  Introduces catalog reading.

btovar:
The dictionary sounds like a good idea. I wouldn't do it as a jx, though, let's do a proper `struct`.

work_queue/src/work_queue.c
```diff
@@ -278,6 +278,9 @@ struct work_queue_worker {
 struct work_queue_factory_info {
 	char *name;
 	int   connected_workers;
+	int   max_workers;
+	int   queried;
```
btovar, line 282: `queried` gives me the idea that we talked directly to the factory to ask for something.
maybe `seen_at_catalog`?

work_queue/src/work_queue.c
```diff
@@ -851,31 +847,76 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
 }
 
-static void remove_factory(struct work_queue_factory_info *f)
+static int factory_trim_workers(struct work_queue *q, struct work_queue_factory_info *f)
+{
+	// Iterate through all workers and shut idle ones down
+	char *factory_name = f->name;
+	struct work_queue_worker *w;
+	char *key;
+	int i = 0;
```
btovar, line 856: int trimmed_workers = 0;  instead of i

work_queue/src/work_queue.c
```diff
@@ -851,31 +847,76 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
+			hash_table_nextkey(q->worker_table, &key, (void **) &w) ) {
+		if( !strcmp(factory_name, w->factory_name) &&
+			!itable_size(w->current_tasks) ) {
+			shut_down_worker(q, w);
+			i++;
+		}
+	}
+
+	// Trimmed some but not enough workers.
+	if (f->connected_workers > f->max_workers) f->trimmed = 1;
```
btovar, line 869: f->trimmed should be set to 1, regardless.

work_queue/src/work_queue.c
```diff
@@ -851,31 +847,76 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
+	debug(D_WQ, "Trimmed %d workers from %s", i, factory_name);
+	return i;
+}
+
+static struct work_queue_factory_info* create_factory_info(struct work_queue *q, const char *name)
```
btovar, line 874: `static struct work_queue_factory_info *create_factory_info`
Rest of the code usually puts the * next to the name of function.

work_queue/src/work_queue.c
```diff
@@ -851,31 +847,76 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
-		}
-		f->name = xxstrdup(name);
-		f->connected_workers = 0;
-		hash_table_insert(q->worker_table, f->name, f);
+		debug(D_WQ, "factory %s not recorded", name);
+		return;
 	}
 
-	return;
+	f->max_workers = jx_lookup_integer(j, "max_workers");
```
btovar, line 912: What if factory did not send max_workers?

work_queue/src/work_queue.c
```diff
@@ -851,31 +847,76 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
+	f->queried = 1;
+	// Trim workers if appropriate.
+	if (f->connected_workers > f->max_workers && !f->trimmed) {
```
btovar, line 915: We can simplify to `if(!f->trimmed)`, as the other condition is checked in `factory_trim_workers`.

work_queue/src/work_queue.c
```diff
@@ -914,6 +956,16 @@ void update_read_catalog_factory(struct work_queue *q, time_t stoptime) {
 	}
 
 	jx_delete(jexpr);
+
+	// Remove outdated factories
+	hash_table_firstkey(q->factory_table);
+	while ( hash_table_nextkey(q->factory_table, factory_name, (void **) f) ) {
+		if ( !(*f)->queried && !(*f)->connected_workers ) {
```
btovar, line 963: `if( !f->seen_at_catalog && f->connected_workes < 1) {`

work_queue/src/work_queue.c
```diff
@@ -893,8 +934,9 @@ void update_read_catalog_factory(struct work_queue *q, time_t stoptime) {
 
 	hash_table_firstkey(q->factory_table);
 	while ( hash_table_nextkey(q->factory_table, factory_name, (void **) f) ) {
```
btovar, line 936: Check other uses of `hash_table_nextkey`. You are missing the & in both key and value.

work_queue/src/work_queue.c
```diff
@@ -914,6 +956,16 @@ void update_read_catalog_factory(struct work_queue *q, time_t stoptime) {
 	}
 
 	jx_delete(jexpr);
+
+	// Remove outdated factories
+	hash_table_firstkey(q->factory_table);
+	while ( hash_table_nextkey(q->factory_table, factory_name, (void **) f) ) {
```
btovar, line 962: Same as above

work_queue/src/work_queue.c
```diff
@@ -914,6 +956,16 @@ void update_read_catalog_factory(struct work_queue *q, time_t stoptime) {
...
+		if ( !(*f)->queried && !(*f)->connected_workers ) {
+			hash_table_remove(q->factory_table, *factory_name);
```
btovar, line 964: `hash_table_remove(q->factory_table, factory_name);`

work_queue/src/work_queue.c
```diff
@@ -914,6 +956,16 @@ void update_read_catalog_factory(struct work_queue *q, time_t stoptime) {
...
+			remove_factory_info(q, *factory_name);
```
btovar, line 965: As above, that `*factory_name` looks suspicious.

btovar (changes requested):

work_queue/src/work_queue.c
sghuang19: The notion here is:

- After trimming the workers, if the we succeed to reduce the number of workers as desired, we consider this round of draining is completed. So we should actually restore `trimmed` to 0.
- If we didn't succeed to reduce that many workers, we set `trimmed = 1`, so that in this round of draining, we won't attempt to trim it again.

In short, `trimmed` is just for a single round of draining. If the draining is completed, we should set it back to 0.

work_queue/src/work_queue.c
btovar: It communicates that connected_workers is not considered a boolean, but that it is a counter.

work_queue/src/work_queue.c
sghuang19: Consider the following case:

`connected_workers = 15`, `6` of them are idle, `max_workers = 10`. This will trigger draining, and after trimming, `10` workers are left, `1` of them is still idle.

Then, one more worker connected and starts to work. `max_workers` is not changed, but this will trigger draining again.

If we use `if (f->connected_workers > f->max_workers) f->trimmed = 1;`, `trimmed` will not be set in first round of trimming, and we are able to trim one more idle worker in the second round.

If we use `f->trimmed = 1;`, `trimmed` will be set in the first round, thus we won't be able to trim that one last idle worker in the second round.

btovar: Ah good point. I would like to eliminate any such hidden interactions by reducing the state we keep, so I suggest:
- Remove f->trimmed completely.
- Trigger a factory trim only when max_workers is reduced for that factory.
- When a new worker sends the "info factory..." message, immediately shutdown the worker if `connected_workers >= max_workers`
- Do not schedule tasks to workers of factories with `connected_workers > max_workers`
- On task retrieval, if no other current task in the worker, shutdown the worker if `connected_workers > max_workers`

sghuang19: Sounds good, the first three points need to be done, last two are already written.

work_queue/src/work_queue.c
```diff
@@ -827,6 +855,127 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
+	// Iterate through all workers and shut idle ones down
+	char *factory_name = f->name;
+	struct work_queue_worker *w;
+	char *key;
+	int trimmed_workers = 0;
+
+	hash_table_firstkey(q->worker_table);
+	while( f->connected_workers > f->max_workers &&
+			hash_table_nextkey(q->worker_table, &key, (void **) &w) ) {
+		if( !strcmp(factory_name, w->factory_name) &&
```
btovar, line 869: `w->factory_name && !strcmp(factory_name, w->factory_name) &&`

strcmp will segfault if w->factory_name is not set otherwise.

work_queue/src/work_queue.c
```diff
@@ -827,6 +855,127 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
+			!itable_size(w->current_tasks) ) {
```
btovar, line 870: `itable_size(w->current_tasks) < 1`

It is counter, not a boolean. Also, we protect against bugs if we ever get a negative number.

work_queue/src/work_queue.c
```diff
@@ -827,6 +855,127 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
 	return timeout;
 }
 
+static int factory_trim_workers(struct work_queue *q, struct work_queue_factory_info *f)
+{
```
btovar, line 859:
```
if(!f || !f->name) {
    return 0;
}
```

work_queue/src/work_queue.c
```diff
@@ -827,6 +855,127 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
+		hash_table_remove(q->factory_table, name);
+	} else {
+		debug(D_WQ, "Failed to remove unrecorded factory %s", name);
+	}
+}
+
+static void update_factory(struct work_queue *q, struct jx *j)
+{
+	const char *name;
+	if ( !(name = jx_lookup_string(j, "factory_name")) ) return;
```
btovar, line 911:
```
name = jx_lookup_string(j, "factory_name");
if(!name) {
   return;
}
```

work_queue/src/work_queue.c
```diff
@@ -827,6 +855,127 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
+	struct work_queue_factory_info *f;
+	if ( !(f = hash_table_lookup(q->factory_table, name)) ) {
+		debug(D_WQ, "factory %s not recorded", name);
```
btovar, line 914: `create_factory_info` instead?

work_queue/src/work_queue.c
```diff
@@ -827,6 +855,127 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
+		return;
+	}
+
+	f->seen_at_catalog = 1;
+	int max_workers = jx_lookup_integer(j, "max_workers");
+	int old_max_workers = f->max_workers;
+	if (max_workers) {
```
btovar, line 921: Won't work if `max_workers` is zero.  Check `jx_lookup_guard`.

work_queue/src/work_queue.c
```diff
@@ -3847,6 +3987,13 @@ static int check_hand_against_task(struct work_queue *q, struct work_queue_worke
 		return 0;
 	}
 
+	struct work_queue_factory_info *f;
+	if ( w->factory_name &&
+			(f = hash_table_lookup(q->factory_table, w->factory_name)) &&
```
btovar, line 3992: Separate these statements.
```
if(w->factory_name) {
  f = ...
  if(f...
```

work_queue/src/work_queue.c
```diff
@@ -4384,6 +4531,15 @@ static int receive_one_task( struct work_queue *q )
 		if( task_state_is(q, taskid, WORK_QUEUE_TASK_WAITING_RETRIEVAL) ) {
 			w = itable_lookup(q->worker_task_map, taskid);
 			fetch_output_from_worker(q, w, taskid);
+			// Shutdown worker if appropriate.
+			struct work_queue_factory_info *f;
+			if ( w->factory_name &&
```
btovar, line 4536: as above

btovar (changes requested):

work_queue/src/work_queue.c
sghuang19: Solving this by making default factory_name `"unknown"`

btovar: I don't recommend setting a default factory name. It will get confusing, and we have to remember that "unknown" is a special value. Let's just check whether w->factory_name is different from NULL.

work_queue/src/work_queue.c
sghuang19: I'm not sure what's the difference between your code and mine?

work_queue/src/work_queue.c
sghuang19: I don't think we need to handle this. All new factories are added when workers connect to the manager. And when we query the catalog, we only get those factories that are already in the table.

work_queue/src/work_queue.c
sghuang19: Fixed. And I am wondering if this could cause some problems in the heap, as we are shutting down workers as we iterate the workers table.

work_queue/src/work_queue.c
```diff
@@ -851,31 +858,88 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
 	return timeout;
 }
 
-static void remove_factory(struct work_queue_factory_info *f)
+static int factory_trim_workers(struct work_queue *q, struct work_queue_factory_info *f)
 {
-	free(f->name);
-	free(f);
+	char *factory_name;
+	if ( !f || !(factory_name = f->name) ) return 0;
```
btovar, line 864: If f->name is null, that's a bug. Let's do:
```
if(!f) {
    return 0;
}

assert(f->name);
```
Below let's use `f->name` instead of `factory_name`. Less things to keep track of.

work_queue/src/work_queue.c
```diff
@@ -851,31 +858,88 @@ static int get_transfer_wait_time(struct work_queue *q, struct work_queue_worker
...
-	if ( !(f = hash_table_lookup(q->factory_table, name)) ) {
-		// Create a new factory
-		debug(D_WQ, "new factory %s detected", name);
-		f = malloc(sizeof(*f));
-		if (!f) {
-			debug(D_NOTICE, "Cannot allocate memory for factory %s.", name);
-			return;
-		}
+	if ( (f = hash_table_lookup(q->factory_table, name)) ) return f;
+	if ( (f = malloc(sizeof(*f))) ) {
```
btovar, line 899: No need to check malloc here. We don't check anywhere else, and if malloc fails, it is not clear what to do anyway.

work_queue/src/work_queue.c
btovar: Mine is easier to read and debug. Assignments inside `if`s are sometimes confused with buggy `==`.

work_queue/src/work_queue.c
btovar: Do not assign inside the `if`. You can make more readable with 
```
struct work_queue_factory_info *f = hash_table_lookup(q->factory_table, name);
if(f) {
 ...

```

work_queue/src/work_queue.c
```diff
@@ -913,7 +978,21 @@ void update_read_catalog_factory(struct work_queue *q, time_t stoptime) {
 		debug(D_WQ, "Failed to retrieve factory info from catalog server(s) at %s.", q->catalog_hosts);
 	}
 
-	jx_delete(jexpr);
+	// Remove outdated factories
+	struct hash_table *outdated_factories = hash_table_create(0, 0);
```
btovar, line 982: It is more efficient and clear to put outdated factories in a list:
```
list_push_tail(outdated_factories, f);

and

while(list_size(outdated_factories) > 0) {
   f = list_pop_head(outdated_factories);
  ...
}
```

work_queue/src/work_queue.c
```diff
@@ -943,23 +1022,23 @@ void update_read_catalog(struct work_queue *q)
...
 		return;
 
 	// If host and port are not set, pick defaults.
 	if(!q->catalog_hosts) q->catalog_hosts = xxstrdup(CATALOG_HOST);
 
 	// Update the catalog.
-	update_write_catalog(q, foreman_uplink);
+	// Only write if we have a name.
+	if (q->name) update_write_catalog(q, foreman_uplink);
 	update_read_catalog(q);
```
btovar, line 1042: Let's mirror what we do for `update_write_catalog`, e.g.:
if(q->fetch_factory) update_read_catalog(q);

work_queue/src/work_queue.c
```diff
@@ -943,23 +1022,23 @@ void update_read_catalog(struct work_queue *q)
 {
 	time_t stoptime = time(0) + 5; // Short timeout for query
 
-	update_read_catalog_factory(q, stoptime);
+	if (q->fetch_factory) {
```
btovar, line 1025: Let's do this check in `update_catalog`, see below.

work_queue/src/work_queue.c
```diff
@@ -5968,7 +6059,8 @@ void work_queue_delete(struct work_queue *q)
 		struct work_queue_factory_info *f;
 		hash_table_firstkey(q->factory_table);
 		while(hash_table_nextkey(q->factory_table, &key, (void **) &f)) {
-			remove_factory(f);
+			free(f->name);
```
btovar, line 6062: Don't you need `remove_factory_info` here? It seems that the memory assigned to `f` leaks here.

btovar (changes requested):

work_queue/src/work_queue.c
sghuang19: I did that inside `update_read_catalog`

```c
	if (q->fetch_factory) {
		update_read_catalog_factory(q, stoptime);
	}
```

So that we could extend this to other usages beyond reading factory conveniently.

work_queue/src/work_queue.c
sghuang19: I should use `remove_factory_info`, but this will not cause mem leak. As far as I know, `hash_table_reomve` will free the memory allocated for the struct, so long as I freed the string inside the struct, this should be fine.

work_queue/src/work_queue.c
btovar: I see. Then let's put if(q->name)  as the first statement of `update_write_catalog`, so that both functions have the same structure.

sghuang19:
All done!

work_queue/src/work_queue.c
btovar: It does not, a free is need.

sghuang19: updated `remove_factory_info` to fix this issue.

PR #2918, opened 2022-07-06
WQ: Multiple Serverless Coprocesses

David-Simonetti-ND:
The worker now accepts an argument for number of coprocess instances to start (defaults to 1 if not specified) which the worker will attempt to start when it is initializing. When sending out RemoteTasks to be done, the worker searches for a ready coprocess instance not currently running a task, and if one is not found the task is put back onto the queue. 
There is also some general refactoring in that all coprocess functions require a coprocess struct to be passed to them, because now there may exist multiple coprocesses on one worker. 
Nothing about resource measurement or enforcement has been implemented yet.


dthain:
@David-Simonetti-ND is all of your work from the summer pushed to this branch?  It looks like it hasn't been updated in a few weeks.

David-Simonetti-ND:
Oops, I didn't push the last couple weeks of work. I'll work on getting that in sometime in the next couple days.

dthain:
It looks like the RemoteTask test is hanging indefinitely in the github test run,
and so it eventually gets cancelled for running too long.

work_queue/src/work_queue_coprocess.c
```diff
@@ -257,51 +211,164 @@ char *work_queue_coprocess_run(const char *function_name, const char *function_i
...
+	{
+		struct work_queue_coprocess *curr_coprocess = &coprocess_info[i];
+		if (curr_coprocess->state == WORK_QUEUE_COPROCESS_DEAD || curr_coprocess->state == WORK_QUEUE_COPROCESS_UNINITIALIZED) {
+			continue;
+		}
+		struct rmsummary *resources = rmonitor_measure_process(curr_coprocess->pid);
+
+		debug(D_WQ, "Measuring resources of coprocess with pid %d\n", curr_coprocess->pid);
+		debug(D_WQ, "cores: %lf, memory: %lf, disk: %lf, gpus: %lf\n",    	resources->cores, 
+																			resources->memory + resources->swap_memory,
```
btovar, line 306: remove empty lines

work_queue/src/work_queue_coprocess.c
```diff
@@ -257,51 +211,164 @@ char *work_queue_coprocess_run(const char *function_name, const char *function_i
...
+																			resources->disk,
+																			resources->gpus);
+		debug(D_WQ, "Max resources available to coprocess:\ncores: %"PRId64 " memory: %"PRId64 " disk: %"PRId64 " gpus: %"PRId64 "\n",  
+																				curr_coprocess->coprocess_resources->cores.total,
```
btovar, line 310: remove empty lines

work_queue/src/work_queue_coprocess.c
```diff
@@ -257,51 +211,164 @@ char *work_queue_coprocess_run(const char *function_name, const char *function_i
...
+int work_queue_coprocess_enforce_limit(struct work_queue_coprocess *coprocess) {
+	if (coprocess == NULL || coprocess->state == WORK_QUEUE_COPROCESS_DEAD || coprocess->state == WORK_QUEUE_COPROCESS_UNINITIALIZED) {
+		return 1;
+	}
+	else if (
+		coprocess->coprocess_resources->cores.inuse  > coprocess->coprocess_resources->cores.total || 
+		coprocess->coprocess_resources->memory.inuse > coprocess->coprocess_resources->memory.total ||
+		coprocess->coprocess_resources->disk.inuse   > coprocess->coprocess_resources->disk.total ||
+		coprocess->coprocess_resources->gpus.inuse   > coprocess->coprocess_resources->gpus.total) {
+		fprintf(stdout, "Coprocess with pid %d has exceeded limits, killing coprocess\n", coprocess->pid);
```
btovar, line 331: debug message? error message?

work_queue/src/work_queue_worker.c
```diff
@@ -229,7 +229,13 @@ static int tlq_port = 0;
 
 static char *coprocess_command = NULL;
 static char *coprocess_name = NULL;
-static int coprocess_port = -1;
+static int number_of_coprocess_instances = -1;
```
btovar, line 232: why not zero?

work_queue/src/work_queue_worker.c
```diff
@@ -2395,6 +2417,7 @@ static void show_help(const char *cmd)
 	printf( " %-30s Set the percent chance per minute that the worker will shut down (simulates worker failures, for testing only).\n", "--volatility=<chance>");
 	printf( " %-30s Set the port used to lookup the worker's TLQ URL (-d and -o options also required).\n", "--tlq=<port>");
 	printf( " %-30s Start an arbitrary process when the worker starts up and kill the process when the worker shuts down.\n", "--coprocess <executable>");
+	printf( " %-30s Specify the number of coprocesses the worker should maintain.\n", "--num_coprocesses=<number>");
```
btovar, line 2420: Something like?

Specify the number of coprocesses for serverless functions that the worker should maintain. Default is 0.

We don't want people activating this one thinking it increases the concurrency of all the functions.

work_queue/test/TR_work_queue_remotetask.sh
```diff
@@ -0,0 +1,103 @@
...
+import_config_val CCTOOLS_PYTHON_TEST_EXEC
+import_config_val CCTOOLS_PYTHON_TEST_DIR
+PONCHO_PACKAGE_SERVERIZE=$(cd "$(dirname "$0")/../../poncho/src/"; pwd)/poncho_package_serverize
+
+export PYTHONPATH=$(pwd)/../src/bindings/${CCTOOLS_PYTHON_TEST_DIR}:$PYTHONPATH
+
+STATUS_FILE=wq.status
+PORT_FILE=wq.port
+
+run_wq_worker_coprocess()
```
btovar, line 15: I think you should be able to modify the regular run wq worker by setiing  --coprocess 0 as the default. In this way, if we ever have more than one test using the coprocess we don't need to duplicate run_wq_worker_coprocess

work_queue/test/network_function.py
```diff
@@ -0,0 +1,147 @@
+#! /usr/bin/env python3
```
btovar, line 1: The name network_function does not communicate what it does. serverless_function, coprocess_function?

btovar (changes requested):

poncho/src/poncho/package_serverize.py
```diff
@@ -50,7 +50,9 @@ def remote_wrapper(event, q=None):
 
 def send_configuration(config):
     config_string = json.dumps(config)
-    print(len(config_string) + 1, "\\n", config_string, flush=True)
+    config_cmd = f"{len(config_string) + 1}\\n{config_string}\\n"
```
btovar, line 53: I'm not sure you need the double slash. I think it ends up working because `write` interprets the result "\n". I'd suggest we use a single slash here so we don't get confused in the future.

btovar (changes requested):

poncho/src/poncho/package_serverize.py
David-Simonetti-ND: I just tested it, and the double slash is required because there is a level of indirection here. The python code in package_serverize is stored as a string, so if it has single slashes like
`config_cmd = f"{len(config_string) + 1}\n{config_string}\n"`
when the serverless function is created, the \n are treated as literal newlines, and we get this:
`config_cmd = f"{len(config_string) + 1}
{config_string}
"`

btovar:
Ready to merge?

David-Simonetti-ND:
Yup! Ready to merge.

dthain:
Hooray!

PR #2919, opened 2022-07-07
wq: WQ periodically checks catalog server.

sghuang19:
Let WQ manager periodically reads from catalog server for further usage. Similar to `update_catalog()`.


sghuang19:
@btovar I added this function, borrowed some code from other files, what do you think?

btovar:
Looks good to me.

sghuang19:
Is it okay if we simply call `read_catalog` inside of `update_catalog`? It'll be more nested but convenient. @btovar

btovar (changes requested):

PR #2932, opened 2022-07-26
wq: equal resource proportions mode

btovar:
Adds q.tune("force_proportional_resources", 1), which makes adjust
max_proportion so that an integer number of the given task category fit
the worker. In other words, it exchanges external for internal
fragmentation of resources. For automatic allocation modes such as
CATEGORY_ALLOCATION_MODE_MAX, this is beneficial, as tasks are much less
likely to fail for resource exhaustion.
    
For example, if a worker has 12GB, and the task asks for 3.5GB, then
force_proportional_resources will assign 4GB per task (three tasks
completely cover the worker, other than only 3*3.5GB=10.5GB, with 1.5GB
of external fragmentation). If the task asks
for 7GB, then it assigns 12GB to the task.

Advantages:
    
Eliminates avoidable resource exhaustion when using
ALLOCATION_MODE_MAX. For example, if the maximum task seen has 1 core,
4GB of memory, and a worker has 2 cores, 4GB of memory, without
force_proportional_resources, a future task may fail by exhausting the 1
core. But this avoidable, as the task will be assigned %100 of the
memory, so no other task will be running by that worker anyway.
    
This aggressive ALLOCATION_MODE_MAX has been very successful in reducing
the retries of accumulation tasks in topEFT, where tasks get larger and
larger.
    
Disadvantages:
    
If there are many different sizes of small and large tasks, throughput
may be reduced, as small tasks won't be able to run in the 'leftovers'
that large tasks previously left.

### DIFF
--- a/doc/manuals/work_queue/index.md
+++ b/doc/manuals/work_queue/index.md
@@ -1915,6 +1915,38 @@ of file transfer time on overall performance. See
 [work_queue_graph_workers(1)](../man_pages/work_queue_graph_workers.md) for
 detailed information.
 
+## Specialized and Experimental Settings
+
+The behaviour of Work Queue can be tuned by the following parameters. We advise
+caution when using these parameters, as the standard behaviour may drastically
+change.
+
+| Parameter | Description | Default Value |
+|-----------|-------------|---------------|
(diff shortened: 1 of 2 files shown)


dthain (approved):
This is a good idea, and I am almost inclined to make it the default behavior.

PR #2946, opened 2022-08-10
DataSwarm Reorganization

dthain:


dthain:
@btovar please have a look over the organization here.  What I've done is to separate the code into manager, worker, tools, etc and name appropriately.  The tests are now passing in the current form.

If this looks good, then I would like to merge and then proceed to remove deprecated code in a second PR, and then do some refactoring in a third PR.

Issue #2533, opened 2021-02-11
Project: Binderize Work Queue Applications


dthain:
Project: Figure out how to make a "one button" deployment of a Work Queue application from source using Binder and Jupyter.

1 - Read up on Binder and Jupyter and learn how to use them generally.

2 - Create a repository with the standard Work Queue example application and show that you can deploy it via Binder into a running Jupyter notebook.

3 - Figure out how to get workers started that talk to that manager.  Can they run alongside in the same notebook?  Do we need to provide instructions to run them somewhere else?

4 - Extend the examples to include the Coffea, TopEFT, and SHADHO.

btovar:
Closing as wq is now in maintenance mode.

PR #2947, opened 2022-08-10
wq resources revamp

btovar:
- adds the `work_queue_status` API call, which works like the `work_queue_status` command, but can be called inside the code. (In fact, the status messages to the command are now constructed using this call.)
- adds the `application_info` method to python where applications can include statistics they want displayed.
- adds the StatusDisplay, which shows statistics as the manager runs. Currently only supports jupyter, but I plan to add text tables too. Activate by setting `WorkQueue` parameter like `status_display_interval=10` (update at least 10s apart).

Some examples: dynamic chunksize computation. coffea sends the stats shown in the application info box. 
image

mark when resources available are not enough for the tasks:
image

Example of display info implementation:

```python
class MyWQ(wq.WorkQueue):
    def application_info(self):
        app_stats = {"current_time": time.time()}
        return {"application_info": {"value": app_stats, "units": "s"}}
```

For  #2533

### DIFF
--- a/dttools/src/category.c
+++ b/dttools/src/category.c
@@ -428,7 +428,7 @@ int64_t category_first_allocation_max_seen(struct histogram *h, int64_t top_reso
     double rounded = max_seen;
     double bucket_size = histogram_bucket_size(h);
 
-    rounded = histogram_round_up(h, rounded + floor(bucket_size/2));
+    rounded = histogram_round_up(h, rounded + floor(bucket_size/3));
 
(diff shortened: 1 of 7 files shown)


dthain:
This looks very nice, and I especially like that you can pass through application-level information.

Where is the right place to document this, so that others can find it?

btovar:
I think we need to update topcoffea, coffea, and work queue manuals.

dthain (approved):

Issue #2949, opened 2022-08-15
poncho_package_analyse: packages vs dependencies


btovar:
poncho_package_analyse and other tools should respect the conda specification for ease of use. This means that any occurrence of "packages" should be renamed to "dependencies".  Currently, the documentation also uses "packages".

BarrySlyDelgado:
Merged #2956

PR #2956, opened 2022-08-19
Updates Poncho and doc to reflect Conda's specification format

BarrySlyDelgado:
This addresses #2949. the key "packages" is changed to "dependencies". pip packages are under "dependencies" as well within their own dict. The doc manual has been updated to reflect this change.

### DIFF
--- a/doc/manuals/poncho/index.md
+++ b/doc/manuals/poncho/index.md
@@ -67,7 +67,7 @@ Once created, this package can be moved to another machine for execution.
 Then, to run a program in the environment:
 
 ```
-poncho_package_run -e package.tar.gz python example.py
+poncho_package_run -e package.tar.gz -- example.py
 ```
 
 ## Specification File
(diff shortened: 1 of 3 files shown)


dthain:
Will this break existing tools (Parsl, Coffea, TopEFT) that are already using Poncho?

BarrySlyDelgado:
@dthain This would affect tools that use  `poncho_package_create`  Already created specifications would have to be updated to the new format to use poncho_package_create. So, I believe that topEFT would have to be modified.

dthain:
So we can't just break our most important customers.
And we must consider the case that multiple versions will exist simultaneously.
The way to proceed here is to permit the new keywords while retaining backwards compatibility with the old ones.
Then, update Parsl/Coffea to use the new keywords.
Finally, remove support for the old keywords once everyone has transitioned.

BarrySlyDelgado:
I will change it to allow both formats for the specifications.

BarrySlyDelgado:
Poncho_package_create now accepts both specification formats. All existing tools that use Poncho should still operate correctly.

doc/manuals/poncho/index.md
```diff
@@ -57,14 +63,11 @@ Then to create a complete package from the specification:
...
 
-The outputed tarball will be put into a directory named `envs` if a specific 
-cache directory is not specified.
-
 Once created, this package can be moved to another machine for execution.
 Then, to run a program in the environment:
 
 ```
-poncho_package_run -e package.tar.gz -- example.py
+poncho_package_run -e package.tar.gz python example.py
```
btovar, line 70: Let's keep the -- here, e.g.: `poncho_package_run -e package.tar.gz -- python example.py`

btovar (changes requested):

btovar (approved):

PR #2980, opened 2022-09-15
Add tmpdir option to condor_submit_workers

colinthomas-z80:


btovar:
@colinthomas-z80 This idea looks good! Could you make these changes to `work_queue_submit_common` rather than `condor_submit_workers`? That way all the scripts would be able to specify a scratch directory. `condor_submit_workers` would still be special in that its default starts with /tmp, rather than the current directory as all others.

Also, to conform with work_queue_factory, let's call this option `--scratch-dir`

PR #2998, opened 2022-09-26
Fixed typo causing error in package_create

BarrySlyDelgado:
A typo in the an the IF branch when converting the PONCHO spec file caused an error in which empty environments would be created.


btovar:
ready to merge?

BarrySlyDelgado:
@btovar Yes!

Issue #2994, opened 2022-09-20
Infra: Integrate Doxygen into ReadTheDocs


dthain:
Our technical API documentation is generated automatically by Doxygen from the code, and it serves as a good reference guide to everything that can be done with Chirp, Work Queue, Dataswarm, and other things:
http://ccl.cse.nd.edu/software/manuals/api/html/index.html

However, it is built by a separate process than both the code itself and our manuals in readthedocs, and so is hard to find and often out of date.

Figure out how to get the ReadTheDocs service to generate our Doxygen API documentation, and make it part of the RTD online manual.

PR #3006, opened 2022-09-28
readthedocs + doxygen

btovar:
readthedocs now runs the doxygen config file. The generated docs are hosted in readthedocs.
As part of the doc generation, a theme is cloned from github (default theme is ugly.)


btovar:
For #2994

Issue #2993, opened 2022-09-20
DS Task Ordering


dthain:
If you run one of the example DS applications, you will see that, by default, DS tasks are selected to run in the opposite order of how they are submitted.  This isn't exactly *wrong*, because tasks run concurrently and can complete in *any* order.  But it does seem unexpected.  Modify `ds_submit` so that freshly submitted tasks go to the back of the ready queue, instead of the front.

Easy first issue for a new student.

dthain:
@JosephDuggan7 I think this would be a good first issue for you to work on.

Make sure you that pull the latest version of cctools from git (it's changing frequently) and then figure out how to run one of the examples in `dataswarm/src/examples` with a local worker.  You should see that tasks often complete in reverse order.

Then, look for `ds_submit` in `dataswarm/src/manager` and figure out where tasks are added to the ready queue..

btovar:
Also, tasks can't be added blindly to the beginning/start of the queue, as the priority of the task should be taken into account.

PR #3008, opened 2022-09-28
Fixed task ordering issue

JosephDuggan7:
Tasks will now go to the back of the ready list if they have a priority of 0 (default priority) rather than inserting themselves before other tasks with the same priority.


dthain:
Fixes #2993

PR #3009, opened 2022-09-29
Removed obsolete int_sizes.h and replaced with inttypes.h throughout.

dthain:


dthain:
This was a bridge too far, I think...

PR #3021, opened 2022-10-06
TaskVine() to Manager() etc.

btovar:


dthain:
Looks like you converted `TaskVineFuture` into `ManagerFuture` by accident.

btovar:
That was on purpose. TaskVine() -> Manager(), TaskVineFutures() -> ManagerFutures().  It seems more regular that way.

Issue #2988, opened 2022-09-20
Vine: PONCHO Integration


dthain:
I think we now have a way to cleanly integrate Python environments via PONCHO in the TaskVine framework.

Here is the problem:

`poncho_package_run package.tgz command` really does three things:
1 - untar a package
2 - specialize the package to the local host
3 - run the command in the package environment

If two instances want to run concurrently, then either they must unpack the package in separate places (wasteful) or lock a common directory (unreliable) as they perform steps 1-2.

What we want to happen is for the worker to do steps 1-2 **once**, and then keep the resulting directory for use by any number of tasks.

I think this can be easily added to TaskVine as follows:

Take a look at the `VINE_UNPACK` flag, which is used in `vine_cache.c`.  If the flag is given, then a tarball (or other package) is unpacked once after transfer, and the unpacked version is used by all consumers thereafter.

Extend this by adding a new flag `VINE_PONCHO_UNPACK` which performs steps 1-2 for a poncho package, leaving the unpacked directory for use by the application in step 3.

This will require adding some arguments to `poncho_package_run` to do only steps 1-2 or step 3 as requested.

dthain:
@BarrySlyDelgado this one is a higher priority than the persistent caching one.

BarrySlyDelgado:
Looking at this now, Is there a reason that under the  `VINE_UNPACK` flag when the cache_file suffix is `tgz` we don't unzip the file but when the `suffix` is `tar.gz` we do? That is, using `tar xf` vs `tar xzf`.

dthain:
@BarrySlyDelgado how is this going?

BarrySlyDelgado:
@dthain Writing out a test, had to get familiar with the interface. Though, I have a question regarding the caching. From my perspective, a user specifies the input file `env.tar.gz` with the flag `VINE_PONCHO_UNPACK`. The command is something along the lines of `poncho_package_run -u <unpack-directory> -e env.tar.gz python test.py` what goes into `<unpack-directory>`. With work queue it was somethng like ${work_queue_sandbox}. for vine it  should be whatever `cache_path` is  with `vine_cache.c` but I'm not sure what that is.

dthain:
From the user's perspective, it should look like this:

`vine_task_add_input_file(task,"package.tar.gz","package",VINE_PONCHO_UNPACK)`

This means that the package will be unpacked into the cache directory,
[...]

PR #3026, opened 2022-10-07
Initial work on integrating PONCHO into taskvine

BarrySlyDelgado:
For #2988, Notes:

- Added new flags to `poncho_package_run` to facilitate integration
1. `-n | --no-command`, This tells PONCHO not to run a command passed into the argument, passing in no command as an argument also does this. Thus, `poncho_package_run -n -e e.tar.gz python test.py` and `poncho_package_run -e e.tar.gz` will do the same thing.
2. `-k | --keep-alive`, This tells PONCHO to not activate the cleanup function on exit keeping the environment in place for the worker to use
3.  `-r | --remove-locking`, This tells PONCHO to unpack the environment without the use of file locking.

- One current issue is that activating the environment within poncho_package_run does not propagate up to the parent environment. It would probably be better to have it be done within `poncho_package_run`
- However, with `--keep-alive` the environment is still in place unpacked for the worker to use. Thus, added to `vine_cache.c` is the ability for the worker to activate the environment itself.
- Since we're using `keep-alive` PONCHO does not cleanup on exit, thus we might need the worker to do so.


dthain:
Does cleanup just involve deleting the directory tree, or is there more to it than that?

(The worker is responsible for deleting things no longer needed.)

btovar:
@BarrySlyDelgado When using -u DIR, the directory is not removed. I'd rather use -u DIR than add the --keep-alive facility, as now we have to figure out the tmp directory created. I think that the worker has to decide that DIR, it shouldn't be up to poncho.

btovar:
Also, I don't think there is a use for turning off locking. Either two instances are writing to two different directories, in which case locking has no effect, or they are writing to the same dir, in which case we are going to have hard to debug race conditions if locking is off.

btovar:
I'd suggest not to add the `-n | --no-command`. If -u is given, but no command is given, then we simply unpack. If -u or a command are missing, then we print an error message and exit.

taskvine/src/manager/taskvine.h
```diff
@@ -44,6 +44,7 @@ typedef enum {
 	VINE_NOCACHE  = 0, /**< Do not cache file at execution site. (default) */
 	VINE_CACHE    = 1, /**< Cache file at execution site for later use. */
 	VINE_UNPACK   = 2, /**< Unpack this archive (.tar .tgz .zip) into a directory on arrival. */
+	VINE_PONCHO_UNPACK, /**< Unpacks a Poncho environment into a directory on arrival. */
```
dthain, line 47: Note that this enumeration is a bitfield where multiple flags can be set at once,
so each line needs a distinct value which is a power of two.

taskvine/src/worker/vine_cache.c
```diff
@@ -228,6 +228,32 @@ int unpack_or_rename_target( struct cache_file *f, const char *transfer_path, co
 		debug(D_VINE,"unpacking %s to %s via command %s",transfer_path,cache_path,command);
 		unix_result = system(command);
 		free(command);
+	} else if(f-> flags & VINE_PONHO_UNPACK){
+		command = string_format("poncho_package_run -k -r -e %s --unpack-to %s", trasnfer_path, cache_path);
```
dthain, line 232: Don't forget to free the command.

taskvine/src/worker/vine_cache.c
```diff
@@ -228,6 +228,32 @@ int unpack_or_rename_target( struct cache_file *f, const char *transfer_path, co
...
+		debug(D_VINE,"unpacking %s to %s via command %s", transfer_path, cache_path, command);
+		unix_result = system(command);
+		if(unix_result!=0){
+			debug(D_VINE,"command failed: %s",strerror(errno));
+			return 0;
+		}
+		command = string_format("%s/bin/activate", transfer_path);
```
dthain, line 239: Likewise

taskvine/src/worker/vine_cache.c
dthain: Actually, you could just make this a single command line string:
`poncho_package_run ... && activate && set_env`
since it all needs to run to completion anyway..

BarrySlyDelgado:
'set_env' is not present in all environments so the commands in `vine_cache.c`  were split into two to check if `set_env` exists.

- removed `no-command` and `remove-locking`.
- Added` free(command)` to `VINE_PONCHO_UNPACK` block
- Fixed bitfield enumeration

Issue #3004, opened 2022-09-27
Vine: API Renaming


dthain:
A number of the API names are clunky, or at least inconsistent with outside use of terms.  Our use of `specify` throughout is a bit of a mouthful.  I don't know of any other APIs that look like that.

Considering renaming as follows:

`vine_task_set_X` for setting individual properties.  If called multiple times, values are overridden.
`vine_task_get_X` for reading back that property.
`vine_task_add_X` for adding an item to a task.  If called multiple times, you get multiple results.
`vine_task_enable_X` for turning on a feature that has continuing effects.  (logging, monitoring, etc)

And then to separate input and output files for clarity;

- `vine_task_add_input_file`
- `vine_task_add_input_url`
- `vine_task_add_input_buffer`
- `vine_task_add_input_command`

- `vine_task_add_output_file`
- `vine_task_add_output_command`

Or course, this should be done with one giant `sed` script to change the code, headers, examples, etc all at once.

@btovar, what do you think?

btovar:
One that gives me pause is `ds_task_enable_X`, as currently it is used to enable and configure. As long as all the things we want to enable have at most one char* argument I think it is ok.

dthain:
I don't have a strong feeling about that one, what do you propose?

To clarify, I'm just talking about changing function names at this point, not signatures.

btovar:
Ah yes, I keep forgetting the name of the property is not argument.

dthain:
I was in there already and made the following API changes:

`ds_task_specify_{file/url/buffer/etc}`

becomes:

```
ds_task_specify_input_file
ds_task_specify_input_buffer
ds_task_specify_input_url
ds_task_specify_input_command
ds_task_specify_input_piece

ds_task_specify_output_file
ds_task_specify_output_buffer
```

dthain:
I'll let you do the big rename of specify->get/add/etc

dthain:
Done!

PR #3027, opened 2022-10-07
Vine api rename

btovar:
For #3004 

specify to add, set, and enable. 
I still need to go through each exposed function and maybe propose a better name (e.g. more regular names). But I think we can merge this one before the better names in order to reduce conflicts with other prs.


dthain:
I am impressed that this passes CI on the first try!

btovar:
It seems we need better tests, ha!

btovar:
ok to merge?

dthain:
Looks good!

PR #3032, opened 2022-10-17
Vine api rename2

btovar:


dthain:
Note this is good:
`get_output` -> `get_stdout`

But these are not:
`add_output_buffer` -> `add_stdout_buffer`
`get_output_buffer` -> `get_stdout_buffer`

because output buffers refer to multiple arbitrary files produced by the task

PR #3039, opened 2022-10-19
TaskVine Peer Transfer

colinthomas-z80:
Initial stages of implementing peer transfer in taskvine

### DIFF
--- a/taskvine/src/manager/vine_current_transfers.c
+++ b/taskvine/src/manager/vine_current_transfers.c
@@ -1,0 +1,70 @@
+/*
+Copyright (C) 2022- The University of Notre Dame
+This software is distributed under the GNU General Public License.
+See the file COPYING for details.
+*/
+
+#include "vine_manager.h"
+#include "vine_current_transfers.h"
+
+#include "debug.h"
+
+// add a current transaction to the transfer table
+char *vine_current_transfers_add(struct vine_manager *q, struct vine_worker_info *to, char *source)
+{
+    cctools_uuid_t uuid;
+    cctools_uuid_create(&uuid);
+
+    char *transfer_id = strdup(uuid.str);  
+    struct vine_transfer_pair *t = malloc(sizeof(struct vine_transfer_pair));
+    t->to = to;
+    t->source = source;
+
+    hash_table_insert(q->current_transfer_table, transfer_id, t);
+    return transfer_id;
+}
+
+// remove a completed transaction from the transfer table - i.e. open the source to an additional transfer
+int vine_current_transfers_remove(struct vine_manager *q, char *id)
+{
+    if(hash_table_remove(q->current_transfer_table, id)) return 1;
+    return 0;
+}
+
+
+// count the number transfers coming from a specific source
+int vine_current_transfers_source_in_use(struct vine_manager *q, char *source)
+{
+    char *id;
+    struct vine_transfer_pair *t;
+    int c = 0;
+    HASH_TABLE_ITERATE(q->current_transfer_table, id, t)
+    {
+    	if(strcmp(source, t->source) == 0) c++;
+    }
+    return c;
+}
+
+// remove all transactions involving a worker from the transfer table - if a worker failed or is being deleted intentionally
+int vine_current_transfers_wipe_worker(struct vine_manager *q, struct vine_worker_info *w)
+{
+    return 1;
+}
+
+void vine_current_transfers_print_table(struct vine_manager *q)
+{ 
+    char *id;
+    struct vine_transfer_pair *t;
+    debug(D_VINE, "-----------------TRANSFER-TABLE--------------------");
+    HASH_TABLE_ITERATE(q->current_transfer_table, id, t)
+    {	    	
+    	debug(D_VINE, "%s : source=%s", id, t->source);
+    }
+    debug(D_VINE, "-----------------END-------------------------------");
+}
+
+
+
+
+
+
--- a/taskvine/src/manager/vine_current_transfers.c
+++ b/taskvine/src/manager/vine_current_transfers.c
@@ -9,6 +9,14 @@ See the file COPYING for details.
 
 #include "debug.h"
 
+static struct vine_transfer_pair *vine_transfer_pair_create(struct vine_worker_info *to, char *source)
+{
+    struct vine_transfer_pair *t = malloc(sizeof(struct vine_transfer_pair));
+    t->to = to;
+    t->source = source;
+    return t;
+}
+
 // add a current transaction to the transfer table
 char *vine_current_transfers_add(struct vine_manager *q, struct vine_worker_info *to, char *source)
 {
@@ -16,9 +24,7 @@ char *vine_current_transfers_add(struct vine_manager *q, struct vine_worker_info
     cctools_uuid_create(&uuid);
 
     char *transfer_id = strdup(uuid.str);  
-    struct vine_transfer_pair *t = malloc(sizeof(struct vine_transfer_pair));
-    t->to = to;
-    t->source = source;
+    struct vine_transfer_pair *t = vine_transfer_pair_create(to, source);
 
     hash_table_insert(q->current_transfer_table, transfer_id, t);
     return transfer_id;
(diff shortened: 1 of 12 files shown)


taskvine/src/worker/vine_cache.c
```diff
@@ -198,6 +199,52 @@ static int do_command( struct vine_cache *c, const char *command, const char *ca
...
+*/
+static int do_worker_transfer( struct vine_cache *c, const char *source_url, const char *cache_path, char **error_message)
+{	
+	int port_num;
+	char *addr, *port, *path;
+	char source_mutable[4096];
+	
+	strcpy(source_mutable, source_url);
+	
+	// expect the form: worker://addr:port/path/to/file
```
dthain, line 214: You can use `sscanf` to parse all this stuff out in one go.
Note that the format specifiers can be used sort of like a regex.

taskvine/src/worker/vine_cache.c
```diff
@@ -198,6 +199,52 @@ static int do_command( struct vine_cache *c, const char *command, const char *ca
...
+	{
+		debug(D_VINE, "Could not connect with worker at: %s", source_url);
+	}
+
+	if(!vine_transfer_get_any(worker_link, c, path, time(0) + 120))
+	{
+		return 0;
+	}
+
+	// rename file to our expected cache name (probably not the best way to do this)
```
dthain, line 240: The program `mv` is basically a big wrapper around the system call `rename`.
You can just call `rename` directly here and save some trouble.

taskvine/src/worker/vine_cache.c
```diff
@@ -198,6 +199,52 @@ static int do_command( struct vine_cache *c, const char *command, const char *ca
...
+	
+	// expect the form: worker://addr:port/path/to/file
+	strtok(source_mutable, "/:");
+	addr = strtok(NULL, "/:");
+	port = strtok(NULL, "/:");
+	path = strtok(NULL, "");
+
+	if(path == NULL)
+	{ 
+		debug(D_VINE, "Improperly formed worker URL: %s", source_url);
```
dthain, line 222: You should do something with `error_message` in this case, because the caller is expecting it to be set when something goes wrong.

dthain (commented):
You are in the right place, let's just tighten up a few things.
Then you can move on to fixing up the manager side.

taskvine/src/manager/vine_manager_put.c
```diff
@@ -348,15 +348,26 @@ Send a single input file of any type to the given worker, and record the perform
...
+	// search for workers with our file
+	char *id, *new_source;
+	struct vine_worker_info *peer;
+	HASH_TABLE_ITERATE(q->worker_table, id, peer){
+		if(hash_table_lookup(peer->current_files, f->cached_name))
+		{
+			debug(D_VINE, "Signaling this file to be requested from worker: %s:%d", peer->transfer_addr, peer->transfer_port);
+			debug(D_VINE, "Original source: %s", f->source);
+			new_source = string_format("worker://%s:%d/%s", peer->transfer_addr, peer->transfer_port, f->cached_name);
+			f->source = new_source;
```
dthain, line 365: Changing the file object on the fly probably isn't a good idea, since later uses of the file will want to know the original source.  Try making a new temporary `vine_file` object with the desired properties, and call `vine_manager_put_input_file` on that.

taskvine/src/worker/vine_cache.c
```diff
@@ -198,6 +199,43 @@ static int do_command( struct vine_cache *c, const char *command, const char *ca
...
 }
 
+/*
+Transfer a single input file from a worker url to a local file name. 
+
+*/
+static int do_worker_transfer( struct vine_cache *c, const char *source_url, const char *cache_path, char **error_message)
+{	
+	int port_num;
+	char addr[99], path[4096];
```
dthain, line 209: VINE_LINE_MAX is the simplest thing here.

dthain:
Almost there!

dthain:
Nice!

taskvine/src/worker/vine_cache.c
```diff
@@ -196,6 +198,43 @@ static int do_command( struct vine_cache *c, const char *command, const char *ca
...
+	int port_num;
+	char addr[VINE_LINE_MAX], path[VINE_LINE_MAX];
+	int stoptime;	
+	struct link *worker_link;
+
+	// expect the form: worker://addr:port/path/to/file
+	sscanf(source_url, "worker://%99[^:]:%d/%s", addr, &port_num, path);
+
+	stoptime = time(0) + 15;
+	worker_link = link_connect(addr, port_num, stoptime);
```
dthain, line 216: Don't forget to `link_close` on all paths out of this function!

dthain:
@colinthomas-z80 see comments above.

dthain:
Yup, you on on the right track.  I like the separate module with the key operations called out.

taskvine/src/manager/vine_current_transfers.c
```diff
@@ -0,0 +1,75 @@
...
+#include "debug.h"
+
+// add a current transaction to the transfer table
+char *vine_current_transfers_add(struct vine_manager *q, struct vine_worker_info *to, char *source)
+{
+    cctools_uuid_t uuid;
+    cctools_uuid_create(&uuid);
+
+    char *transfer_id = strdup(uuid.str);  
+    struct vine_transfer_pair *t = malloc(sizeof(struct vine_transfer_pair));
```
dthain, line 19: Make functions `vine_transfer_pair_create/delete` to manage this structure.

taskvine/src/manager/vine_current_transfers.h
```diff
@@ -0,0 +1,27 @@
...
+
+int vine_current_transfers_remove(struct vine_manager *q,char *id);
+
+int vine_current_transfers_source_in_use(struct vine_manager *q,char *source);
+
+int vine_current_transfers_wipe_worker(struct vine_manager *q, struct vine_worker_info *w);
+
+void vine_current_transfers_print_table(struct vine_manager *q);
+
+void vine_current_transfers_delete( struct vine_transfer_pair *p );
```
dthain, line 27: delete can be private to the module, since it is only used internally.

taskvine/src/manager/vine_current_transfers.h
```diff
@@ -0,0 +1,27 @@
...
+This software is distributed under the GNU General Public License.
+See the file COPYING for details.
+*/
+
+#include "vine_worker_info.h"
+#include "uuid.h"
+
+#define VINE_FILE_SOURCE_MAX_TRANSFERS 1
+
+struct vine_transfer_pair {
```
dthain, line 12: Put this structure definition in the C file, since it isn't used by callers of the module.

taskvine/src/manager/vine_current_transfers.h
```diff
@@ -0,0 +1,27 @@
...
+    struct vine_worker_info *to;
+    char   *source;
+};
+
+char *vine_current_transfers_add(struct vine_manager *q,struct vine_worker_info *to, char *source);
```
dthain, line 17: Most string parameters to functions should be `const char *` since you aren't going to modify the contents of those strings.

dthain (changes requested):
Looking good, almost there?

colinthomas-z80:
> Looking good, almost there?

There is a small problem with the transfer IDs not coming back from the worker. Then I figure we need a way to turn the feature on and off. After that I just need to do some cleaning/refactoring.

taskvine/src/manager/vine_current_transfers.h
colinthomas-z80: delete is referenced in vine_manager when we call hash_table_delete( , vine_current_transfers_delete)

dthain:
Don't rely on domain names, those can be messed up in many complicated ways.
Just have the manager use `link_address_remote` to observe the remote IP address
of the existing connection to the worker.

taskvine/src/manager/vine_current_transfers.h
colinthomas-z80: related to previous comment, this is an argument to the hash table delete helper function used in vine_manager

dthain (approved):

dthain:
For some reason, OSX has intermittent failures related to port availability.  I'm rerunning that test...

taskvine/src/worker/vine_transfer_server.c
```diff
@@ -96,6 +97,7 @@ void vine_transfer_server_stop()
 
 void vine_transfer_server_address( char *addr, int *port )
 {
+	domain_name_cache_guess(addr);
```
dthain, line 100: This is the problem: `domain_name_cache_guess` fills in its argument with the guessed domain name.
`addr` is a string only long enough to hold an IP address.

dthain (changes requested):

dthain:
I can reproduce on my laptop, let me see what's going on...

PR #3041, opened 2022-10-25
WQ: Bucketing Algorithms for Resource Classification

tphung3:
Implementation of fast bucketing and exhaustive bucketing (under development)


dttools/src/bucketing_fast.c
```diff
@@ -0,0 +1,240 @@
...
+    struct list* break_point_list = list_create();
+
+    /* create list and push (0, n-1) to list */
+    struct list* bucket_range_list = list_create();
+    bucketing_bucket_range* init_range = bucketing_bucket_range_create(min, max, s->sorted_points);
+    list_push_tail(bucket_range_list, init_range);
+
+    bucketing_bucket_range* lo_bucket_range;
+    bucketing_bucket_range* hi_bucket_range;
+    bucketing_cursor_w_pos** break_point = malloc(sizeof(*break_point));
```
dthain, line 95: I think this code will be easier to read/write if you remove a level of indirection.  Try this:
`
bucketing_cursor_w_pos * break_point = 0;
list_get(lc,&break_point);
 . . .break_point->pos . . .
```

File: dttools/src/bucketing_fast.c [hunk 2]
```diff
@@ -0,0 +1,240 @@
...
+int compare_break_points(bucketing_cursor_w_pos* p1, bucketing_cursor_w_pos* p2)
+{
+    return p1->pos - p2->pos;
+}
+
+int bucketing_fast_break_bucket(bucketing_bucket_range* range, bucketing_cursor_w_pos** break_point)
+{
+    double min_cost = -1;
+    double cost;
+    bucketing_cursor_w_pos** tmp_break_point = malloc(sizeof(*tmp_break_point));
```
dthain, line 152: Same sort of thing here.

tphung3:
I just removed the redundant cursor that prevents list_delete. Code should run by now. I'm checking for correctness

dthain:
Also, I think you added some constants to `category.h` that should be added to the PR.

tphung3:
Fast bucketing is doing what it is supposed to do. Moving on to exhaustive now..

dthain:
@tphung3 add and push `category.h` here so you can get a clean CI build.

tphung3:
I added some fixes and correctness checks to exhaustive and fast bucketing. 
I also added sort of a wrapper called `bucketing_manager` that plugs the base algorithms to WQ. 
The file `bucketing_manager.h` describes the APIs to the bucketing algorithms. 
It tries to use as much of existing stuffs in `category.h` and `rmsummary.h` as possible. 
Some part of the APIs may feel weird so any feedback is welcome @btovar @dthain.
TODO: 
- change names of bucketing algorithms?
- implement and test bucketing manager

dthain:
@tphung3 overall, this is looking quite good, and I appreciate your attention to detail, commenting, debugging, etc.  A few little items:

1 - I have no strong opinion about the names.  I think at this point it's best just to make sure that they are mentioned in the documentation, and have the docs briefly describe the goal of each one, and point to the relevant publication.

2 - The header files tend to mix public API operations (e.g. functions called by other modules) with internal implementation details.  It would be preferable if the header files only contained the absolute minimum needed to use the modules, and the remaining prototypes and structures were put inside the relevant .c files.

3 - A few style things: give your enumerations qualified names like BUCKET_OP_ADD and BUCKET_OP_PREDICT, rather than add or predict, so that we don't have namespace clashes.  Also, please give each typedef name a suffix of `_t` so that it is clear that it is a type.

4 - There are a couple cases where you call `malloc` and then return zero if it fails.  I think here it's better to call `xxmalloc` and not check the result.  We really have no good options if we run out of memory.

5 - Please add a bit of description about the modes to `work_queue.h` and then in the corresponding place for `taskvine.h`-- I think it works in the same place there as well.

6 - Add a test case something like `dttools/test/TR_test_bucketing.sh` that makes use of your existing test executable, so we can get it as part of the regular CI.

dthain:
@tphung3  I think you may give yourself a headache by merging a whole bunch of commits from master into this branch  Suggest that you revert to `128e33a` and fix the CI build on that commit.  Then, if @btovar agrees, merge the PR and fix any (minimal) conflicts at that point.

tphung3:
My local and remote repos have the same sequence of commits but different hash value per commit for some reason. So [39a855e](https://github.com/cooperative-computing-lab/cctools/pull/3041/commits/39a855e398d08e5f4fdb2d518d2e411950bfadc0) is to correct the hash values so I can push my local changes (it doesn't add or remove code). Not sure why this happens though..

btovar:
@tphung3 is this ready to merge?

tphung3:
Not yet. I'm debugging the manager test.

tphung3:
I added some fixes to the manager and its test. The dangling pointer bug is due to the inconsistent bookkeeping of internal data structures by the manager, so from now the manager keeps its own copy of each datum and callers are expected to free whatever the manager takes in as parameters or returns. If all is well then this PR is ready to merge.

dthain:
Well, it does have to build.  :)

btovar (approved):

work_queue/src/work_queue.c
```diff
@@ -4589,6 +4589,8 @@ static int send_one_task( struct work_queue *q )
 		// Skip task if min requested start time not met.
 		if(t->resources_requested->start > now) continue;
 
+		//prepare_task(t);
```
dthain, line 4592: Dead code?

dthain (commented):
If I understand correctly, the new allocation mode is listed in the enumeration, but it isn't actually hooked up in `category.c`, right?

tphung3:
> If I understand correctly, the new allocation mode is listed in the enumeration, but it isn't actually hooked up in `category.c`, right?
Yes that's correct

dttools/src/.gitignore
```diff
@@ -36,3 +36,5 @@ jx_merge_test
 mq_poll_test
 mq_wait_test
 mq_store_test
+bucketing_base_test
```
tphung3, line 39: @dthain here's the latest changes

work_queue/src/work_queue.c
tphung3: Yes the fix is added. Thanks!

dthain (approved):

PR #3046, opened 2022-10-31
Vine: Mini Tasks

dthain:
This is a quick experiment, and the code has some deficiencies down in the weeds.
But it does show the key idea of files feeding into commands, producing files for tasks.
`vine_example_chained.c`shows an example use of creating an input url:
```
struct vine_file *u = vine_file_url(CCTOOLS_URL,"package.tar.gz",VINE_CACHE);
```

Feeding that to a (clumsy) unpack command:
```
struct vine_file *f = vine_file_command("mkdir %%; tar xvzf $REQUIRES -C %%","cctools",VINE_CACHE,u);
```

And then running a task that uses it:
```
struct vine_task *t = vine_task_create("ls -lR cctools");
vine_task_add_input(t,f);
```

Then the manager sends the files along with dependency info:
```
022/10/31 16:18:16.48 vine_worker[157485] vine: rx: puturl http://ccl.cse.nd.edu/software/files/cctools-7.4.14-source.tar.gz url-0-bf382b9892a1d475bdc8f3f5a551b86f 0 777 1 0
2022/10/31 16:18:16.48 vine_worker[157485] vine: rx: putcmd mkdir%20%25%25;%20tar%20xvzf%20url-0-bf382b9892a1d475bdc8f3f5a551b86f%20-C%20%25%25 cmd-0-440112e79a72e6fe373a7baf5658c369 0 777 1 url-0-bf382b9892a1d475bdc8f3f5a551b86f
```

And the worker executes:
```
2022/10/31 16:18:16.48 vine_worker[157485] vine: cache: transferring http://ccl.cse.nd.edu/software/files/cctools-7.4.14-source.tar.gz to url-0-bf382b9892a1d475bdc8f3f5a551b86f
2022/10/31 16:18:16.48 vine_worker[157485] vine: executing: curl -sSL --stderr /dev/stdout -o "/tmp/worker-102744-157485/cache/url-0-bf382b9892a1d475bdc8f3f5a551b86f.transfer" "http://ccl.cse.nd.edu/software/files/cctools-7.4.14-source.tar.gz"
2022/10/31 16:18:16.60 vine_worker[157485] vine: renaming /tmp/worker-102744-157485/cache/url-0-bf382b9892a1d475bdc8f3f5a551b86f.transfer to /tmp/worker-102744-157485/cache/url-0-bf382b9892a1d475bdc8f3f5a551b86f
2022/10/31 16:18:16.61 vine_worker[157485] vine: cache: created url-0-bf382b9892a1d475bdc8f3f5a551b86f with size 13290783 in 124963 usec
2022/10/31 16:18:16.61 vine_worker[157485] vine: tx: cache-update url-0-bf382b9892a1d475bdc8f3f5a551b86f 13290783 124963
2022/10/31 16:18:16.61 vine_worker[157485] vine: cache: creating cmd-0-440112e79a72e6fe373a7baf5658c369 via shell command
2022/10/31 16:18:16.61 vine_worker[157485] vine: executing: mkdir /tmp/worker-102744-157485/cache/cmd-0-440112e79a72e6fe373a7baf5658c369.transfer; tar xvzf url-0-bf382b9892a1d475bdc8f3f5a551b86f -C /tmp/worker-102744-157485/cache/cmd-0-440112e79a72e6fe373a7baf5658c369.transfer
2022/10/31 16:18:16.87 vine_worker[157485] vine: renaming /tmp/worker-102744-157485/cache/cmd-0-440112e79a72e6fe373a7baf5658c369.transfer to /tmp/worker-102744-157485/cache/cmd-0-440112e79a72e6fe373a7baf5658c369
2022/10/31 16:18:16.87 vine_worker[157485] vine: cache: created cmd-0-440112e79a72e6fe373a7baf5658c369 with size 39068670 in 262831 usec
2022/10/31 16:18:16.87 vine_worker[157485] vine: tx: cache-update cmd-0-440112e79a72e6fe373a7baf5658c369 39068670 262831
```


dthain:
This isn't a serious pull request (yet), I just want to play around with the idea.

One of the things that gets complicated is that shell commands must have some way of referring to their inputs and outputs, which are (internally) stored as unique cache names.  So, for example here:

```
struct vine_file *f = vine_file_command("mkdir %%; tar xvzf $REQUIRES -C %%","cctools",VINE_CACHE,u);
```

The symbol `%%` refers to the cache_name of the file that will be created by this command.  The symbol `$REQUIRES` refers to the cache name of the required file `u`.  (These can be changed, I just picked something easy to hack in.)

dthain:
Some issues to ponder:
- Is this "mini-task" really an improvement over first-class tasks?  e.g. should we be concerned with resources, sandboxing, etc of the mini-tasks?
- Who should perform the substitution of the names, the manager or the worker?  Regardless, it is still necessary for the worker to have the full dependency information in order to put things in proper order.  (I already made that mistake once.)
- Syntactically, what's the easiest way to deal with multiple file dependencies?  In this example, the command type can only have a single file dependency.  We will want these trees to be both broader and deeper.

btovar:
I think I prefer the file abstraction not have any dependencies like this one:

```C
struct vine_file *f = vine_file_command("mkdir %%; tar xvzf $REQUIRES -C %%","cctools",VINE_CACHE,u);
```
It is a subtle way to leak the abstraction, as a I have to tell the fetch how to do a transformation, but I can't think of an alternative yet... 

I am likely to need more than one dependency (e.g, for poncho I need poncho_package_run and the actual tarball), so I would make it variadic.

I would like to think of all the `vine_file_*` as simple fetchs, e.g., at most they read a file locally, do curl, xrdcp, etc. If something needs to happen to the file, then I would create an environment abstraction, e.g.:

```C
struct vine_file *l = vine_file("mylocal.db");
struct vine_file *p = vine_file("myfancyunpack");
struct vine_file *u = vine_file_url(CCTOOLS_URL,"package.tar.gz");

struct vine_environment *e = vine_environment_create(VINE_CACHE);

vine_environment_add_file(e, l);
vine_environment_add_file(e, u);
vine_environment_add_file(e, p);
vine_environment_add_file_transform(e,
    "mkdir %%; ./myfancyunpack package.tar.gz  %%",
   my_chosen_unpack_name,
   u,
   p);
```

`vine_environment_add_file_transform` would be variadic.

I'm not sure about the flags, but maybe we would only need to say that the env is cached, and that would apply to all the files and transformations in the environment.

This is how I would  internally define a poncho environment, as:

```C
struct vine_environment vine_environment_poncho_create(const char *env_name, vine_file *tarball) {

  struct vine_environment *e = vine_environment_create(VINE_CACHE);
  vine_environment_wrapper("./poncho_package_run -u env_name  -- %%");   # where %% is the task command line

  poncho_package_run = ... find me a poncho package run ...;
  struct vine_file *p = vine_file(poncho_package_run);

  vine_environment_add_file(e, tarball);
  vine_environment_add_file(e, p);

  vine_environment_add_file_transform(e,  "./poncho_package_run -u %%  -e myponcho.tar.gz", env_name, u, p);  # where %% is cache name

 return e;
}
```

dthain:
Summing up our conversation just now:
- Environments look like a good way to capture multiple dependencies and a task transformation into a single bundle.  Good for sharing commonalities among multiple tasks.
- However, they don't capture the parameterized production of individual files, e.g. unpack or xrdcp.  We need something else for that.
- I like "transform" as an alternative to "command".
- Both transformations and environments look and execute like mini-tasks, but have some limitations on scheduling and locality, and so it makes sense to keep them separate from individual tasks.

dthain:
Another deficiency that is also coming to light is that we haven't cleanly separated the source of data from its mapping into the file.  For example this isn't right because it associates the remote_name with the task in stead of the file:
```
f = vine_file_url(url, remote_name,flags)
vine_task_add_input(t,f)
```
It really should be more like this:
```
f = vine_file_url(url)
vine_task_add_input(t,f,remote_name,flags)
```

dthain:
Our flags primarily control manager-worker interactions:
- caching or not
- watching or not
- return failure on success/failure

So I think they can be safely omitted from the sources.  And then the cache name can uniquely identify the source and content of the data, independent of the handling.

Then we can do this:

```
struct vine_file * vine_file_url( url );
struct vine file * vine_file_transform(cmd,...);
```
And put them together like this:
```
f = vine_file_transform("gunzip $1 > $$",vine_file_url("something.gz"),vine_file_local("/usr/bin/gunzip"),0);
```

btovar:
One practical advantage of adding the flags to things such as `vine_file_url`, is that it makes it easy to know which files should be deleted from the cache. Checking the wq source code right now there is this ugly case where if one task marks a file with cache=True, and another with cache=False, then the file will be deleted from the worker regardless. (I.e., I don think we check if some other task marked the file with cache=True when cleaning up tasks.)

It is not incorrect by itself as we are just giving preference to cache=False. I think that giving preference to cache=True makes more sense, and for that the manager probably has to do some bookkeeping.

dthain:
So now here is an example of a file created by delivering both the command executable and an input file:
```
struct vine_file *tar = vine_file_local("/usr/bin/tar");
struct vine_file *url = vine_file_url(CCTOOLS_URL);
struct vine_file *file = vine_file_command("mkdir $0; ./$1 xvzf $2 -C $0",tar,url,0);

struct vine_task *task = vine_task_create("ls -lR cctools");
vine_task_add_input(task,file,"cctools",VINE_CACHE);
```

This works as desired, albeit there are a few code smells to fix.

Upsides:
- It is relatively straightforward to specify a complex input inline in a functional style.
- It should now be easy to make client side wrappers like this:`vine_file_xrootd(url) = vine_file_command( "./$1 $2 -O $0 , "xrdcp",url )`
- File dependencies are naturally scheduled with the corresponding task.

Downsides:
- Commands are currently run "unsafely" in the cache directory, without any serious resource management.
- Users might be tempted to do expensive operations in commands, instead of limiting to basic download/unpack sorts of things.

dthain:
I think there are a few ways forward from here, in order of complexity:
1 - Leave file commands as they are, a lightweight and possible dangerous transformation handled as a side effect of a task.
2 - Improve the implementation of file commands at the worker, by running them in a sandbox directory, but leaving them as second-class citizens.
3 - Make the file commands fully-fledged tasks that have all the properties of regular tasks (namespace, resources, etc) but are attached as children of ordinary tasks, and only executed by the worker as a side effect of running a first-class task.
4 - Make these transformations first-class tasks, so that the manager schedules them, tracks the output file produced, and then schedules the task that needs them to the location of the produced file.

@btovar what do you think?

btovar:
I tend to like 2 or 3, execute them as a side effect if the end result of the transformation is not already there. In that way, we get to execute them in a defined sandbox, and we do not pollute the stats and logic of first-class tasks at the manager.

btovar:
I think that C deleting the minitask is a little unexpected, as the minitask is created without referring to the task. So, either:

1. C shouldn't delete the minitask either. Minitask are deleted when the files they produce are no longer needed. We determine this when we get something akin to "invalidate cached file" for those files. This has the advantage that minitask can be shared among tasks.
2. The minitask is created directly with a relation to the regular task, e.g. `mt = t.add_minitask(..); mt.add_input(...)`. This has the advantage of clear ownership. In python, the __del__ of mt wouldn't delete the C pointer, as that would happen in the C delete of the main task.

This two modes have the advantage that the C and python implementations have the same garbage collection strategies.

dthain:
The problem with the latter approach is that the minitask is created by the same multi-call API as a regular task:
```
minitask = Task(cmd)
minitask.add_input(...)
minitask.add_output(...)

t = Task(cmd)
t.add_input_mini_task( t, minitask, name, flags )
```

btovar:
For that, minitasks have to be created with their own call, like SetupTask(), even if behind the scenes it is the same structure than task, and we just set some attribute. In C, it would be the same, vine_create_setup_task or something like that.

dthain:
To make that work, we would have to come up with a more Pythonic way of representing an entire task, something like:

```
t = Task( { cmd: xyz, inputs: [ ...], outputs: [ ] } )
```

Which would have some value, but then becomes a type checking problem.

dthain:
Actually, here is a simpler alternative:  When attaching a minitask m to a task t, simply clone m for cleanup by the C runtime, and leave the original for cleanup by Python.  What do you think?

btovar:
I think it will hide mysterious bugs if the minitask is attached and then modified.

I don't think we need to modify the Task like that. If we are willing to add any Task as a minitask, then SetupTask would be the same, only that we know not to clean it up.

dthain:
Hmm, I'm trying to decide which path results in more trouble:
- Add a flag or subclass to the Python side, indicating not to delete, because C will. Objects can be modified after attaching.  No difficulties with aliasing.  **But** user could crash Python if they attempt to modify a minitask after deleting the parent task.
- Add a flag to the C side, indicating not to delete, because Python will.  Objects can still be modified after attaching.  User cannot crash Python because they objects exists as long as Python says it will.  **But** looks like a bit of a hack from the C side, could be difficult to maintain.
- Copy the C object when it is attached.  Easy to implement, and works as well as clone is maintained effectively.  Keeps ownership of objects clear between C and Python.  **But** results in strangeness if users modifies object after attaching.

btovar:
Why you don't like the decoupling?  Neither C nor Python delete when a parent task goes out of scope. Minitasks are objects shared by Tasks, and deleted explicitly when `invalidated`. I'm afraid that any other option will give us weird constraints and memory bugs to deal in the future.

Further, it is something we will need to do anyway if we are thinking of chaining more than minitasks.

btovar:
Also, one way to think about it is to implement the decoupling of Task/Minitask as a primitive operation, and we can think on how to handle the garbage collection policy on top of that.

dthain:
Come on over and let's talk about it.  I think we are confusing each other.

Issue #3045, opened 2022-10-31
Vine: change python bindings to use  some_flag=True


btovar:
Right now we have things like:  `add_input_file(..., flags=VINE_UNPACK|VINE_CACHE, ...`, which is rather ugly in python. Instead, let's make each flag part of the api of the function:  `add_input_file(..., cache=True, unpack=True)`.

btovar:
Completed in #3050

PR #3050, opened 2022-11-09
taskvine: more pythonic file flags

btovar:
Instead of doing flags = VINE_CACHE | VINE_UNPACK, now we do  cache=True, unpack=True, etc.


btovar:
For #3045

dthain:
No objection here, but the set of flags is in flux as we move features around for mini-tasks.

PR #3054, opened 2022-11-11
Taskvine Serverless

David-Simonetti-ND:
This PR ports over all the serverless functionality from WQ to TaskVine. Still need to fix worker/task resources.


taskvine/src/worker/vine_coprocess.c
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (C) 2022- The University of Notre Dame
```
btovar, line 2: We do want the hyphen. 2022-

btovar (approved):

PR #3055, opened 2022-11-16
Vine: starting coprocesses via tasks

David-Simonetti-ND:
This PR is a major work in progress, but makes significant steps to being able to start a coprocess on a worker from the manager. Added a new kind of task, a "duty" task, that the manager expects the worker to run for as long as it is connected to the manager (as a task is something finite to complete, is duty is something that keeps going forever). 
The way a duty is created in Python is such
`t.set_duty("duty description here")`
I imagine formatting it as such for coprocesses, sending the required information from the manager to the worker.
`t.set_duty("coprocess_name:$name\ncoprocess_memory:$memory...")`
The worker checks if incoming tasks are duties before executing them, and if they are, it creates a new coprocess instance instead. The worker then informs the manager it now has a coprocess running and the resources the coprocess has.
Things left to do:
Starting coprocess instances at worker initialization is now broken, should be fixed
Manager should place "duty" tasks in a separate queue, as it now will never have an empty queue as it will wait for the duties to finish for forever
There seems to be a bug where the manager double counts the resources used for a RemoteTask.


dthain:
@David-Simonetti-ND this is looking great!
What remains to be done before you think it is ready for review?

One minor naming issue: you have several functions in `vine_manager` that operate on a `struct vine_manager *`:

```
vine_task_install_duty
vine_task_send_duty
...
```

These should be named `vine_manager_install_duty` etc since they are essentially methods on the manager object.

David-Simonetti-ND:
@dthain There is a segfault I still have to hunt down, but duties are able to be installed and removed across all workers, including workers that just connected. I should get it ready for review sometime tomorrow.

dthain (approved):

PR #3056, opened 2022-11-16
WQ: Integrating Bucketing Algorithms into WorkQueue and TaskVine Resource Management

tphung3:
@btovar @dthain 
This PR is opened for a discussion on how to merge bucketing algorithms to category. There are a couple of issues that I need feedback on if possible.
1. The `histogram` facilities seem to assume that buckets are of fixed sizes (`size = upper_end_val - lower_end_val`). If implemented I'd probably set the bucket size for cores to be 1 and for memory and disk to be 32 or 64 MBs (instead of 256 MBs like Ben did). The upside to this is that the quality of prediction is better as the buckets are more fine-grained and bucketing algorithms can also be used for future WQ/Vine functions that don't consume as much memory as regular tasks. The downside is for very large tasks (32-64 GBs) there will be ~1000-2000 buckets. Which leads to my second point.
2. I can add improvements to the current implementation by capping the number of points a bucketing state can hold to 1000 or 2000. This would take 16*2000 = 32KBs at all times (bucketing point only contains 2 doubles.) It also makes sense to cap at 2000 as old points have a polynomially reduced significance overtime in the design of bucketing and thus can be removed (First task at 2000 points has less than (5e-6)% of being chosen). (Another improvement is to rebucket every k tasks as Ben suggested, unrelated to this discussion though, but both improvements can be quickly implemented.)
3. It seems that the **only** place in work queue that tasks' resource reports are processed is in `accumulate_summary`, which is in `wait_internal -> receive_one_task->fetch_output_worker->accumulate_task->accumulate_summary`. Is this correct @btovar ? Also I checked `send_one_task` to find the place where the allocation for next task is computed but I can't find it anywhere. Any hints on this?


btovar:
1. For benchmarks, fine-grained buckets might make sense, but for workflows in the wild the small bucket sizes we have found to be counterproductive, as trying to optimize with small bucket sizes causes unnecessary retries when the workflow is about to finish. I'd leave the 256MB value, as it can be `q.tune()` down for the workflows that need it. Further, note that wq won't run more tasks in a worker than the worker's number of cores,  as we are moving to a default where resources of a worker are allocated proportionally to the tasks running there, i.e., from the manual:

> Unspecified cores, memory and disk will get a default according to the proportion the specified resources will use at a worker, rounding-up. Thus, if task specifies that it will use two cores, but does not specify memory or disk, it will be allocated %50 of memory and disk in 4-core workers, or %25 in 8-core workers. When more than one resource is specified, the default uses the largest proportion.

and the `q.tune()` parameter `force-proportional-resources`, which I would like to make the default in taskvine:

> force-proportional-resources: When task with requirement r of resources is allocated in a worker,divide the resources of the worker evenly so that only a whole number of tasks with requirement r fit in the worker.

Also, wq regularly runs on workers with more than 128GB of memory, so I have even been tempted to increase the bucket size from 256MB...

2. When using histograms I don't think the number of points is a consideration?  Also, see `q.tune("category-steady-n-tasks", n)` 

3. Correct. They are accumulated in that function, and then the predictions are retrieved with `category_dynamic_task_{max,min}_resources`

tphung3:
@btovar  I'm writing a test program to run the bucketing algorithms in WQ and it seems that the manager retries exhausted tasks automatically instead of returning those tasks to users. Is there a way to force return exhausted tasks to users (I need to correctly account badput/goodput for efficiency calculations)?

The second problem I'm having is it seems that the manager computes the allocation predictions for tasks right after submit time instead of right before dispatch time, which is bad as the manager doesn't have to predict so early and could have had more information about completed tasks. The proposed changes would be in the `send_one_task` function where a `prepare_task` function would do as above. Any thoughts on this?

btovar:
When using `CATEGORY_ALLOCATION_MODE_FIXED`, the task should be returned without a retry. The `category_next_label` function returns `CATEGORY_ALLOCATION_ERROR` when trying to compute a next allocation and the task is returned when  `work_queue_wait` is called. 

For the second point, there are two related concepts. The next allocation for the task is chosen with the call ``category_next_label`. This happens before `start_one_task`. The actual allocation the task gets is chosen inside `start_one_task`, once we know which worker is going to execute the task.  Which point would your `prepare_task` function cover?

Note that the theoretical next allocation for the task might be smaller than the actual allocation given at a worker. This is done to avoid silly retries, when for example, the task would use all the cores of a worker. In such cases we want to assign all the memory to the task too, as it wouldn't make sense to stop the task if our original allocation for memory was too small.

dthain:
Thinking about how we write this up for HPDC....

It would seem that there are really two resource allocation steps:
1 - Predicting the next allocation from the corpus of prior results.
2 - Adapting that prediction to the facilities actually available.

btovar:
Yes, that's it!
The first allocation works with labels to the task: "first allocation", and "max allocation". This comes from predicting from previous results. Then the label gets translated to the worker selected, assuming the worker is of at least the needed size.

btovar:
The reason the second step is needed is because the optimization would get preoccupied with optimizing very small ranges (e.g. assigning 100MB or 120MB of memory to a task. If the worker had 128GB of memory, such an optimization was actually wasteful.)

tphung3:
1. I see what you are referring to and I can think of a way to force the task to return when the category is in `GREEDY/EXHAUSTIVE BUCKETING` mode (Thanks!). However it is a good feature to retry tasks transparently, and all I need is to account for resource efficiency of tasks. So would it make sense to add a field in `struct work_queue_task` that tracks this (e.g. `struct rmsummary* resources_efficiency`)? It would simply require correctly updating efficiency of tasks inside `receive_one_task` and I also avoid breaking the task-retry transparency.
2. I just double check and the manager makes allocation at dispatch time (`start_one_task`) instead of submit time (`work_queue_submit`) so my bad...

btovar:
So, that information is already in the transactions file. You could read the resources that cause a resource exhaustion. Would that be enough?

What information would the new field contain?

tphung3:
That'd be enough. Hmm..
The new field would contain the efficiency of a task. For example, t->resources_efficiency-cores returning 0.75 means the task's core efficiency is 75%.

btovar:
Sounds good. Would this efficiency be computed according to the desired allocation, or the allocation actually given to the task at a worker? Is it for the last successful attempt or for all attempts regardless of failure, or includes only failures that exhausted resources?

tphung3:
I think the efficiency field should reflect as close to what truly happens as possible, so 1) the actual allocation given at a worker, and 2) all attempts regardless of failure.

tphung3:
Bucketing integration is ready for review now @btovar @dthain

dthain (commented):
I will defer to @btovar on the overall structure of the work.
Two minor things:
- Remove the generated files chirp.py and taskvine/src/examples from this PR.
- Make similar changes to vine_manager.c and vine_txn_log.c as you did in work_queue.c

tphung3:
@btovar Where would I put the integration tests in the repository? `dttools/test` is for the algorithms themselves and `taskvine/test/` doesn't look like the place either.

btovar:
There are such tests in `work_queue/test`. See `TR_work_queue_allocations.sh`.

tphung3:
The build-osx test seems to fail because of a task pulling a tarball from ccl website. Otherwise @btovar can you provide some feedbacks please?

dttools/src/bucketing_manager.c
```diff
@@ -15,16 +15,25 @@ static char* int_to_string(int n)
...
 }
 
+/* Add default resource types with default values per resource type to manager
+ * See this function's definition for more info
+ * @param m the bucketing manager */
+static void bucketing_manager_add_default_resource_types(bucketing_manager_t* m)
+{
+    bucketing_manager_add_resource_type(m, "cores", 0, 1, 10, 2, 10, 1);
+    bucketing_manager_add_resource_type(m, "memory", 0, 1000, 10, 2, 10, 1);
+    bucketing_manager_add_resource_type(m, "disk", 0, 1000, 10, 2, 10, 1);
```
btovar, line 25: gpus?

dttools/src/bucketing_manager.c
```diff
@@ -52,23 +68,51 @@ void bucketing_manager_delete(bucketing_manager_t* m)
...
 
-    /* only add resource type if it doesn't exist, do nothing otherwise */
+    bucketing_state_t* b; 
+    /* only add resource type if it doesn't exist, warn otherwise */
     if (!hash_table_lookup(m->res_type_to_bucketing_state, r))
     {
-        bucketing_state_t* b = bucketing_state_create(default_value, num_sampling_points, increase_rate, max_num_buckets, m->mode);
+        if (set_default)
+        {
+            if (!strncmp(r, "cores", strlen("cores")))
```
btovar, line 87: Use `strcmp` and drop the `strlen`. (Here and below.)

dttools/src/bucketing_manager.c
```diff
@@ -52,23 +68,51 @@ void bucketing_manager_delete(bucketing_manager_t* m)
...
+            {
+                b = bucketing_state_create(1, 10, 2, 10, m->mode, 1);
```
btovar, line 89: These parameters should be named, maybe as const at the top of the file, like `bucket_default_....`

dttools/src/category.c
```diff
@@ -133,6 +136,14 @@ void category_specify_first_allocation_guess(struct category *c, const struct rm
 	rmsummary_merge_max(c->first_allocation, s);
 }
 
+int category_in_bucketing_mode(struct category* c)
+{
+    if (c->allocation_mode == CATEGORY_ALLOCATION_MODE_GREEDY_BUCKETING ||
```
btovar, line 141: BUCKETING_MODE_GREEDY, etc?

taskvine/src/manager/vine_manager.c
```diff
@@ -4862,6 +4879,8 @@ int vine_set_category_mode(struct vine_manager *q, const char *category, vine_ca
 		case CATEGORY_ALLOCATION_MODE_MAX:
 		case CATEGORY_ALLOCATION_MODE_MIN_WASTE:
 		case CATEGORY_ALLOCATION_MODE_MAX_THROUGHPUT:
+		case CATEGORY_ALLOCATION_MODE_GREEDY_BUCKETING:
+		case CATEGORY_ALLOCATION_MODE_EXHAUSTIVE_BUCKETING:
```
btovar, line 4883: When to use `BUCKETING_` and when to use `CATEGORY_`?

btovar (changes requested):

dttools/src/bucketing.c
```diff
@@ -227,7 +288,7 @@ double bucketing_predict(bucketing_state_t* s, double prev_val)
     /* in sampling phase */
     if (s->in_sampling_phase)
     {
-        /* if new or no resource, return default value */
+        /* if new or empty resource, return default value */
         if (prev_val == -1 || prev_val == 0)
```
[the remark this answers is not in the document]
tphung3: The next else if in [here](https://github.com/cooperative-computing-lab/cctools/blob/d2b0dc17a2b70de7091f35876d343bf0aa3c01a7/dttools/src/bucketing.c#L296) guards prev_val from negative values.

dttools/src/bucketing.h
```diff
@@ -1,9 +1,15 @@
 #ifndef BUCKETING_H
 #define BUCKETING_H
 
-#include "category.h"
 #include "list.h"
 
+/* all modes of bucketing */
+typedef enum
+{
```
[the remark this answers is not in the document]
tphung3: Is it a strict requirement?

dttools/src/bucketing_exhaust.c
```diff
@@ -239,7 +239,12 @@ static struct list* bucketing_exhaust_get_buckets(bucketing_state_t* s, int n)
         
     /* fill candidate values with linear increase */
     for (int i = 0; i < n; ++i) 
-        candidate_vals[i + steps] = max_val * (i + 1) / (n * 1.0);
+    {
+        if (i == n - 1)
+            candidate_vals[i + steps] = max_val;
+        else
+            candidate_vals[i + steps] = max_val * (i + 1) / (n * 1.0);
```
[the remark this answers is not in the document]
tphung3: No by the for loop at [line 326](https://github.com/cooperative-computing-lab/cctools/blob/d2b0dc17a2b70de7091f35876d343bf0aa3c01a7/dttools/src/bucketing_exhaust.c#L326)

dttools/src/bucketing.c
```diff
@@ -181,6 +204,44 @@ void bucketing_state_delete(bucketing_state_t* s)
         list_delete(s->sorted_buckets);
         free(s);
     }
+    else
+        warn(D_BUCKETING, "Ignoring command to delete empty bucketing state\n");
+}
+
+void bucketing_state_tune(bucketing_state_t* s, const char* field, void* val)
+{
+    if (!s)
```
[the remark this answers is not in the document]
btovar: The spacing matches the rest of the code, which makes easy to read quickly, and to make global changes in the rare cases they are needed.

As for adding the braces, it makes clear the intention, it makes easy to add other statements to the if in the future, it guards against macros that expand to multiple lines, and protects against confusing lines where indentation matches (e.g., someone just coming from python.)

dttools/src/bucketing.c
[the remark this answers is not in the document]
btovar: I don't think we need that second guard. Simplify by changing 292 to preval <= 0.

From the resource_monitor perspective all negative values are the same as -1.

dttools/src/bucketing.h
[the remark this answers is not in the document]
btovar: You should match the style of the rest of the code as much as possible.

dttools/src/category.c
tphung3: BUCKETING_MODE_GREEDY is used internally to decide which algorithm to compute the next bucketing state [here](https://github.com/cooperative-computing-lab/cctools/blob/d2b0dc17a2b70de7091f35876d343bf0aa3c01a7/dttools/src/bucketing.c#L111). Category code I think should still use CATEGORY_ALLOCATION_...

taskvine/src/manager/vine_manager.c
tphung3: BUCKETING_ is internal to bucketing code (decide how to recompute bucketing state), CATEGORY_ is used to connect bucketing code to category code.

dttools/src/bucketing.c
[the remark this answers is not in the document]
tphung3: The second guard also serves to flag error in bucketing_predict as bucketing_predict should never predict a negative value in [here](https://github.com/cooperative-computing-lab/cctools/blob/d2b0dc17a2b70de7091f35876d343bf0aa3c01a7/dttools/src/bucketing_manager.c#L248)

dttools/src/bucketing.c
```diff
@@ -142,32 +140,47 @@ bucketing_bucket_t* bucketing_bucket_create(double val, double prob)
...
     if (max_num_buckets < 1 && mode == BUCKETING_MODE_EXHAUSTIVE)
+    {
         warn(D_BUCKETING, "The maximum number of buckets for exhaustive bucketing must be at least 1\n");
+        max_num_buckets = 1;
+    }
     
     if (mode != BUCKETING_MODE_GREEDY && mode != BUCKETING_MODE_EXHAUSTIVE)
+    {
         fatal("Invalid bucketing mode\n");
+        mode = BUCKETING_MODE_GREEDY;
```
btovar, line 176: fatal exits the program. Do you mean warn? If you mean fatal, then line 176 can be removed.

dttools/src/bucketing.c
```diff
@@ -142,32 +140,47 @@ bucketing_bucket_t* bucketing_bucket_create(double val, double prob)
...
+    }
 
     if (update_epoch < 1)
+    {
         fatal("Update epoch for bucketing cannot be less than 1\n");
+        update_epoch = 1;
```
btovar, line 182: Same as above.

dttools/src/bucketing.c
```diff
@@ -204,44 +217,46 @@ void bucketing_state_delete(bucketing_state_t* s)
...
 
-    if (!val)
-    {
+    if (!val){
         fatal("No value to tune field %s in bucketing state to\n", field);
         return;
     }
 
-    if (!strncmp(field, "default_value", strlen("default_value")))
+    if (!strncmp(field, "default_value", strlen("default_value"))){
```
btovar, line 239: `if(...) {` here and below.

dttools/src/bucketing.c
```diff
@@ -204,44 +217,46 @@ void bucketing_state_delete(bucketing_state_t* s)
...
+    }
+    else if (!strncmp(field, "mode", strlen("mode"))){
         s->num_sampling_points = *((bucketing_mode_t*) val);
-    else if (!strncmp(field, "update_epoch", strlen("update_epoch")))
+    }
+    else if (!strncmp(field, "update_epoch", strlen("update_epoch"))){
         s->update_epoch = *((int*) val);
-    else
+    }
+    else{
```
btovar, line 257: `else {` here and below.

dttools/src/bucketing.h
```diff
@@ -4,15 +4,13 @@
...
-{
+typedef enum{
     BUCKETING_MODE_GREEDY,
     BUCKETING_MODE_EXHAUSTIVE
 } bucketing_mode_t;
 
 /* Bucketing has two operations, add and predict */
-typedef enum
-{
+typedef enum{
```
btovar, line 13: `typedef enum {`

dttools/src/bucketing_exhaust.c
```diff
@@ -234,16 +234,20 @@ static struct list* bucketing_exhaust_get_buckets(bucketing_state_t* s, int n)
     double candidate_vals[steps + n];
 
     /* fill candidate values with logarithmic increase */
-    for (int i = 0; i < steps; ++i)
+    for (int i = 0; i < steps; ++i){
```
btovar, line 237: `...) {`

btovar (changes requested):

btovar:
Looks good! Just some more aesthetic changes.

tphung3:
@btovar This PR is now ready to be merged.

work_queue/src/work_queue.c
```diff
@@ -8008,7 +8009,13 @@ void work_queue_accumulate_task(struct work_queue *q, struct work_queue_task *t)
 		case WORK_QUEUE_RESULT_TASK_MAX_RUN_TIME:
 		case WORK_QUEUE_RESULT_DISK_ALLOC_FULL:
 		case WORK_QUEUE_RESULT_OUTPUT_TRANSFER_ERROR:
-			if(category_accumulate_summary(c, t->resources_measured, q->current_max_worker)) {
+			if (t->result == WORK_QUEUE_RESULT_SUCCESS)
```
btovar, line 8012: Add braces to all these ifs.

dttools/src/bucketing.c
```diff
@@ -183,6 +219,46 @@ void bucketing_state_delete(bucketing_state_t* s)
     }
 }
 
+void bucketing_state_tune(bucketing_state_t* s, const char* field, void* val)
+{
+    if (!s){
```
btovar, line 224: Add space after ), e.g.: `if(!s) {`

btovar (changes requested):

dttools/src/bucketing_greedy.c
```diff
@@ -88,6 +88,8 @@ static void bucketing_bucket_range_delete(bucketing_bucket_range_t* range)
         bucketing_cursor_w_pos_delete(range->hi);
         free(range);
     }
+    else
```
btovar, line 91: Add braces

btovar (changes requested):

btovar (approved):

PR #3058, opened 2022-11-28
adds unponcho mini task

BarrySlyDelgado:
added the function for a poncho unpack mini task

### DIFF
--- a/taskvine/src/bindings/python3/taskvine.binding.py
+++ b/taskvine/src/bindings/python3/taskvine.binding.py
@@ -177,7 +177,7 @@ class FileUntgz(File):
     def __init__(self,subfile):
         self._file = vine_file_untgz(vine_file_clone(subfile._file))
 
-class FileUnponcho(File)
+class FileUnponcho(File):
     ##
     # Create a file by unpacking a poncho package.
     #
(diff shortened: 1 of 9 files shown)


dthain (changes requested):
Please make sure that you do a minimal build and test before pushing, I see at least one obvious typo that would prevent a good build.

Also, please add the corresponding bits to the Python interface in taskvine.binding.py.

BarrySlyDelgado:
The minitask and tasks from `vine_example_unponcho.c` complete successfully however, it seems that tasks are not reusing the unpacked package from the minitask this is evident from the absence of `logmsg directory ${UNPACK_TO} is not empty. Not expanding environment file again` in the output. Investigating this now.

BarrySlyDelgado:
Initially thought it wasn't working as intended but after a lot of gdb I realized it was actually working. The issue was that I didn't realize that doing `pp_run -e <directory>` doesn't produce the log message i thought it did but rather `pp_run -u <directory> -e <env>` does. Added an example with and without unponcho which is nicely just a change in 2 lines. working on python interface and this should be wrapped up.

BarrySlyDelgado:
@dthain This should be ready now.

dthain (approved):

dthain:
LGTM

PR #3067, opened 2022-12-02
Vine: Remove unpack flags in favor of mini-tasks.

dthain:


btovar:
Suggest changing the suffix untgz to untar.
Since we are using the `tar` executable and `tar` detects which file type it operates on, it  vine_file_untar could be fed .tar, tar.gz, and others.

dthain:
Wow, I had no idea that tar could do that.  I s that new or has it been there all along?
The only complication I see is that the mini-task sets up its own namespace.
So, UnTgz presents the files as input.tar.gz, while Untar presents it as input.tar:
https://github.com/cooperative-computing-lab/cctools/blob/master/taskvine/src/manager/vine_file.c#L181

btovar:
As far as I can see, it ignores the extension, so you may say something like `tar -C target_dir -xf input_tar_file`.

I think it has been like that for a long while, as the ability to pipe stdin to tar has been there for a long time.

btovar (approved):

PR #3069, opened 2022-12-02
Taskvine: New Cachenames

BarrySlyDelgado:
Adds content based cachenames for files in Taskvine.

- Currently hashing a files content is very slow (~4secs) to submit a single task to a worker. 
- Can use other options other than md5


dthain (changes requested):

BarrySlyDelgado:
Added generating cachenames for URLs. Here are somethings to consider:
- If we cannot get any relevant meta data from the header should the file be cached at all?(The argument against caching it would be that using things like just the filename is contrary to the problem were trying to solve here)
- Should we just use the  hashes given to us? My implementation is a bit lazy and just hashes the line.
- Using You need extra info with Last-Modified  as two items can have the same timestamp.  right now we are just hashing the timestamp + the URL source

taskvine/src/manager/vine_file.c
```diff
@@ -37,6 +40,107 @@ In each of the other file types, a similar approach is taken,
...
+	char line[STR_MAX];
+	char hash_src[STR_MAX];
+	char * buffer;
+	unsigned char digest[MD5_DIGEST_LENGTH];
+
+	char tmp[] = "curl_tmp.XXXXXX";
+	mkstemp(tmp); 
+
+	char * command = string_format("curl -I -sSL --stderr /dev/stdout -o \"%s\" \"%s\"", tmp, f->source);
+	result = do_command(command, &buffer);
```
dthain, line 78: This seems overly complicated.  Just use `popen` to run the command and read its output with `fgets` and `sscanf`.  That way you don't have to generate a temporary file and clean it up.

dthain:
These are good questions.  I think it would be helpful if you can write out in prose all of the possible cases and indicate how the hash will be computed.  Big that in a big comment at the top.

Some observations about the code:
- You are running the command, sending its output to a file, and then opening the file and reading it.  (But not closing it or deleting it.). It would be much simpler to just popen() the command and read its output with fgets() and sscanf().
- You are returning with a hash as soon as you match any one header.  This probably isn't right b/c multiple headers may be present in any order.  Better to accumulate information and then choose what to use in the hash at the bottom.
- If you open a file, you better close it!

dthain (changes requested):

taskvine/src/manager/taskvine.h
```diff
@@ -38,6 +38,14 @@ expected events.
 #define VINE_RANDOM_PORT  0                  /**< Indicates that any port may be chosen. */
 #define VINE_WAITFORTASK  -1                 /**< Timeout value to wait for a task to complete before returning. */
 
+/** **/
```
dthain, line 41: Yes, this is the right idea, but it is an implementation detail and not part of the public API, so don't put it in `taskvine.h`, just put it at the top of `vine_file.c`

taskvine/src/manager/vine_file.c
```diff
@@ -103,11 +103,28 @@ should not be cached. That is, generating a cachename could lead to possible con
...
+	}
+	return 0;
+}
+
+char * vine_task_to_json(const struct vine_file *f){
 	char * buffer;
 	char * file_buffer;
 	struct vine_file * file;
-	unsigned char digest[MD5_DIGEST_LENGTH];
 	buffer = string_format("{\ncmd = \"%s\"\n", f->source);
```
dthain, line 128: I think it is an accident that `f->source` contains the command.
Better to use `f->mini_task->cmd` instead.

taskvine/src/manager/vine_file.c
```diff
@@ -103,11 +103,28 @@ should not be cached. That is, generating a cachename could lead to possible con
...
 	if(f->mini_task->input_files){
 		buffer = string_combine(buffer, "inputs = ");
```
dthain, line 130: One thing that we have to be careful of here is that lists need to appear in the same order every time, so that we get consistent hashes of mini tasks.  At least drop a comment here so we don't forget to do it.

taskvine/src/manager/vine_file.c
```diff
@@ -248,21 +263,21 @@ char *make_cached_name( const struct vine_file *f )
...
 		case VINE_FILE:
-			return string_format("file-%d-%s-%s", cache_file_id, hash, source_enc);
+			return string_format("file-%d-md5-%s-%s", cache_file_id, hash, source_enc);
 			break;
 		case VINE_EMPTY_DIR:
-			return string_format("file-%d-%s-%s", cache_file_id, md5_string(digest), source_enc);
+			return string_format("file-%d-md5-%s-%s", cache_file_id, md5_string(digest), source_enc);
 			break;
 		case VINE_MINI_TASK:
 			/* XXX This should be computed from the constituents of the mini task */
```
dthain, line 272: If the comment doesn't apply any more, get rid of it.

taskvine/src/manager/vine_file.c
```diff
@@ -299,7 +314,13 @@ struct vine_file *vine_file_create(const char *source, const char *remote_name,
 	if(vine_hack_do_not_compute_cached_name) {
   		f->cached_name = xxstrdup(f->source);
 	} else {
-		f->cached_name = make_cached_name(f);
+		char * cache_name = make_cached_name(f);
+		if(cache_name == 0 && type == VINE_URL){
+			struct vine_file *new_file = retrieve_url(f);
```
dthain, line 319: And what if `retrieve_url` fails?

dthain (changes requested):

taskvine/src/manager/vine_file.c
```diff
@@ -38,20 +40,76 @@ including a hash and a name where one is known, or another
 unique identifier where no name is available.
 */
 
+
+/* creates a hash for a single file or directory */
+
+const char * merkle(char * src){
```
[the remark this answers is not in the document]
BarrySlyDelgado: This is resolved.

taskvine/src/manager/vine_file.c
```diff
@@ -38,20 +41,195 @@ including a hash and a name where one is known, or another
...
+dependencies which have their own cachenames. There is the  possibility of generating the cachename for the minitask from the cachenames of the files
+that the mini task depends on. However, certain commands have a level of dynanicism in which it we cannot use this method to adequately predict the identity
+of the resulting file. It could be a possibility to just let the user decide whether a command can have a cachename. However, what happens if they are wrong?
+
+When a cachename cannot be generated - There is an argument that if cachename cannot be generated given the present information, that the file
+should not be cached. That is, generating a cachename could lead to possible conflicts on the worker side.
+
+
+*/
+const char * make_mt_cached_name(const struct vine_file *f){
```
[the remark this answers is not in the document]
BarrySlyDelgado: This is resolved.

taskvine/src/manager/vine_file.c
```diff
@@ -38,20 +41,195 @@ including a hash and a name where one is known, or another
...
+	char line[STR_MAX];
+	char hash_src[STR_MAX];
+	char * buffer;
+	unsigned char digest[MD5_DIGEST_LENGTH];
+
+	char * command = string_format("curl -I -sSL i --verbose --stderr /dev/stdout \"%s\"", f->source);
+	FILE * stream = popen(command, "r");
+	while(fgets(line, STR_MAX, stream) != NULL){
+		if(sscanf(line, "Content-MD5: %s", hash_src)){
+			val = 3;
```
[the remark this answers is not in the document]
BarrySlyDelgado: resolved.

taskvine/src/manager/taskvine.h
BarrySlyDelgado: moved.

dthain:
Last thing now is to do something when retrieve_url fails.

dthain (approved):

dthain:
Almost there, looks like you have one more conflict...

BarrySlyDelgado:
Forgot about cache names for outputs. pretty sure that's what is causing the build error. fixing it now.

BarrySlyDelgado:
Build Test passes now, Output file cache name is currently generated by a random number with the prefix "of". There is a check for this prefix in `vine_task_add_input` where we can tell if an input file's cache name could not be generated but nothing is done there at the moment as it caused the build tests to fail.

dthain:
The place to check for failures (under the current regime) is`vine_manager_put_input_file_if_not_cached`.
Are you checking there?

dthain (changes requested):
Good, almost there.  One minor item that bugs me.  Use "output" instead of "of" for the cache file name.  Avoid abbreviations when a plain word is explanatory.

dthain (approved):

dttools/src/md5.c
```diff
@@ -367,6 +369,88 @@ const char *md5_string(unsigned char digest[16])
 	str[32] = 0;
 	return str;
 }
+const char * md5_dir_rec(char * src){
```
btovar, line 372: It seems there is a combination of spaces and tabs when indenting these functions. Just make sure the style matches the rest of the file (tabs).

btovar (changes requested):

dttools/src/md5.c
BarrySlyDelgado: Fixed.

taskvine/src/manager/vine_file.c
```diff
@@ -38,24 +41,190 @@ including a hash and a name where one is known, or another
...
+		if(val < FOUND_LM && sscanf(line, "last-modified: %s", hash_src)){
+			val = FOUND_LM;
+		}
+	}
+	int exit_status = pclose(stream);
+	char * prefix;
+	if(exit_status != 0) return 0; // Curl executes but retuns an error
+	switch(val){
+		case FOUND_NONE:
+			return 0;
```
btovar, line 179: Add `break;` to all these cases, otherwise the compiler may complain.

btovar (changes requested):

dttools/src/md5.c
```diff
@@ -406,46 +423,31 @@ const char * md5_dir_rec(char * src){
 						return 0;
 					}
 				}
+			free(path);
 		}
 		closedir(dir);
 		return hash;
 	}
-	else if(S_ISREG(info.st_mode)){
```
btovar: I think this removal is incorrect. The branches should stay the same (if/else).

The only part I suggested to move before the if is: `unsigned char digest[MD5_DIGEST_LENGTH];`

dttools/src/md5.h
```diff
@@ -54,11 +54,17 @@ which  must be converted to a human readable form with @ref md5_string.
 
 int md5_file(const char *filename, unsigned char digest[MD5_DIGEST_LENGTH]);
 
+/** Converts either a file or directory to a string of the its hashed contents.
+@param src The path to source file or a directory
+@returns A static pointer to the string;
+*/
+
```
btovar, line 61: remove line so that it is clear the comment is for the declaration below.

taskvine/src/manager/vine_file.c
```diff
@@ -128,68 +128,72 @@ struct vine_file * retrieve_url(const struct vine_file *f){
...
-		case FOUND_NONE:
+		case VINE_FOUND_NONE:
 			return 0;
-		case FOUND_LM:
+			break;
+		case VINE_FOUND_LM:
 			prefix = string_format("LM-md5-");
 			buffer = string_combine(hash_src, f->source); // + server
 			md5_buffer(buffer,strlen(buffer),digest);
                         return string_combine(prefix, md5_string(digest));
```
btovar, line 185: break;

taskvine/src/manager/vine_file.c
```diff
@@ -128,68 +128,72 @@ struct vine_file * retrieve_url(const struct vine_file *f){
...
-		case FOUND_ET:
+			break;
+		case VINE_FOUND_ET:
 			prefix = string_format("ET-md5-");
 			md5_buffer(hash_src, (strlen(hash_src)), digest); // + server
                         return string_combine(prefix, md5_string(digest));
```
btovar, line 190: break;

btovar (changes requested):

taskvine/src/manager/vine_file.c
BarrySlyDelgado: Did I misplace the`break`?

btovar: It is good practice to add a break to each case. Thus:
```C
			return 0;
			break;
		case FOUND_LM:
			break;
		case VINE_FOUND_LM:
			prefix = string_format("LM-md5-");
			buffer = string_combine(hash_src, f->source); // + server
			md5_buffer(buffer,strlen(buffer),digest);
                        return string_combine(prefix, md5_string(digest));
                        break;
```
etc.

One reason is that a future change may remove one of the `return` statements, and the ensuing fall-through will cause a bug difficult to detect.

BarrySlyDelgado: I believe I've added them for each of the four cases.

btovar: Yes, I see that now. I think I got confused and saw an added `break` to a case that was deleted.

dthain:
This PR is now rolled into #3120

PR #3073, opened 2022-12-10
Updated python abstractions and created examples

JosephDuggan7:


dthain:
@JosephDuggan7 the examples here look nice, but the PR contains a number of binary objects and log files that shouldn't be checked in, please remove them.

dthain (changes requested):
Please remove the unneeded files from the PR.

JosephDuggan7:
Removed unneeded files

taskvine/src/examples/vine_example_pair.py
```diff
@@ -0,0 +1,28 @@
+import taskvine as vine
```
btovar, line 1: This example fails with:
```
TypeError: 'PythonTaskNoResult' object is not iterable
```

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1980,10 +1983,15 @@ def fpairs(fn, s):
...
-                t = self.wait_for_tag(str(i), 10)
+                for key, value in tasks.items():
+                    if value == i:
+                        t_id = key
+                        break
+                t = self.wait_for_task_id(t_id, 10)
 
                 if t:
-                    results[tasks[vine_task_get_id(t)]] = vine_task_get_stdout(t)
+                    print(t.output)
```
btovar, line 1993: debugging statement?

btovar (changes requested):

JosephDuggan7:
Removed print statement, however I am not encountering any errors with that example

PR #3080, opened 2023-01-10
Vine Worker - Persistent Caching

BarrySlyDelgado:
This adds persistent cache for the worker.

The cache directory is created using the user's UID. It was previously created with the UID and PID which would have acaused the cache to be different most times. The worker creates the cache directory if it doesn't alreadt exists.

This adds two functions vine_cache_load and vine_init_update

- vine_cache_load lods all the files into the worker's cache
- vine cache_init_update sends cache updates to the manager for all files found within the worker's persistent cache

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -405,7 +405,11 @@ static vine_msg_code_t vine_manager_recv_no_retry(struct vine_manager *q, struct
 		result = handle_cache_update(q, w, line);
 	} else if (string_prefix_is(line, "cache-invalid")) {
 		result = handle_cache_invalid(q, w, line);
-	} else if (string_prefix_is(line, "transfer-address")) {
+	} 
+//	else if (string_prefix_is(line, "worker-init")) {
+//		result = handle_worker_init(q, w, line);
+//	} 
+	else if (string_prefix_is(line, "transfer-address")) {
 		result = handle_transfer_address(q, w, line);
 	} else if( sscanf(line,"GET %s HTTP/%*d.%*d",path)==1) {
 	        result = handle_http_request(q,w,path,stoptime);
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -20,6 +20,7 @@ See the file COPYING for details.
 #include "vine_blocklist.h"
 #include "vine_txn_log.h"
 #include "vine_perf_log.h"
+#include "vine_current_transfers.h"
 
 #include "cctools.h"
 #include "envtools.h"
@@ -130,6 +131,9 @@ void vine_disable_monitoring(struct vine_manager *q);
 static void aggregate_workers_resources( struct vine_manager *q, struct vine_resources *total, struct hash_table *features);
 static struct vine_task *vine_wait_internal(struct vine_manager *q, int timeout, const char *tag, int task_id);
 static void release_all_workers( struct vine_manager *q );
+
+static void vine_manager_send_duty_to_workers(struct vine_manager *q, const char *name);
+static void vine_manager_send_duties_to_workers(struct vine_manager *q);
 
 /* Return the number of workers matching a given type: WORKER, STATUS, etc */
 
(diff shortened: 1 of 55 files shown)


taskvine/src/worker/vine_cache.c
```diff
@@ -69,9 +70,62 @@ struct vine_cache * vine_cache_create( const char *cache_dir )
...
+void vine_cache_load(struct vine_cache *c)
+{
+	DIR *dir = opendir(c->cache_dir);
+	if(dir){
+		debug(D_VINE, "loading cache at: %s", c->cache_dir);
+		struct dirent *d;
+		while((d=readdir(dir))){
+			debug(D_VINE, "found %s in cache at: %s",d->d_name, c->cache_dir);
+			struct stat info;
+			char *cache_path = vine_cache_full_path(c,d->d_name);
```
dthain, line 84: If you create it, delete it.

taskvine/src/manager/vine_manager.c
```diff
@@ -405,7 +405,11 @@ static vine_msg_code_t vine_manager_recv_no_retry(struct vine_manager *q, struct
 		result = handle_cache_update(q, w, line);
 	} else if (string_prefix_is(line, "cache-invalid")) {
 		result = handle_cache_invalid(q, w, line);
-	} else if (string_prefix_is(line, "transfer-address")) {
+	}
```
dthain, line 408: This seems incomplete.  Did you mean to finish this, or is it meant to be part of a future PR?

taskvine/src/manager/vine_worker_info.h
```diff
@@ -47,6 +47,7 @@ struct vine_worker_info {
 	int  transfer_port_active;
 
 	/* Worker condition that may affect task start or cancellation. */
+	int  worker_init;                       // if 1, worker does not accept tasks.
```
dthain, line 50: Together `worker_init` and  `draining` are likely to be confusing to a future reader. Instead, combine these into a single state variable with names something like `INITIALIZING, WORKING, DRAINING` and that makes it much clearer that a worker has a lifecycle in which it may or may not accept tasks.

taskvine/src/worker/vine_cache.c
```diff
@@ -69,9 +70,62 @@ struct vine_cache * vine_cache_create( const char *cache_dir )
 }
 
 /*
-Delete the cache manager structure, though not the underlying files.
+Load exsiting cache directory into cache structure.
```
dthain, line 73: Typo

taskvine/src/worker/vine_cache.c
```diff
@@ -69,9 +70,62 @@ struct vine_cache * vine_cache_create( const char *cache_dir )
...
+	DIR *dir = opendir(c->cache_dir);
+	if(dir){
+		debug(D_VINE, "loading cache at: %s", c->cache_dir);
+		struct dirent *d;
+		while((d=readdir(dir))){
+			debug(D_VINE, "found %s in cache at: %s",d->d_name, c->cache_dir);
+			struct stat info;
+			char *cache_path = vine_cache_full_path(c,d->d_name);
+			if(stat(cache_path, &info)==0){
+				if(!strncmp(d->d_name, "file", strlen("file"))){
```
dthain, line 86: If the stored object is a directory and not just a plain file, then we need to measure the size recursively, and not just the simple size from info.st_size, right?  (I'm sure we have a function to do that already.)

taskvine/src/worker/vine_worker.c
```diff
@@ -1454,7 +1465,7 @@ static void workspace_delete()
 	is inside the workspace.  Abort if we really cannot clean up.
 	*/
 
-	unlink_recursive(workspace);
+//	unlink_recursive(workspace);
```
dthain, line 1468: Don't check in dead code, just delete it.

dthain (changes requested):
On the right track, a few things to fix up.
I think your build failure on Linux is due to a memory leak picked up by valgrind.

taskvine/src/worker/vine_cache.c
```diff
@@ -69,9 +70,79 @@ struct vine_cache * vine_cache_create( const char *cache_dir )
...
+	if(dir){
+		debug(D_VINE, "loading cache at: %s", c->cache_dir);
+		struct dirent *d;
+		while((d=readdir(dir))){
+			debug(D_VINE, "found %s in cache at: %s",d->d_name, c->cache_dir);
+			struct stat info;
+			int64_t nbytes, nfiles;
+			char *cache_path = vine_cache_full_path(c,d->d_name);
+			if(stat(cache_path, &info)==0){
+				if(!strncmp(d->d_name, "file", strlen("file"))){
```
dthain, line 87: Can you make this even simpler?
It seems like the same action is taken no matter what the name pattern is.
And for that matter, the manager might change the naming pattern, and I'm not sure that the worker should care.

dthain (changes requested):

taskvine/src/worker/vine_cache.c
BarrySlyDelgado: This has been condensed.

dthain:
@BarrySlyDelgado this is looking good, are you ready to merge?

BarrySlyDelgado:
It should be once this build  passes. Had to add checks for ./ and ../ in loading the cache which i somehow deleted.

dthain (approved):

PR #3081, opened 2023-01-11
Adds statx to parrot

btovar:
As observed in NDCMS+parrot in rh8,  utilities like ls and stat now use the call statx.

Local service redirects to statx.
cvmfs uses stat, and then copies relevant info to statx structures.

The rest of the services use a dummy entry, as with stat.


dthain:
Ben, this is looking good overall.  Please add two things:
1 - A test case that explicitly hits the new and old system calls.
2 - Some commentary that explains whether and why a filesystem driver should implement stat as a special case of statx, or the other way around, or both separately.

dthain (approved):

PR #3084, opened 2023-01-13
wq factory runos simplification

btovar:
Simplified to always run the worker contained in the image. This is a transitional change while we determine the best way to declare images in the factory for general use (not only in ND).

I also rewrote a shorter version runos in bash. The old runos was a python2 program, and python2 is not available in the rhel8 installations at ND. We could update it to python3, but then it wouldn't work in the machines still running rhel7. (So we end up needing a runos for runos...)


dthain:
IMHO, simple is good here.  Just make a narrow feature that meets the immediate need.

dthain (approved):

PR #3088, opened 2023-01-27
Vine transaction log modifications for tracking peer transfers

colinthomas-z80:


colinthomas-z80:
Grabbing the last transfer ID submitted through the global variable bothers me a bit but for logging purposes it does the job.

PR #3089, opened 2023-01-27
Vine: modify transaction log to track peer transfers

colinthomas-z80:
Added a global variable in vine_current_transfers for logging purposes. it does what is needed; getting the last transfer ID to print to the transactions log.

Also fixed a peer transfer issue when a mini_task specification would get scheduled as a peer transfer. As it is now, I don't think mini task specifications will be able to be peer transferred, as the peer transfer method changes all file types to VINE_URL. As it should be... since we are substituting a URL. However; changing a mini task to type URL removes any indication to the manager that it is a mini task. So it just becomes a text file containing the task specification that the worker does nothing with.

### DIFF
--- a/doc/manuals/install/index.md
+++ b/doc/manuals/install/index.md
@@ -101,7 +101,7 @@ the software repository and build it:
 ```sh
 git clone git://github.com/cooperative-computing-lab/cctools.git cctools-src
 cd cctools-src
-./configure --with-base-dir $CONDA_PREFIX --prefix $HOME/cctools
+./configure --with-base-dir $CONDA_PREFIX --prefix $CONDA_PREFIX
 make
 make install
 ```


taskvine/src/manager/vine_manager.c
```diff
@@ -316,8 +316,9 @@ static int handle_cache_update( struct vine_manager *q, struct vine_worker_info
 		remote_info->in_cache = 1;
 		
 		vine_current_transfers_remove(q, id);
-
-		vine_txn_log_write_cache_update(q,w,size,transfer_time,cachename);
+		
+	   /* Do not write a cache-update to the log without an ID . such as a mini-task or buffer */
+	   if(strcmp(id, "X") != 0)vine_txn_log_write_cache_update(q,w,size,transfer_time,cachename, id);
```
dthain, line 321: This doesn't look right: we **do** want cache-update messages for mini-tasks and other things in the transaction log, at the very least so we can visualize when the worker is taking these actions.  (See Barry's latest viz graph as an example.)

taskvine/src/manager/vine_manager_put.c
```diff
@@ -287,7 +287,8 @@ static vine_result_code_t vine_manager_put_input_file(struct vine_manager *q, st
 		q->stats->bytes_sent += total_bytes;
 
 		// Write to the transaction log.
-		vine_txn_log_write_transfer(q, w, t, f, total_bytes, elapsed_time, 1);
+		char *id = vine_current_transfers_get_last_id(q);
+		if(t->task_id != 0) vine_txn_log_write_transfer(q, w, t, f, total_bytes, elapsed_time, 1, id);
```
dthain, line 291: Hmm, I don't get this either.  Why would we not want to log files transferred as inputs to a mini-task?

taskvine/src/manager/vine_txn_log.h
```diff
@@ -21,8 +21,8 @@ void vine_txn_log_write(struct vine_manager *q, const char *str);
 void vine_txn_log_write_task(struct vine_manager *q, struct vine_task *t);
 void vine_txn_log_write_category(struct vine_manager *q, struct category *c);
 void vine_txn_log_write_worker(struct vine_manager *q, struct vine_worker_info *w, int leaving, vine_worker_disconnect_reason_t reason_leaving);
-void vine_txn_log_write_transfer(struct vine_manager *q, struct vine_worker_info *w, struct vine_task *t, struct vine_file *f, size_t size_in_bytes, int time_in_usecs, int is_input );
-void vine_txn_log_write_cache_update(struct vine_manager *q, struct vine_worker_info *w, size_t size_in_bytes, int time_in_usecs, const char *name );
+void vine_txn_log_write_transfer(struct vine_manager *q, struct vine_worker_info *w, struct vine_task *t, struct vine_file *f, size_t size_in_bytes, int time_in_usecs, int is_input, char *id );
```
dthain, line 24: const char *id

taskvine/src/manager/vine_manager.c
```diff
@@ -2655,22 +2656,31 @@ static int vine_manager_transfer_capacity_available(struct vine_manager *q, stru
...
-					f->substitute = vine_file_substitute_url(f,peer_source);
-					free(peer_source);
-					found_match = 1;
-					break;
-				} else {
-					free(peer_source);
-				}
+		/* If the file is a mini task specification, check its file dependencies for availability */
+		/* Do not peer transfer the mini task itself */
+		if(f->type==VINE_MINI_TASK) {
```
dthain, line 2662: I think this makes sense: we shouldn't peer-transfer the outputs of mini tasks.
But the control flow is confusing, each time I read it is a head scratcher.
Write it in the positive sense like this:
```
if(transfer transfer capacity available) {
  // do something
} else {
  // do something else
}
```

And if one of those is empty and falls through, then put int a big ol comment about what it is supposed to do.

dthain (commented):
Hmm, there is a lot going on here that it's not clear to me whether if fixes a problem or creates a new one.  Can you elaborate on why your are making these new changes.  See comments inline.

taskvine/src/manager/vine_manager.c
colinthomas-z80: Yes the scheduling order did become a bit confusing. The purpose here was to maintain the priority of scheduling peer transfers. However we need to make sure we don't peer transfer mini-tasks due to complications later down the line (transferring mini task input files is fine).

dthain:
Ok, I think I finally understand what's making this uncomfortable to write.  Once we introduced the notion of peer transfers, the meaning of log events implicitly changed.  Here is what we want:
- A TRANSFER log event should describe only a significant data transfer between the manager and the worker.  This includes regular file and buffer transfers.  It doesn't include urls, mini-tasks, and the other oddities, since those types just involve sending a tiny quantity of metadata to the worker.
- A CACHE-UPDATE event describes the worker adding something to its own cache that happened at the direction of the manager.  This includes mini-tasks, urls, and output files of tasks.  The manager wants to find out when these things completed, otherwise it won't know that they exist.

So, logging a transfer transaction at the end of `vine_manager_put_input_file` isn't the right thing, because it covers too many disparate cases.  What you should do is log a transfer transaction in the FILE case and in the BUFFER case, and not in the others.  And then those transfers don't need an ID, because they are simple manager->worker transfers.  And then you don't need the global var hack.

PR #3091, opened 2023-01-30
Workqueue __del__ issue fixed on last instance of a python task

mcarbona1:
Previously there was an issue of __del__ being called on a python task after os and sys had gone out of scope, causing the test case in the manual to error out at the end. By giving del two arguements (os and sys) with default arguments being the required libraries, Python will preserve references to the libraries, allowing __del__ to be called successfully.

### DIFF
--- a/work_queue/src/bindings/python3/work_queue.binding.py
+++ b/work_queue/src/bindings/python3/work_queue.binding.py
@@ -52,6 +52,7 @@ if cctools_tmpdir:
 else:
     staging_directory = tempfile.mkdtemp(prefix='wq-py-staging-')
 
+
 def cleanup_staging_directory():
     try:
         shutil.rmtree(staging_directory)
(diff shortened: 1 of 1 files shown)


btovar:
@mcarbona1 If I understand correctly, this ensures that there is a reference to the modules, so they are not  garbage collected. 

I wouldn't modify the signature (parameter list) of `__del__`, as its use is well documented in python. Instead, let's think of other solutions to perform this cleanup.

One idea is to make sure all these temp directories are created in the staging directory. The staging directory is cleaned up when the python exists (see `atexit`). Then, remove from the code the error message that uses `sys.stderr` when there is an exception with `rmtree`. In this way, the `__del__` can collect temp directories while the application is running, and we ensure all of the directories are collected with the `atexit` cleanup.

mcarbona1:
@btovar Sounds good! I'll take a look and see if I can get it working by tonight or sometime tomorrow!

mcarbona1:
@btovar I've removed the previous solution and checked the code, it looks like the worker files are already created in the temp file as far as I can tell ad are removed when the atexit function is called, so we should be good there. I just added an if statement in del so it would still report errors if its called before python is shut down (and therefore sys is still defined). If we don't want that it would be easy to pass in del instead, just let me know which implementation you think is best!

btovar (approved):
Looks good, thanks!

PR #3092, opened 2023-02-01
Vine: Serverless running on condor

David-Simonetti-ND:
Getting serverless to work on condor required changing the method of communication between coprocess and worker from using TCP to using pipes, as for whatever reason the task responsible for invoking the function call on the coprocess would not connect.

### DIFF
--- a/poncho/src/poncho/package_serverize.py
+++ b/poncho/src/poncho/package_serverize.py
@@ -55,42 +55,24 @@ def send_configuration(config):
     sys.stdout.flush()
 
 def main():
-    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    try:
-        # modify the port argument to be 0 to listen on an arbitrary port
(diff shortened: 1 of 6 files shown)


btovar:
Pipes made more sense in this case anyway, right?

btovar (commented):
There are debugging lines commented. They need to be removed before merging.

David-Simonetti-ND:
I added two extra bugfixes, one was that the resource allocation for the coprocess was always set to the default value, but now it is the resource values set for the duty in the manager. Another bugfix was that the coprocesses were never set to the correct process group, so the worker would stall forever trying to kill them once the manager finished.
Should be ready for merge.

dthain:
@David-Simonetti-ND since this is a major addition to the system, I would like to see a new test case in `taskvine/test` that creates a duty, installs it, invokes it, and gets the result back.

dthain (approved):

PR #3096, opened 2023-02-03
docker image spec and script for static parrot+cvmfs

btovar:


dthain (approved):
This looks like the right approach, good idea keeping it in the cctools repo.

PR #3097, opened 2023-02-06
Fix __del__ behavior in taskvine

mcarbona1:
Taskvine had the same del issues that workqueue had so I applied the same fix in the vine bindings. As far as I can tell the python task temp dirs are created in the same way as work queue so everything should be cleaned up regardless at exit!


btovar:
Thanks!

btovar (approved):

PR #3098, opened 2023-02-07
Vine: concurrent worker transfers

colinthomas-z80:


dthain (changes requested):
Yes, nice and simple.  However, there is one wrinkle -- as child processes exit, they will accumulate as zombies b/c nobody is cleaning them up.  The transfer server needs to periodically wait for children and clean then up when they are finished.  See the dttools process module for some helpers:
https://cctools.readthedocs.io/en/latest/api/html/process_8h.html

taskvine/src/worker/vine_transfer_server.c
```diff
@@ -51,11 +53,38 @@ static void vine_transfer_handler( struct link *lnk, struct vine_cache *cache )
...
+			if(lnk)
+			{
+				vine_transfer_handler(lnk,cache);
+				link_close(lnk);
+			}
+			_exit(0);
+		} 
+		else 
+		{
+			if(child_count < VINE_TRANSFER_PROC_MAX_CHILD) {
```
dthain, line 79: Hmm, this isn't quite right, you are skipping the `waitpid` and going back to the `accept`, which is going to result in an odd accumulation of zombies.

Try it this way:
- `count< limit` Do a non-blocking `waitpid` and go back to `accept`
- `count>=limit` ?  Do a blocking waitpid until a child finishes.

dthain (changes requested):

dthain (approved):

PR #3099, opened 2023-02-07
Vine logs per run

btovar:
By default, logs and staging files are written to:

```C
"vine-run-info/%Y-%m-%dT%H:%M:%S/logs/{debug,performance,transactions}"
"vine-run-info/%Y-%m-%dT%H:%M:%S/staging"
```

The prefix can be changed with:

```C
vine_set_runtime_info_path(path);   // where path is "vine-run-info" by default 
```

If the env variable `VINE_RUNTIME_INFO_DIR` is set, then logs and staging are written to that directory.

Logs are enabled by default and can't be disabled or renamed. The only aspect that can change is where they are written to (as above).

I added `Task.submit_finalize` to python, which is called by the manager in `submit` before the task is really submitted. This allows the task to know the staging directory of the manager and do the serialization of its files correctly.

The global staging directory in python was removed. Tasks use the one from their manager, and factories use one per factory.

### DIFF
--- a/doc/manuals/taskvine/index.md
+++ b/doc/manuals/taskvine/index.md
@@ -1556,56 +1556,66 @@ Pair: [2, 4, 6, 8, 4, 8, 12, 16, 6, 12, 18, 24, 8, 18, 24, 32]
 Tree: 8
 ```
 
+
 ## Logging facilities
 
-We can observe the lifetime of the queue through three different logs:
+A taskvine manager produces three logs: `debug`, `performance`, and
+`transactions`. These logs are always enabled, and appear in the current
+working directory in the sudirectories:
+
+```sh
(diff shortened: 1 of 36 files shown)


btovar:
Done with logs, working on staging.

btovar:
@David-Simonetti-ND Of interest is that `RemoteTask` do their serialization in their own `submit_finalize` function, rather than the manager doing it for them.

David-Simonetti-ND (approved):

David-Simonetti-ND:
The way RemoteTask did it originally was a bit of a workaround, so now that all tasks having one last chance to change something before being submitted it a great change! This gives us more flexibility with tasks, and I'm thinking of the possibility for the environment needed for a PythonTask/RemoteTask to be generated on the fly and added in submit_finalize.

PR #3101, opened 2023-02-08
Vine txn standard

btovar:


dthain:
It's ok to change the log format, but then we also need to change the consumers of that log in the same PR.
@BarrySlyDelgado @colinthomas-z80 does this affect your viz tools?

btovar:
Yes, I decided to change this before we build more tools on the log.
I'll submit a pr for Barry's tools.

dthain:
Something that Colin and I discussed (and I think this change reflects) is that a TRANSFER event should only be used for a substantial data movement between the manager and worker, and not just for metadata describing some other kind of transfer.

btovar:
Yes, I agree! Let me look for those.

colinthomas-z80:
Feel free to change. I need to rework my graphing tools anyway

BarrySlyDelgado:
Should the worker_id be used in WAITING_RETRIEVAL/DONE/RETRIEVED lines too? The transfer change helps as I originally had to loop through the tr log twice once tasks were already mapped to workers. The changes to the tool should be minor.

btovar:
Barry, yes, I have commits for that too. I saw that in your script, which made obvious that using worker-address was a mistake.

BarrySlyDelgado (approved):

colinthomas-z80 (approved):

PR #3110, opened 2023-02-10
bucketing mem to memory

btovar:


tphung3 (approved):
Yeah it should be memory

tphung3 (approved):

PR #3111, opened 2023-02-13
Add --tex to vine_plot_txn_log

btovar:
Useful to keep the text in the graphs in the same style as the papers.


BarrySlyDelgado (approved):
Looks Good.

PR #3114, opened 2023-02-13
Change catalog update protocol to tcp by default.

dthain:


dttools/src/catalog_query.c
```diff
@@ -290,11 +290,18 @@ char *catalog_query_compress_update(const char *text, unsigned long *data_length
...
+more likely to make it through various network
+translation devices.
+*/
+
 static int catalog_update_protocol()
 {
 	const char *protocol = getenv("CATALOG_UPDATE_PROTOCOL");
 	if(!protocol) {
-		return 1;
+		return 0;
```
btovar, line 304: Also change the default below. (line 310)

btovar (changes requested):
man/m4/catalog_update.m4  and   man/m4/catalog_server.m4   (with their respective .md) also need to be updated.

btovar (approved):

PR #3116, opened 2023-02-14
Vine: Separate File and Mount Objects

dthain:
- Split vine_file into vine_file and vine_mount.
- Modify vine_file to be reference-counted, so the same object can be used with multiple tasks.
- Change task dependencies to be a list of input mounts and output mounts. 
- Update examples to use shared temp task objects.

Note that this change touches a lot of objects, but in a fairly straightforward way.  It now makes it easy to express applications a little more naturally, will help content-addressible-storage (because we only declare files once) and puts us on the path towards declaring files through the manager, instead of standalone objects.


taskvine/src/manager/vine_manager_put.c
```diff
@@ -348,8 +349,8 @@ static vine_result_code_t vine_manager_put_input_file_if_not_cached(struct vine_
 	/* If so, check that it hasn't changed, and return success. */
 	if(remote_info) {
 		if(f->type==VINE_FILE && (info.st_size!=remote_info->size || info.st_mtime!=remote_info->mtime)) {
-			debug(D_NOTICE|D_VINE,"File %s has changed since it was first cached!",f->source);
-			debug(D_NOTICE|D_VINE,"You may be getting inconsistent results.");
+			//debug(D_NOTICE|D_VINE,"File %s has changed since it was first cached!",f->source);
```
btovar, line 352: Leftover comments?

taskvine/src/manager/vine_mount.c
```diff
@@ -0,0 +1,37 @@
+
```
btovar, line 1: Copyright

taskvine/src/manager/vine_mount.c
```diff
@@ -0,0 +1,37 @@
...
+	/* Add a reference each time a file is connected. */
+	m->file->refcount++;
+	
+	return m;
+}
+
+void vine_mount_delete( struct vine_mount *m )
+{
+	if(!m) return;
+	vine_file_delete(m->file);
```
btovar, line 28: Deleting a file when the count of mounts go away may lead to weird segfault bugs. E.g., users for whatever reason may keep a list with the references to `vine_file`.  Those references will be free() under them when the mounts go away.

taskvine/src/manager/vine_mount.c
```diff
@@ -0,0 +1,37 @@
...
+	if(m->remote_name) free(m->remote_name);
+	free(m);
+}
+
+struct vine_mount * vine_mount_clone( struct vine_mount *m )
+{
+	if(!m) return 0;
+	return vine_mount_create(vine_file_clone(m->file),m->remote_name,m->flags,vine_file_clone(m->substitute));
```
btovar, line 36: If deleting vine_file_delete above, then these vile_file_clone will produce a memory leak.

btovar (changes requested):
It is nice that flags become a parameter of the mount, and are not one for the file object.

taskvine/src/manager/vine_mount.c
dthain: I think the refcounting works within the universe of file/task create/delete/connect. If the user does this:

```
file = vine_file_local(...)
task = vine_task_create(...)
vine_task_add_input(task, file,...)
```

Then at this point the file will have **two** references: one from the original file create, and as second one from vine_task_add_input.  So, then they can delete the file and task in any order, and will only go away on the last reference:

```
vine_task_delete(task);
vine_file_delete(file);
```

Likewise, if you clone a task, then you get a third reference to the underlying files.

Now, you are correct that it will break if the user keeps around multiple pointers to the same file.  To fix that, we need to move the file management into the manager object itself, and refer to them with unique identifiers.  We aren't there yet, but that's where I want to go...

taskvine/src/manager/vine_mount.c
dthain: The intent here is that cloning a mount means making a copy of the mount, but adding a ref to the underlying file objects.  Did I miss something?

taskvine/src/manager/vine_manager_put.c
dthain: This warning message is broken, but that's a different problem for another day.

btovar (approved):

btovar (approved):

PR #3132, opened 2023-02-16
vine xrootd

btovar:
adds vine_file_xrootd, FileXrootD.

Example is functional, but is missing the poncho recipe (i.e., assumes uproot and xrootd on execution site). The example uses PythonTask.

### DIFF
--- a/taskvine/src/manager/vine_file.c
+++ b/taskvine/src/manager/vine_file.c
@@ -150,4 +150,52 @@ struct vine_file * vine_file_unstarch( struct vine_file *f )
 }
 
 
+static char * find_x509_proxy()
+{
+	const char *from_env = getenv("X509_USER_PROXY");
+
+	if(from_env) {
+		return xxstrdup(from_env);
+	} else {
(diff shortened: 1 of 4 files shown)


taskvine/src/manager/vine_file.c
```diff
@@ -215,4 +215,51 @@ struct vine_file * vine_file_unstarch( struct vine_file *f )
...
+{
+	if(!proxy) {
+		char *proxy_filename = find_x509_proxy();
+		if(proxy_filename) {
+			proxy = vine_file_local(proxy_filename);
+			free(proxy_filename);
+		}
+	}
+
+	char *command = string_format("%sxrdcp %s output.root", proxy ? "X509_USER_PROXY=proxy509.pem " : "", source);
```
dthain, line 250: Perhaps use `vine_add_env_var`?

btovar: Ah, yes, much better.

dthain (approved):

PR #3136, opened 2023-02-16
Vine: Change loop using vine_tune

BarrySlyDelgado:
This adds some options within the manager that enables us to change the control flow for the manager loop using vine tune.
These options are:
"receives" - This changes the max number of receives the manager does on each loop. The default is one. However, this either 1 receive_all_tasks_from_worker or 1 receive_one_task. The minimum is 1 as setting it to 0 would tasks never be received. 

"prefer-receives" - If at least one  task is received `continue;` and restart the loop.

"receive-all-from-worker" - this enables receive_all_tasks_from_worker if set to zero do receive_one_task.

The new default config is:
    q->receives = 1;
    q->prefer_receives = 0;
    q->receive_all_from_worker = 1;
Screenshot from 2023-02-16 18-14-57

To use the old config:
    q->receives = 1;
    q->prefer_receives = 1;
    q->receive_all_from_worker = 0;

Screenshot from 2023-02-16 18-01-57

this can be changes using `vine_tune()`

### DIFF
--- a/work_queue/src/work_queue_coprocess.c
+++ b/work_queue/src/work_queue_coprocess.c
@@ -256,12 +256,17 @@ struct work_queue_coprocess *work_queue_coprocess_find_state(struct work_queue_c
 	return NULL;
 }
 
-struct work_queue_coprocess *work_queue_coprocess_initalize_all_coprocesses(int coprocess_cores, int coprocess_memory, int coprocess_disk, int coprocess_gpus, struct work_queue_resources *total_resources, char *coprocess_command, int number_of_coprocess_instances) {
+struct work_queue_coprocess *work_queue_coprocess_initalize_all_coprocesses(int coprocess_cores, int coprocess_memory, int coprocess_disk, int coprocess_gpus, struct work_queue_resources *total_resources, struct work_queue_resources *coprocess_resources, char *coprocess_command, int number_of_coprocess_instances) {
 	if (number_of_coprocess_instances <= 0) return NULL;
 	int coprocess_cores_normalized  = ( (coprocess_cores > 0)  ? coprocess_cores  : total_resources->cores.total);
 	int coprocess_memory_normalized = ( (coprocess_memory > 0) ? coprocess_memory : total_resources->memory.total);
 	int coprocess_disk_normalized   = ( (coprocess_disk > 0)   ? coprocess_disk   : total_resources->disk.total);
 	int coprocess_gpus_normalized   = ( (coprocess_gpus > 0)   ? coprocess_gpus   : total_resources->gpus.total);
+
+	coprocess_resources->cores.total = coprocess_cores_normalized;
+	coprocess_resources->memory.total = coprocess_memory_normalized;
+	coprocess_resources->disk.total = coprocess_disk_normalized;
+	coprocess_resources->gpus.total = coprocess_gpus_normalized;
 
 	struct work_queue_coprocess * coprocess_info = malloc(sizeof(struct work_queue_coprocess) * number_of_coprocess_instances);
 	memset(coprocess_info, 0, sizeof(struct work_queue_coprocess) * number_of_coprocess_instances);
@@ -279,7 +284,7 @@ struct work_queue_coprocess *work_queue_coprocess_initalize_all_coprocesses(int 
 	return coprocess_info;
 }
 
-void work_queue_coprocess_shutdown_all_coprocesses(struct work_queue_coprocess *coprocess_info, int number_of_coprocess_instances) {
+void work_queue_coprocess_shutdown_all_coprocesses(struct work_queue_coprocess *coprocess_info, struct work_queue_resources *coprocess_resources, int number_of_coprocess_instances) {
 	if (number_of_coprocess_instances <= 0) return;
 	work_queue_coprocess_shutdown(coprocess_info, number_of_coprocess_instances);
 	for (int coprocess_num = 0; coprocess_num < number_of_coprocess_instances; coprocess_num++){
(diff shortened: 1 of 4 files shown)


dthain (changes requested):
Yes this is the right idea but `q->receive` could mean a lot of things.  Give the variables and tune string painfully verbose names like `mainloop_receive_task_limit` or something like that.  Also be sure to update the comment at the top of `vine_wait_internal` so it describes the code correctly.

BarrySlyDelgado:
An interesting note on why the build tests failed: Previously, since only one tasks out put would be retrieved per loop, and once received the loop would restart and mark that task done. Since the new default is to only restart on sending a single task, the task will not get marked done by the time the manager exits. 

An option here is to mark all the tasks you receive as done right after receiving them.

btovar:
Tasks are marked as done only after returned to the user with `m.wait()`.  Marking them as done would also lose them. The manager should not exit until all tasks are marked as done, regardless on when the loop restarts, so probably there is another bug.

taskvine/src/manager/vine_manager.c
```diff
@@ -4289,6 +4350,7 @@ static struct vine_task *vine_wait_internal(struct vine_manager *q, int timeout,
 		// return if manager is empty and something interesting already happened
 		// in this wait.
 		if(events > 0) {
+			if(task_state_any(q, VINE_TASK_RETRIEVED) && t == NULL) continue;
```
BarrySlyDelgado, line 4353: This fixed it. what was happening was that t gets set and marked done at the top and this branch breaks  if done `int done = !task_state_any(q, VINE_TASK_RUNNING) && !task_state_any(q, VINE_TASK_READY) && !task_state_any(q, VINE_TASK_WAITING_RETRIEVAL);` so t would be returned as `NULL`

doc/manuals/taskvine/index.md
```diff
@@ -1792,6 +1792,10 @@ change.
 | resource-submit-multiplier | Assume that workers have `resource x resources-submit-multiplier` available.<br> This overcommits resources at the worker, causing tasks to be sent to workers that cannot be immediately executed.<br>The extra tasks wait at the worker until resources become available. | 1 |
 | wait-for-workers        | Do not schedule any tasks until `wait-for-workers` are connected. | 0 |
 | wait-retrieve-many      | Rather than immediately returning when a task is done, `m.wait(timeout)` retrieves and dispatches as many tasks<br> as `timeout` allows. Warning: This may exceed the capacity of the manager to receive results. | 0 |
+| main-loop-max-receives | Sets the max number of receives for the manager to do in the main loop. This is receives of either `receive_one_task` or `receive_all_tasks_from_worker` | 1 |
+| main-loop-prefer-receives | If set to 1, main loop restarts after receiving at least one task | 0 |
```
btovar, line 1796:
```
prefer-receives | If set to 1, receive all completed tasks from all workers before dispatching new tasks to workers. If set to 0 then...
```
Let's avoid "main-loop" and "restarts", as these terms do not have any meaning unless you are familiar with the code.

doc/manuals/taskvine/index.md
```diff
@@ -1792,6 +1792,10 @@ change.
...
+| retrieve-all-waiting-tasks-from-worker | If set to 1, use `receive_all_tasks_from_worker`. If set to 0, use `receive_one_task` | 1 |
```
btovar, line 1797:
```
retrieve-completed-per-worker | If 0, then retrieve one completed result per worker before deciding to dispatch new tasks or connect new workers. If 1, retrieve all completed results from a worker instead. | 1 |
```

The attributes in the manager should have the same names, but with _ instead of -.

btovar (changes requested):

doc/manuals/taskvine/index.md
btovar: Also, does `wait-retrieve-many` has any relevance? Should it be deleted?

doc/manuals/taskvine/index.md
```diff
@@ -1792,9 +1792,9 @@ change.
...
 | wait-for-workers        | Do not schedule any tasks until `wait-for-workers` are connected. | 0 |
 | wait-retrieve-many      | Rather than immediately returning when a task is done, `m.wait(timeout)` retrieves and dispatches as many tasks<br> as `timeout` allows. Warning: This may exceed the capacity of the manager to receive results. | 0 |
-| main-loop-max-receives | Sets the max number of receives for the manager to do in the main loop. This is receives of either `receive_one_task` or `receive_all_tasks_from_worker` | 1 |
-| main-loop-prefer-receives | If set to 1, main loop restarts after receiving at least one task | 0 |
-| retrieve-all-waiting-tasks-from-worker | If set to 1, use `receive_all_tasks_from_worker`. If set to 0, use `receive_one_task` | 1 |
+| max-receives | Sets the max number of receives for the manager to do in the main loop. This is receives of either `receive_one_task` or `receive_all_tasks_from_worker` | 1 |
```
btovar, line 1795: Barry, please describe these options without referring to the main loop. E.g. `If set to 1, main loop restarts after receiving at least one task ` does not mean much to a user. We should describe here what prefer-receives, etc., in lay terms.

btovar (changes requested):

btovar (approved):

btovar:
@BarrySlyDelgado could you check that the removal of code repetition didn't break anything?

BarrySlyDelgado:
The changes are not entirely the same. I think i know whats causing it though
Screenshot from 2023-03-07 15-28-06

btovar:
sounds good!

PR #3143, opened 2023-02-17
Vine: Remove always-true condition causing misleading debug messages

tphung3:
In code, `info.st_mtime` refers to the last modified timestamp of the local file and `remote_info->mtime` refers to the last modified timestamp of the remote copy of the local file. The latter is always newer than the former as it is written to the local filesystem of the remote machine, thus the two debug messages are always printed. Removing the comparison condition removes the issue.


dthain:
We have an underlying problem that the mtime is never sent over to the remote machine.
Not sure if it is necessary at this point since we are using content hashes.

tphung3:
Oh so the remote machine checks the mtime of the local file? I thought it was the manager from the file name `vine_manager_put`

dthain (approved):

PR #3144, opened 2023-02-20
delete cache on startup/shutdown, uid+pid workspace

BarrySlyDelgado:


taskvine/src/worker/vine_worker.c
```diff
@@ -2187,7 +2187,14 @@ int main(int argc, char *argv[])
 	debug(D_VINE, "VINE_SANDBOX set to %s.\n", workspace);
 	setenv("VINE_SANDBOX", workspace, 0);
 
+	struct stat info;
```
dthain, line 2190: As a general rule, don't check something with `stat` and then act.
Because you will have to deal with any failures that result from acting anyway.
So just act! 
In this case, `trash_file` won't complain if the cache doesn't exist.

dthain (changes requested):

dthain (approved):

dthain:
It looks like `trash_setup` was never called before the first `trash_empty`

dthain (approved):

PR #3149, opened 2023-02-21
Vine: Simplified Output Buffer API

dthain:
Now works like this:
```
a = FileBuffer("input data")
b = FileBuffer()
...
t.add_input(a,"input.txt")
t.add_output(b,"output.txt")
...
print(b.contents())
```

### DIFF
--- a/taskvine/src/bindings/python3/taskvine.binding.py
+++ b/taskvine/src/bindings/python3/taskvine.binding.py
@@ -73,7 +73,7 @@ class File(object):
     #
     # @param self       A file object.
     
-    def length(self):
+    def __len__(self):
         return vine_file_length(self._file)
     
 ##
--- a/taskvine/src/bindings/python3/taskvine.binding.py
+++ b/taskvine/src/bindings/python3/taskvine.binding.py
@@ -138,13 +138,8 @@ class FileBuffer(File):
     #
     # @param self       The current file object.
     # @param buffer     The contents of the buffer, or None for an empty output buffer.
-
     def __init__(self, buffer=None):
-        if not buffer:
-            self._file = vine_file_buffer(None,0)
-        else:
-            buffer = bytes(buffer,"ascii")
-            self._file = vine_file_buffer(buffer, len(buffer))
+        self._file = vine_file_buffer(buffer)
 
 ##
 # \class FileMiniTask
--- a/taskvine/src/bindings/python3/taskvine.binding.py
+++ b/taskvine/src/bindings/python3/taskvine.binding.py
@@ -138,6 +138,13 @@ class FileBuffer(File):
     #
     # @param self       The current file object.
     # @param buffer     The contents of the buffer, or None for an empty output buffer.
+    #
+    # For example:
+    # @code
+    # # The following are equivalent
+    # >>> s = "hello pirate ♆"
+    # >>> FileBuffer(bytes(s, "utf-8"))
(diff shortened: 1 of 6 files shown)


taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -59,7 +59,23 @@ def __del__(self):
...
+    
+    def contents(self):
+        return vine_file_contents(self._file)
 
+    ##
+    # Return the length of a file object.
+    #
+    # @param self       A file object.
+    
+    def length(self):
```
btovar, line 76: Suggest the more pythonic `def __len__(self)`. Code then calls `len(File(...))`

btovar (changes requested):

taskvine/src/bindings/python3/taskvine.binding.py
dthain: That's a good idea.  I was tempted to put in `__str__` instead of `contents`, but wasn't ready to decide if the default string should be the file contents, the source, the remote name, or...

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -116,20 +132,19 @@ def __init__(self, url):
...
-    # @param buffer     The contents of the buffer.
-
-    def __init__(self, name, buffer):
-        name = str(name)
-        buffer = str(buffer)
-        self._file = vine_file_buffer(name, buffer, len(buffer))
+    # @param buffer     The contents of the buffer, or None for an empty output buffer.
 
+    def __init__(self, buffer=None):
+        if buffer is None:
```
btovar, line 143: `if not buffer:`

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -116,20 +132,19 @@ def __init__(self, url):
...
+            self._file = vine_file_buffer(None,0);
```
btovar, line 144: drop ;

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -116,20 +132,19 @@ def __init__(self, url):
...
+        else:
+            buffer = str(buffer)
```
btovar, line 146: I don't think you want str, as that calls a method on buffer which may not give you what you want. Suggest for now:
```
buffer = bytes(buffer, "ascii")
```

In the future we may want to use memory views:  https://stackoverflow.com/questions/16998686/expose-a-vector-as-a-memoryview-using-swig

btovar (changes requested):

btovar (approved):

btovar:
If you want, I can investigate how to pass that bytes object to swig.  For now, if you want to merge, I think enclosing it in a str(bytes(...)) is very ugly but should work.

dthain:
Yes, please take a look; I am quite fuzzy on how SWIG interacts with Python types.

btovar:
@dthain added some commits with the typemap.  The following works:

```python
>>> import taskvine as vine
>>> s = "hello pirate ♆"
>>> f = vine.FileBuffer(bytes(s, "utf-8"))
>>> print(f.contents())
hello pirate ♆
```

btovar:
ready to merge?

dthain:
go for it

PR #3150, opened 2023-02-21
Vine: Added logging and graphing capabilities for duties

David-Simonetti-ND:
Couple of changes in this pull request:
1)
Changed the API for sending duties to workers (vine_manager_send_duties_to_workers). Now accepts a timeout, and the maanger will send as many duties out to workers that it can until that timeout is reached. Sending a duty also creates an entry in the transaction log.
2)
Changed when duties are sent to workers. Instead of having to send the duty as soon as a worker connects, the manager will accept workers without immediately giving them a duty. The manager then waits until the end of the main loop before attempting to send duties until it reaches its timeout. This is beneficial in a few ways. First, the manager will not go overcapacity and send out more duties to workers than it can fulfill. Second, if there is a mixed workload of RemoteTasks and regular tasks, not all workers will need the duties. Third, it allows some tasks to start running on workers while the newly connected workers wait for the duties.
3)
Added a message sent from the worker to the manager that indicates when the duty has started. Since the manager logs the time the duty is sent out, we can reconstruct the timeline of duties from the txn log now.
4)
Updated vine_plot_txn_log to plot the minitasks that occur for duties, when they were not being shown before
5)
Updated vine_plot_txn_log to properly recognize failed tasks and to plot them with a red bar on the resulting graph.


taskvine/src/tools/vine_plot_txn_log
```diff
@@ -145,7 +145,7 @@ def parse_tasks(log):
                 manager_info["Duties Sent"].append(time - manager_info["Start"])
```
btovar, line 145: David, I suggest to change the transaction line for duties. "SEND" sounds like an event to me. E.g:

```
# subject target event arguments
DUTY dutyid event *args
```
It looks to me that you just need to swap target and event.

btovar (changes requested):

taskvine/src/manager/vine_manager.c
```diff
@@ -3890,14 +3891,30 @@ static int vine_manager_send_duty_to_worker(struct vine_manager *q, struct vine_
 	vine_manager_send(q,w, "duty %lld %lld\n",  (long long) strlen(name), (long long)t->task_id);
 	link_putlstring(w->link, name, strlen(name), time(0) + q->short_timeout);
 	vine_result_code_t result = start_one_task(q, w, t);
+
+	if(q->txn_logfile)
+	{
+		struct buffer B;
+		buffer_init(&B);
+		buffer_printf(&B, "DUTY SEND %d %s", t->task_id, w->workerid);
```
btovar, line 3899: buffer_printf(&B, "DUTY %d SEND %s", t->task_id, w->workerid);

btovar (changes requested):

taskvine/src/manager/vine_manager.c
```diff
@@ -3890,14 +3891,30 @@ static int vine_manager_send_duty_to_worker(struct vine_manager *q, struct vine_
 	vine_manager_send(q,w, "duty %lld %lld\n",  (long long) strlen(name), (long long)t->task_id);
 	link_putlstring(w->link, name, strlen(name), time(0) + q->short_timeout);
 	vine_result_code_t result = start_one_task(q, w, t);
+
+	if(q->txn_logfile)
```
dthain, line 3895: Factor this out as a new function in `vine_txn_log.c`.  (That's where all transactions get written)

taskvine/src/manager/vine_manager.c
```diff
@@ -3940,6 +3958,30 @@ void vine_manager_remove_duty( struct vine_manager *q, const char *name ) {
...
+	int duty_id = 0;
+	char status[VINE_LINE_MAX];
+
+	int n = sscanf(line, "duty-update %d %s", &duty_id, status);
+
+	if(n != 2) {
+		return VINE_MSG_FAILURE;
+	}
+
+	debug(D_VINE, "Duty %d started on worker %s\n", duty_id, w->workerid);
```
dthain, line 3971: Likewise

btovar (changes requested):

btovar:
Ready to be merged?

btovar (approved):

btovar:
David, added a vine_duty_state_t, thinking that in the future we will keep an explicit state for the duties. I changed the functions to use that state.

Also, I change the transactions to: `DUTY duty_id worker-id state`,  with the state at the end rather than the worker-id. This is because we may want to give states an argument, and it is more natural for this argument to immediately follow the state.

David-Simonetti-ND:
Looks good to me! Ready for merge.

PR #3151, opened 2023-02-22
Vine: File Replica Table

colinthomas-z80:
still debugging worker file transfers.


dthain:
You are going in the right direction here.  The result should be to pull some complexity out of `vine_manager` and into `vine_remote_file_table`.

colinthomas-z80:
Turns out we are sharing a vine_mount pointer in the case of mini tasks, which was the cause of the difficulties. Some thought may need to go in to the immutable nature of input files.

dthain:
As. iWork on some of the documentation, the terminology that seems to come more naturally are 'file' (the original objects) and `file_replicas` (multiple copies of the files.). Once you have this working, please rename the module and associated structures to `file_replica_table`, `file_replica` and so forth to correspond.

dthain:
Rerunning the failed test to see what we get...

dthain:
And in a similar vein, `vine_remote_file_info` should become `vine_file_replica`

taskvine/src/manager/vine_manager.c
```diff
@@ -2684,20 +2684,17 @@ static int vine_manager_transfer_capacity_available(struct vine_manager *q, stru
...
-						found_match = 1;
-						break;
-					} else {
-						free(peer_source);
-					}
-				}
-			}
+		if(m->file->type != VINE_MINI_TASK) 
+		{
+			vine_file_delete(m->substitute);
```
dthain, line 2689: This might go away with the separate bug fix, but it looks dangerous to delete without setting m->substitute to null.

dthain (changes requested):
looks good, one thing to fix

taskvine/src/manager/vine_manager.c
colinthomas-z80: Yes that was an artifact from the previous implementation. I don't think it had any functional effect. I will remove it

dthain (approved):

PR #3156, opened 2023-02-23
Catalog: Perform TCP Updates in Background Process

dthain:
Works, however it currently has the problem that update processes will accumulate as zombies.
Need to flesh out the `process` module so that we can easily reap only relevant processes of a given type.


dthain:
Simpler solution here: use double-fork to separate update process from parent.

Issue #3146, opened 2023-02-20
Vine: Attach Files to Manager


dthain:
Right now, file objects are created standalone and then attached to tasks, which makes it difficult to do things like de-duplication, automatic cleanup, and so forth.  Instead, files should be created through the manager object.  I think this can be done quite easily now:

1 - Create a new hash table `file_table` in the manager, mapping cache names to file objects.
2 - Create manager methods like `vine_declare_file(m, path)` `vine_declare_url(m, url)` and so forth.  These simply create the corresponding objects and enter them into the table.
3 - In `vine_delete`, iterate over the table and delete the objects.
4 - Update our examples to use the new interface and forget about deleting the objects.
5 - At some point soon, move all the `vine_file_X` methods back into `vine_file.h` and keep them private from the user.

This gets us a little value right away, in that file objects no longer need to be manually deleted.  And then it sets us up for some bigger wins once we make tasks persistent in the same way.

@btovar do you want to take a crack at this one?  I think the API was your idea to begin with.

PR #3157, opened 2023-02-23
Vine declare file

btovar:
For #3146

Files are now registered with the manager with `vine_declare_*(m, ...)`. E.g.: `vine_declare_temp(m)`
The manager checks for duplicates, and only keeps one copy.


dthain:
This is the right idea.  A few thoughts:

1 - We generally have `vine_manager_X` for internal operations on the `vine_manager` object, but `vine_X` for public API functions the the manager (`vine_submit`, `vine_hungry`, etc). While I see the rationale behind `vine_manager_file_X`, it's quite a mouthful for something we are going to say frequently.  I think a verb is needed.  Maybe `vine_declare_X` like you suggested before?

2 - No need to add an additional reference inside `vine_manager_declare_file`.  I think we only need to add refs when attaching a pointer to an object that we control.  (For example, we add a reference in `vine_task_add_{input/output}` because `vine_task_delete` -> `vine_mount_delete` -> `vine_file_delete`

3 - Once we have all the files entered into a file table, then the manager can delete the file table and all the files in it.  (So the user never has to delete files.)

dthain:
Almost there, I think we want all the file creations to be manager methods and not object constructors, like this:
```
m = vine.Manager(9123);
f = m.declareFile(path);
f = m.declareURL(url);
```

And then the public file constructors go away, except for a and object that wraps a generic `vine_file`, something like this:

```
def declareFile( self, path ):
    return vine.File( vine_declare_file( self._manager, path ) )
```

Would that work?

btovar:
Let me give it a go.

dthain (approved):

btovar:
Ok, here we go....

PR #3161, opened 2023-02-24
Workqueue: Fixes for BenC running serverless with Parsl

David-Simonetti-ND:
Couple of fixes here:
1)
poncho_package_serverize broke backwards compatibility with work_queue serverless in a recent commit, so I readded the network code that works for work_queue. This adds a now mandatory command line option to specify --version {work_queue, task_vine}
2)
Added functionality for the worker to check in its PATH for the coprocess file location 
3)
Backported Matt's change of making the coprocess enter the correct task sandbox when executing a RemoteTask. 
4)
Performance improvement for the coprocess when lots of input data is sent, as a debug print line was removed.


dthain:
@David-Simonetti-ND does poncho_package_serverize work the same way, regardless of whether the coprocess is working with Work Queue versus TaskVine?

David-Simonetti-ND:
Because the coprocess in WQ uses TCP but it uses pipes in TaskVine, poncho_package_serverize uses different network code depending if the output file is intended to be used with WQ or TV. The generated code is not compatible with both WQ and TV.

poncho/src/poncho_package_serverize
```diff
@@ -7,13 +7,14 @@ if __name__ == "__main__":
     parser.add_argument('--function', action='append', help='Name of the function in filename to be serverized')
     parser.add_argument('--pack', help='Optional argument. If given, will create an environment to run the network function at this path')
     parser.add_argument('--dest', help='Destination path for the output network function')
+    parser.add_argument('--version', choices=['work_queue', 'task_vine'], help='Version of serverless (work_queue or task_vine)', required=True)
```
dthain, line 10: A very little thing: we use `taskvine` consistently instead of `task_vine`

dthain (changes requested):

poncho/src/poncho_package_serverize
btovar: And related, use `vine` instead of `tv`.

poncho/src/poncho/package_serverize.py
```diff
@@ -112,10 +112,129 @@ def main():
...
+                        p.join()
+                        response = json.dumps(q.get()).encode("utf-8")
+                    elif exec_method == "direct":
+                        response = json.dumps(globals()[function_name](event)).encode("utf-8")
+                    else:
+                        p = os.fork()
+                        if p == 0:
+                            response =globals()[function_name](event)
+                            os.write(write, json.dumps(response).encode("utf-8"))
+                            os._exit(-1)
```
btovar, line 210: Is there a reason why to exit with an error code? Why not  exit(0)?

poncho/src/poncho/package_serverize.py
```diff
@@ -112,10 +112,129 @@ def main():
...
+'''
+import socket
+import json
+import os
+import sys
+import threading
+import queue
+def remote_execute(func):
+    def remote_wrapper(event, q=None):
+        if q:
```
btovar, line 130: Better to decide to load because of the nature of event, not if q was given, e.g.  `if isinstance(event, str):` and surround with try: except:.

btovar: Also, is remote_wrapper used anywhere?

David-Simonetti-ND: q is a thread safe data structure that is used to retrieve the result of the function in the case thread execution is used. Since the thread method is being removed with Matt's upcoming PR, this code will be removed there.

David-Simonetti-ND:
@btovar could you have another look at this PR? I need it to get merged to fix things for BenC.

btovar (approved):

PR #3164, opened 2023-02-27
Vine: Declare Task (WIP)

dthain:
@btovar what do you think about this approach?

```
f = vine_declare_url(m, url)
t = vine_declare_task(m,cmd)
vine_task_add_input( t, url, "data.txt" )
vine_submit(m,t)

/* optional if you need to reduce state during run */
vine_drop_file(m,f)
vine_drop_task(m,t)

/* cleans up all tasks and files regardless */
vine_delete(m)
```

### DIFF
--- a/dttools/src/histogram.c
+++ b/dttools/src/histogram.c
@@ -49,20 +49,13 @@ struct histogram *histogram_create(double bucket_size) {
 
 void histogram_clear(struct histogram *h) {
 
-	uint64_t key;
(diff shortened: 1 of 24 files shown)


btovar:
We want a Task object to subclass. Such thing really simplifies the code in coffea.

dthain:
Abandoning this in favor of Python-flavored task constructors, followed by renaming of `m.submit` to `m.declare_task`

PR #3166, opened 2023-02-27
Vine: Serverless sandbox fixed

mcarbona1:
This is the fix for the serverless sandbox issues that caused output file transfers to fail. By sending the sandbox to the coprocess, the duty can then change directory before doing anything else. This makes sure the work is done in the proper directory and that the worker knows where to look when it's time to transfer the files back.


btovar (commented):
The biggest issue is that threads do not have their own working directory. Thus, this modification won't work if more than one function runs concurrently in the coprocess. Thus, since only one function can be executed, this effectively makes the "thread" method the same as the "direct" method. Using os.chdir in a thread will make some debugging horror stories.

I suggest getting rid of the "thread" method. If more than one function has to run concurrently, either use the fork method, or have the worker spawn multiple coprocesses.

btovar:
Also, I'm not sure that the coprocess won't get confused when using "thread" and "direct" after the os.chdir().  It might be that the mounts given to the coprocess will also need to be mounted in the task sandboxes.  Along these lines, at the end of the function execution the sandbox is deleted, and the current working directory of the coprocess is gone.

dthain:
Agreed -- the sandbox directory is fundamental to the taskvine abstractions.
And that seems to be incompatible with the threads model.

David-Simonetti-ND (commented):
I agree, getting rid of the thread execution method makes sense. As Python only allows one thread to run at any given time, running multiple coprocesses on one worker would be the way to achieve concurrency.

mcarbona1:
Ok, I've removed the thread options from the package_serverize.py code and the taskvine bindings so it no longer will accept the thread option as well as edited some comments to remove the mention of threads. As far as I can tell there is no mention of serverless threading in the manual yet so I don't think anything has to change there.

poncho/src/poncho/package_serverize.py
```diff
@@ -78,18 +78,13 @@ def main():
...
-                    if exec_method == "thread":
-                        # create a forked process for function handler
-                        q = queue.Queue()
-                        p = threading.Thread(target=globals()[function_name], args=(event_str, q))
-                        p.start()
-                        p.join()
-                        response = json.dumps(q.get())
-                    elif exec_method == "direct":
+                    if exec_method == "direct":
+                        os.chdir(sandbox)
```
btovar, line 82: Here you want to re-establish the previous working directory after the function is execute, regardless of any error. Please look at  the  `try: except: finally:` statements.

For the fork one this is not needed, as the process that switch working directory exits just after execution.

btovar (changes requested):

mcarbona1:
Ok, I think this is pretty much what we discussed @btovar, let me know if there are any other changes you want me to make!

btovar:
Closing because of conflicts. Re-doing one commit.

PR #3167, opened 2023-02-27
Vine: Updated task declaration for Python

JosephDuggan7:


dthain:
Yes, this is on the right track!  Now, a few things to make it complete:
- As klaxon suggested, provide the option to specify via keyword arguments.
- Go through taskvine.binding.py and all all of the various task methods with keywords.
- A problem to think through: each `add_input` and `add_output` can have optional mount options: cache, watch, only-on-failure, and so forth.  Is there a way to add those options without making the common case unnecessarily complicated?
-

btovar:
I'd suggest making the first argument  the keyword "command=", so that one can pass a unix command as before, or use **dict to fill up other keywords.

For optional mount points and options, I'd suggest reversing the key-values of inputs and outputs so that the keys of the dictionaries are the vine file objects. If the value is a string, then it is taken as the remote name. If the value is dictionary-like object, then we set the options such as cache, watch, etc. as needed. If arguments are setup correctly, we may able also expand this dictionary with **dict.

JosephDuggan7:
Tasks now declared with keywords that are filled with a **dict.  More **dicts used for optional mounts.  I still need to add the other possible methods.

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -82,12 +82,29 @@ class Task(object):
...
         self._task = vine_task_create(command)
         if not self._task:
             raise Exception("Unable to create internal Task structure")
+        
+        if "inputs" in task_info.keys():
+            for key, value in task_info["inputs"].items():
+                if isinstance(value, str):
+                    self.add_input(key, value, cache=True)
+                elif isinstance(value, dict):
+                    self.add_input(key, **value)
```
btovar, line 98:
```python
else:
 # raise an error, otherwise we silently ignore unexpected types.
```

taskvine/src/examples/vine_example_blast.py
```diff
@@ -36,13 +36,18 @@
...
 
     for i in range(10):
-        t = vine.Task("blastdir/ncbi-blast-2.13.0+/bin/blastp -db landmark -query query.file")
-
-        t.add_input(query_buffer, "query.file", cache=True)
-        t.add_input(blast, "blastdir", cache=True )
-
-        t.add_input(landmark, "landmark", cache=True )
-        t.set_env_var("BLASTDB", value="landmark")
+        t = vine.Task(
```
btovar, line 39: neat!

btovar (changes requested):

taskvine/src/bindings/python3/taskvine.binding.py
JosephDuggan7: Fixed.

JosephDuggan7:
All methods now can be done through task declaration.

doc/manuals/taskvine/index.md
```diff
@@ -225,6 +225,13 @@ Here is how to describe the files needed by this task:
     # when the name at manager is the same as the exection site, we can write instead:
     t.add_input_file("my-file",     cache = False)
     t.add_output_file("my-file.gz", cache = False)
+
+    # files can also be described when the task is declared:
+     t = vine.Task(
+        command = "./gzip < my-file > my-file.gz",
+        input_files = {"/usr/bin/gzip" : {remote_name : "gzip, cache : True}, "my-file" : {remote_name : "my-file", cache : False}},
```
btovar, line 232: Given this is a doc example, I suggest having one key per line, as:
```python
input_files = {
    gzip_file: {
        "remote_name": "..."
   ...
}

File: doc/manuals/taskvine/index.md [hunk 4]
```diff
@@ -225,6 +225,13 @@ Here is how to describe the files needed by this task:
...
+        output_files = {"my-file.gz : {remote_name : "my-file.gz", cache = False}}
```
btovar, line 233: Same as above. Also, keys should be declared files.

taskvine/src/examples/vine_example_blast.py
```diff
@@ -36,13 +36,18 @@
...
+            command = "blastdir/ncbi-blast-2.13.0+/bin/blastp -db landmark -query query.file",
+            inputs = {query : "query.file", blast : "blastdir", landmark : "landmark"},
+            env = {"BLASTDB" : "landmark"}
+        )
+
+        # t = vine.Task("blastdir/ncbi-blast-2.13.0+/bin/blastp -db landmark -query query.file")
```
btovar, line 45: This comment should be removed.

btovar (changes requested):

doc/manuals/taskvine/index.md
JosephDuggan7: Fixed

doc/manuals/taskvine/index.md
```diff
@@ -1418,8 +1481,10 @@ either their `taskid` or `tag`. For example:
 === "Python"
     ```python
     # create task as usual and tag it with an arbitrary string.
-    t = vine.Task(...)
-    t.set_tag("my-tag")
```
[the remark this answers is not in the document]
JosephDuggan7: Fixed

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -82,12 +82,121 @@ class Task(object):
...
-    def __init__(self, command):
+    # @param task_info  Optional dictionary containing specified task parameters.
+    def __init__(self, command, **task_info):
         self._task = None
 
         self._task = vine_task_create(command)
         if not self._task:
             raise Exception("Unable to create internal Task structure")
+        
+        if "coprocess" in task_info.keys():
```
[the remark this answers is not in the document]
JosephDuggan7: Fixed

PR #3169, opened 2023-02-28
Adding Filetrace

jdolak:
This is my draft of filetrace so far. Currently there are two known problems:
1 . If a file changes from from write to read, filetrace will count all occurrences as the later (Im working on a fix for this today)
2. When trying to see how many bytes were written I am using the output of `read()`, however, the file path produced by this does not match the paths already present. This can be seen as the entries with '?' in the second output file


dthain:
Good, this is on the right track.

I especially like that you included some detailed examples , but it would be better to include the examples as plain-text in the document, instead of embedded images.

I'll let you develop it a bit further before accepting the PR.

jdolak:
@dthain Apologies, I had changes in a branch that I forgot to merge, should be fixed now.

dthain:
Hmm, I'm still not seeing the images as part of the PR...?
Take a look at the "files changed" tab.

dthain:
Looks good, thank you!

Issue #3159, opened 2023-02-24
Vine: Function, Library, Coprocess


dthain:
Per our discussion today, rename the following components throughout vine:

A **Library** is a special type of **Task** that can be installed on all workers.
A **FunctionCall** is a special type of **Task** that invokes a function in a library.
A **Coprocess** is a running instance of a Library at a Worker.

dthain:
This is done!

PR #3170, opened 2023-03-01
Vine: serverless rename and create_library_from_functions

David-Simonetti-ND:
In this PR
1)
Fixing issue #3159: General renaming of Duty to Library and RemoteTask to FunctionCall. Also decided to rename the network function to Library code file, but open to suggestions for a better name. Network function does not sound descriptive enough for what it is to me.
2)
Creation of a LibraryTask that inherits from Task. I needed a way to attach the name of the Library to the individual instance, so I added a name field. It also allows install_library to check that the task being installed is a proper Library Task.
3)
Changed the way LibraryTasks are created. There are two methods now:
a) Command line method. 
The user calls poncho_package_serverize in order to generate Library code, creates an environment for that code either manually or with package_serverize again. They then call create_library_from_files with the name of the Library, the path to the generated output of package_serverize, and the path to their environment file. An example is below

```
t = m.create_library_from_files("library_name", "path_to_library_code.py", "path_to_env.tar.gz")
m.install_library(t)
```

The function then takes care of wrapping all the extras around a task and creates a Library Task.
b) In manager file method. 
The user defines Python functions in the manager file, such as 
```
def my_sum(a, b):
        return a + b
def my_mult(a, b):
        return a * b
def my_div(a, b):
        return a / b
```
They can then call create_library_from_functions to install a library like so, directly referencing the functions in the same file.

```
t = m.create_library_from_functions("library_name", my_sum, my_mult, my_div)
m.install_library(t)
```

The manager then creates a hash of the function names and source code, and checks in a cache directory to see if a library has already been created for the given functions. If not, it creates the appropriate library code, an environment to run the functions, and places them in the cache. On future runs, the library code and environment can be reused. In this way, it avoids the user having to interact with the command line to get started working with Libraries and FunctionCalls, and should streamline the learning experience for users.


btovar:
pretty neat!

dthain:
This is very well done!

```
t = m.create_library_from_functions("library_name", my_sum, my_mult, my_div)
m.install_library(t)
```

dthain (commented):
I very much like the general approach here.

My main high level comment would be to keep in mind that a Library could be any kind of program that attaches to the worker as a coprocess and obeys the invocation protocol.

So, from that perspective:
- Make sure that you can add additional input file dependencies to a library.  Perhaps it needs some C libraries or a large input dataset.
- Make sure that the poncho package (and poncho_package_run) is optional, since the Library might not even be written in Python.

Q: What is the namespace of t->library_id?  Is it in the same space as taskid, or is it a different thing?

David-Simonetti-ND:
Couple changes
As Ben suggested last week, I moved the network code in package_serverize out of a string and into a function so that its easier to look at.
Regarding adding extra input files, Library Tasks are a superclass of Task, so extra inputs can be added via the normal task API. I feel like an interface specifically for Library Tasks is not needed.

I renamed `create_library_from_files` to `create_library_from_serverized_files` to indicate that this a python specific function. There is now a generic `create_library_from_command` that accepts a name and an executable path, and creates a Library Task from that information. This should allow for non Python Libraries.

To answer your question, the namespace of t->library_id is the same as tasks, so in the set of all Library Tasks and regular Tasks in a single manager everything gets a unique ID. (ex. tasks might get ids 1-1000, and libraries get ids 1001-1200).

The PR is not quite done yet, as I want to add the gradient descent example program using the new create_library_from_functions. After that, it should be good to go.

poncho/src/poncho_package_serverize
```diff
@@ -2,8 +2,8 @@
 from poncho import package_serverize as serverize
 import argparse
 if __name__ == "__main__":
-    parser = argparse.ArgumentParser(description="Creates a network function based on an input file and function")
-    parser.add_argument('--src', help='Source file that contains the function to be serverized')
+    parser = argparse.ArgumentParser(description="Creates Library code based on an input file and function. Allows the running of FunctionCalls")
```
btovar, line 5: FunctionCall instances?  I'm thinking that someone may look for the class FunctionCalls.

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1602,25 +1608,98 @@ def submit(self, task):
...
     # @param self   Reference to the current manager object.
-    # @param task   A task description created from @ref taskvine::Task.
-    # @param name   Name of the duty to be installed.
-    def install_duty(self, task, name):
-        self._duty_table[name] = task
-        vine_manager_install_duty(self._taskvine, task._task, "duty_coprocess:" + name)
+    # @param task   A Library Task description created from create_library_from_functions or create_library_from_files
+    def install_library(self, task):
+        if not isinstance(task, LibraryTask):
+            raise("Please provide a LibraryTask as the task argument")
```
btovar, line 1618: raise TypeError(...

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1602,25 +1608,98 @@ def submit(self, task):
...
+    # Turn a list of python functions into a Library
+    #
+    #
+    # @param self            Reference to the current manager object.
+    # @param name            Name of the Library to be created
+    # @param function_list   List of all functions to be included in the library
+    def create_library_from_functions(self, name, *function_list):
+        # ensure poncho python library is available
+        if poncho_available == False:
+            raise("The poncho module is not available. Cannot create Library.")
```
btovar, line 1641: raise ModuleNotFoundError(...

btovar (approved):

David-Simonetti-ND:
Made a couple of changes
1) Added the gradient descent example in the examples folder
2) As per our discussion at the meeting on Wednesday, I had another go at resource allocations. Instead of having the resource allocations on the Library Task be used to determine how many FunctionCalls to schedule to a specific worker, the manager now counts the resource allocations on the Library Task as consuming the resources of the worker, and considers the total worker resources when scheduling new FunctionCalls.

This presents a problem, however, as currently each Library Task can only run 1 FunctionCall task at any given time. So now the manager thinks it can send for example 4 tasks to a worker where the Library coprocess can only run 1 at a time, with 3 tasks sitting there not running.

I'm struggling to find a solution to this. The hackiest idea is to just have the worker spawn one Library coprocess per core that it has, but this doesn't account for the fact that Libraries can be an arbitrary executable and it might not make sense if the Library is capable of harnessing multiple cores. 
The more robust idea I had was to send a message to the coprocess upon startup specifying how many cores that the worker had access to, and the coprocess would respond with a maximum number of concurrent tasks it could run (in this case for Python serverless, it would always be one to one). The Library coprocess would be able to run multiple FunctionCalls at once using a sort of "pool" of forked worker processes with a size equal to its max concurrency, but I was running into race conditions trying to coordinate between the Library and the worker.

dthain:
This sounds like a longer conversation best had in person.   Please look at your schedule and pick a time that you, me, and @btovar can talk in detail.

David-Simonetti-ND:
@btovar @dthain I looked at everyone's schedule and it seems like there a slot from 12:00-1:00pm tomorrow. We can also discuss after the chalk talk is over.
I will be in San Fransisco all of next week for Spring break, so I would have to meet virtually if we decide next week.

poncho/src/poncho/package_serverize.py
```diff
@@ -14,226 +14,221 @@
 import ast
 import tarfile
 import hashlib
+import inspect
 
 shebang = "#! /usr/bin/env python3\n\n"
```
btovar, line 19: used anywhere?

poncho/src/poncho/package_serverize.py
```diff
@@ -14,226 +14,221 @@
...
+                response = {
+                    "Result": func(*args, **kwargs),
+                    "StatusCode": 200
+                }
+            except Exception as e:
+                response = { 
+                    "Result": str(e),
+                    "StatusCode": 500 
+                }
+            if not q:
```
btovar, line 44: More natural:
```python
if q:
   q.put(response)

return response
```
In that way the function always returns the same and a potential caller does not have to think about that.

poncho/src/poncho/package_serverize.py
```diff
@@ -14,226 +14,221 @@
...
-                        elif p < 0:
-                            print('Network function: unable to fork', file=sys.stderr)
-                            response = { 
-                                "Result": "unable to fork",
-                                "StatusCode": 500 
-                            }
+            while True:
+                # wait for message from worker about what function to execute
+                line = input()
+                print(f"Network function received task", file=sys.stderr, flush=True)
```
btovar, line 66: Is this a debug message?  If it is it needs to be removed.

poncho/src/poncho/package_serverize.py
```diff
@@ -14,226 +14,221 @@
...
-                            os._exit(0)
-                        elif p < 0:
-                            print('Network function: unable to fork', file=sys.stderr)
-                            response = { 
-                                "Result": "unable to fork",
-                                "StatusCode": 500 
-                            }
+            while True:
+                # wait for message from worker about what function to execute
+                line = input()
```
btovar, line 65: try ... except EOFError  wrap around this line. On EOF, probably break out of the loop?

poncho/src/poncho/package_serverize.py
```diff
@@ -14,226 +14,221 @@
...
+                print(f"Network function received task", file=sys.stderr, flush=True)
+                if len(line) >= 0:
```
btovar, line 67: Isn't this always True if `input` didn't fail?

poncho/src/poncho/package_serverize.py
```diff
@@ -14,226 +14,221 @@
...
-                            chunk = os.read(read, 65536).decode("utf-8")
-                            all_chunks = [chunk]
-                            while (len(chunk) >= 65536):
+                            p = os.fork()
+                            if p == 0:
+                                response =globals()[function_name](event)
+                                os.write(write, json.dumps(response).encode("utf-8"))
+                                os._exit(0)
+                            elif p < 0:
+                                print('Network function: unable to fork', file=sys.stderr)
```
btovar, line 96: print(f'unable to fork to execute {line}')  or something like that.

poncho/src/poncho/package_serverize.py
```diff
@@ -14,226 +14,221 @@
...
+                                response = { 
+                                    "Result": "unable to fork",
+                                    "StatusCode": 500 
+                                }
+                            else:
                                 chunk = os.read(read, 65536).decode("utf-8")
-                                all_chunks.append(chunk)
-                            response = "".join(all_chunks)
-                            os.waitid(os.P_PID, p, os.WEXITED)
+                                all_chunks = [chunk]
```
btovar, line 103: max_read = 65536
And use in the lines below.

poncho/src/poncho/package_serverize.py
```diff
@@ -14,226 +14,221 @@
...
+                response = {
+                    "Result": func(*args, **kwargs),
+                    "StatusCode": 200
+                }
+            except Exception as e:
+                response = { 
+                    "Result": str(e),
+                    "StatusCode": 500 
+                }
+            if not q:
```
btovar, line 140: as above.

poncho/src/poncho/package_serverize.py
```diff
@@ -14,226 +14,221 @@
...
-            "name": name(),
-            "port": s.getsockname()[1],
-            }
-    send_configuration(config)
-    while True:
-        s.listen()
-        conn, addr = s.accept()
-        print('Network function: connection from {}'.format(addr), file=sys.stderr)
+            s.close()
+            print(e)
```
btovar, line 158: Since it is an error, do you want it in stderr?

poncho/src/poncho/package_serverize.py
```diff
@@ -307,10 +305,10 @@ def sort_spec(spec):
     return json.dumps(sorted(conda_deps) + nested_deps, sort_keys=True).encode("utf-8")
 
 def search_env_for_spec(envpath):
+    env_spec = None
     if os.path.exists(envpath) and envpath.endswith(".tar.gz"):
-        print("Cached environment found, checking if it is compatiable with new network function")
+        print("Cached environment found, checking if it is compatiable with new library code")
```
btovar, line 310: You may want to standardize all messages to use "library code" rather than "network function".

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+    # Turn a list of python functions into a Library
+    #
+    #
+    # @param self            Reference to the current manager object.
+    # @param name            Name of the Library to be created
+    # @param function_list   List of all functions to be included in the library
+    def create_library_from_functions(self, name, *function_list):
+        # ensure poncho python library is available
+        if poncho_available == False:
+            raise TypeError("The poncho module is not available. Cannot create Library.")
```
btovar, line 1613: ModuleNotFoundError?

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+        # positional arguments are the list of functions to include in the library
+        # create a unique hash of a combination of function names and bodies
+        functions_hash = package_serverize.generate_functions_hash(function_list)
+
+        # create path for caching library code and environment based on function hash
+        library_cache_path = f"./vine-library-cache/{functions_hash}"
```
btovar, line 1619:
```python
library_cache_path = f"{self.cache_directory}/vine-library-cache/{functions_hash}"
```

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+        library_code_path = f"{library_cache_path}/library_code.py"
+        library_env_path = f"{library_cache_path}/library_env.tar.gz"
+
+        # library cache folder doesn't exist, create it
+        if not os.path.exists("vine-library-cache"):
```
btovar, line 1624:
```python
from pathlib import Path   # (somewhere above)

Path( library_cache_path ).mkdir(mode=0o755, parents=True, exist_ok=True)
```

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+            os.mkdir("vine-library-cache")
+        # if the library code and environment exist, move on to creating the Library Task
+        if os.path.isfile(library_code_path) and os.path.isfile(library_env_path):
+            pass
+        else:
+            # check if the cache entry exits, and create it if it doesn't exist 
+            if not os.path.exists(library_cache_path):
```
btovar, line 1631: not needed if using the above mkdir method.

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+                os.mkdir(library_cache_path)
+            print("No cached Library code and environment found, regenerating...")
+            # create library code and environment
+            package_serverize.serverize_library_from_code(library_cache_path, function_list, name)
+            # enable correct permissions for library code
+            os.chmod(library_code_path, 0o775)
```
btovar, line 1637: Not needed if using above mkdir.

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+        
+        # create Task to execute the Library
+        t = LibraryTask("./poncho_package_run -e package python ./library_code.py", name)
```
btovar, line 1640:
```python
t = LibraryTask("python ./library_code.py", name)
f = self.declare_poncho(self.declare_file(library_env_path))
t.add_environment(f)    # adds the poncho_package_run
```

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+        # declare the environment
+        f = self.declare_poncho(self.declare_file(library_env_path))
+        t.add_input(f, "package")
+        # declare the library code as an input
+        f = self.declare_file(library_code_path)
+        t.add_input(f, "library_code.py")
+        # add poncho_package_run to be able to execute the Library
+        f = self.declare_file(shutil.which('poncho_package_run'))
```
btovar, line 1648: poncho_package_run does not need to be added anymore (as of yesterday), as poncho tarballs now have it included.

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+    # Turn Library code created with poncho_package_serverize into a Library Task
+    #
+    #
+    # @param self            Reference to the current manager object.
+    # @param library_path    Path of the file which contains the library code - the output of poncho_package_serverize
+    # @param env_path        Path of the environment file which contains the environment capable of running the Library
+    def create_library_from_serverized_files(self, name, library_path, env_path):
+        if poncho_available == False:
+            raise ModuleNotFoundError("The poncho module is not available. Cannot create library.")
+        t = LibraryTask(f"./poncho_package_run -e package python ./library_code.py", name)
```
btovar, line 1661: As above.. remove poncho_package_run, add t.add_environment

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+        f = self.declare_poncho(self.declare_file(env_path))
+        t.add_input(f, "package")
+        f = self.declare_file(library_path)
+        t.add_input(f, "library_code.py")
+        f = self.declare_file(shutil.which("poncho_package_run"))
```
btovar, line 1666: Not needed...

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1580,104 @@ def submit(self, task):
...
+        t.add_input(f, "poncho_package_run")
+        return t
+
+    # Create a Library task from arbitrary inputs
+    #
+    #
+    # @param self            Reference to the current manager object.
+    # @param executable_path Path of the file which contains the library executable
+    # @param name            Name of the library to be created
+    def create_library_from_command(self, executable_path, name):
```
btovar, line 1676: Add an argument for an optional environment.  If not None, add the environment to the the task as above.

taskvine/src/examples/vine_example_gradient_descent.py
```diff
@@ -0,0 +1,125 @@
...
+import random
+
+# this is the function we will pack into a Library Task. It accepts a list of training data, a list of testing data, and returns the optimized set of weights.
+# the model used is a simple linear regression with polynomial basis functions.
+def batch_gradient_descent(train_data, test_data):
+    import random
+    import time
+    import numpy as np
+    # the number of parameters/basis functions in the model (M).
+    NUMBER_OF_PARAMS = 100
```
btovar, line 27: lowercase so it looks more pythonic?

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -2209,13 +2293,13 @@ def set_fn_args(self, args=[], kwargs={}):
...
-    # "thread". Fork creates a child process to execute the function, direct
-    # has the worker directly call the function, and thread spawns a thread to
-    # execute the function
+    # @param remote_task_exec_method  Can be either of "fork" or "direct".
+    # Fork creates a child process to execute the function and direct
+    # has the worker directly call the function.
     def set_exec_method(self, remote_task_exec_method):
-        if remote_task_exec_method not in ["fork", "direct", "thread"]:
-            print("Error, vine_exec_method must be one of fork, direct, or thread")
+        if remote_task_exec_method not in ["fork", "direct"]:
```
btovar, line 2300: Code still seems to support the thread method (e.g. q.put).  Do you want to remove it?

btovar (changes requested):

poncho/src/poncho/package_serverize.py
David-Simonetti-ND: Inspect is used to get the source code from the library_source_code function and place it into the output Library code file.

poncho/src/poncho/package_serverize.py
David-Simonetti-ND: The q is only used in threading execution modes, and Matt's upcoming PR is getting rid of the thread execution method (including this q variable). So his PR will be fixing this.

poncho/src/poncho/package_serverize.py
David-Simonetti-ND: I will change all taskvine messages to use Library code, but leave all the work_queue messages as network function to keep consistency.

taskvine/src/bindings/python3/taskvine.binding.py
David-Simonetti-ND: Matt's PR will be removing it, so I'll wait for his PR to officially remove all the code related to it. This is just a temporary fix that ensures it can't be used.

poncho/src/poncho/package_serverize.py
David-Simonetti-ND: I made the process exit on EOF. The only case where EOF should happen is if the worker closes the write end of the pipe, and in that case the coprocess can't do much of use so it should just terminate itself.

David-Simonetti-ND:
@btovar went ahead and made the changes you requested!

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1581,99 @@ def submit(self, task):
...
+    #
+    #
+    # @param self            Reference to the current manager object.
+    # @param library_path    Path of the file which contains the library code - the output of poncho_package_serverize
+    # @param env_path        Path of the environment file which contains the environment capable of running the Library
+    def create_library_from_serverized_files(self, name, library_path, env_path):
+        if poncho_available == False:
+            raise ModuleNotFoundError("The poncho module is not available. Cannot create library.")
+        t = LibraryTask("python ./library_code.py", name)
+        f = self.declare_poncho(self.declare_file(env_path))
```
btovar, line 1656: With `declare_poncho` in python you can simply pass the tarball. Also, I'd allow the user to pass any environment that they want, like:
```python
...
# param env   Either a vine file or a poncho tarball filename that expands to an environment to run the library.

def create_library_from_serverized_files(self, name, library_path, env):
...
if env:
  if isinstance(env, str):
    env = self.declare_poncho(env, cache=True)
  t.add_environment(env)
```
Note that in the python bindings passing a str to declare_poncho does the right thing (i.e. declares a local file if needed.)

taskvine/src/bindings/python3/taskvine.binding.py
```diff
@@ -1574,25 +1581,99 @@ def submit(self, task):
...
+        t.add_input(f, "library_code.py")
+        return t
+
+    # Create a Library task from arbitrary inputs
+    #
+    #
+    # @param self            Reference to the current manager object.
+    # @param executable_path Path of the file which contains the library executable
+    # @param name            Name of the library to be created
+    # @param env_path        Optional argument to include a path to an environment to run the Library in
```
btovar, line 1668: Same as above. Change to env, where env can be a str or a vine file.

btovar (changes requested):

David-Simonetti-ND:
@btovar ready for re-review!

btovar (approved):

PR #3171, opened 2023-03-01
adds cache directory at vine_runtime_info_path/

btovar:
Adds a cache directory to the runtime dirs.

The motivation is to move directories such as `./vine-library-cache/` (#3170) to   m.cache_directory + `library-cache`


dthain (approved):
Yep, we are going to need a place to drop all sorts of things, like URLs downloaded at the manager side.

Is there a difference between `cache` and `staging` ?

btovar:
Staging is for temp files that have lifetime of task/workflow. (E.g. resource summary files, or serialized python function definitions.)

btovar:
Also, staging dirs are removed at the end of the manager (or atexit). It should be considered as the "/tmp" of a manager.

PR #3180, opened 2023-03-07
work_queue: Adding coprocesses-total default

David-Simonetti-ND:
More fixes for BenC.
Adds a default behavior for work_queue coprocess. Now, if coprocesses-total is not specified, the entire resources of the worker is split evenly among all coprocesses (the number of which is the number of cores on the worker).
coprocess_resources now contains the aggregated sum of resources commited to coprocesses for the worker.


David-Simonetti-ND:
@btovar This is ready for review!

David-Simonetti-ND:
@btovar Ready to merge!

PR #3181, opened 2023-03-07
Vine: file_flags and mount_flags

colinthomas-z80:
Left cache flags in the mount_flags for now


btovar (approved):

btovar:
Ready to merge?

colinthomas-z80:
It is a starting point but I think some further changes are in order. Standby for discussion today.

taskvine/src/examples/vine_example_peer_transfer.c
```diff
@@ -24,58 +12,51 @@ with all the same tasks on the worker.
...
 
-const char *query_string = ">P01013 GENE X PROTEIN (OVALBUMIN-RELATED)\n\
-		QIKDLLVSSSTDLDTTLVLVNAIYFKGMWKTAFNAEDTREMPFHVTKQESKPVQMMCMNNSFNVATLPAE\n\
-		KMKILELPFASGDLSMLVLLPDEVSDLERIEKTINFEKLTEWTNPNTMEKRRVKVYLPQMKIEEKYNLTS\n\
-		VLMALGMTDLFIPSANLTGISSAESLKISQAVHGAFMELSEDGIEMAGSTGVIEDIKHSPESEQFRADHP\n\
-		FLFLIKHNPTNTIVYFGRYWSP\n\
-";
-
-#define BLAST_URL "https://ftp.ncbi.nlm.nih.gov/blast/executables/blast+/LATEST/ncbi-blast-2.13.0+-x64-linux.tar.gz"
+#define BLAST_URL "https://mirrors.slackware.com/slackware/slackware-iso/slackware-12.0-iso/slackware-12.0-install-d1.iso"
```
btovar, line 15: I suggest creating a local large file with `dd` and make that the input to the tasks.  The first workers will get the file from the manager, but others would get it from the peers.  (We do not want to mess up with the download stats of slackware, or cause unnecessary traffic of large files.)

taskvine/src/examples/vine_example_peer_transfer.c
```diff
@@ -24,58 +12,51 @@ with all the same tasks on the worker.
...
 
 #define LANDMARK_URL "https://ftp.ncbi.nlm.nih.gov/blast/db/landmark.tar.gz"
```
btovar, line 17: Maybe remove this one? We can make the example targeted to the peers sharing one big file.

taskvine/src/manager/vine_task.c
```diff
@@ -435,23 +435,23 @@ void vine_task_add_output( struct vine_task *t, struct vine_file *f, const char
 	list_push_tail(t->output_mounts, m);
 }
 
-void vine_task_add_input_file(struct vine_task *t, const char *local_name, const char *remote_name, vine_file_flags_t flags)
+void vine_task_add_input_file(struct vine_task *t, const char *local_name, const char *remote_name, vine_mount_flags_t flags)
 {
-	struct vine_file *f = vine_file_local(local_name);
+	struct vine_file *f = vine_file_local(local_name, (vine_file_flags_t)flags);
```
dthain, line 440: Careful!  A cast is usually a sign of a problem.

`vine_file_flags_t` and `vine_mount_flags_t` are not the same thing.

Here, we are creating a `vine_file_local` which takes file flags, and attaching it to a task with `vine_task_add_input` which takes mount flags.

The `vine_file_local` must be created with **file** flags, which haven't been given here.

Since this is a shortcut method that doesn't allow one to specify the same file used multiple times, I think the right thing to do here is pass in file flags indicating no-share and no-cache.

(Similar comments apply to the other `add_input_X` methods.)

dthain (changes requested):

taskvine/src/manager/vine_task.c
colinthomas-z80: We intend to remove these "add_input_X" methods, so I cast the flags to get rid of the warnings for the time being, since the VINE_CACHE symbols are still in both enumerations it still works.

dthain: That's a grenade waiting to go off.  :). Please put in some explicit default values there.

colinthomas-z80:
To summarize the current state:

-  caching flags are still in vine_mount_flags_t and caching decisions are still being made using that value
-  peer transfer decisions are being made using the vine_file_flags_t value
-  users define file flags on file declaration and they are passed down to the vine_file object
-  declarations that indirectly create a file through a minitask set the specified flags to the output of the minitask.
-  I've left vine_file_chirp alone because I'm not sure what it would mean to cache/share this. It gets default flags.
-  What default flags should be is up for debate, but it either must be NOSHARE or minitasks must be NOSHARE by default.

What remains:
- cache flags must be moved to the vine_file_flags_t and caching decisions need to be made using this value
- use of deprecated vine_task_add_input_X will set default flags for the files they create internally

dthain (approved):

btovar (approved):

BarrySlyDelgado (approved):

colinthomas-z80:
Yes go ahead

PR #3191, opened 2023-03-09
Vine: Add complex Blast example in C

tphung3:


dthain (changes requested):
This looks nice: I just merged the python equivalent in #3177.

Please pull and rebase on that, then make sure that the C and Python versions have equivalent comments and printfs, so that it is very easy for the reader to see their equivalence.

Then let's merge it.

tphung3:
The PR should be ok to merge now.

dthain (approved):

PR #3192, opened 2023-03-10
Cleanup add files

btovar:
Moved move vine_task_add_{input,output}_.* to vine_task.h, as these functions never declare a file. They are useful at the worker where a file struct needs to be created without declaring to a manager.


btovar:
The CI should fail because of a memory leak that will be fixed in another pr.

taskvine/src/manager/taskvine.h
```diff
@@ -49,8 +49,8 @@ typedef enum {
 } vine_mount_flags_t;
 
 typedef enum {
-	VINE_PEER_SHARE = 1, /**< Schedule this file to be shared between peers where available. (default) **/
-	VINE_PEER_NOSHARE = 2  /**< Do not schedule this file to be shared between peers. **/
+	VINE_PEER_SHARE = 0, /**< Schedule this file to be shared between peers where available. (default) */
```
colinthomas-z80, line 52: This will not work with comparing flags with "|="

colinthomas-z80 (changes requested):
Issues with sharing mini tasks and bit comparison of file flags

taskvine/src/manager/taskvine.h
btovar: My motivation was that for the examples 0 was used, but it is not part of the enumeration. We want 0 to mean something, I think, and it is natural to be the default. If we use 1,2,4, etc. we can do something similar to |=. Would something like the flags in dttools/src/debug.h work for what you are thinking?

E.g., we can change:

`if((m->file->flags |= VINE_PEER_SHARE))`  to   `if(!(m->file->flags  & VINE_PEER_NOSHARE))`. That & checking set bits is also more in line with the rest of the code.

btovar: Or we can make VINE_PEER_NOSHARE=0 the default.  What do you think?

colinthomas-z80: Since we are combining cache flags and sharing flags there will be occasions where that would attempt something like `10000000 & 10100000'

taskvine/src/examples/vine_example_blast.c
```diff
@@ -88,19 +88,19 @@ int main(int argc, char *argv[])
...
 
 	printf("Declaring files...");
-	struct vine_file *blast_url = vine_declare_url(m, BLAST_URL, 0);
-	struct vine_file *landm_url = vine_declare_url(m, LANDMARK_URL, 0);
+	struct vine_file *blast_url = vine_declare_url(m, BLAST_URL, VINE_PEER_SHARE);
+	struct vine_file *landm_url = vine_declare_url(m, LANDMARK_URL, VINE_PEER_SHARE);
 
-	struct vine_file *software = vine_declare_untar(m, blast_url, 0);
-	struct vine_file *database = vine_declare_untar(m, landm_url, 0);
+	struct vine_file *software = vine_declare_untar(m, blast_url, VINE_PEER_SHARE);
```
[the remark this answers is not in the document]
btovar: Would https://github.com/cooperative-computing-lab/cctools/blob/ae1249ff26fd2fba73b4aa2ee10e689bea9c9865/taskvine/src/manager/vine_file.c#L149 address this?

PR #3194, opened 2023-03-13
Edit main loop using vine_tune

BarrySlyDelgado:
Accidentally deleted the original pull request.  However, some of the changes had large spaces between task retrievals after the bulk of tasks were complete. Also, all tasks weren't marked done until the end .
Screenshot from 2023-03-07 15-28-06

There was a break in the beginning of the main loop that returned a task as soon as it completed. We also check if there are any ready tasks. If not, We'll retrieve the max number of tasks.
Screenshot from 2023-03-13 09-50-24


dthain (approved):

dthain:
@btovar what do you think?

taskvine/src/manager/vine_manager.c
```diff
@@ -4673,7 +4674,7 @@ int vine_tune(struct vine_manager *q, const char *name, double value)
 		q->wait_for_workers = MAX(0, (int)value);
 
 	} else if(!strcmp(name, "max-retrievals")) {
-		q->max_retrievals = MIN(-1, (int)value);
+		q->max_retrievals = MAX(1, (int)value);
```
btovar, line 4544: -1 means infinite retrievals, so this one should be `MAX(-1,...`

btovar (changes requested):

btovar (approved):

PR #3195, opened 2023-03-13
Adds caching flags for vine files

BarrySlyDelgado:


dthain (changes requested):
This is the right idea, but you have a nice opportunity to factor things to simplify code.
For example, `vine_random_name` calls `string_cookie` six or seven times.  Just call it once at the top of the function.   Likewise, `vine_cached_name` needs to generate a random name in several places -- just have it call `vine_random_name` as needed.

btovar (approved):

PR #3197, opened 2023-03-14
Rename file flags

btovar:
`vine_mount_flags_t` and `vine_file_flags_t` had repeated flags related caching. This commit removes all caching related flags in `vine_mount_flags_t`. The "no flags" option in vine_mount_flags_t is the new `VINE_TRANSFER_ALWAYS`. Removing VINE_CACHE from vine_mount_flags_t showed that this mount flag was used in vine_task_add_{in,out}put in all the examples, which was incorrect.
    
file flags where renamed to `VINE_CACHE_NEVER`, `VINE_CACHE`, `VINE_CACHE_ALWAYS`, and `VINE_PEER_NOSHARE`. The main ideas are that `VINE_CACHE&VINE_CACHE_ALWAYS == VINE_CACHE`, and that peer transfer has to be disabled on a per file basis, as we assume that the users want peer transfers as they explicitly have to enable them.

### DIFF
--- a/taskvine/src/bindings/python3/taskvine.binding.py
+++ b/taskvine/src/bindings/python3/taskvine.binding.py
@@ -98,10 +98,8 @@ class Task(object):
             pass
 
     @staticmethod
-    def _determine_mount_flags(cache=False, watch=False, failure_only=False, success_only=False):
-        flags = VINE_NOCACHE
-        if cache:
-            flags |= VINE_CACHE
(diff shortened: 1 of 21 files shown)


dthain:
@BarrySlyDelgado @colinthomas-z80 please weigh in on this so we can make a decision and move on to application evaluation..

BarrySlyDelgado (approved):

colinthomas-z80 (approved):

PR #3208, opened 2023-03-16
Adds execution environments to tasks

btovar:
An environment is a vine file that expands to a directory. The directory is expected to a have a `bin/run_in_env` wrapper.
Environments nest, applied in the order they are defined.

I modified starch to have the `bin/run_in_env`, and update the chirp example to show how it works for chirp. (It worked!)

Something nice about this approach is that we can design general mini task that have as an output a directory with the bin/run_in_env wrapper. (E.g. for apptainer, or other execution wrappers.)

@BarrySlyDelgado 
I think we should do something similar to poncho. The poncho environment should include bin/poncho_package_run, and a script bin/run_in_env that knows how to run poncho_package_run for that environment.

### DIFF
--- a/makeflow/src/starch
+++ b/makeflow/src/starch
@@ -94,9 +94,9 @@ extract() {
 }
 
 run() {
-    # make the extraction directory easily available in run.sh
+    # make the extraction directory easily available in bin/run.sh
     export SFX_DIR
-    $SFX_DIR/run.sh "$@"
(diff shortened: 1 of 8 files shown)


dthain (approved):
Clever, I like this unification of abstractions!

BarrySlyDelgado:
Yes, I made something similar earlier. Though I added an option to poncho_package_run that does it instead of another script in bin/. Though I think having the other script is better.

PR #3216, opened 2023-03-17
change cache flags for tarballs from always to never

tphung3:
Tarballs should not be cached I think

### DIFF
--- a/taskvine/src/examples/vine_example_blast.c
+++ b/taskvine/src/examples/vine_example_blast.c
@@ -104,15 +104,18 @@ int main(int argc, char *argv[])
 	vine_enable_peer_transfers(m);
 
 	printf("Declaring files...");
-	//These files do not change across vine runs, thus we cache them with VINE_CACHE_ALWAYS,
+	
+    //These archival files are only needed once right after download as they will be 
+    //transformed to the actual software files in `vine_declare_untar` below.
+	//These declarations only register the files with the manager, but do not
+	//associate them with any task (yet).
+	struct vine_file *blast_url = vine_declare_url(m, BLAST_URL, VINE_CACHE_NEVER);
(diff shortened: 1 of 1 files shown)


dthain:
I'm confused -- don't we want the tarballs to be widely replicated and stored between runs?

btovar:
Yes, I think they should be cached. In fact that's one of the main points this example shows: expensive to transfer of files is only done once per worker.

tphung3:
We want the thing that the tarball unpacks to be cached, not the tarball right?

dthain:
Ah, I seem to recall that we had some difficulty caching and sharing the output of a mini-task.  @colinthomas-z80 does this work now?

btovar:
For tarballs we never want to share the output, as expanded it may have thousands of files. Mini task also specialize to the task sandbox.

colinthomas-z80:
@btovar if we refer to vine_file.c, taking vine_declare_untar as an example, does specifying caching/sharing flags actually do anything at all?

tphung3:
> For tarballs we never want to share the output, as expanded it may have thousands of files. Mini task also specialize to the task sandbox.

But the untar minitask on its own will inflate those thousands of files anyway right?

btovar:
@colinthomas-z80  The flags given to the mini task are the flags that it's output gets. I don't think the mini task gets any flags for itself.

@tphung3 Yes, but it will do so locally. We don't have to send all those files uncompressed over the network.

tphung3:
@btovar I sort of see what you are saying. So we cache the output (1000s of files) of a minitask on a worker, but share (via peer transfer) the input (tarball) of that minitask to other workers.

btovar:
Yes, I think that's the current idea. I know Colin is trying to figure out how to share the output of a mini task when it makes sense, which should produce some interesting usage patterns.

PR #3232, opened 2023-03-23
vine library: small changes to function docs, example with command line arguments

btovar:
Changed env_path to env to match rest of code.
Added command line arguments (e.g. --iterations) to example to easily test with one worker from command line.
Added missing cache=True to envs and library codes.


David-Simonetti-ND (approved):
Looks really good! I like the changes to the gradient descent example. Should make it much more flexible to run now.

PR #3264, opened 2023-04-11
Conda from vars

btovar:
Recent changes to the conda build process cause the configure script to pickup the incorrect version of python. There are two version of python, the one we want to build against, and the one in the build environment. In previous versions of the build environment these were the same.

This pr picks up the version from conda environment variables when available. It also removes python2 and golang from the configure script, as their wq bindings are out of date and we won't update them.

### DIFF
--- a/configure
+++ b/configure
@@ -208,10 +208,6 @@ do
 		--globus-flavor)
 			globus_flavor=${arg}
 			;;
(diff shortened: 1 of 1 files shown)


dthain (approved):
+1 for consolidating on python/python3
+1 for removing golang

PR #3265, opened 2023-04-11
Reproducibility edits to BGD example

David-Simonetti-ND:
Changed around some resource allocations in the example and allowed the number of tasks to be passed as an argument on the command line.

### DIFF
--- a/taskvine/src/examples/vine_example_gradient_descent.py
+++ b/taskvine/src/examples/vine_example_gradient_descent.py
@@ -91,16 +91,16 @@ def main(number_of_params, max_iterations, min_error, learning_rate, num_tasks):
 
     # specify resources used by Library Task
     t.set_cores(1)
-    t.set_disk(0)
-    t.set_memory(0)
+    t.set_disk(2000)
+    t.set_memory(2000)
 
     # install the Library Task on all workers that will connected to the manager
     m.install_library(t)
 
-    # create our data by sampling 100 points off a sin curve and adding noise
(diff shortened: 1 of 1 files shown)


dthain:
Hold on - I want to make sure that we get the right things into the right places here.

The `examples` directory is meant to be ...  examples of how to use TaskVine, and may change as we add features, find better ways to express things, etc.  It's fine that you are adding command-line switches and so forth.  But surely cores(0) and disk(0) isn't the final form of what we want the examples to look like.

However, to the extent that our goal is to capture the figures in the paper, just put the exact code of `vine_example_gradient_descent.py` that you used into the taskvine-paper repository.  So then that repository is a faithful representation of what's in the paper, warts and all.  And the software repository can evolve as needed.

David-Simonetti-ND:
I see, things make a lot more sense now

btovar:
Including the python examples makes sense, but not so much for the modified C examples. I was trying to avoid the need of any compilation and for the compiled examples to be included in the conda distribution. The reasoning was that compilation is harder to get right for every case and we do not want the reviewers to need to modify a makefile.

Were any of the figures produced from the C examples?

dthain:
I'm pretty sure the figures were all done with python.

David-Simonetti-ND:
@btovar This is a small change that just adds an extra option to the bgd example for number of tasks. Should be safe to merge.

btovar:
Is it different from the one already in taskvine-paper?

PR #3280, opened 2023-04-18
vine: adds apptainer example

btovar:


btovar:
I'm still in awe that this worked with so naturally and in such a self-contained way.

PR #3288, opened 2023-04-21
dill to cloudpickle

btovar:
Needed because cloudpickle can serialize some of the objects that dask needs. In the internal use of PythonTask, dill and cloudpickle should be interchangable.

@David-Simonetti-ND I don't think this change affects coprocesses, as I don't think your code relies on particulars of dill.
@tphung3 Is dill used with the parsl integration? I think parsl does its own serialization.


BarrySlyDelgado (approved):

David-Simonetti-ND (approved):
Looks good! Serverless doesn't use dill at all so there is no change there.

Issue #3282, opened 2023-04-20
Vine: Peer to Peer Transfer Fixups


dthain:
Currently, worker-to-worker transfers operate correctly when manually enabled, and (I think) only for URL transfers.  Fix up two things so that we can use them ubiquitously:
- [ ] Worker-to-worker transfers should operate correctly for any file type (local, buffer, mini-task output) and not only URLs.
- [ ] Worker-to-worker transfers should be enabled by default and controlled with an `enable/disable` pair in the API.
- [ ] Add a note about enabling and disabling peer transfers in the documentation.

colinthomas-z80:
Resolved in #3290

PR #3290, opened 2023-04-25
Vine: Peer Transfer Fixups

colinthomas-z80:
Initial work on issue #3282 

Normal input files sent by the manager are properly considered for peer transfers now

### DIFF
--- a/doc/manuals/taskvine/index.md
+++ b/doc/manuals/taskvine/index.md
@@ -724,6 +724,28 @@ that peer transfers are not permitted:
     ```
     vine_declare_file(m,"myfile.txt",VINE_CACHE|VINE_PEER_NOSHARE)
     ```
+Automatic sharing of files between workers, or peer transfers, are enabled by default
+in TaskVine. If communication between workers is not possible or not desired, peer transfers
+may be globally disabled:
+
+=== "Python"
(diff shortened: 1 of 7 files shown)


dthain:
That change makes sense to me.  Do peer transfers also work for mini-tasks and temporary files?  Those work a little differently b/c they originate at a worker and not at the manager.

colinthomas-z80:
Mini tasks still need additional work. Temp files are good to go.

dthain:
RTM?

colinthomas-z80:
@dthain still working on an issue, also have to add python bindings and document

dthain (changes requested):
I can see `remove_one_manager` but it doesn't seem to get called anywhere.  Am I missing something?

colinthomas-z80:
@dthain I had originally used that as a way of adding/removing a manager entry to the transfer table, but later on I figured tracking the manager transfers was just unnecessary overhead since we are only making sequential transfers anyway.

dthain (approved):
Make sure you link and  close the relevant issue.

btovar (changes requested):
Please remove the function duplicates from manager.py

btovar (approved):

PR #3314, opened 2023-04-30
sort directory for hashing

BarrySlyDelgado:


dthain (changes requested):
Must handle case when sort_dir fails.

dthain (approved):

PR #3322, opened 2023-05-05
Add metadata to directory checksum

BarrySlyDelgado:


taskvine/src/manager/vine_checksum.c
```diff
@@ -35,20 +35,29 @@ static char *vine_checksum_dir( const char *path, ssize_t *totalsize )
...
+
 		char *subpath = string_format("%s/%s",path,entries[i]);
+		if(stat(subpath, &info)) return 0;
+
+		char *mode = string_format("%o",info.st_mode);
+		char *mtime = ctime(&info.st_mtime);
+		char *meta = string_format("%s:%s",mode,mtime);
 		char *subhash = vine_checksum_any(subpath,totalsize);
-		char *line = string_format("%s:%s\n",entries[i],subhash);
+		char *line = string_format("%s:%s:%s:\n",entries[i],meta,subhash);
```
dthain, line 53: Just do one big `string_format` with all the fields here, that way you can save three `malloc/free` pairs in the inner loop/.

dthain (changes requested):

Issue #3317, opened 2023-05-01
vine: re-declaring a file using the same file name creates two different file objects when there should be only one


tphung3:
When a regular file is declared using its file name and with its cache flag set to `never` or `workflow`, then a re-declaration of the same file creates a different file object as the TaskVine name of the file is randomly generated.

A re-declaration of a file should be detected and mapped to the already generated file object, and this can be done by (either/or):
1. Maintain a table that maps a file's path (entry name) to its TaskVine name
2. A file's TaskVine name is not randomly generated but hashed using its entry name

btovar:
Another option is to just give a warning if the flags are different, and just keep going with the first declaration. Files are supposed to be immutable, so changing their names mid run breaks a design principle and it may be a lot of work for little gain.

dthain:
The problem here is that this code sequence creates two different objects:
```
a = m.declare_file("xyz")
b = m.declare_file("xyz")
```

Because local files (when not cached forever) create a random unique number, to avoid the cost of hashing on content.

tphung3:
Only VINE_FILE and VINE_URL are possibly de-duplicable by maintaining a mapping from source to file object. VINE_BUFFER is not without looking at the buffer's content. VINE_TEMP and VINE_MINI_TASK are unclear to me how the de-duplication happens. 

Deduplicating only VINE_FILE and VINE_URL is probably confusing and inconsistent, so it's probably the best to not change anything and let the user/app do the deduplication.

We can also have something in between where TaskVine maintains a mapping from source to file object for VINE_FILE and VINE_URL and throws a warning whenever it detects possible duplication.

tphung3:
Yes that's correct. The executor maintains a dictionary that maps a file path to a TaskVine File object and checks the dictionary for every file in a submitted task.

dthain:
I believe that this is fixed at two different levels:
- In TV, declaring the same file twice should result in the same underlying hash.
- When used with Dask/Parsl the upper level DAG processor will declare each file once.
-

PR #3323, opened 2023-05-08
metadata cache names for vine_files that are not CACHE_ALWAYS

BarrySlyDelgado:
This is for #3317. VINE_FILEs are given cache names based on their meta data when they are declared with a flag that is not CACHE_ALWAYS. Previously, they were just given a random cache name.

### DIFF
--- a/taskvine/src/manager/vine_cached_name.c
+++ b/taskvine/src/manager/vine_cached_name.c
@@ -279,7 +279,7 @@ char *vine_meta_name(const struct vine_file *f, ssize_t *totalsize ){
 
 	if(path_disk_size_info_get(f->source, &size, &number_of_files)) return 0;
 
-	char *meta = string_format("%s-%ld-%s", f->source, size, mtime);
+	char *meta = string_format("%s-%"PRIu64"-%s", f->source, size, mtime);
 
 	char *metahash = md5_of_string(meta);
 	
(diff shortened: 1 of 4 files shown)


taskvine/src/manager/vine_cached_name.c
```diff
@@ -261,7 +263,34 @@ char *make_mini_task_cached_name(const struct vine_file *f)
...
+	int64_t number_of_files;
+
+	if(stat(f->source, &info)) return 0;
+	char *mtime = ctime(&info.st_mtime);
 
+	if(path_disk_size_info_get(f->source, &size, &number_of_files)) return 0;
+
+	char *meta = string_format("%s-%ld-%s", f->source, size, mtime);
+
+	md5_buffer(meta,strlen(meta),digest);
```
dthain, line 285: Can simplify with `md5_of_string` here.

dthain (changes requested):

taskvine/src/manager/vine_cached_name.c
```diff
@@ -261,7 +263,33 @@ char *make_mini_task_cached_name(const struct vine_file *f)
...
+	if(f->type != VINE_FILE) return 0;
+
+	struct stat info;
+	int64_t size;
+	int64_t number_of_files;
+
+	if(stat(f->source, &info)) return 0;
+	char *mtime = ctime(&info.st_mtime);
+
+	if(path_disk_size_info_get(f->source, &size, &number_of_files)) return 0;
```
dthain, line 280: Last thing: don't use `path_disk_size_info_get` here because it is quite expensive for large directory trees.  (And if we cared about detecting changes, we would use CACHE_FOREVER anyway.)  Just use `info.st_size` instead.

dthain (changes requested):

dthain:
@BarrySlyDelgado when you make changes, please re-request the review so I know it's been updated.

dthain (approved):

PR #3326, opened 2023-05-09
Vine: Scheduling Optimizations

colinthomas-z80:


dttools/src/rmsummary.c
```diff
@@ -699,18 +699,29 @@ void rmsummary_bin_op(struct rmsummary *dest, const struct rmsummary *src, rm_bi
...
 
+/* This function may be called many times, therefore it is worthwhile to omit/inline
+ * function calls and remove indirect jumps to rmsummary_bin_op */
 void rmsummary_merge_override(struct rmsummary *dest, const struct rmsummary *src)
 {
 	if(!src) {
 		return;
 	}
 
-	rmsummary_bin_op(dest, src, override_field);
```
btovar, line 713: Instead of this, what about declaring `rmsummary_bin_op` as static inline?

dthain: Let's have a look at the performance effect before reaching a conclusion..

btovar: If the call to `rmsummary_bin_op` is really a problem, I would first convert it to a macro rather than copy its contents where it is needed. (I.e., copy the contents with a macro instead.)

Also, the `rmsummary_bin_op` compares more than just cores, memory, disk and gpus.  You may want to add a version that just compares the resources needed in taskvine, as I think we do in category.h.

taskvine/src/manager/vine_schedule.c
```diff
@@ -145,26 +145,36 @@ static struct vine_worker_info *find_worker_by_files(struct vine_manager *q, str
...
 	struct vine_file_replica *remote_info;
 	struct vine_mount *m;
 
-	HASH_TABLE_ITERATE(q->worker_table,key,w) {
+	HASH_TABLE_ITERATE(q->worker_table, key, w) {
 		if( check_worker_against_task(q, w, t) ) {
 			task_cached_bytes = 0;
+			has_all_files = 1;
+
 			LIST_ITERATE(t->input_mounts,m) {
```
btovar, line 157: @colinthomas-z80 Maybe you should also skip over files that have VINE_NO_CACHE. If a task has no inputs marked as cached, then any worker would do.

dthain:
@colinthomas-z80 this approach makes sense to me.  The macros likely remove substantial overhead without breaking the abstraction.

One naming thing:`rm_summary_merge_vine` tells us who is using it, but not why.  Maybe call it `_basic` or `_minimal` or something like that so it's obvious that the function just does less.  (And we might end up using it in WQ as well.)

dthain:
@colinthomas-z80 what's the status on this one?

colinthomas-z80:
@dthain I think this is at a good point to merge in and continue work with smaller, more focused PRs

dthain (approved):

PR #3329, opened 2023-05-09
poncho: add option to package directory directly

tphung3:
This new feature allows poncho to package a specified conda directory directly, which may include editable packages on a shared file system.

### DIFF
--- a/poncho/src/poncho/package_create.py
+++ b/poncho/src/poncho/package_create.py
@@ -97,77 +97,82 @@ def _copy_run_in_env(env_dir):
         f.write('exec "${env_dir}"/bin/poncho_package_run -e ${env_dir} "$@"\n')
     os.chmod(f'{env_dir}/env/bin/run_in_env', 0o755)
 
+def pack_env_with_conda_dir(spec, output, ignore_editable_packages=False):
+    # remove trailing slash if present
+    spec = spec[:-1] if spec[-1] == '/' else spec 
+    try:
+        logger.info('packaging the environment...')
+        os.makedirs(f'{spec}/bin/', exist_ok=True)
+        os.makedirs(f'{spec}/env/bin/', exist_ok=True)
+        _copy_run_in_env(spec)
+        os.rename(f'{spec}/env/bin/poncho_package_run', f'{spec}/bin/poncho_package_run')
+        os.rename(f'{spec}/env/bin/run_in_env', f'{spec}/bin/run_in_env')
+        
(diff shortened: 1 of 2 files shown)


poncho/src/poncho/package_create.py
```diff
@@ -98,51 +98,74 @@ def _copy_run_in_env(env_dir):
...
+def pack_env(spec, output, conda_executable=None, download_micromamba=None, ignore_editable_packages=None):
+    # pack a conda directory directly
+    if not os.path.isfile(spec) and spec != "-":
+        # remove trailing slash if present
+        spec = spec[:-1] if spec[-1] == '/' else spec 
+        try:
+            logger.info('packaging the environment...')
+            _copy_run_in_env(spec)
+            os.makedirs(f'{spec}/bin/', exist_ok=True)
+            os.rename(f'{spec}/env/bin/poncho_package_run', f'{spec}/bin/poncho_package_run')
```
btovar, line 110: This function is already too long. Let's split it in your new case, and the case that was there before.

btovar (changes requested):

BarrySlyDelgado (approved):

poncho/src/poncho/package_create.py
```diff
@@ -97,77 +97,82 @@ def _copy_run_in_env(env_dir):
...
+        os.rename(f'{spec}/env/bin/poncho_package_run', f'{spec}/bin/poncho_package_run')
+        os.rename(f'{spec}/env/bin/run_in_env', f'{spec}/bin/run_in_env')
+        
+        conda_pack.pack(prefix=f'{spec}', output=str(output), force=True, ignore_missing_files=True, ignore_editable_packages=ignore_editable_packages)
+        logger.info('to activate environment run poncho_package_run -e {} <command>'.format(output))
+        return output
+    except Exception as e:
+        raise Exception(f"Error when packing a conda directory.\n{e}")
+    finally:
+        os.remove(f'{spec}/bin/run_in_env')
```
btovar, line 117: Why do you remove run_in_env and poncho?

btovar (changes requested):

poncho/src/poncho/package_create.py
tphung3: Hmm on second thought maybe poncho_package_run is already there and I would accidentally remove it.. run_in_env is ok to remove as it's already in the created tarball.

btovar (approved):

PR #3341, opened 2023-05-11
Replace centos8 with almalinux8

btovar:
centos8 does not reflect rhel8 as previous centos versions did. Instead we can use almalinux as a replacement for centos.

This pr replaces centos8 with almalinux8.  The version of mysql used to compile parrot does not compile anymore in almalinux9, though, which is something to consider for bxgrid when the campus upgrades to rhel9.

alma9 was added to the script that generates the images, but cctools does not compile against it yet.


btovar:
Next release will have alma8 instead of centos8.  The script that deploys to ND campus has been updated with this change.

Issue #3103, opened 2023-02-08
Poncho: Package Preparation Primarily in Python Please


dthain:
Right now, Poncho requires you to use the Unix command line in order to create a poncho package.  But there are many uses (like what @David-Simonetti-ND is doing) where we want to create the environment directly from Python.  The first step is just something like this for normal tasks:

```
p = [ "matplotlib","tensor flow","pytorch"]
pkgfile = poncho.MakePackageFromList(p)
task.specify_environment(pkgfile)
```

But then you can do the same thing for duties, so that it is easy to define the environment once.

And then, you might even make the construction of the package a mini-task!

dthain:
@mcarbona1 I think this is a good next step for you, and you can test it out with David.

David-Simonetti-ND:
I think a lot of the machinery needed for this already exists. I don't even think there needs to be a manual list of packages, as if we have the user's function to creation an environment for, there are methods in the poncho python library (scan_pkgs and analyze_function in package_analyze) that can get a full list of packages from the installed conda environment, and then output a yml description of dependencies for that user function. 
Once we have the yml description, package_create can turn that into an actual tarball, and that can be added as an environment for the task.

Some rough pseudocode for what could be done.
```
from poncho import package_analyze as analyze
from poncho import package_create as create

# function we wish to create an environment for
def user_function():
    # lines of code, imports, etc here

# get info about current conda environment
analyze.scan_pkgs()

# read in the code for this file and parse it 
fnc_code = ast.parse(user_function)
# get imports we need
imports = analyze.analyze_function(fnc_code)

env_yml = analyze.export_env(imports)

create.pack_env(env_yml, output_file)

```

dthain:
Yes, the pieces for this are already here, but I want to keep two things separate:
1 - Scanning a function to produce a dependency list.
2 - Packing a dependency list into a package.

The reason for this is that we may want to attach packages to ordinary executables, and not only to server less functions.

dthain:
A related problem I see is that `declare_poncho` incorrectly takes a filename as an argument and assumes that it is a local file source.  Instead, it should take a vine file object as a source, so that it can draw packages from anywhere.

btovar:
I think declare_poncho has both in python. You can pass a vine file, but if you are lazy, you can pass a string a declare_poncho declare_file's that for you.

dthain:
Ok, that makes more sense to me.

btovar:
[...]

PR #3349, opened 2023-05-16
create poncho package from dict or string (WIP)

BarrySlyDelgado:
This addresses #3103. This allows users to create poncho packages in python directly in python using `create_env_from_dict` which accepts a dict object or a string corresponding to the poncho specification. It hashes the spec to create the name of the resulting tarball. currently I think the spec should be sorted for consistency.

### DIFF
--- a/doc/manuals/poncho/index.md
+++ b/doc/manuals/poncho/index.md
@@ -255,14 +255,14 @@ will attempt to pack an environment of the same name from the users local conda 
 
 The poncho module allows users to create poncho packages within python itself.
 
-### createing packages with a specification
+### creating packages with a specification
 
 Poncho packages can either be created by dicitionary or string representations of a poncho specification.
 The function `dict_to_env` creates the corresponding environment and returns the path to the environment.
 The function contains various options to facilitate environment creation:
 	
 	- cache(default=True): caches the environment in the directory set by `cache_path` 
-	- cache_path(default=None): Path to cache and retrieve generated environments.
+	- cache_path(default='.poncho_cache'): Path to cache and retrieve generated environments.
 	- force(default=False): forces poncho_package_create to recreate the environment.
 
 If no cache path is specified, cached environments will be stored in the directory `.poncho_cache`.
--- a/taskvine/src/examples/vine_example_poncho_in_python.py
+++ b/taskvine/src/examples/vine_example_poncho_in_python.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+
+# Example on how to execute python code while creating an poncho package within python.
+
+# A poncho specification can be given by a string or dictionary representation.
+# The function package_create.dict_to_env(spec) will create the corresponding 
+# poncho package. The package will be cached when cache=True. The default location
+# of the poncho cache is `.poncho_cache` but can be modiied by providing a cache path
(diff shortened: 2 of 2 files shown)


poncho/src/poncho/package_create.py
```diff
@@ -24,6 +24,20 @@
 
 conda_exec = 'conda'
 
+def create_env_from_dict(spec, conda_executable=None, download_micromamba=False, ignore_editable_packages=False):
```
btovar, line 27: @BarrySlyDelgado Giving that pack_env transforms the json file again to a dict, wouldn't it be better to have `create_env_from_dict` as the "main" function? Then the one that uses json can json.loads it and pass it to the new `create_env_from_dict`.

dthain:
Ping?

poncho/src/poncho/package_create.py
BarrySlyDelgado: done.

BarrySlyDelgado:
should be ready to merge.

dthain:
Ok, will the necessary package be created every time this is called,
or is it cached in some way?

dthain:
I think this needs more cooking before it is a complete feature that others will want to use:
- Should cache packages created based on hash of contents.
- Needs an entry in the manual

BarrySlyDelgado:
Added the Functionality that allows poncho packages to be cached. The created environment has a a name generated by the contents of the specification. The user provided specification is sorted before creating the hash for consistency. users can provide a directory to serve as a cache like so: `env = dict_to_env(spec, cache_path=<DIR_NAME>)` if an environment is to be cached but no cache path was provided, thee cache path will be `.poncho_cache`. Users can force the recreation of environment even if the environment is present within the cache like so:  `env = dict_to_env(spec, force=True)`. The Poncho Manual has been updated to include the details of this change.

doc/manuals/poncho/index.md
```diff
@@ -243,3 +243,37 @@ file will be passed through `tar` and extracted into a temporary
...
+If the input to poncho\_package\_create is the path to a conda env directory, poncho\_package\_run
+will package that directory as a poncho package. If the input is neither a file or directory poncho\_package\_run 
+will attempt to pack an environment of the same name from the users local conda environments.
+
+
+## Creating Poncho Packages Within Python
+
+The poncho module allows users to create poncho packages within python itself.
+
+### createing packages with a specification
```
dthain, line 258: Spelling

doc/manuals/poncho/index.md
```diff
@@ -243,3 +243,37 @@ file will be passed through `tar` and extracted into a temporary
...
+
+Poncho packages can either be created by dicitionary or string representations of a poncho specification.
+The function `dict_to_env` creates the corresponding environment and returns the path to the environment.
+The function contains various options to facilitate environment creation:
+	
+	- cache(default=True): caches the environment in the directory set by `cache_path` 
+	- cache_path(default=None): Path to cache and retrieve generated environments.
```
dthain, line 265: `default=.poncho_cache` ?

dthain (changes requested):

dthain (approved):

dthain:
RTM?

BarrySlyDelgado:
Yes.

PR #3350, opened 2023-05-18
Vine: Worker-Worker Authentication

dthain:
When password auth is available, use it for worker-worker connections.


colinthomas-z80 (approved):

dthain:
Still want to think about renaming this, but will consider that late.r..

PR #3357, opened 2023-05-26
vine: use reduced rmsummary op in work queue

colinthomas-z80:


colinthomas-z80:
I did not think very long about the title...

Issue #3363, opened 2023-05-30
Vine: Manual is inconsistent


tphung3:
1. In section `Running a TaskVine Application`, a script in Python is provided (`functions.py`) but the followed description uses `taskvine_example.py` instead, which is non-existent.

2. Also in the subsection `Submitting Workers to a Batch System`, these scripts work only with WorkQueue, not TaskVine. This will be fixed in an upcoming PR.

3. Subsection `Apptainer Custom Environment` has an empty reference to `run_command_in_apptainer.sh`.

dthain:
1 and 2 look fixed to me.
Is 3 still a problem?

tphung3:
only 2 is fixed in PR #3365 , 1 and 3 are not.

dthain:
@btovar do you happen to have `run_command_in_apptainer.sh` sitting in a directory somewhere?
It looks like you wrote it into the docs but didn't add the file.

btovar:
Yes, sorry about that! Opened https://github.com/cooperative-computing-lab/cctools/pull/3371

PR #3365, opened 2023-05-30
Vine: add wrapper scripts to batch systems

tphung3:
Wrapper scripts are updated to use `vine_worker` and have names changed to avoid reusing WQ wrapper scripts. WQ's scripts should probably have their names changed too (e.g., from `condor_submit_workers` to `work_queue_condor_submit_workers`)


tphung3:
Fix 2nd point in this issue https://github.com/cooperative-computing-lab/cctools/issues/3363

btovar:
I'd recommend not renaming the wq scripts as several users know the current names.

dthain:
Agreed that we should not rename the WQ scripts.  (That would also requiring changing all the documentation, examples, test scripts, etc)

Re the vine side of things, what would you think of consolidating all of the `vine_X_submit` scripts into one single `vine_submit_workers` script?

dthain:
That's a good point.  In years past, I would have been reluctant to make this script depend upon Python.  But in a Conda world, I think it is the better way to go.

dthain:
Oh, so my point was more about making one self-contained script, instead of having multiple scripts that invoke each other.

dthain:
Suggestion on how to proceed in order to bring us up to date:
- Drop torque and pbs, since we have not used or tested those in quite some time.
- Keep condor, sge, and slurm, since that covers the hpc space pretty well today.
- Combine all the scripts into one file called `vine_submit_workers` including batch.submit.common.
- Keep your existing check of `-T type` as the first argument, and use that to call `condor_parse_args`, `sge_parse_args` and so forth...
-

dthain:
That's looking nice, you are on the right track.

tphung3:
If all is well this PR is ready to merge.

dthain:
Script looks good, please update the corresponding documentation:
https://cctools.readthedocs.io/en/stable/taskvine/#running-a-taskvine-application

And add a manpage here:
https://github.com/cooperative-computing-lab/cctools/tree/master/doc/man/m4

tphung3:
> Script looks good, please update the corresponding documentation: https://cctools.readthedocs.io/en/stable/taskvine/#running-a-taskvine-application
> 
> And add a manpage here: https://github.com/cooperative-computing-lab/cctools/tree/master/doc/man/m4

Doc changes in this PR https://github.com/cooperative-computing-lab/cctools/pull/3368

dthain (approved):

PR #3368, opened 2023-06-02
Vine: Fix more vine doc

tphung3:
Fix more docs and add man page for `vine_submit_workers`

### DIFF
--- a/doc/manuals/taskvine/index.md
+++ b/doc/manuals/taskvine/index.md
@@ -524,7 +524,7 @@ If you have access to a HTCondor pool, you can use this shortcut to submit ten
 workers at once via HTCondor:
 
 ```sh
-$ vine_condor_submit_workers MACHINENAME 9123 10
+$ vine_submit_workers -T condor MACHINENAME 9123 10
 
 Submitting job(s)..........
 Logging submit event(s)..........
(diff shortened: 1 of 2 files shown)


doc/manuals/taskvine/index.md
```diff
@@ -707,9 +705,9 @@ software packages and reference datasets.
 
 === "Python"
     ```python
-    f = m.declare_file("myfile.txt",cache="never")
-    f = m.declare_file("myfile.txt",cache="workflow") # (default)
-    f = m.declare_file("myfile.txt",cache="always") # (default)
+    f = m.declare_file("myfile.txt",cache="never") # (default)
```
dthain, line 708: workflow caching is the default.

dthain (approved):

doc/manuals/taskvine/index.md
tphung3: Do we want to change the default value here then?
https://github.com/cooperative-computing-lab/cctools/blob/cdc25542f936df263b75f72e14b321ad9d5960df/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py#L1355

tphung3: For now let's match the code as other code depends on this behavior. Can't think which is beneficial in the long run though..

Issue #3130, opened 2023-02-16
Vine: Tune up Hash Methods


dthain:
The computation of file hashes should be improved in several ways:
- [x] `task_to_json` creates a string representation of a task, which is hashed to give a unique task name.  This needs to be looked at carefully and include all elements that control the output of the task.  For example, environment variables should be part of the string.
- [x] `md5_dir` computes a directory hash by having strings containing the name and hash of each file in a directory.  This should be extended to include metadata (permissions, mtime, etc) so that metadata changes force a new object creation.
- [x] `md5_dir` should sort the entries in each directory before hashing, so that values are independent of accidental ordering.

Each one of these is an independent improvement and should be a separate focused PR.

BarrySlyDelgado:
I've completed the metadata hashing and sorting

BarrySlyDelgado:
Completed with #3372

PR #3372, opened 2023-06-06
Task to json: add env variables

BarrySlyDelgado:
For #3130. Adds environment variables to task json.


taskvine/src/manager/vine_task.c
```diff
@@ -845,6 +845,16 @@ char * vine_task_to_json(struct vine_task *t)
 		}
 		buffer = string_combine(buffer, "\n");
 	}
+
+	if(t->env_list){
```
btovar, line 849: @BarrySlyDelgado I suggest using `dttools/src/buffer.h` here, as it is more efficient and the intent is clearer.

dthain (approved):

PR #3373, opened 2023-06-06
WQ:  Attempt scheduling a fixed number of tasks to expedite retrievals

colinthomas-z80:


dthain:
Please rebase on master for a clean build.

work_queue/src/work_queue.c
```diff
@@ -7391,6 +7397,9 @@ int work_queue_tune(struct work_queue *q, const char *name, double value)
 	} else if(!strcmp(name, "wait-for-workers")) {
 		q->wait_for_workers = MAX(0, (int)value);
 
+	} else if(!strcmp(name, "attempt-schedule-depth")) {
```
dthain, line 7400: Make sure to add this new tuning parameter to the docs in `taskvine.h`

work_queue/src/work_queue.c
```diff
@@ -4569,29 +4570,39 @@ static int send_one_task( struct work_queue *q )
 	struct work_queue_task *t;
 	struct work_queue_worker *w;
 
+	int tasks_considered = 0;
 	timestamp_t now = timestamp_get();
 
-	// Consider each task in the order of priority:
-	list_first_item(q->ready_list);
-	while( (t = list_next_item(q->ready_list))) {
```
dthain, line 4576: I'm a little uncomfortable with the implied free/malloc that comes from pop/push every time we consider one item.  Also, the multiple exit points that all must call `list_push_tail` looks like a maintenance hazard.

What would you think of introducing a new call `list_rotate` that moved the head item to the tail (without free/malloc) returning its value?

So then you would have this:
```
while( t= list_rotate(q->ready_list) ) {
    // check various conditions and return if needed

   // finally
   list_pop_tail
   commit_task_to_worker
}
```

dthain (changes requested):

dthain (approved):

colinthomas-z80:
@dthain yes!

PR #3375, opened 2023-06-06
Poncho: Stop logger from having a global effect

tphung3:
`logging.basicConfig` has a global effect which might affect higher applications (e.g., this sort of hijacks the global logging config and stops them from setting their own configurations, happened with parsl+vine). This fix removes such behavior and only adds a handler (object telling messages where to go) to the locally created logger.

### DIFF
--- a/poncho/src/poncho/package_create.py
+++ b/poncho/src/poncho/package_create.py
@@ -19,8 +19,12 @@ import re
 from platform import uname
 from packaging import version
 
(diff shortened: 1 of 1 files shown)


btovar (approved):

BarrySlyDelgado (approved):

tphung3:
This PR is ready to merge.

Issue #3284, opened 2023-04-20
Vine: Measure and Tune Dispatch Rates


dthain:
Now that we have fast serverless tasks running in TaskVine+Parsl, the scheduling of tasks is likely to be the performance constraint, if we intend to have millions of tasks running on thousands of nodes.  Let's understand the performance of the current TaskVine scheduler and if necessary, see what algorithmic improvements can be made to get dispatch rates that are reasonable stable with respect to the number of tasks and workers.
- [x] Create a simple dispatch benchmark for null tasks.
- [x] Write up a high level pseudo-code description of the current scheduling loop.
- [x] Propose some improvements to our approach.

dthain:
@colinthomas-z80 please summarize here what you have found and keep some running notes as you go.

colinthomas-z80:
Separate from the issue of scheduling tasks to workers, we have observed that having an exceptionally large waiting task queue will cause the manager to spend unnecessary time iterating through tasks to schedule, when it should be fetching results so the workers will become available again. 

That is to say, with 10k+ tasks submitted, the manager will dispatch tasks to all workers, yet continue to iterate through the 10k tasks trying to find one that will fit the busy workers. In throughput testing cases, and perhaps some practical cases, the tasks will finish before the manager is even done iterating through the list. Therefore it would be more effective for the manager to retrieve these tasks and perform other bookkeeping rather than iterate through tasks that cannot be scheduled. 

A simple test where we only attempt to schedule the waiting task at the top of the list shows much better retained throughput as the size of the task queue grows. This however would severely limit workers from being effectively packed in the case of diverse tasks (3 core and 1 core tasks submitted to be run in parallel). Attempting to make a quick judgement about the resources available in the cluster compared to the task requirements in the queue is perhaps not possible. 

One possible method would be to attempt scheduling a fixed number of tasks, and if none succeed, then we assume no workers are available and results should be fetched. This method shows good throughput results by considering 100 tasks. The implications of this on other aspects of the workflow still need to be studied.

dthain:
A slight variation: just keep a cursor in the list, and examine ~100 before going back through the main loop.  Then, next time, pick up where you left off.  That way, you eventually make it through the entire list, just not all in one scheduling pass.

dthain:
Alternate discussed today: add `list_rotate` option that moves the head to the tail, then you can just consider one item and keep going, up until N.

colinthomas-z80:
[...]

PR #3380, opened 2023-06-16
WQ: Further mainloop optimizations

colinthomas-z80:


colinthomas-z80:
@benclifford This change improved my performance in the ranges of 10-20k tasks in queue. But yes it still does not hold up in the larger queue sizes. I believe the remaining issue is in the way we retrieve tasks.

colinthomas-z80:
Task retrieval is shaped up a bit, now the suspect seems to be logging overhead as seen in the profile on #3284

dthain:
Looks like this is making progress but maybe not ready to merge yet.
Let me know when you think it's ready.

colinthomas-z80:
@dthain Before this is done I wanted to hear your thoughts on adding some sort of "high throughput" mode which disables the performance log.

dthain:
Some quick thoughts -
- Ideally, the performance log shouldn't come at an enormous overhead.  Check to see if we are doing something unnecessarily expensive in writing to the performance log, like fflush/fsync.
- The performance log should only be enabled if specifically requested by `vine_enable_perf_log`.  Perhaps the parsl-taskvine executor has that on by default.

colinthomas-z80:
@dthain It is not necessarily writing the performance log that is expensive, rather collecting the information about waiting/ready tasks when the queue is large. We do an unnecessarily expensive operation collecting that information which could perhaps be reduced to a single full iteration of the task list, which will still present a problem as the queue size grows. 

This is also pertaining to work queue specifically at the moment. Even if the performance log is not specified we still collect the information before we break out of `log_queue_stats`

btovar:
@colinthomas-z80 An alternative is to increase the interval at which stats are written (e.g in work_queue.c change to 15*ONE_SECOND), and remove the force write in line 7090, (change to log_queue_stats(q, 0), and remove the enclosing if).

dthain:
Ok, so our guiding principles here are:
- Make minimal changes to WQ to meet needs and keep it working.
- Put our main efforts into TaskVine going forward.

So, if the underlying problem is that we are inefficiently iterating over the entire ready queue in order to count task states when writing the performance lag, then I think we should do the following:

1 - Make a few safe and easy modifications to WQ: just traverse the list once, and do it less frequently.
2 - Modify TaskVine to properly track the count of tasks in each state.  Since all task state changes (wisely) go through `change_task_state`, it ought to be straightforward to update some integer counters there, as long as we are disciplined to ensure that there is no other way to change the state.

btovar:
Regarding point 2, I think that's how it was in wq 1, and why we changed it to the current implementation in wq 2.  As code evolved, it became really hard to maintain the state twice (i.e., the real state, and the stats state).  I think point 1 (iterating the list once and doing it less frequently) may go a long way.

dthain:
Let's proceed with doing the easy thing for WQ and see how far it gets us.

I'm willing to try the more complex thing for TV if it necessary to meet our performance goals.

colinthomas-z80:
Ready to merge if there are no further concerns

work_queue/src/work_queue.c
```diff
@@ -4670,6 +4668,24 @@ static int receive_one_task( struct work_queue *q )
 	struct work_queue_worker *w;
 	uint64_t taskid;
 
+	if((t = q->last_waiting_task)) {
+		w = itable_lookup(q->worker_task_map, t->taskid);
+		fetch_output_from_worker(q, w, t->taskid);
+
```
dthain, line 4674: This code duplicates the code below, please make this a common function call.

dthain (changes requested):

dthain:
What's the performance effect of this change?

dthain:
Ok, how much does this help performance?

colinthomas-z80:
@dthain Although it is subjective depending on the hardware, This gives me a peak around 500 tasks/sec with 12k tasks in the queue. At queue sizes up to 60k-70k I will have around 200 tasks/sec. The most I have tested is 110k tasks which resulted in a 160 tasks/sec average. So the drop off is much less severe than before. 

Before we were seeing 300 ish tasks/sec with 10k in the queue, and less than 100 tasks/sec with 20k+ in the queue. 

Once I start working on taskvine I hope to have some more visual depictions of the performance.

work_queue/src/work_queue.c
```diff
@@ -4663,33 +4661,43 @@ static void print_large_tasks_warning(struct work_queue *q)
...
-			// Shutdown worker if appropriate.
-			if ( w->factory_name ) {
-				struct work_queue_factory_info *f;
-				f = hash_table_lookup(q->factory_table, w->factory_name);
-				if ( f && f->connected_workers > f->max_workers &&
-						itable_size(w->current_tasks) < 1 ) {
-					debug(D_WQ, "Final task received from worker %s, shutting down.", w->hostname);
-					shut_down_worker(q, w);
-				}
+	if(!(t = q->last_waiting_task)) {
```
dthain, line 4681: I'm going to be persnickety here since this is deep in the core of WQ.

Simplify the test at the top to make it as clear as possible:
```
t  = q->last_waiting_task;
if(!t) {
...
}
```

work_queue/src/work_queue.c
```diff
@@ -4663,33 +4661,43 @@ static void print_large_tasks_warning(struct work_queue *q)
...
+		itable_firstkey(q->tasks);
```
dthain, line 4682: Use ITABLE_ITERATE.

work_queue/src/work_queue.c
```diff
@@ -4663,33 +4661,43 @@ static void print_large_tasks_warning(struct work_queue *q)
...
+		while( itable_nextkey(q->tasks, &taskid, (void **) &t) ) {
+			if( task_state_is(q, taskid, WORK_QUEUE_TASK_WAITING_RETRIEVAL) ) {
+				break;
```
dthain, line 4685: This doesn't look right.  Doesn't it just end up at `return 0` no matter what?

colinthomas-z80: Indeed

colinthomas-z80:
==== Iteration 1 ====
Will run 10 tasks to target 120 seconds runtime
Submitting tasks / invoking apps
warning: using plain-text when communicating with workers.
warning: use encryption with a key and cert when creating the manager.
All 10 tasks submitted ... waiting for completion
Submission took 0.008 seconds = 1235.690 tasks/second
Runtime: actual 4.432s vs target 120s
Tasks per second: 2.256
==== Iteration 2 ====
Will run 270 tasks to target 120 seconds runtime
Submitting tasks / invoking apps
All 270 tasks submitted ... waiting for completion
Submission took 0.105 seconds = 2569.648 tasks/second
Runtime: actual 2.376s vs target 120s
Tasks per second: 113.648
==== Iteration 3 ====
Will run 13637 tasks to target 120 seconds runtime
Submitting tasks / invoking apps
All 13637 tasks submitted ... waiting for completion
Submission took 5.136 seconds = 2654.964 tasks/second
Runtime: actual 26.247s vs target 120s
Tasks per second: 519.573
==== Iteration 4 ====
Will run 62348 tasks to target 120 seconds runtime
Submitting tasks / invoking apps
All 62348 tasks submitted ... waiting for completion
Submission took 25.965 seconds = 2401.229 tasks/second
Runtime: actual 325.878s vs target 120s
Tasks per second: 191.323
Cleaning up DFK
The end

colinthomas-z80:
Compared to cctools current 

      ==== Iteration 1 ====
      Will run 10 tasks to target 120 seconds runtime
      Submitting tasks / invoking apps
      warning: using plain-text when communicating with workers.
      warning: use encryption with a key and cert when creating the manager.
      All 10 tasks submitted ... waiting for completion
      Submission took 0.008 seconds = 1261.672 tasks/second
      Runtime: actual 4.412s vs target 120s
      Tasks per second: 2.266
      ==== Iteration 2 ====
      Will run 271 tasks to target 120 seconds runtime
      Submitting tasks / invoking apps
      All 271 tasks submitted ... waiting for completion
      Submission took 0.088 seconds = 3077.867 tasks/second
      Runtime: actual 2.372s vs target 120s
      Tasks per second: 114.231
      ==== Iteration 3 ====
      Will run 13707 tasks to target 120 seconds runtime
      Submitting tasks / invoking apps
      All 13707 tasks submitted ... waiting for completion
      Submission took 4.890 seconds = 2802.854 tasks/second
      Runtime: actual 91.802s vs target 120s
      Tasks per second: 149.310
      Cleaning up DFK
      The end

dthain:
Looking good!  Ready to merge now?

colinthomas-z80:
Yes I believe so

dthain (approved):

PR #3387, opened 2023-07-05
vine: mainloop improvements

colinthomas-z80:
WIP

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -1166,17 +1166,14 @@ static int expire_waiting_tasks(struct vine_manager *q)
 {
 	struct vine_task *t;
 	int expired = 0;
-	int count;
-
+	
+	int tasks_considered = 0;
 	double current_time = timestamp_get() / ONE_SECOND;
-	count = task_state_count(q, NULL, VINE_TASK_READY);
-
-	while(count > 0)
+	while( (t = list_rotate(q->ready_list)) )
 	{
-		count--;
-
-		t = list_pop_head(q->ready_list);
-
+		if(tasks_considered > q->attempt_schedule_depth) {
+			return expired;
+		}
 		if(t->resources_requested->end > 0 && t->resources_requested->end <= current_time)
 		{
 			vine_task_set_result(t, VINE_RESULT_MAX_END_TIME);
@@ -1186,11 +1183,9 @@ static int expire_waiting_tasks(struct vine_manager *q)
 			vine_task_set_result(t, VINE_RESULT_MAX_RETRIES);
 			change_task_state(q, t, VINE_TASK_RETRIEVED);
 			expired++;
-		} else {
-			list_push_tail(q->ready_list, t);
-		}
+		} 
+		tasks_considered++;
 	}
-
 	return expired;
 }
 
@@ -2918,20 +2913,16 @@ and mark it as done.
 static int receive_one_task( struct vine_manager *q )
 {
 	struct vine_task *t;
-	uint64_t task_id;
 
-	ITABLE_ITERATE(q->tasks,task_id,t) {
-		if( t->state==VINE_TASK_WAITING_RETRIEVAL ) {
-			struct vine_worker_info *w = t->worker;
-
-			/* Attempt to fetch from this worker. */
-			if(fetch_output_from_worker(q, w, task_id)) {
-				/* If we got one, then we are done. */
-				prune_worker(q, w);
-				return 1;
-			} else {
-				/* But if not, the worker pointer is no longer valid. */
-			}
+	if((t = list_peek_head(q->waiting_list))) {
+		struct vine_worker_info *w = t->worker;
+		/* Attempt to fetch from this worker. */
+		if(fetch_output_from_worker(q, w, task_id)) {
+			/* If we got one, then we are done. */
+			prune_worker(q, w);
+			return 1;
+		} else {
+			/* But if not, the worker pointer is no longer valid. */
 		}
 	}
 
@@ -3261,6 +3252,8 @@ struct vine_manager *vine_ssl_create(int port, const char *key, const char *cert
 	q->next_task_id = 1;
 	
 	q->ready_list = list_create();
+	q->waiting_list = list_create();
+	q->retrieved_list = list_create();
 
 	q->tasks          = itable_create(0);
 	q->libraries         = hash_table_create(0, 0);
@@ -3745,6 +3738,16 @@ static vine_task_state_t change_task_state( struct vine_manager *q, struct vine_
 		case VINE_TASK_READY:
 			vine_task_set_result(t, VINE_RESULT_UNKNOWN);
 			push_task_to_ready_list(q, t);
+			break;
+		case VINE_TASK_WAITING_RETRIEVAL:
+			list_push_head(q->waiting_list, t);
+			break;
+		case VINE_TASK_RETRIEVED:
+			struct vine_task *tmp = NULL;
+			while((tmp = list_rotate(q->waiting_list))){
+				if((t = tmp)) list_pop_tail(q->waiting_list);
+			}
+			list_push_head(q->retrieved_list, t);
 			break;
 		case VINE_TASK_DONE:
 		case VINE_TASK_CANCELED:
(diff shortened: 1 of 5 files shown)


taskvine/src/manager/vine_perf_log.c
```diff
@@ -52,7 +52,7 @@ void vine_perf_log_write_update( struct vine_manager *q, int force )
 	struct vine_stats s;
 
 	timestamp_t now = timestamp_get();
-	if(!force && (now - q->time_last_log_stats < ONE_SECOND)) {
+	if(!force && (now - q->time_last_log_stats < (ONE_SECOND*5))) {
```
dthain, line 55: Please make this adjustable via vine_tune

dthain (changes requested):
Looks good overall, one thing to make tunable.

dthain:
Status on this one?

colinthomas-z80:
@dthain I will add the tune option and then it should be good to go

colinthomas-z80:
do not merge yet, investigating issue

colinthomas-z80:
bug fixed in #3412

taskvine/src/manager/vine_manager.h
```diff
@@ -100,6 +100,8 @@ struct vine_manager {
 
 	struct itable *tasks;           /* Maps task_id -> vine_task of all tasks in any state. */
 	struct list   *ready_list;      /* List of vine_task that are waiting to execute. */
+	struct list   *waiting_list;      /* List of vine_task that are waiting to execute. */
```
dthain, line 103: Fix up comments here.

taskvine/src/manager/vine_manager.c
```diff
@@ -4769,9 +4781,35 @@ void vine_get_stats(struct vine_manager *q, struct vine_stats *s)
 	// s->workers_able computed below.
 
 	//info about tasks
-	s->tasks_waiting      = task_state_count(q, NULL, VINE_TASK_READY);
-	s->tasks_with_results = task_state_count(q, NULL, VINE_TASK_WAITING_RETRIEVAL);
-	s->tasks_on_workers   = task_state_count(q, NULL, VINE_TASK_RUNNING) + s->tasks_with_results;
+	struct vine_task *t;
```
dthain, line 4784: Now that we have separate lists for ready, waiting, retrieved, can you just use `list_size` to return the count of tasks in each state, rather than iterating over everything?

taskvine/src/manager/vine_perf_log.c
```diff
@@ -52,7 +53,7 @@ void vine_perf_log_write_update( struct vine_manager *q, int force )
 	struct vine_stats s;
 
 	timestamp_t now = timestamp_get();
-	if(!force && (now - q->time_last_log_stats < ONE_SECOND)) {
+	if(!force && (now - q->time_last_log_stats < (ONE_SECOND*VINE_PERF_LOG_INTERVAL))) {
```
dthain, line 56: Make this a variable `q->perf_log_interval` modifiable by `vine_tune` instead of a macro.

I am certain that we will run into issues at scale where we will want to increase/decrease this interval periodically.

dthain (changes requested):
Almost there, last few things.

dthain:
@colinthomas-z80 don't forget this one...

taskvine/src/manager/vine_manager.c
```diff
@@ -3746,6 +3743,18 @@ static vine_task_state_t change_task_state( struct vine_manager *q, struct vine_
 			vine_task_set_result(t, VINE_RESULT_UNKNOWN);
 			push_task_to_ready_list(q, t);
 			break;
+		case VINE_TASK_WAITING_RETRIEVAL:
+			list_push_head(q->waiting_list, t);
+			break;
+		case VINE_TASK_RETRIEVED:
+		{
+			struct vine_task *tmp = NULL;
+			while((tmp = list_rotate(q->waiting_list))){
```
dthain, line 3752: Too clever.  How about `list_remove` instead?

taskvine/src/manager/vine_manager.c
```diff
@@ -3746,6 +3743,18 @@ static vine_task_state_t change_task_state( struct vine_manager *q, struct vine_
 			vine_task_set_result(t, VINE_RESULT_UNKNOWN);
```
dthain, line 3743: It seems that all of these cases take on the responsibility of **pushing** to a data structure, but not removing from the old one.  With the exception of `RETRIEVED` which does both.  Is there a reason for the inconsistency?

taskvine/src/manager/vine_manager.c
```diff
@@ -4233,6 +4242,8 @@ static struct vine_task *vine_wait_internal(struct vine_manager *q, int timeout,
 				if(temp->state==VINE_TASK_RETRIEVED) {
 					t = temp;
 				}
+			} else if((t = list_peek_head(q->retrieved_list))) {
```
dthain, line 4245: This doesn't seem right, because this case removes a retrieved task from the list, while the very next case also finds a retrieved task, but **doesn't** remove it from the list.

dthain (changes requested):
On the right track, but we need extra care in ensuring each task is in one list at a time.

taskvine/src/manager/vine_manager.c
colinthomas-z80: It could be worth some cleanup work. But the motivation for how it is now is that we remove a task from the ready list when we commit it to the worker by calling list_pop_tail, which can be completed without having to seek through the task list.

In theory we could call list_pop_tail inside change_task_state since nothing unexpected should happen between the two events, but it would not be as clear in the code to know the task we just scheduled was positioned at the end of the list.

We are also ignoring the VINE_TASK_RUNNING state transition inside change_task_state, since there has been no apparent utility for a separate running_list data structure. So when a task changes to VINE_TASK_WAITING_RETRIEVAL, it is not currently in any of the lists so it does not need to be removed from anywhere.

dthain:
Please stop by to talk, I think it's easier to explain in person...

colinthomas-z80:
@dthain I moved the mechanism of removing/adding to any task state list to change_task_state. 

There are a few separate scenarios where we might change tasks to VINE_TASK_RETRIEVED so that motivates keeping list operations in a single place.

dthain:
Ah, I think you took my advice in the opposite direction.

Your code assumes that a task always marches through the states linearly from READY to DONE.
But that's not the case b/c it can also do these things:
- RUNNING -> CANCELLED
- READY -> RUNNING
- READY -> DONE

So, I don't think it is quite right for `change_task_state` to do the pops as well as the pushes.

I think what you want here is for each action in the code to remove a task from a data structure, consider what to do with it, and then either put it back or call `change_task_state`.

Under that model, `change_task_state`would only push things onto the data structures.

taskvine/src/manager/vine_manager.c
```diff
@@ -1212,6 +1209,7 @@ static int enforce_waiting_fixed_locations(struct vine_manager *q)
 		t = list_pop_head(q->ready_list);
 		if(t->has_fixed_locations && !vine_schedule_check_fixed_location(q, t)) {
 			vine_task_set_result(t, VINE_RESULT_FIXED_LOCATION_MISSING);
+			list_remove(q->ready_list, t);
```
dthain, line 1212: `list_remove` not needed, already removed in line 1209

taskvine/src/manager/vine_manager.c
```diff
@@ -2609,6 +2608,18 @@ static void reap_task_from_worker(struct vine_manager *q, struct vine_worker_inf
...
 
+	switch(t->state)
+	{
+		case VINE_TASK_RUNNING:
+			list_remove(q->running_list, t);
+			break;
+		case VINE_TASK_WAITING_RETRIEVAL:
+			list_remove(q->waiting_retrieval_list, t);
+			break;
+		default:
```
dthain, line 2619: Hmm, it's odd that only two cases are handled in this switch.

If it is possible that the other cases may occur, then handle them.

If it's not possible for other cases to occur here (based on where it was called) then add an assert or some error checking to ensure that we maintain the invariant.

taskvine/src/manager/vine_manager.c
```diff
@@ -3261,6 +3268,9 @@ struct vine_manager *vine_ssl_create(int port, const char *key, const char *cert
 	q->next_task_id = 1;
 	
 	q->ready_list = list_create();
+	q->running_list = list_create();
```
dthain, line 3271: It seems to me that `running_list` doesn't need to be a list, because we don't really care about the order.  Tasks get added to this when they start running, and then get removed individually when they finish.  If you convert this to a `struct table *running_table` then you can avoid the cost of linear `list_remove` operations.

dthain (changes requested):
Almost there

dthain (approved):

PR #3391, opened 2023-07-07
Vine: Asynchronous worker transfers

BarrySlyDelgado:
Tested with 1 worker so far.
- vine_sandbox_ensure checks if all files are ready for a process. Returns 1 if all files are present, Returns 0 on failure and Returns -1 when a transfer is still running.
- vine_cache_wait checks to see if a transfer process has completed.
- vine_cache_ensure forks if a file is not present. If the file is a mini task I think we should first ensure each of its dependencies are present.


dthain (commented):
Yes, this is the right idea.  However, instead of using 1/0/-1, make an enumeration that describes complete/failed/running so that is clearer for the reader to understand what's happening.

You are correct that ensure() called on a mini-task should check all of its inputs.

taskvine/src/worker/vine_cache.c
```diff
@@ -376,6 +380,46 @@ static int do_transfer( struct vine_cache *c, const char *source_url, const char
...
+				// pid is still going
+			}
+			else if(result<0){
+				debug(D_VINE, "wait4 on pid %d returned an error: %s",(int)f->pid,strerror(errno));	
+			}
+			else if(result>0){
+				if(!WIFEXITED(status)){
+					exit_code = WTERMSIG(status);
+					debug(D_VINE, "transfer process (pid %d) exited abnormally with signal %d",f->pid, exit_code);
+					f->complete = -1;
```
dthain, line 406: Another case where a variable looks like a boolean, but isn't.
Make this an enumeration that explains what all the possible values mean.

taskvine/src/worker/vine_process.c
```diff
@@ -55,15 +55,26 @@ extern char * workspace;
 extern struct list *library_list;
 extern struct hash_table *library_ids;
 
-static int create_sandbox_dir( struct vine_process *p )
+static int create_sandbox_dir( struct vine_process *p, int mini_task)
 {
 	p->cache_dir = string_format("%s/cache",workspace);
-  	p->sandbox = string_format("%s/t.%d", workspace,p->task->task_id);
+
```
dthain, line 61: Don't duplicate the `string_format` calls here.
Instead do something like this:

```
const char *type = mini_task ? "m" : "t";
```

Then just use those strings when calling string_format.

taskvine/src/worker/vine_sandbox.c
```diff
@@ -59,6 +59,28 @@ static int ensure_input_file( struct vine_process *p, struct vine_mount *m, stru
 	return result;
 }
 
+/*
+Ensures that each input file is present.
+*/
+
+int vine_sandbox_ensure(struct vine_process *p, struct vine_cache *cache, struct link *manager)
```
dthain, line 66: This function currently returns a boolean, presumably true if all the files are present.
But I think there are three possibilities:
1 - All files are downloaded and present.  The task should run now.
2 - Some files have not been downloaded yet.  The task should not run yet.
3 - One or more files failed to download.  The task should be returned as a failure.

So this function needs a better return value that is handled properly where it is called.

dthain: Or possibly case 3 should allow the task to proceed to execution, and then fail in `vine_sandbox_stagein`.

dthain (changes requested):

taskvine/src/worker/vine_cache.h
```diff
@@ -30,11 +30,19 @@ typedef enum {
 } vine_cache_type_t;
 
 typedef enum {
+	VINE_FILE_UNKNOWN,
```
dthain, line 33: This is looking much better.  However, there is likely to be some confusion of names b/c we already have a number of `VINE_FILE_X` symbols.  Suggest that you call the field `f->status` and the values `VINE_FILE_STATUS_X`

taskvine/src/worker/vine_cache.h
```diff
@@ -30,11 +30,19 @@ typedef enum {
...
 	VINE_FILE_READY,
 	VINE_FILE_FAILED,
 	VINE_FILE_PROCESSING,
 } vine_file_status_type_t;
 
+typedef enum {
```
dthain, line 39: Seems like this is duplicating `vine_file_status_t` you could just re-use that type.

dthain (changes requested):

dthain:
Barry, per our discussion today, only the transfer should happen in the child process.  The manipulation of data structures (particularly on completion) should happen in the parent process.  (Some in `vine_cache_wait`). Also, make the child process a distinct function, so it's very clear where the limited action takes place.

taskvine/src/worker/vine_cache.c
```diff
@@ -35,24 +38,19 @@ struct vine_cache {
 	char *cache_dir;
 };
 
-struct cache_file {
```
dthain, line 38: The definition of `cache_file` should stay inside of `vine_cache` b/c we don't want other modules manipulating it outside of the API.

taskvine/src/worker/vine_cache.c
```diff
@@ -376,99 +374,160 @@ static int do_transfer( struct vine_cache *c, const char *source_url, const char
...
+int vine_cache_wait(struct vine_cache *c, struct link *manager)
+{
+	int status;
+	int exit_code;
+	struct cache_file *f;
+    	char * cachename;
+	HASH_TABLE_ITERATE(c->table, cachename, f){
+		if(f->pid && f->status==VINE_FILE_STATUS_PROCESSING){
+			int result = waitpid(f->pid, &status, WNOHANG);
+			if(result !=0 && f->type==VINE_CACHE_MINI_TASK){
```
dthain, line 390: This seems like a strange place to put this, shouldn't it be part of the `result>0` clause below with the rest of the success case?

taskvine/src/worker/vine_cache.c
```diff
@@ -376,99 +374,160 @@ static int do_transfer( struct vine_cache *c, const char *source_url, const char
...
-			debug(D_VINE,"cache: transferring %s to %s",f->source,cachename);
 			result = do_transfer(c,f->source,cache_path,&error_message);
 			break;
 
 		case VINE_CACHE_MINI_TASK:
-			debug(D_VINE,"cache: creating %s via mini task",cachename);
-			result = do_mini_task(c,f->mini_task,manager,&error_message);
+			result = do_mini_task(c,f,&error_message);
 			break;
 	}
```
dthain, line 528: Nothing is done with the error message here, at the very least send it to the debug stream.

dthain (changes requested):
Almost there!

dthain:
Ok, this is going to need some serious testing.  What's the biggest workflow that you have run with this?

BarrySlyDelgado:
I've ran it with `vine_example_future_matrix` using 10 workers each with 4 cores the workflow had 512 tasks using SGE. temp files should be transferred asynchronously. I also done it with the blast example but only with one worker which includes minitasks and url transfers.

dthain:
Go bigger yet!

dthain (approved):

taskvine/src/worker/vine_cache.c
```diff
@@ -376,99 +387,163 @@ static int do_transfer( struct vine_cache *c, const char *source_url, const char
 	return result;
 }
 
+/* 
+scan through transfer processes and mark them as completed as they finish.
+*/
+
+int vine_cache_wait(struct vine_cache *c, struct link *manager)
```
dthain, line 394: Please factor this function into several smaller functions that:
- iterate over the table
- process one entry
- evaluate the exit status
- extract the outputs

I think that will make the behavior more transparently clear.

taskvine/src/worker/vine_cache.c
```diff
@@ -376,99 +387,163 @@ static int do_transfer( struct vine_cache *c, const char *source_url, const char
...
 	if(!f) {
 		debug(D_VINE,"cache: %s is unknown, perhaps it failed to transfer earlier?",cachename);
-		return 0;
+		return VINE_FILE_STATUS_FAILED;
 	}
 
-	if(f->complete) {
-		/* File is already present in the cache. */
-		return 1;
+	/* File is already present in the cache. */
```
dthain, line 473: As a general rule, when you are examining the states of an enumeration, use a `switch` and address all of the cases.
This is a defensive measure that will make it easier to maintain the code if/when the enumeration adds or removes cases.

dthain (changes requested):

taskvine/src/worker/vine_cache.c
```diff
@@ -376,99 +387,181 @@ static int do_transfer( struct vine_cache *c, const char *source_url, const char
...
+	pid_t pid = fork();
+
+	if(pid == -1) {
+		debug(D_VINE,"failed to fork transfer process");
+		return VINE_FILE_STATUS_FAILED;
+	}
+	if(pid > 0){
+		f->pid = pid;
+		f->status = VINE_FILE_STATUS_PROCESSING;
+		if(f->type==VINE_CACHE_TRANSFER) debug(D_VINE,"cache: transferring %s to %s",f->source,cachename);
```
dthain, line 528: Again, use a switch.

taskvine/src/worker/vine_cache.c
```diff
@@ -376,99 +387,181 @@ static int do_transfer( struct vine_cache *c, const char *source_url, const char
...
+			vine_cache_handle_exit_status(status,cachename,f,manager);
+			vine_cache_check_outputs(f,cachename,c,manager);
+		}
+	}
+}
+
+void vine_cache_check_outputs(struct cache_file *f, char *cachename, struct vine_cache *c, struct link *manager)
+{
+	int64_t nbytes, nfiles;
+	char *cache_path = vine_cache_full_path(c,cachename);
```
dthain, line 420: Memory leak

BarrySlyDelgado:
I think I have to remove the `stagein` check included by PR #3450 since in this PR we are already checking if a minitask has its inputs in vine_cache_ensure and won't proceed with task execution if it doesn't. The `stagein` call will in the `process_execute_and_wait function` may be causing the tests to fail as it would be trying to link into the task sandbox twice?

dthain:
Well, be careful to mind the general principle:

`vine_sandbox_stagein` creates the sandbox and links the inputs to their proper places.

It has to be called somewhere.
 
And wherever it is called, the results must be checked.

Issue #3402, opened 2023-07-12
Vine: FunctionCall + Library Task doesn't work.


tphung3:
Here's a program with a manager executing functions in a serverless manner.
```python
#!/usr/bin/env python
import ndcctools.taskvine as vine
import argparse

def divide(dividend, divisor):
    import math
    return dividend/math.sqrt(divisor)

def main():
    parser = argparse.ArgumentParser(
        prog="vine_example_function_call.py",
        formatter_class=argparse.ArgumentDefaultsHelpFormatter
    )

    parser.add_argument(
        "--disable-peer-transfers",
        action="store_true",
        help="disable transfers among workers.",
        default=False,
    )

    q = vine.Manager(9123)

    args = parser.parse_args()

    if args.disable_peer_transfers:
        q.disable_peer_transfers()
    else:
        q.enable_peer_transfers()

    function_lib = q.create_library_from_functions('test-library', divide)
    q.install_library(function_lib)

    for i in range(1, 5):
        s_task = vine.FunctionCall('test-library', 'divide', 1, i**2)
        q.submit(s_task)

    total_sum = 0
    while not q.empty():
        t = q.wait(5)
        if t:
            x = t.output
            total_sum += x
            print(total_sum)


if __name__ == '__main__':
    main()
```

Running this program with `vine_worker localhost 9123` (local worker) results in two errors (in temporal order):
- `vine_example_function_call.py`: (above code)
```
cctools_list_seek: list.c:201[DEVELOPMENT]: Assertion 'cur' failed.
Aborted (core dumped)
```
- `vine_worker`: 
```
vine_worker: creating workspace /tmp/worker-1000-45242
vine_worker: using 4 cores, 11591 MB memory, 5443 MB disk, 0 gpus
connected to manager localhost:9123 via local address 127.0.0.1:37928
kill_all_tasks: vine_worker.c:907[DEVELOPMENT]: Assertion 'itable_size(procs_running)==0' failed.
Aborted (core dumped)
```
It seems that the worker code doesn't kill all tasks properly when exiting.

dthain:
Curious what the `-d all` output is on the worker.  Is it exiting of its own accord?

tphung3:
Nice catch! I assume worker finishes all work and gets into trouble but not quite.
Worker fails to read from the manager.
```
[...]

PR #3409, opened 2023-07-18
TaskVine: Fix serverless documentation

tphung3:
The documentation shows the API of FunctionCall as `FunctionCall(lib_name,func_name, *args)` while the true API in python binding is `FunctionCall(func_name, lib_name, *args)`, partially causing issue https://github.com/cooperative-computing-lab/cctools/issues/3402. This PR fixes the docs and is RTM.

### DIFF
--- a/doc/manuals/taskvine/index.md
+++ b/doc/manuals/taskvine/index.md
@@ -1474,7 +1474,7 @@ that invoke the library and functions by name:
 
 === "Python"
     ```python
-    t = vine.FunctionCall("my_mul","my_library",20,30);
+    t = vine.FunctionCall("my_library","my_mul",20,30);
     t.set_cores(1)
     t.set_memory(100)
     t.set_disk(100)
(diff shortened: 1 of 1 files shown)


btovar:
@tphung3 I'd suggest changing the API to `FunctionCall(lib_name,func_name, *args)` as it feels more natural.

tphung3:
If all is well then the FunctionCall API will be `FunctionCall(lib_name: str, fn_name: str, *args, **kwargs )`.

tphung3:
This PR is RTM. PR https://github.com/cooperative-computing-lab/cctools/pull/3410 will follow this API.

dthain:
I agree with the API fix, but this PR only seems to contain a single spelling fix.  Is something missing?

tphung3:
No nothing is missing. It used to be that the API is wrong but the doc is right. Midway in this PR I changed the doc to be wrong so it's consistent. Now the API is right on the other PR so there's no significant change to the doc. (too many words haha)

tphung3:
This PR is RTM.

dthain (approved):

PR #3410, opened 2023-07-18
TaskVine: Fix serverless functionality

tphung3:
Changing a task's state to READY inadvertently pushes that task to the list of ready tasks, which is not what we want if that task is a library task. This fixes the issue https://github.com/cooperative-computing-lab/cctools/issues/3402 and adds test code.
PR https://github.com/cooperative-computing-lab/cctools/pull/3409 should also be coordinated with this PR for FunctionCall API.

The formatting of results from a FunctionCall is a bit weird, as it is in the json format (see `vine_example_function_call.py:49`), which implies that results are json-serializable. Maybe cloudpickle is the better alternative here.


btovar (approved):

dthain:
Thanks for disentangling this.

While I am content to let the function results be json-encoded for now, I'm not thrilled with the idea that the user is responsible for decoding from json.  It seems to me that taskvine should do it for the user.  (And that frees us up to do some other encoding later.)

What do you think?

tphung3:
This PR is RTM.

PR #3416, opened 2023-07-25
dttools: fix too many open files

tphung3:
`path_disk_size_info` used to open directories in advance before they are read. This behavior causes `too many open files` error with a big enough directory.
The change fixes this behavior by only opening directories to read when needed.
This PR is RTM if all is well.

### DIFF
--- a/dttools/src/path_disk_size_info.c
+++ b/dttools/src/path_disk_size_info.c
@@ -81,6 +81,7 @@ int path_disk_size_info_get_r(const char *path, int64_t max_secs, struct path_di
 				debug(D_DEBUG, "error opening directory '%s': (errno) %s\n", tail_name, strerror(errno));
 				result = -1;
 			}
+			free(tail_name);
 			continue;
 		}
 
@@ -93,6 +94,7 @@ int path_disk_size_info_get_r(const char *path, int64_t max_secs, struct path_di
 				result = -1;
 			}
 			closedir(tail_dir);
+			free(tail_name);
 			continue;
 		}
 		s->size_so_far += file_info.st_size;
--- a/dttools/src/path_disk_size_info.c
+++ b/dttools/src/path_disk_size_info.c
@@ -105,11 +105,11 @@ int path_disk_size_info_get_r(const char *path, int64_t max_secs, struct path_di
 			if( strcmp(".", entry->d_name) == 0 || strcmp("..", entry->d_name) == 0)
 				continue;
 
-			char composed_path[PATH_MAX];
+			char* composed_path;
 			if(entry->d_name[0] == '/') {	// absolute path
-				strncpy(composed_path, entry->d_name, PATH_MAX);
+				composed_path = string_format("%s", entry->d_name);
 			} else {						// relative path
-				snprintf(composed_path, PATH_MAX, "%s/%s", tail_name, entry->d_name);
+				composed_path = string_format("%s/%s", tail_name, entry->d_name);
 			}
 
 			if(lstat(composed_path, &file_info) < 0) {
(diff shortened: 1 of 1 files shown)


dttools/src/path_disk_size_info.c
```diff
@@ -165,15 +172,10 @@ void path_disk_size_info_delete_state(struct path_disk_size_info *state) {
...
-		struct DIR_with_name *tail;
-		while((tail = list_pop_tail(state->current_dirs))) {
-			if(tail->dir)
-				closedir(tail->dir);
-
-			if(tail->name)
-				free(tail->name);
-
-			free(tail);
+		char* tail_name;
```
dthain, line 175: `list_clear(list, free)`

dthain (changes requested):

dttools/src/path_disk_size_info.c
```diff
@@ -47,102 +42,114 @@ int path_disk_size_info_get(const char *path, int64_t *measured_size, int64_t *n
...
-	while((tail = list_peek_tail(s->current_dirs))) {
-		struct dirent *entry;
-		struct stat   file_info;
-		while((entry = readdir(tail->dir))) {
+		/* Read entries of opened directory */
+		while((entry = readdir(tail_dir))) {
 			if( strcmp(".", entry->d_name) == 0 || strcmp("..", entry->d_name) == 0)
 				continue;
 
 			char composed_path[PATH_MAX];
```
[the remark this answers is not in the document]
tphung3: Can you explain how string_format can help that code?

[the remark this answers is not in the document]
dthain: `char * path = string_format( ... )`

Then path is not limited to any particular length.  Also reduces stack consumption because you aren't allocating the maximum path length on every recursive function call.

dthain:
And don't forget to free `composed_path` as needed.

dttools/src/path_disk_size_info.c
```diff
@@ -47,102 +42,117 @@ int path_disk_size_info_get(const char *path, int64_t *measured_size, int64_t *n
...
+	 * files that are directories. Directory entries are added to a list to be 
+	 * counted later.
+	 */
+	char* tail_name;
+	struct dirent *entry;
+	struct stat file_info;
+	DIR* tail_dir;
+	int timeout = 0;
+	while((tail_name = list_pop_tail(s->current_dirs))) {
+		tail_dir = opendir(tail_name);
```
btovar, line 75: I'm not sure about this one. The purpose of keeping the open dirs with DIR_with_name is so that the measurement can be interrupted on a timeout and then resumed given the state. If the opendir is called here, then either the previous progress is lost, or we can't interrupt the call.

I'd suggest a minimal change to the previous version.  Keep the DIR_with_name struct, and only open the dir when popping the struct from the list.

btovar (changes requested):

dttools/src/path_disk_size_info.c
dthain: Ah, that is a better approach, thanks!

dttools/src/path_disk_size_info.c
```diff
@@ -130,10 +143,10 @@ int path_disk_size_info_get_r(const char *path, int64_t max_secs, struct path_di
 				}
 			}
 		}
-
 		/* we are done reading a complete directory, and we go to the next in the queue */
 		tail = list_pop_tail(s->current_dirs);
-		closedir(tail->dir);
+		if (tail->dir)
+			closedir(tail->dir);
```
btovar, line 149: add braces here?

btovar (approved):

dthain:
RTM?

tphung3:
Yes

Issue #3417, opened 2023-07-26
Vine: manager hangs when running `vine_example_function_call.py`


tphung3:
The manager hangs when running `taskvine/src/examples/vine_example_function_call.py` with a local worker. From a quick look at debug logs, the manager is not sending function call tasks over to the worker.

tphung3:
Addressed by PR https://github.com/cooperative-computing-lab/cctools/pull/3418.

PR #3418, opened 2023-07-26
TaskVine: fix FunctionCall API, update example, and add test

tphung3:
This PR fixes issue #3417, updates example code, and adds test code.
Do all files with prefix `TR_` get tested? I'm not sure. The test code is very similar to `python_task` though.

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
@@ -937,11 +937,11 @@ class FunctionCall(Task):
     # Create a new FunctionCall specification.
     #
     # @param self       Reference to the current FunctionCall object.
-    # @param fn         The name of the function to be executed on the coprocess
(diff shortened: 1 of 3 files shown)


dthain (approved):
Thanks - Having a good test goes a long way towards preventing these sorts of problems.

dthain:
Possibly `import ndcctools.poncho` instead of `import poncho` inside of `manager.py` ?

tphung3:
I think the python version inside the container is 3.6, while poncho uses `ast.unparse` which is a feature for python 3.9. I'm tracking down where python is installed.

tphung3:
> Possibly `import ndcctools.poncho` instead of `import poncho` inside of `manager.py` ?

Your are right this will work for tests that don't use the serverless feature.

tphung3:
This PR is RTM.

dthain (approved):

PR #3427, opened 2023-07-31
TaskVine + Poncho: Add env option serverless

tphung3:
This PR brings a couple of changes as follows:
- Remove some print statements as it pollutes stdout when using with other modules (e.g., Parsl).
- Decouple the creation of library code `library_code.py` which contains the core functionality of `LibraryTask` and `FunctionCall` from the environment creation/registration along with the task. Now `LibraryTask` can be run without creating/registering environments, which significantly reduces latency when environment is not needed or prepared by the user/application.
- Add `init_command` to `LibraryTask` so users can run `LibraryTask` on remote nodes by prepending the command with custom setup code (e.g., `conda run --name conda_env python ./library_code.py`). This frees `LibraryTask` from its environment and serves as the base case by relying on the shared filesystem.


tphung3:
This PR also corrects the installation path of `poncho` so `import ndcctools.poncho` won't cause an import error.

BarrySlyDelgado (approved):

tphung3:
This PR is RTM.

Issue #3434, opened 2023-08-03
Vine: Serverless Invocation Crash


benclifford:
I see most-of-the-time (but not always) segfaults in taskvine+parsl. I commented on this initially in the Parsl github, here: https://github.com/Parsl/parsl/pull/2849#issuecomment-1662364188

The backtrace there shows a suspicious looking pointer (an ASLR prefix but the low bits all zero).

With assistance from others, I investigated these assignments in the hash table code and added these asserts:

```
benc@parsl-dev-3-11-1240:~/parsl/src/cctools$ git diff
diff --git a/dttools/src/hash_table.c b/dttools/src/hash_table.c
index f5e3fc8bb..7c78482e2 100644
--- a/dttools/src/hash_table.c
+++ b/dttools/src/hash_table.c
@@ -7,6 +7,7 @@ See the file COPYING for details.
 
 #include "hash_table.h"
 
+#include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -233,7 +234,9 @@ int hash_table_nextkey(struct hash_table *h, char **key, void **value)
 {
        if(h->ientry) {
                *key = h->ientry->key;
+               assert(*key == h->ientry->key);
                *value = h->ientry->value;
+               assert(*key == h->ientry->key);
 
                h->ientry = h->ientry->next;
                if(!h->ientry) {
```

here's the log I get. The assert that fails is the *second* assert that I added.

```
vine_worker -d all 127.0.0.1 9000 
2023/08/03 15:03:40.04 vine_worker[21722] debug: vine_worker version 8.0.0 DEVELOPMENT (released 2023-08-03 14:58:40 +0000)
2023/08/03 15:03:40.04 vine_worker[21722] debug: Built by benc@parsl-dev-3-11-1240 on 2023-08-03 14:58:40 +0000
2023/08/03 15:03:40.04 vine_worker[21722] debug: System: Linux parsl-dev-3-11-1240 5.19.0-46-generic #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jun 21 15:35:31 UTC 2 x86_64 GNU/Linux
2023/08/03 15:03:40.04 vine_worker[21722] debug: Configuration: --prefix=/home/benc/parsl/src/cctools/I
vine_worker: creating workspace /tmp/worker-1000-21722
[...]

PR #3435, opened 2023-08-04
Vine:  Fix worker crash when using serverless.

dthain:
Addresses #3434.  The are some other infelicities in the serverless code, but keeping the fix focused for now.

### DIFF
--- a/taskvine/src/worker/vine_process.c
+++ b/taskvine/src/worker/vine_process.c
@@ -52,8 +52,9 @@ Create temporary directories inside as well.
 
 extern char * workspace;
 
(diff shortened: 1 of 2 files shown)


tphung3 (approved):

tphung3:
I think I just saw this error when doing a large scale run. The manager exits fine but the worker fails to read from the manager when it exits.

dthain:
Are you saying that you reproduced the prior problem, or that you found a problem with this PR?

tphung3:
This PR is good, I reproduced the problem. Sorry for the confusion.

Issue #3433, opened 2023-08-02
Build: Test the Documentation


dthain:
CCTools overall has an expansive set of tests and examples that cover a lot of ground:
1 - The integral test suite run by `make test` tests internal functionality for developers and on every checkin.
2 - The external test suite `cctools-packaging-test` tests the end-to-end execution of cctools and related packages, starting from a scratch conda environment.
3 - The source code contains a number of example programs (e.g. `taskvine/src/examples`) that are built but not tested automatically.
4 - The documentation contains a number of install instructions and example programs of interest.

However, a consistent challenge is ensuring that our documentation stays working and correct in the face of change.  It is all too easy for a minor defect in the documentation to result in an end-user getting stuck.  Or, a change to the code itself may render the (previously correct) documentation no longer working.  Or, the tests, examples, and documentation may not even line up!

What we need is a strategy to **test the documentation itself**.  Begin by adding a new set of tests to `cctools-packaging-test` that executes the published documentation for installing the software, and running some basic tests.  For this to work, the documentation must be structured so that the embedded examples are separate files, arranged and edited so that they can in fact be executed.

Once this is working, then we can gradually convert cases 2 and 3 into documented examples, and the tests will consist of nothing but downloading and executing the documentation.

dthain:
Thanks @thieber22 !

PR #3444, opened 2023-08-08
Fix imports in doc example

BarrySlyDelgado:
examples used deprecated import statements "import taskvine as vine" instead of "import ndcctools.taskvine as vine"


dthain:
This is an example of why we should do #3433!

PR #3447, opened 2023-08-08
Vine: fix #3445

### LINKED ISSUE #3445: Vine: Delete After Failed Transfer
I think we have a (possibly rare) case where a failed transfer may result in an incomplete file in the cache.

`vine_cache.c:do_transfer` does at least two kinds of transfers.
1 - When doing a curl transfer, it downloads to a (temporay) .transfer file and then renames it into place.  This is precisely to ensure that failed transfers don't become part of the cache.
2 - But when doing a worker-to-worker transfer, it uses vine_transfer_get_any (the same code as manager-to-worker transfer) and that routine writes files directly into the cache destination.  So if a transfer should fail or timeout, the partial file will still be in the target location, and could get picked up by the cache scan when the worker restarts.

So: `vine_transfer_get_any` (and get_dir and get_file) should explicitly remove the (possibly incomplete) downloaded file/dir when vine_transfer_get_any_internal fails.

tphung3:
This PR fix #3445 by fsyncing and removing invalid files appropriately.
Also transfer timeout is bumped up to 300 seconds, which should serve fine for 5 GBs per transfer at ND.

### DIFF
--- a/taskvine/src/worker/vine_cache.c
+++ b/taskvine/src/worker/vine_cache.c
@@ -323,9 +323,9 @@ static int do_worker_transfer( struct vine_cache *c, const char *source_url, con
 		}
 	}
 	
-	/* XXX A fixed timeout of 300 certainly can't be right! */
+	/* XXX A fixed timeout of 900 certainly can't be right! */
 	
-	if(!vine_transfer_get_any(worker_link, c, path, time(0) + 300))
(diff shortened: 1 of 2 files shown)


dthain (changes requested):

dthain:
FYI I'm accepting this after making a few formatting cleanups and removing the fsync, since that is likely to have a big performance impact...

Issue #3446, opened 2023-08-08
Vine: worker runs task then cache updates the task's environment file.


tphung3:
Running 1000 tasks each attached with a poncho environment (~1GB) results in a situation where around 3/4 of tasks or even all tasks fail because the poncho environment is seemingly truncated.

Bumping the timeout in `do_worker_transfer` from 120 to 300 seconds reduces the number of failed tasks to < 10, and the machine that has failed tasks run the next tasks successfully.

Examining the transaction log shows that the first cache-update message of the environment file happens 2 seconds after a task is run. The transaction log may not capture causality, but it is probably a bug.

tphung3:
Update: The run above is 1000 tasks and 200 workers. I just ran another experiment with 1000 tasks and 100 workers and all 1000 tasks fail.

tphung3:
The latter experiment fails twice with the same 1000 failed tasks, so it's probably not random,

dthain:
Well, it's easier to debug, then!

Starting point would be to collect the manager log and at least one worker log, right?

dthain:
Is this fixed by #3447 ?

tphung3:
Fixed in #3450.

PR #3449, opened 2023-08-09
Vine: make sure file is fsync'ed at worker

### LINKED ISSUE #3448: Vine: vine_transfer needs fsync to work properly
Whenever a file is transferred between workers, the receiving worker must call `fsync` on the file descriptor to ensure that the file is dumped to disk.

Evidence: running a workflow with 1024 tasks, each with a poncho environment tarball ~1GB:
- no `fsync`: `success: 78, failure: 946`
- `fsync`: `success: 1024, failure: 0`.

tphung3:
This PR fixes #3448 and #3446.

### DIFF
--- a/taskvine/src/worker/vine_transfer.c
+++ b/taskvine/src/worker/vine_transfer.c
@@ -226,6 +226,11 @@ static int vine_transfer_get_file_internal( struct link *lnk, const char *filena
 		return 0;
 	}
 	
+    if (fsync(fd)<0) {
+        debug(D_VINE, "Failed to fsync file - %s (%s)\n", filename, strerror(errno));
+        return 0;
+    }
+
 	if(close(fd)<0) {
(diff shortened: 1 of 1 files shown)


dthain:
Still have to close the file on the way out!

tphung3:
This PR is RTM.

btovar:
Shouldn't the fsync come after the fclose?

Just a shot in the dark here, but are we sure is really the fsync that's working, or is fsync causing a delay for long enough that some race condition is minimized?  If the workers are not running on top of a shared filesystem, this fsync is highly suspicious.

dthain:
The fsync operates on an open fd, so it's in the right place.

But I agree that this behavior doesn't yet make sense.  I like the race condition hypothesis!

btovar:
Oh right, needs the open fd.

Issue #3348, opened 2023-05-15
How to allocate more then one Node in makeflow in an HPC cluster 


Dcn303:
Hi everyone
i am working on HPC cluster
i have 2 node with 4 core each .
How will i write a makeflow script to allocate the 4 core of both the node 1 and node 2
.I have a script simple.sh as shown below
`#! /bin/bash `
`N=8 `
`(mpirun -np $N simpleFoam -parallel) 2>&1 | tee log.simpleFoam `
`touch c`
which uses 8 mpi processes to execute
i also have to submit to slurm
i have written the makeflow script(run.makeflow) as show below
`BATCH_OPTIONS= --nodes=2 --ntasks-per-node=4`
`.MAKEFLOW MPI_PROCESS 8`
`.MAKEFLOW CORES 8`
`c : b a `
 `srun $(BATCH_OPTIONS)  ./simple.sh`
`b: a`
 `srun ./decompose.sh`
`a: `
 `srun ./blockMesh.sh`
i have submitted the makeflow to slurm as
`makeflow -T slurm run.makeflow `
the rules
`b: a`
 ` srun ./decompose.sh`
`a: `
  `srun ./blockMesh.sh` works fine and it have also allocated the two node1 and node2
but the `./simple.sh` is not executing
can anyone kindly help me on this regard

btovar:
@Dcn303 One possible issue is that the .MAKEFLOW directives refer to the resources allocated for single rules in the makeflow file, and not to the total of resources that you have available. Since there are no nodes with 8 cores, then it is possible that makeflow is not dispatching them. Thus, one possible solution is to use:

.MAKEFLOW_MPI_PROCESS 4
.MAKEFLOW_CORES 4

Note that makeflow is not designed to run MPI executables that spawn multiple nodes. It only knows how to run one executable per node.

Also, it would be helpful if you can share the output of  `makeflow -dall -T slurm run.makeflow`

Dcn303:
my new run.makeflow as shown below 
`BATCH_OPTIONS= -N 2 --ntasks-per-node=4`
`c: a b`
          `sbatch $(BATCH_OPTIONS) simple.sh  `
`b: a`
          `srun  ./decomposePar.sh`
`a:`
          `srun ./blockMesh.sh`
i am not sure but when i was running this i saw it running on node 1 as well as node 2 and it seems its running fiine .
my simple.sh is same as the above , let me try makeflow -dall -T slurm run.makeflow and let you know

[...]

PR #3450, opened 2023-08-10
Vine: Properly check if minitask has its input files staged in.

tphung3:
Mini tasks or any tasks that are run with `execute_and_wait` have the problem of executing regardless of the status of staged-in files. This PR fixes this and walks back on the `fsync` as it is not relevant. The previous situation where `fsync` apparently fixes the issue in #3348 is random and depends entirely on whether a worker transferring input files to a mini task on another worker fails or not, (e.g., sometimes workers are not evicted and sometimes they are.)

This PR is RTM.

### DIFF
--- a/taskvine/src/worker/vine_process.c
+++ b/taskvine/src/worker/vine_process.c
@@ -234,7 +234,12 @@ int vine_process_execute_and_wait( struct vine_task *task, struct vine_cache *ca
 {
 	struct vine_process *p = vine_process_create(task);
 
-	vine_sandbox_stagein(p,cache,manager);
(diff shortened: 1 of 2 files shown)


tphung3:
Should also fix #https://github.com/cooperative-computing-lab/cctools/issues/3446.

dthain:
Code makes sense to me, but it looks like you have a test failure.
Is it possible the test was not correctly constructed to capture "input missing" ?

dthain:
`vine_test.py` fails because `file:///dummy.tar.gz` is an invalid path.  The test was already broken, you just added the code to find it! Should be something like `file://{working-dir}/dummy.tar.gz` instead.

PR #3463, opened 2023-08-17
vine: avoid linear scheduling operations

colinthomas-z80:
removed remaining usage of task_state_any.

Will discuss if there is anything to do with task_state_any_with_tag


dthain:
`task_state_any_with_tag` is inherently inefficient, yes, but the user only pays the price if waiting on a tag.

Note the macros to measure accumulated time at line 4432, so there is some archaeological evidence that those searches were paying a price that had to be measured!

Looking forward to seeing how performance compares to the old ways...

PR #3469, opened 2023-08-17
Vine: Fix Handling of Failed Library Tasks

dthain:
- A failed library task goes into the RETRIEVED state, and then eventually into the DONE state in one place in vine_wait.
- A created library task goes into the master task table with an extra reference.  It is properly deleted twice, once in change_task_state when it is removed, and again in vine_wait when it goes to DONE.


dthain:
build-conda is failing due to very long condo-install times.  Could be a transient problem?

btovar:
Given that it was terminated in the middle of a test, maybe it is not a transient?

dthain:
I think it was the overall 30 minute timeout firing.

PR #3473, opened 2023-08-21
Vine: add default resources to library tasks

tphung3:
A LibraryTask currently doesn't have resources specified, so when a libtask is sent to a worker, it takes over all resources of that worker (by default), thus no real tasks (function calls) are sent to the worker. This PR made the change to specify a libtask some arbitrary (and hopefully sensible) resource values to prevent the above bug.

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
@@ -1011,5 +1011,32 @@ class LibraryTask(Task):
         Task.__init__(self, fn)
         self.provides_library(name)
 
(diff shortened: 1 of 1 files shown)


dthain:
Aha!  So it worked by accident before because the library was sent before the resource message?

dthain:
Let's discuss whether to set default resource consumption on the libraries or on the functions.  I don't think we should do both.

dthain:
Superseded by #3477

tphung3:
This PR is resolved by https://github.com/cooperative-computing-lab/cctools/pull/3477 thus closed without merge.

PR #3477, opened 2023-08-25
Vine: Partial fix of resources for serverless feature.

tphung3:
This PR made the following changes:
- Set default resources for library tasks so these tasks will have resources accounted for. The values are arbitrary and need further consideration.
- Remove max_proportion adjustment in `vine_manager_choose_resources_for_task`. This code assumes all tasks have the same resource box and thus tries to round resources. This is not true when library task uses few resources (e.g., 1 core) and function call tasks use arbitrary resources.
- Add check for library tasks when sending them to workers. This check is needed in case worker doesn't have available resources.
- Log library sending operation appropriately.
- Make `check_worker_against_task` available to be used in `vine_manager.c`.
- Fix resources for worker and tasks in serverless test script.

Warning: this setup won't work if function calls don't specify the right amount of resources. Also current implementation only allows 1 function call per library. 

Todo: clean design of resources for serverless.

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
@@ -1011,5 +1011,14 @@ class LibraryTask(Task):
         Task.__init__(self, fn)
         self.provides_library(name)
 
+        # Set some default resources to a library task so it doesn't consume whole worker.
+        # Values below are arbitrary and should be changed as appropriate.
+        self._default_cores = 1
+        self._default_memory = 1000
+        self._default_disk = 1000
+
+        self.set_cores(self._default_cores)
+        self.set_memory(self._default_memory)
+        self.set_disk(self._default_disk)
 
 # vim: set sts=4 sw=4 ts=4 expandtab ft=python:
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
@@ -1011,14 +1011,4 @@ class LibraryTask(Task):
         Task.__init__(self, fn)
         self.provides_library(name)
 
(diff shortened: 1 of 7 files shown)


taskvine/src/manager/vine_manager.c
```diff
@@ -2506,21 +2513,15 @@ struct rmsummary *vine_manager_choose_resources_for_task(
...
+		/* If max_proportion > 1, then the task does not fit the worker for the
+		 * specified resources. For the unspecified resources we use the whole
+		 * worker as not to trigger a warning when checking for tasks that can't
+		 * run on any available worker. */
 		if (max_proportion > 1) {
 			use_whole_worker = 1;
 		} else if (max_proportion > 0) {
 			use_whole_worker = 0;
 
-			// adjust max_proportion so that an integer number of tasks fit the
```
dthain, line 2518: Hmm, I think this is an important step for normal tasks.
Maybe make this conditional on `!t->needs_library` so that it doesn't apply to functions?

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
```diff
@@ -1011,5 +1011,14 @@ def __init__(self, fn, name):
         Task.__init__(self, fn)
         self.provides_library(name)
 
+        # Set some default resources to a library task so it doesn't consume whole worker.
```
dthain, line 1014: Let's not set default resources in `task.py`, I think it's going to be very easy to forget that they are here.
You are already setting them in the application code anyway.

dthain (changes requested):
Glad you got it working!
Two little things to fix.

dthain:
ping?

tphung3:
This PR is RTM.

dthain (approved):

PR #3478, opened 2023-08-28
Vine: Serverless: Match on library name, not feature.

### LINKED ISSUE #3441: Vine: Serverless Feature Declaration
(Note this is relative to PR #3436 that hasn't been merged yet.)

As written, the serverless system uses features in an unusual way:
- When the manager sends a LibraryTask to a worker, it explicitly adds that feature to the `vine_worker_info`
- When the worker starts a LibraryTask and reads back the initialization string, it **also** declares the feature itself, and sends an update back to the manager.
- When the worker ends a LibraryTask, it removes the feature.
- When the manager removes a LibraryTask, it manually removes the feature from the `vine_worker_info`.

This is all very inconsistent and needs to be fixed up, because:
- Both the manager and the worker are adding and removing features, which is redundant.
- There isn't an obvious way for a worker to indicate that it has multiple libraries.
- There isn't an obvious way in the manager-worker protocol to **remove** a feature at runtime.

dthain:
Fixes #3441.  A worker "feature" describes the whole worker, and doesn't have a clear way to remove once activated.  This PR no longer uses "features" to indicate that a serverless library is present.  Instead, the manager directly looks for the presence of a library task where l->provides_library matches t->needs_library.

Note that this PR does not (yet) fix the problem of controlling the number of tasks assigned to a library.  That's a separate problem.

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -4202,7 +4202,7 @@ static int vine_manager_send_library_to_worker(struct vine_manager *q, struct vi
 	return 1;
 }
 
-static struct vine_task *vine_manager_find_library_on_worker(
+struct vine_task *vine_manager_find_library_on_worker(
 		struct vine_manager *q, struct vine_worker_info *w, const char *library_name)
 {
 	uint64_t task_id;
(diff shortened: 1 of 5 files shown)


tphung3 (approved):
Looks good. Can you merge this and my PRs so I can try both?

PR #3479, opened 2023-08-29
Vine: Serverless: Track Assignment of Functions to Libraries

dthain:
Each library task now has a certain number of function slots (by default one) assigned by`vine_task_set_function_slots`

Each function task is now assigned a specific library task during scheduling, up to the limit of the number of slots.

The manager will now only dispatch as many function calls as actually fit to the available library slots.

The worker main loop will only dispatch function calls up to the slot limit.

**BUT NOTE** that the library invocation protocol in `vine_process` does not yet admit multiple invocations per worker.


dthain:
And note also here that a function task still requires a resource assignment.
That hasn't been taken away (yet).

PR #3480, opened 2023-08-31
Vine: Library Task communicates with Vine Worker using pipes instead of stdin/stdout.

tphung3:
Currently Library Task is using stdin/stdout to communicate with worker processes to get data and return results. If a function call prints something out to stdout however, then the communication is corrupted. This fix changes Library Task to use proper pipes to talk with vine worker.
There's also some minor code refactoring and comment editing/adding.


dthain (approved):
Thanks for noticing that the WQ coprocess code needs to stay the same, I hadn't thought of that issue in the Taskvine evolution.

tphung3:
This PR is RTM.

PR #3485, opened 2023-09-11
vine, wq: ramp down mode

btovar:
Adds `m.tune("ramp-down-heuristic", 1)`.  When this is enabled and there are more workers than ready tasks, then tasks are allocated all the free resources of a worker.  When looking for workers by files or by time, ties are broken by worst fit.  This mode avoids unnecessary resource exhaustion when there are more workers than tasks. If the monitoring watchdog is not enabled, then this heuristic has no effect. By default it is disabled (set to 0).

It also adds `m.tune("hungry-minimum-factor", 2)`. A manager is hungry if the number of waiting tasks is less than then number of workers times this factor.  By default is set to 2 to avoid ramp-down mode to trigger when the workflow is not really terminating.

With this mode enabled I was able to run the processing+accumulation part of topEFT using the old wq executor in 40 minutes, as the expensive final accumulations are never retried. Together with other advances in taskvine, 30 min processing time should be within reach.


btovar:
Yes, if activated, then are less tasks than workers, then tasks get all free resources of a worker.  Ideally whatever is managing vine would activate this mode when it knows the workflow is about to end. (E.g. in topcoffea, when all the processing tasks are done and we only have accumulation tasks.)

It can be seen as if the queue is starving (e.g. more than hungry), then be aggressive allocating resources for the tasks.

dthain (approved):

PR #3486, opened 2023-09-11
Vine: Function Call Tasks Consume No Resources

dthain:


dthain:
This is a simple change, but want to make sure that I didn't miss some unexpected effect of setting resource values to zero.  (Typically -1 means un-set and zero means zero.)

tphung3 (approved):
Looks right!

PR #3487, opened 2023-09-11
changed conda to use libmamba as its solver

Jbrocket:


dthain:
Good, that's a nice improvement in build performance.
While you are at it please add a note to the installation instructions in `doc/manual/install` suggestion this as an optional speedup.

dthain (approved):

PR #3488, opened 2023-09-13
return file buffer contents as bytes object

btovar:
This seems to work:

```python
import ndcctools.taskvine as vine
import cloudpickle

m = vine.DaskVine(name="test", port=0)

b = m.declare_buffer(cloudpickle.dumps({"a": 0}))
print(type(b.contents()))

c=cloudpickle.loads(b.contents())
print(c)

s="hello pirate ♆"
b2=m.declare_buffer(bytes(s, "utf-8"))
print(bytes.decode(b2.contents(), "utf-8"))
```

```shell
# output
<class 'bytes'>
{'a': 0}
hello pirate ♆
```

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/file.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/file.py
@@ -40,7 +40,7 @@ class File(object):
     #
     # @param self       A file object.
     def contents(self):
-        return cvine.vine_file_contents(self._file)
+        return cvine.vine_file_contents_as_bytes(self._file)
 
     ##
     # Return the size of a file object, in bytes.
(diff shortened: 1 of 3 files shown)


dthain:
This makes sense to me. @tphung what do you think?

tphung3:
Running above code gets me this:
```
<class 'str'>
  File "/scratch365/tphung/serverless_test/sandbox/tmp4.py", line 9, in <module>
    c=cloudpickle.loads(b.contents())
TypeError: a bytes-like object is required, not 'str'
```
I'll rebuild with the master branch code and post results.

btovar:
Indeed, that's the expected result without the current pr.

tphung3:
It's much easier than I thought. Code makes sense. How did you come up with the solution?

btovar:
I remembered that the python C api had some functions to return bytes objects from when I was trying to return arrays for the resource monitor. Once I found a function that may work, I had to look up how to add custom bindings in swig. All of these steps are standard, so luckily it worked at the first try.

dthain (approved):

Issue #3439, opened 2023-08-07
Vine: Library Process Protocol


dthain:
(Note this is relative to PR #3436 that hasn't been merged yet.)

The protocol between the worker and the serverless library has several infelicities that require a coordinated fix between the library code and `vine_process_invoke_function`

- [ ] The output of each function invocation is written as a single **line** with a maximum of `VINE_LINE_MAX`.  It should instead be written as an integer length followed by the actual data.  (Similar to that of the function input.)
- [ ] The function invocation has a default timeout of 5 minutes.  The should be configurable using an existing property of a standard task, like the maximum run time.
- [ ] `vine_process_invoke_function` assumes that the input is text and gets the length with `strlen`.  Measure the size of the file instead so that we can send arbitrary data.
- [ ] The library startup message is very minimal JSON.  Let's add in a few things there for error checking, such as a protocol number, taskvine ID, whether it is capable of forking, etc.
- [ ] The expectations of the protocol should be documented as an appendix to the TaskVine manual, so that others can implement libraries regardless of the language.

dthain:
#3436 has been merged so this now describes the master branch.

dthain:
@tphung3 please take a crack at this one.  In the current state, I think serverless is working for functions that don't violate any of the constraints above, but you should do some testing of your own.  These mixups should make the system much more robust.

dthain:
@JinZhou5042 fixed bullet 5 in #3764

dthain:
@tphung3 I'm giving this one back to you as a quick cleanup item that you can do as part of the function context work.

dthain:
This is a quick fix, finish it up today!

tphung3:
In bullet point 4, what does protocol number mean in this context? I'm interpreting it as the type of programming language a given library is implemented and runs in. @dthain

dthain:
No the worker shouldn't care what language the library is written in, as long as it communicates correctly over the pipe.

Just define a library protocol number that gets bumped anytime you change the expected interaction between the worker and the library.

PR #3490, opened 2023-09-15
Vine: Fixing Library Task

tphung3:
This PR partially fixes issue #3439 and brings the following changes:
- Codes are harder to fail and when they do, more meaningful messages are logged.
- Adds `LibraryResponse` class that conveys the execution of a function call. If the function is programmatically correct, the result will be returned. Otherwise, an exception with detailed trace stack is returned.
- Arbitrarily large inputs/outputs to function calls are supported. Unlimited execution time for function calls is supported.
- All inputs and outputs to function calls are serialized by `cloudpickle`.
- Function call's execution time is 900s by default, and this parameter is changeable per function call with `set_time_max`.
- Function calls' inputs and outputs are all `vine_buffer`.
- Remove the wait for library startup code segment. Now libraries are registered to a library table and worker periodically checks if any library is ready to execute functions instead.

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
@@ -966,6 +966,7 @@ class FunctionCall(Task):
         self._event = {}
         self._event["fn_kwargs"] = kwargs
         self._event["fn_args"] = args
+        self.set_time_max(900)     # maximum run time for function calls is 900s by default.
         self.needs_library(library_name)
         self.output_buffer = None
 
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
@@ -20,6 +20,7 @@ import tempfile
 import textwrap
 import uuid
 import weakref
+import cloudpickle
 
 
 ##
@@ -482,7 +483,7 @@ class Task(object):
     @property
     def output(self):
         if (isinstance(self, FunctionCall)):
-            return json.loads(cvine.vine_task_get_stdout(self._task))['Result']
+            return cloudpickle.loads(cvine.vine_task_get_stdout(self._task))['Result']
         return cvine.vine_task_get_stdout(self._task)
 
     ##
@@ -978,7 +979,7 @@ class FunctionCall(Task):
     # @param manager Manager to which the task was submitted
     def submit_finalize(self, manager):
         super().submit_finalize(manager)
-        f = manager.declare_buffer(json.dumps(self._event))
+        f = manager.declare_buffer(cloudpickle.dumps(self._event))
         self.add_input(f, "infile")
 
     ##
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
@@ -22,9 +22,7 @@ import weakref
 try:
     import cloudpickle
     pythontask_available = True
-except Exception:
-    # Note that the intended exception here is ModuleNotFoundError.
-    # However, that type does not exist in Python 2
+except ModuleNotFoundError:
     pythontask_available = False
 
 ##
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/task.py
@@ -482,8 +482,6 @@ class Task(object):
     # @endcode
     @property
     def output(self):
-        if (isinstance(self, FunctionCall)):
-            return cloudpickle.loads(cvine.vine_task_get_stdout(self._task))['Result']
         return cvine.vine_task_get_stdout(self._task)
 
     ##
@@ -969,6 +967,7 @@ class FunctionCall(Task):
         self._event["fn_kwargs"] = kwargs
         self._event["fn_args"] = args
         self.needs_library(library_name)
+        self.output_buffer = None
 
     ##
     # Finalizes the task definition once the manager that will execute is run.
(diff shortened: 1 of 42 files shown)


dthain:
@BarrySlyDelgado please take a look at the Poncho bits.

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
```diff
@@ -12,15 +12,20 @@
...
 import textwrap
 import uuid
 import weakref
-
+try:
+    import cloudpickle
+    pythontask_available = True
+except Exception:
+    # Note that the intended exception here is ModuleNotFoundError.
+    # However, that type does not exist in Python 2
```
dthain, line 27: We no longer care support Python 2.
Just make the code Python 3 native.

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
```diff
@@ -12,15 +12,20 @@
...
 import os
 import shutil
 import sys
 import tempfile
 import textwrap
 import uuid
 import weakref
-
+try:
+    import cloudpickle
```
dthain, line 23: So, this pythontask_available runaround was due to missing dependencies not in the standard Conda package.

But we can easily add packages to the conda package definition, as well as our developer dependencies.  Is there a strong reason to make cloud pickle optional instead of required?

taskvine/src/worker/vine_worker.c
```diff
@@ -198,6 +198,10 @@ static struct itable *procs_complete = NULL;
 // Table of current transfers and their id
 static struct hash_table *current_transfers = NULL;
 
+// Table of all library processes.
```
dthain, line 201: Suggest that we not introduce a separate table for libraries, since that is one more thing to worry about cleaning up properly.  The total number of processes is small, so it seems sufficient to just iterate over procs_running and act on those that are libraries.

taskvine/src/worker/vine_worker.c
```diff
@@ -1291,6 +1306,87 @@ static int enforce_worker_promises(struct link *manager)
...
+
+	/* Loop through all known libraries and check if they are ready. */
+	ITABLE_ITERATE(libraries_table, library_task_id, library_process)
+	{
+		if (!library_process->library_ready) {
+			/* Skip library processes that haven't been started. */
+			if (!library_process->library_read_link) {
+				continue;
+			}
+			library_link_info.link = library_process->library_read_link;
```
dthain, line 1368: `link_sleep` gives a simpler way to check if a single link is readable or writeable.

dthain (changes requested):
This is looking nice.  A few items to clean up or simplify...

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
tphung3: The only reason I can think of is somebody who only wants the C core of TaskVine, which makes cloudpickle redundant.
Since we want users to use `ndcctools` via conda install, do you know if `cloudpickle` is already in ndcctools dependencies?
@btovar What do you think about this?

taskvine/src/worker/vine_worker.c
tphung3: How do you set the stoptime parameter for `link_sleep`? time(0)+500ms?

BarrySlyDelgado (approved):

taskvine/src/worker/vine_worker.c
dthain: Look like `link_sleep` takes a `time_t` but `link_usleep` takes relative milliseconds:
https://cctools.readthedocs.io/en/latest/api/html/link_8h.html#a88872a9b3d0490d2f690e34486786ac5

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
dthain: I suppose all of the other imports are python standard library.  I don't have a strong feeling about this, just want to make sure you are proceeding deliberately...

taskvine/src/worker/vine_worker.c
tphung3: In terms of performance I think link_poll is better right, since it doesn't block or depends on timeouts?

dthain: They all end up calling `poll` one way or the other, just trying to show a way that is more direct...

tphung3:
This PR is RTM.

taskvine/src/worker/vine_worker.c
```diff
@@ -1191,6 +1192,8 @@ static int process_ready_to_run_now(struct vine_process *p, struct vine_cache *c
 		p->library_process = find_library_for_function(p->task->needs_library);
 		if (!p->library_process)
 			return 0;
+		if (!p->library_process->library_ready)
```
dthain, line 1195: This isn't quite right b/c it will find the first library available, and then if it isn't initialized, will fail.

Just adjust `find_library_for_function` to check for `library_ready` as well as other things.

dthain (changes requested):
One more thing

dthain (approved):

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
btovar: I think we should make it required. It is installed as a conda dependency. The C code parts should work even if you don't have python installed.

dthain:
Well, at least we are finding the missing cloud pickles!
Screenshot 2023-09-18 at 11 06 04 AM

dthain:
@tphung3 see #3495 regarding the failed Conda build.

dthain:
Please merge #3495 and I think that should solve your build problems.

dthain:
Ok, so the `build-conda` platform is now correctly loading cloud pickle and running the tests.  But now the serverless test is failing, so that's your department.  :)

I will fix up the OSX test.

dthain:
Ok, merging on #3502 should fix the MacOS problem for you.

dthain (approved):

PR #3492, opened 2023-09-15
Vine: Rewrite Performance Analyzing Function

JinZhou5042:
The previous function had some little bugs that the x axis and y axis were always confused in the output figure, and the code was a a little bit not easy to read, this is a rewritten version that solved those problems. The input doesn't change while the output image will be written in the log directory.


btovar:
It looks like some commands were renamed to have a py extension.

JinZhou5042:
> It looks like some commands were renamed to have a py extension.

Yes, those files are written by python, the renaming is to make them more explicit.

btovar:
@JinZhou5042 The fact that they are written in python is not really relevant here. They are just a tool that happens to be written in python. For this kind of tools we do not use the extension of the language. Please revert the renames.

JinZhou5042:
> @JinZhou5042 The fact that they are written in python is not really relevant here. They are just a tool that happens to be written in python. For this kind of tools we do not use the extension of the language. Please revert the renames.

Got it, and if I want to make some changes about current PR, the best way is to delete the current PR and submit a new one, or can I just edit on the current PR?

btovar:
You can edit the current pr (e.g. see git revert). We prefer working on the same pr unless the commit history gets to ugly to fix.

JinZhou5042:
> You can edit the current pr (e.g. see git revert). We prefer working on the same pr unless the commit history gets to ugly to fix.

Thanks for the information.

colinthomas-z80 (approved):

dthain:
@JinZhou5042 is this PR complete are are you still working on it?

JinZhou5042:
> @JinZhou5042 is this PR complete are are you still working on it?

I think I need one more day to complete it. So if I submitted a PR, should I explicitly indicate that I have completed it in the comments? And what's the good time to decide to merge it?

dthain:
Sure, just say RTM or "ready to merge" in the comments.

JinZhou5042:
> Sure, just say RTM or "ready to merge" in the comments.

Thanks, I will do that!

JinZhou5042:
RTM

taskvine/src/tools/vine_profile_dispatch
```diff
@@ -8,6 +8,10 @@
 
 import sys
 import matplotlib.pyplot as plt
+import matplotlib.gridspec as gridspec
+import os
+import numpy as np
```
dthain, line 13: Do we really need `numpy` as a dependency here?
It's a huge package that would get added to all of our installation paths.

JinZhou5042: Yes, we do not really need `numpy` as well as `statistics`, I've rewritten those lines with pure python.

colinthomas-z80 (approved):

JinZhou5042:
RTM

dthain (approved):

PR #3494, opened 2023-09-18
Vine: Simplify Cancel

dthain:
## Proposed changes

The design of task cancellation is problematic because `task_cancel` is a distinct way of returning a completed task, which makes reference counting more complicated.  This change simplifies cancellation, so that cancelled tasks just get put into the RETRIEVED state and are then returned from `wait` like any other task.

## Post-change actions

### DIFF
--- a/doc/manuals/taskvine/index.md
+++ b/doc/manuals/taskvine/index.md
@@ -1266,10 +1266,10 @@ environment.
 ### Task Cancellations
 
 This feature is useful in workflows where there are redundant tasks or tasks
-that become obsolete as other tasks finish. Tasks that have been submitted can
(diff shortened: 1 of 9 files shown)


dthain:
This is a breaking API change, which I don't normally like, but I think is necessary to fix up our fundamentals.

@btovar does TopEFT / DaskVine make use of task cancellation, and need an update after this PR?

@tphung3 does Parsl / TaskVine make use of task cancellation, and need an update after this PR?

tphung3:
> @tphung3 does Parsl / TaskVine make use of task cancellation, and need an update after this PR?
Parsl/TaskVine does not make use of task cancellation in its implementation.

btovar:
topEFT does, thanks for the heads up.

PR #3501, opened 2023-09-20
Vine: Make conservative disk estimates when the amount of disk is not given to worker

tphung3:
## Proposed changes

When a task with no resources specified lands on a worker, it is assigned all resources of the worker (see `normalize_resources`). However, later on when the worker checks whether the task can fit the current resources, some resources have been taken away by other tasks, so the current task is too big for the worker and thus it is forsaken.

Example: an unspecified task arrives and is assigned 4 GBs of disk space. It depends on a poncho package of size 300 MBs. Once the poncho package is unpacked, the worker later finds out that its disk capacity is only 4 GBs - 300 MBs, so it forsakens the unspecified task as it thinks it doesn't have enough disk space while the unspecified task should run anyway.

Note that this problem doesn't happen with tasks that are clean, i.e., once their executions are done, their workspaces are wiped and resources are returned. Tasks that have side-effects like minitasks use up some disk space so they interact weirdly with regular tasks with unspecified resources.

This problem goes undetected until now probably because the poncho package is too small (<1 MB) or the task passes the resource check before the worker updates its disk counting.

A check whether library name exists is added to prevent seg fault.

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/worker/vine_worker.c
+++ b/taskvine/src/worker/vine_worker.c
@@ -159,7 +159,7 @@ static struct vine_resources *total_resources = 0;
 
 /* When the amount of disk is not specified, manually set the reporting disk to
  * this percentage of the measured disk. This safeguards the fact that disk measurements
- * are estimates and thus may unncessarily forsaken tasks with unspecified resources. 
+ * are estimates and thus may unncessarily forsaken tasks with unspecified resources.
  * Defaults to 90%. */
 static int disk_percent = 90;
 
@@ -326,8 +326,7 @@ static void measure_worker_resources()
 
 	if (manual_disk_option > 0) {
 		r->disk.total = MIN(r->disk.total, manual_disk_option);
-	}
-	else {
+	} else {
 		/* Set the reporting disk to a fraction of the measured disk to avoid
 		 * unnecessarily forsaking tasks with unspecified resources. */
 		manual_disk_option = floor(r->disk.total * disk_percent / 100);
@@ -1147,7 +1146,7 @@ static int task_resources_fit_now(struct vine_task *t)
 {
 	return (cores_allocated + t->resources_requested->cores <= total_resources->cores.total) &&
 	       (memory_allocated + t->resources_requested->memory <= total_resources->memory.total) &&
-		   (disk_allocated + t->resources_requested->disk <= total_resources->disk.total) &&
+	       (disk_allocated + t->resources_requested->disk <= total_resources->disk.total) &&
 	       (gpus_allocated + t->resources_requested->gpus <= total_resources->gpus.total);
 }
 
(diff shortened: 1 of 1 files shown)


btovar:
@tphung3 the poncho packages I use are around 300MB, and expand to 1.4 GB.  The forsaken result occurs when the worker machine has less resources than originally promised. In your tests, are you specifying the `--disk` parameter?  Workers are pretty much unusable right now without specifying the `--disk` parameter.  We do not observe this issue with the submit scripts, or with the factory because they set `--disk` parameter to some value less than the whole machine.

btovar:
I'd suggest not removing the check for disk space. A better solution is to set the default value the disk to some percentage less than the whole machine. (Say, report 75% of the disk if `--disk` is not specified, and add an option to change this percentage.).

tphung3:
> @tphung3 the poncho packages I use are around 300MB, and expand to 1.4 GB. The forsaken result occurs when the worker machine has less resources than originally promised. In your tests, are you specifying the `--disk` parameter? Workers are pretty much unusable right now without specifying the `--disk` parameter. We do not observe this issue with the submit scripts, or with the factory because they set `--disk` parameter to some value less than the whole machine.

I did not set a disk value for workers.

dthain:
Right, I think our mistake is not in the checking of the resources, but in the assigning 100% of disk space to a task.  It's ignoring the disk space assigned to caching.

In general, we need to take a careful look at how disk space allocation is performed, because we are being careful about counting some aspects, and not so careful in others.

tphung3:
The 90% safeguard to disk estimates now remove the task forsaking behavior.

taskvine/src/worker/vine_worker.c
```diff
@@ -320,6 +326,11 @@ static void measure_worker_resources()
 
 	if (manual_disk_option > 0) {
 		r->disk.total = MIN(r->disk.total, manual_disk_option);
+	} else {
+		/* Set the reporting disk to a fraction of the measured disk to avoid
+		 * unnecessarily forsaking tasks with unspecified resources. */
+		manual_disk_option = floor(r->disk.total * disk_percent / 100);
```
btovar, line 332: Protect for values outise [0,100]:

r->disk.total = ceil(r->disk.total * MIN(100, MAX(disk_percent, 0)));

and let's leave manual_disk_option unset to mark that the disk was not specified from the command line.

taskvine/src/worker/vine_worker.c
```diff
@@ -2088,6 +2097,10 @@ static void show_help(const char *cmd)
 	printf(" %-30s Manually set the amount of disk (in MB) reported by this worker.\n", "--disk=<mb>");
 	printf(" %-30s If not given, or less than 1, then try to detect disk space available.\n", "");
 
+	printf(" %-30s Set the conservative disk reporting percent when amount of disk is unspecified.\n",
```
btovar, line 2100: when --disk is unspecified.

taskvine/src/worker/vine_worker.c
```diff
@@ -1170,7 +1179,7 @@ struct vine_process *find_library_for_function(const char *library_name)
 
 	ITABLE_ITERATE(procs_running, task_id, p)
 	{
-		if (!strcmp(p->task->provides_library, library_name)) {
+		if (p->task->provides_library && !strcmp(p->task->provides_library, library_name)) {
```
btovar, line 1182: should this be included in another pr? It looks unrelated.

btovar (changes requested):

btovar (approved):

Issue #3512, opened 2023-09-22
vine: recovery tasks are created incorrectly


btovar:
In some cases recovery tasks do not seem to be created and this causes segfaults. Looking at the gdb backtrace of those tasks:

```json
{task_id = 13519, type = VINE_TASK_TYPE_STANDARD, command_line = 0x0, tag = 0x0, category ...
```

Note that the type should be RECOVERY (STANDARD is 0 in the enum). The task does not have a command, nor a mount lists (which is what causes the segfault when checking for consistency).

btovar:
Issue seems related to tmp files that are lost twice, i.e., lost again once the recovery task ran.

btovar:
It seems the recovery tasks are deleted unconditionally once ran (which seems ok), but they are not recreated.

btovar:
I think it just needs a t->recovery_task = NULL; after the task is deleted. Let me try that.

btovar:
The task does not know its 'parent' file, that might be a complication.

btovar:
I'll try with a vine_task_reset instead.

dthain:
Actually, I think it's the other way around.
The file object contains a single `file->recovery_task`.
If it is needed, it is submitted.
When done, I think we want it left in place, so it can be submitted again if needed.
So either we *shouldn't* delete it when done, or we should add a reference on submission.

btovar:
I think you are right, does the vine_task_reset make sense to you then?

dthain:
See `vine_manager_consider_recovery_task`

btovar:
Oh I see, we just shouldn't delete it, now I get it, I think.

PR #3513, opened 2023-09-22
vine: don't delete recovery tasks

btovar:
## Proposed changes

Solution for #3512 

## Post-change actions


dthain (approved):

btovar:
One issue with this pr is:

```text
warning: task 16421 has more than one input file named cctools-monitor
```

On resubmission, the monitor is added again.  This makes the task fail at the worker, and the task is retried because the file does not exist, which on re-submission causes the monitor to be added again, etc., which is an infinite loop that keeps increasing the memory usage.

The manager can detect that the monitor file is the same, so the mount list can be cleaned that way. However that leaves the problem of what to do with recovery tasks that systematically fail.

PR #3515, opened 2023-09-22
Vine: Fix disk accounting

tphung3:
## Proposed changes

The formula to compute disk total in the codebase is disk.total = disk_available + disk.inuse (see @vine_resources_measure_locally). By introducing a 90% discount we inadvertently discount the disk inuse, leading to incorrect disk total estimate. This change only discounts the disk_available estimate as disk inuse must be reported as-is.

## Post-change actions

## Additional comments
Note: The disk estimate is very unstable when a poncho tarball unpacks itself, leading to a couple of tasks with unspecified resources being forsaken in the beginning. Once the worker stabilizes then tasks with unspecified resources will be run as usual.

### DIFF
--- a/taskvine/src/worker/vine_worker.c
+++ b/taskvine/src/worker/vine_worker.c
@@ -328,8 +328,8 @@ static void measure_worker_resources()
 		r->disk.total = MIN(r->disk.total, manual_disk_option);
 	} else {
 		/* Set the reporting disk to a fraction of the measured disk to avoid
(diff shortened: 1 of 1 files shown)


dthain (approved):
I'm going to approve this b/c we need a quick fix to get to a working state.
But I am not convinced that we are dealing with disk consumption in a clear
and consistent way.  We need to think about this more carefully.

PR #3516, opened 2023-09-25
vine: clean mount list on reset

btovar:
## Proposed changes

Adds the mount flag VINE_RETRACT_ON_RESET to remove files from the mount lists when the task is reset. This is helpful for mounts that vine adds internally for submission (like monitoring files).

## Post-change actions


btovar:
With this change I was able to have a full run of topEFT that had about 50 recovery tasks.

dthain:
That's great, it sounds like we have all the bugs surrounded.

I see the problem here: we have task options that end up adding files to the task as a side effect at submit time.

I'm pondering a bit whether the best solution here is to remove them when necessary, or to make the adding at submit time idempotent.  Is the name of the file always the same?

btovar:
Some cache names may not be the same as they are output names. The task names come from taskids, so they are different too. Another option is to have a second list of mounts per task that taskvine can manipulate freely and that get clear on reset. I didn't went for this one as it involved more changes in the code.

dthain:
Ok, then reverting the addition is the right way to go.

PR #3537, opened 2023-10-09
Vine: Improve Serverless

tphung3:
## Proposed changes

This PR introduces major fixes and improvements to the serverless facility of TaskVine as follows:
- A function invocation didn't interrupt the vine_worker sleep on the manager link, so each invocation must wait 5 seconds to return. This is because a function invocation isn't a child of the vine_worker. Now the library will send a SIGCHLD signal to the worker upon startup (send after configuration) and function execution completion. Note that the library needs the pid of the vine_worker as the vine_worker is the grandparent of the library instead of the parent. The parent of the library is the shell from exec.
- A function execution will implicitly assume that the necessary parameters are in a file called `infile` and it is supposed to dump results to `outfile`. These files are declared and bound properly to every function invocation.
- To avoid the library spinning on a while loop and wasting 1 core, the library uses the self-pipe trick that turns the SIGCHLD from a function exit into an I/O event from the read end of the pipe.
- The library priorities returning function results over starting functions (start one, get all results if possible.)
- Clean previous function invocation code.
- Library now can run multiple functions. This number is configurable via set_function_slots() API. To do this, the worker sends to the library a message containing the function id, function name, and function sandbox. Once complete, the library sends back to the worker the function id.
- Correct counting of running functions.

## Post-change actions

## Additional comments
Note that the starting and reaping function process is a bit different as the function process structure doesn't reflect what is really happening. For example, `p->pid` of a function process will be 0 (uninitialized).

### DIFF
--- a/taskvine/src/worker/vine_worker.c
+++ b/taskvine/src/worker/vine_worker.c
@@ -589,6 +589,10 @@ static void reap_process(struct vine_process *p, struct link *manager)
 		p->exit_code = 1;
 	}
 
+	if (p->type == VINE_PROCESS_TYPE_FUNCTION) {
+		p->library_process->functions_running--;
+	}
+
 	itable_remove(procs_running, p->task->task_id);
 	itable_insert(procs_complete, p->task->task_id, p);
 }
--- a/taskvine/src/worker/vine_worker.c
+++ b/taskvine/src/worker/vine_worker.c
@@ -490,23 +490,16 @@ static void report_worker_ready(struct link *manager)
 	send_keepalive(manager, 1);
 }
 
-int start_function(struct vine_process *p)
+/* Send a message containing details of a function call to the relevant library to execute it. 
+ * @param p 	The relevant vine_process structure encapsulating a function call. 
+ * @return 		1 if the message is successfully sent to the library, 0 otherwise. */
+static int start_function(struct vine_process *p)
 {
-	uint64_t proc_id;
-	struct vine_process *proc;
-	ITABLE_ITERATE(procs_running, proc_id, proc)
-	{
-		if (proc->type == VINE_PROCESS_TYPE_LIBRARY && proc->library_ready &&
-				!strcmp(proc->task->provides_library, p->task->needs_library)) {
-			char *buffer = string_format("%d %s %s", p->task->task_id, p->task->command_line, p->sandbox);
-			link_printf(proc->library_write_link,
-					time(0) + active_timeout,
-					"%ld\n%s",
-					strlen(buffer),
-					buffer);
-			free(buffer);
-			break;
-		}
+	char *buffer = string_format("%d %s %s", p->task->task_id, p->task->command_line, p->sandbox);
+	ssize_t result = link_printf(p->library_process->library_write_link, time(0) + active_timeout, "%ld\n%s", strlen(buffer), buffer);
+	free(buffer);
+	if (result < 0) {
+		return 0;
 	}
 	return 1;
 }
@@ -522,6 +515,7 @@ static int start_process(struct vine_process *p, struct link *manager)
 	pid_t pid;
 
 	struct vine_task *t = p->task;
+	int ok = 1;
 
 	/* Create the sandbox environment for the task. */
 	if (!vine_sandbox_stagein(p, global_cache)) {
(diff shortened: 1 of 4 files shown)


taskvine/src/worker/vine_worker.c
```diff
@@ -490,6 +490,27 @@ static void report_worker_ready(struct link *manager)
 	send_keepalive(manager, 1);
 }
 
+int start_function(struct vine_process *p)
```
dthain, line 493: I don't think it's necessary to search the process table here b/c p->library_process was already set to a stumble library via `process_ready_to_run_now`.

dthain: Is it possible for this function to fail?  If so, there should be some path by which it returns zero, and then we handle the failure.  If not, just make it return void.

taskvine/src/worker/vine_worker.c
```diff
@@ -520,14 +541,25 @@ static int start_process(struct vine_process *p, struct link *manager)
 		vine_gpus_allocate(t->resources_requested->gpus, t->task_id);
 	}
 
+	/* Starting a function call is different from starting a standard task, so we return early. */
+	if (t->needs_library) {
+		itable_insert(procs_running, p->task->task_id, p);
+		debug(D_VINE, "started function call %d: %s", p->task->task_id, p->task->command_line);
+		p->library_process->functions_running++;
+		return start_function(p);
```
dthain, line 549: I would prefer that all task types flow through the function to the end, so that they get treated mostly the same.

For example, this code here skips the important step at the bottom of inserting `p` into `procs_running`.

dthain, line 549: If `start_function` fails, does `functions_running` have the right value?

taskvine/src/worker/vine_worker.c
```diff
@@ -644,19 +680,49 @@ for later processing.
...
-		if (vine_process_is_complete(p)) {
+		/* Skip functions, we take care of them later by checking if the library sends any message. */
+		if (p->type != VINE_PROCESS_TYPE_FUNCTION && vine_process_is_complete(p)) {
 			/* collect the resources associated with the process */
 			reap_process(p, manager);
 
 			/* must reset the table iterator because an item was removed. */
 			itable_firstkey(procs_running);
 		}
+		/* Check if the library has any function invocations done, noblock. */
```
dthain, line 696: This looks like the right idea to me, but please break it out into one or more well named sub-functions.

taskvine/src/worker/vine_worker.c
```diff
@@ -644,19 +680,49 @@ for later processing.
...
+				link_readline(p->library_read_link, buffer, VINE_LINE_MAX, time(0) + local_timeout);
+				int len_buffer = atoi(buffer);
+				memset(buffer, 0, VINE_LINE_MAX);
+
+				// now read the buffer, which is the task id of the done function invocation.
+				link_read(p->library_read_link, buffer, len_buffer, time(0) + 30);
+				uint64_t done_task_id = (uint64_t)strtoul(buffer, NULL, 10);
+				memset(buffer, 0, VINE_LINE_MAX);
+				debug(D_VINE, "Received result for function %lu", done_task_id);
+				itable_firstkey(procs_running);
```
dthain, line 712: No need to iterate over procs_running, just look up by task_id in procs_table.

dthain (changes requested):
Good, this is coming along nicely.
I am excited to see what performance you are able to get.
Since this is the inner loop of the worker, I am going to nitpick to make sure everything is watertight...

taskvine/src/worker/vine_worker.c
tphung3: This makes sense. Fixed!

taskvine/src/worker/vine_worker.c
tphung3: I inserted the process to procs_running in the beginning, but this makes sense. Fixed!

tphung3: No. Fixed!

dthain (approved):

dthain:
Please update the manual to reflect that the library task should define the proper number of function slots.

PR #3550, opened 2023-10-17
WQ: Regenerate serverless script

tphung3:
## Proposed changes

The serverless script in wq test is stale. This PR updates it.

## Post-change actions


dthain (changes requested):
This is likely to become stale again as things change.


How about modifying the test script to generate it automatically using `poncho_package_serverize` ?

dthain:
The problem is the old Python on Ubuntu 20.04. See `TR_vine_python_serverless.sh` for an idea on how to check and skip.

dthain (approved):

PR #3554, opened 2023-10-23
added nopen

jdolak:
## Proposed changes

Added files associated with nopen

Still having issues with running different tests:
- mosaic : works locally but not on condor
- gutenburg : failes on both local and condor
- blast : does not seem to be affected at all

## Post-change actions


nopen/Makefile
```diff
@@ -0,0 +1,39 @@
+CC=			gcc
```
dthain, line 1: You can `include ../config.mk` to get the settings from the top level `configure`

nopen/convert_montage.sfx
dthain, line 1: Don't check in large binary files, just arrange for this to be created as needed via a test script.

nopen/lib-new-open.so
dthain, line 1: Likewise, don't check in binary files.

taskvine/src/worker/vine_process.c
dthain, line 1: Do a `make format` here to satisfy the code-checking lint.

dthain, line 1: Add a new command-line option to the worker to turn this feature on and off.

dthain (changes requested):

dthain:
Also, you may want to rebase on master to catch up with some of the latest changes to the worker.

dthain:
Some things we discussed today for next steps:
- Make all of the configuration happen on the manager side, not the worker side, so it goes in one place.
- Look at `__attribute__((constructor))` to get a function that runs when `nopen` starts.
- Use that to get the CWD in effect when it starts.
- Add coverage for `stat` and `unlink` calls.
- Add attributes to the rules file for read, write, stat, delete, and notify.

Issue #3531, opened 2023-10-05
Vine: Manager Stuck if Workers with VINE_TEMP Time Out


btovar:
One of my runs got into an endless loop because temporary results were lost as workers idled out. By the time those results were regenerated, other workers had idled out, etc.   I think it is expected for workers not to idle out when they hold VINE_TEMP files as they are not really idle (i.e., they are part of the distributed cache).

dthain:
Hmm, this is a tricky problem with competing goals:
- Workers can't stick around forever: we need a garbage collection mechanism for when the manager just disappears.  It doesn't have to be 15 minutes, but it can't be infinite either.
- Temporary files permit a manager to take a risk by leaving a file in place, and then coming back to it later.  If the bet fails, then the manager has to pay the price of regenerating it.  If we have to regenerate too many files, then the manager doesn't make progress.
- By design, the worker doesn't really know the intentions of the manager.  And while it does (sort of) know which files are temporaries, it doesn't have a full understanding of how they will be used, how many times they will be used, etc...

dthain:
@colinthomas-z80 your thoughts?

btovar:
It may be helpful to divide the problem in two:

A. How to deal with temp files when the worker is still connected to the manager and idles out.
B. How to deal with the cache when the worker disconnects from the manager for any reason.

This issue is just A. B is more complicated but only matters for files cached forever. In any other case it wouldn't be surprising for the worker to clean non-forever files on disconnection.

Automatically converting to non-temp files may not work, as the manager may not have the local disk necessary to host the files, or the workers live in some cloud and the manager does not want to pay for the transfers.

My ideal solution would be for the manager to communicate to the worker not to idle out if it has temp files. When files are no longer needed, the manager application can call `vine_file_remove` and let the worker idle out when needed. I'm ok with the manager sending a temp flag when transmitting the file information.

Currently I have to set the timeout for 1h to be on the safe side, but this just like saying "don't garbage collect", but this wastes tons of resources.

colinthomas-z80:
Could the worker ask the manager permission to time out? If the manager doesn’t exist it will proceed as normal, but if the manager is still around it may respond and tell the worker to stick around, or transfer the temporary data back before exiting

colinthomas-z80:
Or the manager could just send keep-alives to workers with important data

dthain:
[...]

PR #3555, opened 2023-10-23
vine: conditional worker timeout

colinthomas-z80:
## Proposed changes

#3531 

This is one implementation where the worker signals the manager it has reached its configured timeout. With this information the manager makes the decision whether or not to release the worker shortly after. 

## Post-change actions

### DIFF
--- a/taskvine/src/manager/vine_file_replica_table.c
+++ b/taskvine/src/manager/vine_file_replica_table.c
@@ -61,10 +61,27 @@ struct vine_worker_info *vine_file_replica_table_find_worker(struct vine_manager
 }
 
 /*
-Determine if this file is cached *anywhere* in the system.
+Count number of replicas of a file in the system.
 XXX Note that this implementation is another inefficient linear search.
 */
 
+int vine_file_replica_table_count_replicas(struct vine_manager *q, const char *cachename)
+{
+	char *key;
+	struct vine_worker_info *w;
+	struct vine_file_replica *r;
+	int c = 0;
+	
+	HASH_TABLE_ITERATE(q->worker_table, key, w)
+	{
+		r = hash_table_lookup(w->current_files, cachename);
+		if (r) c++;
+	}
+	return c;
+}
+
+/*
+ */
 int vine_file_replica_table_exists_somewhere(struct vine_manager *q, const char *cachename)
 {
 	char *key;
--- a/taskvine/src/manager/vine_file_replica_table.c
+++ b/taskvine/src/manager/vine_file_replica_table.c
@@ -71,11 +71,12 @@ int vine_file_replica_table_count_replicas(struct vine_manager *q, const char *c
 	struct vine_worker_info *w;
 	struct vine_file_replica *r;
 	int c = 0;
-	
+
 	HASH_TABLE_ITERATE(q->worker_table, key, w)
 	{
 		r = hash_table_lookup(w->current_files, cachename);
-		if (r) c++;
+		if (r)
+			c++;
 	}
 	return c;
 }
--- a/taskvine/src/manager/vine_file_replica_table.h
+++ b/taskvine/src/manager/vine_file_replica_table.h
@@ -27,6 +27,8 @@ struct vine_worker_info *vine_file_replica_table_find_worker(struct vine_manager
 
 int vine_file_replica_table_exists_somewhere( struct vine_manager *q, const char *cachename );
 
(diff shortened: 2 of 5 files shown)


btovar:
I suggest making the simpler pr of the worker checking if it has "tmp-.." files before idling out, and resetting its timer if it does. My reasoning is that this current pr practically removes the concept of workers idling-out (e.g. the stat is not even filled anymore), and that requires a deeper discussion. E.g., perhaps if the worker can't idle out by itself, then it shouldn't even request idling-out, and the timeout is only used when the worker is not connected to a manager.

It also worries me that workers tends to be created in chunks, and thus they may request idle-out in chunks (I see idle-disconnection en masse in the topeft logs).  This may create  quadratic checks (all worker, all the files) where the manager is only doing disconnection requests.

btovar:
Where does the manager sends the message to the worker to idle-out? Should there be a message, or should the manager simply release drained workers?

colinthomas-z80:
Right, I just hooked it into the drain functionality. Which seemed desirable since we don't want to send any more tasks to a worker between the period they request to idle out and when we finally release them.

btovar:
I think that what is making noise to me is the change of meaning of the timeout. The original timeout triggers a disconnection when the worker is not being useful to the manager. Under this interpretation a worker should not disconnect if it has tmp files.

The new interpretation is that the timeout also affects the lifetime of the worker, even if is useful to the  manager. If there is lull in the workflow, then files will be drained, workers released, and then when new tasks arrive, new workers need to be created and files would need to be transfer again. E.g., say a worker is marked draining=1; even when new tasks arrive that state cannot be reversed. The drain mechanism was thought for factories that want to reduce the number of workers they maintain, not for the workers to decide their lifetime.

The current pr would work not for the idle timeout, but for workers with a hard deadline.

colinthomas-z80:
So I think this pr is roughly equivalent to the original timeout method. Workers end up exiting around the timeout value, regardless of whether they might be useful in the future. The difference is that we won’t lose temp files. 

Whether it is a good idea for the workflow to release all of those workers is a different question. Currently the workers would have timed out and exited anyway. A worker in the draining state is no different from a worker that timed out, we are just releasing it in a safer way.

btovar:
I think we first need to fix the bug according to the current definition for timeout.  Then we can think of the policy on how the manager decides to drain workers, etc. Such policy may take into account the timeout of the workers, although my hunch is that it shouldn't be relevant.

The worker timeout is kind of a crutch, an indirect way of deciding whether the worker is providing value. Thus, it cannot take precedence over a direct way, e.g., is the worker running tasks or has tmp files? The manager should look at its worker collection and decides to consolidate resources if needed, but it shouldn't be based on kind of an artifact of the implementation.

btovar:
My previous comments were based on a misunderstanding of this pr. @colinthomas-z80 convince me this is a good approach for the general case.

btovar:
Is there chance for race conditions when:
- All workers with a copy of a tmp file idle-out around the same time?
- A worker receives a task before its idle-out message is processed?

colinthomas-z80:
Yes I think it might be vulnerable to the first case if q->max_retrievals is configured to be more than 1. Otherwise we will only accept one timeout message per main loop, and we will sequentially flag and let go the workers. 

The drain procedure does check if a worker is running a task before releasing it. However we might lose an output generated by that task.

btovar:
I think the second case can be avoided by not setting the draining flag is the worker has a task. This check should take constant time.

For the first, perhaps instead of setting the draining flag, we can simple release the worker. That would immediately update the file table.  So, on handling the message, release the worker if it is not running a task, and does not have unique copy of a file. After releasing the worker, increment the idle worker stat.

btovar (approved):

dthain:
Make sure to bump `VINE_PROTOCOL_VERSION` since you have changed the interaction on the wire.

btovar:
I don't think the protocol changed. The worker sent that message before, that's how the idle-disconnect stat was populated.

dthain:
ok

colinthomas-z80:
I did change the message though

btovar:
Also remember to increment the stat for idle workers.

btovar (approved):
Thanks @colinthomas-z80!

dthain:
Looks like GitHub upgraded and you need to remove the `--experimental-solver=libmamba` now.

PR #3560, opened 2023-10-26
TaskVine: Fix library deployment method

tphung3:
## Proposed changes

Before this PR the manager tries to send a library to all known workers, which takes a long time and breaks the logic of the internal wait loop. Now the manager sends the library as necessary only.

## Post-change actions


dthain:
Could you summarize the principle of operation here?
Previously, the idea was to make library deployment the lowest priority activity, and then to only deploy one library per worker.
I'm sure one-library-per-worker is no longer what we want.
But not sure if we want to increase the priority of deploying libraries.

tphung3:
The idea here is to piggyback the library with a function call to send to the worker. Instead of sending all libraries to all known workers at once, we send one function call to one worker as usual, and if that worker doesn't alreay have the library running, then we send the library (lazily).

dthain:
Per our conversation, please do some big runs with this new configuration (large number of tasks and workers) and plot the results to make sure it has the desired effect...

dthain:
Another fundamental problem that I see here is that we are not checking for the compatibility of the library task with the worker, nor are we checking to see if the worker has the appropriate resources available for the library.  For example, perhaps the library needs a GPU.

We need a scheme by which library tasks are scheduled to workers in a "mostly normal" way so that we are checking all of the properties that are important to the task.

tphung3:
A quick experiment shows that the proposed way delivers minimal improvements, even though it is technically more sound than the previous way of delivering libraries to workers. This might be because the scenario when a manager spends a lot of time to send libraries to a large number (100s) of workers isn't real in practice: instead the manager sends to small chunks of workers (in 20-30) as they arrive. Some numbers are provided below to back up this claim. Note that the experiment setup is exactly the same, except the way to deliver libraries:

With 20,000 function calls, each running for 5 seconds, and 100 16-core workers, each can run 16 concurrent function tasks:
- old: 105.88 seconds
- new: 104.21 seconds

Same set of function calls, and 400 4-core workers, each can run 4 concurrent function tasks:
- old: 178.83 seconds
- new: 175.46 seconds

On the second thought, if a user has rounds of experiments within one workflow, which installs one library then installs another library after 500 workers are connected, then the old way of sending libraries will clearly block the main wait loop. Maybe another experiment can confirm this.

Tagging relevant people @dthain @btovar.

taskvine/src/manager/vine_manager.c
```diff
@@ -2592,6 +2591,25 @@ static vine_result_code_t start_one_task(struct vine_manager *q, struct vine_wor
...
+	/* Send library first if this is a function call and worker doesn't have the library. */
+	int ok = 1;
+	if (t->needs_library && !hash_table_lookup(w->libraries, t->needs_library)) {
+		if (vine_manager_send_library_to_worker(q, w, t->needs_library)) {
+			hash_table_insert(w->libraries,
+					t->needs_library,
+					vine_task_copy(hash_table_lookup(q->libraries, t->needs_library)));
+			debug(D_VINE, "Sending library %s to worker %s\n", t->needs_library, w->workerid);
+		} else {
+			debug(D_VINE, "Cannot send library %s to worker %s\n", t->needs_library, w->workerid);
```
dthain, line 2603: This debug could be misleading b/c `vine_manager_send_library_to_worker` may return zero if the library simply does not fit the worker, or is incompatible.  In short, it's not really a failure, it's just the manager realizing (late in the game) that it doesn't make sense to schedule this task here.

taskvine/src/manager/vine_manager.c
```diff
@@ -2723,6 +2728,11 @@ static void commit_task_to_worker(struct vine_manager *q, struct vine_worker_inf
 	if (result != VINE_SUCCESS) {
 		debug(D_VINE, "Failed to send task %d to worker %s (%s).", t->task_id, w->hostname, w->addrport);
 		handle_failure(q, w, t, result);
+	} else {
+		if (t->needs_library) {
+			struct vine_task *library_task = hash_table_lookup(w->libraries, t->needs_library);
```
dthain, line 2733: Use `vine_manager_find_library_on_worker` to find **one** of the running libraries on the worker with an available slot, and then store it in `t->library_task` so that you can correctly un-do the binding when the task ends.

dthain (changes requested):
I think that this approach can work, but it is a mistake to remove the `t->library_task` pointer.  The reason for this pointer is to associate a function call task with one *specific* library task on that host during scheduling.  (This is because a host could run more than one library task at once.). Then the `t->library_task` pointer is used to perform the needed reference counting on the matched library.

Second, I'm not sure that is necessary to introduce the `w->libraries` table since the library task is already represented in `w->current_tasks`.   And every table of aliases becomes a new complication to keep track of.  You can use  the existing `vine_manager_find_library_on_worker` to find a suitable library, and then keep that binding in `t->library_task`

dthain:
Per our recent discussions, please fix up this PR as follows:

1 - Libraries should consume resources in the same way as normal tasks, using the same routines.  (Use explicit resources if given, otherwise fill the worker.) 
2 - It is allowed to have multiple libraries per worker, so function call tasks should continue to track what library they are connected to via `t->library_task`
3 - Libraries should be deployed when needed by function calls, and then removed when the need no longer exists.

I think this can be done with relatively minor changes.  And then let's focus on getting this working with Parsl applications at large scale, and evaluate the performance.  (And probably there will be some bugs to fix along the way.)

taskvine/src/manager/vine_manager.c
```diff
@@ -2723,6 +2725,10 @@ static void commit_task_to_worker(struct vine_manager *q, struct vine_worker_inf
 	if (result != VINE_SUCCESS) {
 		debug(D_VINE, "Failed to send task %d to worker %s (%s).", t->task_id, w->hostname, w->addrport);
 		handle_failure(q, w, t, result);
```
dthain, line 2727: Double check this: under certain circumstances, `handle_failure` can call `reap_task_from_worker` and so you might be a `function_slots_inuse` decrement without a corresponding increment.

taskvine/src/manager/vine_manager.c
```diff
@@ -2813,11 +2819,8 @@ static void reap_task_from_worker(
 	and disassociate the task from the library.
 	*/
 
-	if (t->library_task) {
+	if (t->needs_library) {
 		t->library_task->function_slots_inuse--;
```
dthain, line 2823: Minimum of zero here, for reasons given above?

taskvine/src/manager/vine_schedule.c
```diff
@@ -86,6 +67,57 @@ int vine_schedule_in_ramp_down(struct vine_manager *q)
 	return 0;
 }
 
+/* Check if worker resources are enough to run the task.
+ * @param q     Manager info structure.
+ * @param w     Worker info structure.
+ * @param t     Task info structure.
+ * @param tr    Task resources.
+ * @return 1 if yes, 0 otherwise. */
+int check_worker_have_enough_resources(
```
dthain, line 76: I like how you broke this out separately here.

Put in some big fat comments at the top here explaining that the check takes into account killing unneded libraries.

taskvine/src/manager/vine_schedule.c
```diff
@@ -86,6 +67,57 @@ int vine_schedule_in_ramp_down(struct vine_manager *q)
...
+		struct vine_manager *q, struct vine_worker_info *w, struct vine_task *t, struct rmsummary *tr)
+{
+	struct vine_resources *worker_net_resources = vine_resources_copy(w->resources);
+
+	/* Pretend to reclaim resources from empty libraries. */
```
dthain, line 81: This is potentially an expensive loop, so make it conditional on `t->needs_library`.
And then break it out into a separate function something like `subtract_empty_libraries` or similar.

dthain (changes requested):
Ok, I think you have convinced me that this is a workable idea.  Since it is core to the scheduling loop, please make a few changes to clarify for the next reader.

dthain:
Looks like the OSX build just timed out and preahps will succeed with a simple retry.

taskvine/src/manager/vine_manager.c
tphung3: You're right. The solution is to increment the function slots no matter what, then `handle_failure` can decrement it if failure happens.

taskvine/src/manager/vine_schedule.c
tphung3: I think this check is needed for any type of task, as a regular task can kick out library tasks if they are empty.

dthain: Ah, I see you are right.  So we should be cognizant that this change increases the complexity of the inner scheduling loop.  And may have an impact on scheduling large numbers of tasks.  @colinthomas-z80 what do you think?

tphung3: Do we have a script that benchmarks TaskVine against a large number of tasks? That'd be helpful here.

colinthomas-z80: I usually use vine_benchmark with something like `submit 0 0 0 100000`

dthain: Discuss with the person sitting next to you!

colinthomas-z80: And if I understand this correctly, this iterates through currently running tasks rather than queued? 

I think it is safe to say that most of the time there are not a hugely significant amount of tasks running.

tphung3:
It's the second time the OSX build fails. Can this be related to the recent OSX build changes @dthain?

taskvine/src/manager/vine_schedule.c
colinthomas-z80: @dthain Office hours :)

dthain:
Please rebase on master to get a workaround for the OSX build.

dthain (approved):

dthain:
RTM?

tphung3:
Yes it is

PR #3562, opened 2023-10-31
Vine: add cpu time to txn stats

colinthomas-z80:
## Proposed changes
Quick addition to sum the worker running time. It will be an interesting statistic when considering ramp-down performance. I may add a running time vs idle time as well. 

## Post-change actions


colinthomas-z80:
After some further thought I realize I only calculated the combined worker runtime rather than individual CPU. Perhaps both would be useful

dthain (approved):

Issue #3563, opened 2023-11-02
Vine: Edge Cases for Temporary Files


dthain:
Per our discussion yesterday, several challenges and ideas regarding handling of temporary files.  Currently, temporary files are associated with the tasks that create them, so if a file is lost, it gets recreated.  This is good so far, but causes trouble when the failure rate becomes sufficiently high.

So, we need one or more operation options like this:

- [x] Set a minimum replica count for temporary files.  When a temporary file is created (or also lost?) immediately send out `put url` replication requests to workers to increase the replica count.  Probably should be a configurable option since this could significantly increase cost.

- [x] Make temporary file handling an option for the user to set.  In "conservative" mode, just treat all temporary files like normal local files, so that they are brought home to the manager.  Slower, yes, but also more reliable.  (They can still be cached on workers, of course.)

- [ ] Make one or both of these options adaptive.  When "enough" failures happen, the system should automatically become more conservative to avoid cascading failures.

@colinthomas-z80 other ideas?

colinthomas-z80:
I am thinking a user can declare a replica count, perhaps in an overload of declare_temp. The manager will coordinate its distribution. Maybe on handling the cache update for the first copy? Making sure each replica is stored on a different hostname.

I don't know how convenient a "conservative" mode would be. This would essentially just make it a normal file, and if the user explicitly declared a temp I would assume there are disk space issues, or perhaps something else that makes it undesirable to come back to the manager. 

Adaptive methods are alluring but they might be difficult to configure. One workflow may tolerate more failures than another so the user would have to think of the exact threshold to switch to conservative. Then, once we switch to conservative mode, do we bring everything back all at once? Or only new files?

dthain:
I am all in favor of doing simple things first before complicated things!

Specifying the replica count in the file object makes sense to me.

colinthomas-z80:
Exploring the implementation of this has me thinking about another topic I am considering opening an issue about. That is the disconnect in terms of data structures, between vine file objects and the manager's understanding of the worker's cache. 

That is, we declare files at the manager and keep around the objects for local use, reference counting, etc... 

Yet the hash tables tracking worker's cache files only contain cachenames, which is a field of struct vine_file, however there is no backwards reference possible from cachename->vine_file. The consequence is that we often do linear searches through the worker's hash tables looking for a cachename to count its replicas. 

[...]

PR #3564, opened 2023-11-06
Vine: temp file replication

colinthomas-z80:
## Proposed changes
Related to #3563 

At the moment, upon a cache update, the manager will check if a temp file was created.

If that is the case, the manager will look for up to q->temp_replica_count workers to replicate the file to.

It uses the new message, `puturl_now,` to tell a worker to bring a remote file to the cache without the need to run a task

## Post-change actions

### DIFF
--- a/taskvine/src/manager/vine_file_replica_table.c
+++ b/taskvine/src/manager/vine_file_replica_table.c
@@ -60,6 +60,34 @@ struct vine_worker_info *vine_file_replica_table_find_worker(struct vine_manager
 	return 0;
 }
 
+struct vine_worker_info **vine_file_replica_table_find_replication_targets(struct vine_manager *q, struct vine_worker_info *w, const char *cachename, int *count)
+{
+	char *id;
+	struct vine_worker_info *peer;
+	struct vine_file_replica *remote_info;
+	
+	int found = 0;
+	struct vine_worker_info **workers = malloc(sizeof(struct vine_worker_info)*(q->temp_replica_count));
+
+	HASH_TABLE_ITERATE(q->worker_table, id, peer)
+	{
+		if (found == q->temp_replica_count) break;
+		if (!peer->transfer_port_active)
+			continue;
+
+		// generate a peer address stub as it would appear in the transfer table
+		char *peer_addr = string_format("worker://%s:%d", peer->transfer_addr, peer->transfer_port);
+		if (!(remote_info = hash_table_lookup(peer->current_files, cachename)) && (strcmp(w->hostname, peer->hostname))) {
+			debug(D_VINE, "found replication target : %s", peer_addr); 
+			workers[found] = peer;
+			found++;
+		}
+		free(peer_addr);
+	}
+	*count = found;
+	return workers;
+}
+
 /*
 Count number of replicas of a file in the system.
 XXX Note that this implementation is another inefficient linear search.
--- a/taskvine/src/manager/vine_file_replica_table.c
+++ b/taskvine/src/manager/vine_file_replica_table.c
@@ -79,17 +79,11 @@ struct vine_worker_info **vine_file_replica_table_find_replication_targets(
 		if (!peer->transfer_port_active)
 			continue;
 
-		if (!(remote_info = hash_table_lookup(peer->current_files, cachename)) &&
-				(strcmp(w->hostname, peer->hostname))) {
-			debug(D_VINE, "found replication target : %s", peer->transfer_addr);
-			workers[found] = peer;
-			found++;
-		}
 		char *peer_addr = string_format("worker://%s:%d", peer->transfer_addr, peer->transfer_port);
 		if (!(remote_info = hash_table_lookup(peer->current_files, cachename)) &&
 				(strcmp(w->hostname, peer->hostname))) {
-			if (vine_current_transfers_worker_in_use(q, peer_addr) < q->worker_source_max_transfers) {
-				debug(D_VINE, "found replication target : %s", peer->transfer_addr);
+			if ((vine_current_transfers_worker_in_use(q, peer_addr) < q->worker_source_max_transfers) && (vine_current_transfers_dest_in_use(q, peer) < q->worker_source_max_transfers)) {
+				debug(D_VINE, "found replication target : %s, current load %d", peer->transfer_addr, vine_current_transfers_dest_in_use(q, peer));
 				workers[found] = peer;
 				found++;
 			}
--- a/taskvine/src/manager/vine_file_replica_table.c
+++ b/taskvine/src/manager/vine_file_replica_table.c
@@ -60,25 +60,28 @@ struct vine_worker_info *vine_file_replica_table_find_worker(struct vine_manager
 	return 0;
 }
 
-struct vine_worker_info **vine_file_replica_table_find_replication_targets(struct vine_manager *q, struct vine_worker_info *w, const char *cachename, int *count)
+struct vine_worker_info **vine_file_replica_table_find_replication_targets(
+		struct vine_manager *q, struct vine_worker_info *w, const char *cachename, int *count)
 {
 	char *id;
 	struct vine_worker_info *peer;
 	struct vine_file_replica *remote_info;
-	
+
 	int found = 0;
-	struct vine_worker_info **workers = malloc(sizeof(struct vine_worker_info)*(q->temp_replica_count));
+	struct vine_worker_info **workers = malloc(sizeof(struct vine_worker_info) * (q->temp_replica_count));
 
 	HASH_TABLE_ITERATE(q->worker_table, id, peer)
 	{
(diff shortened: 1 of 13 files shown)


colinthomas-z80:
Proof of concept

dthain:
Please describe the changes above in the nice form provided.  :)

colinthomas-z80:
image

Working in a basic case

dthain:
Nice!  Let me observe that "now" doesn't really mean "synchronously right away" but rather "without waiting for a dependency to require it".   The transfer could still be delayed for reasons of capacity, concurrency limits at the worker, etc.

A question:  Should all url requests be done "now" ?  Is there an upside to having "normal" transfers wait until something requests their presence?

@BarrySlyDelgado what do you think?

BarrySlyDelgado:
The answer is likely that it depends. I think it is mostly beneficial if you are not worried about resource utilization. For example, with limited disk space downloading files without premeditated necessity could fill up space quickly. However, without any worries about certain resources, if a worker is free to download a specific file from a remote source, it might as well do so as it could save start up time for task executions in the future. In regard to concurrency limitations, scenarios could arise in which a worker may be instructed to download a file "now" reaching a set concurrency limit and could possibly block a task from running though it may be a limited case.

colinthomas-z80:
Had a successful large scale run where worker loss was handled without recovery tasks. 

Still need to refine further as it is not distributing the files evenly and definitely overloading specific workers, both as the sender and receiver.

dthain:
That's great!  Balance is a tricky issue.  This is not quite as simple as it first appears...

dthain:
Please rebase on master to get a workaround for the OSX build.

colinthomas-z80:
oops

btovar:
replicate your pr!

taskvine/src/manager/vine_file_replica_table.c
```diff
@@ -60,6 +60,45 @@ struct vine_worker_info *vine_file_replica_table_find_worker(struct vine_manager
...
+		struct vine_manager *q, struct vine_worker_info *w, const char *cachename, int *count)
+{
+	char *id;
+	struct vine_worker_info *peer;
+	struct vine_file_replica *remote_info;
+
+	int found = 0;
+	struct vine_worker_info **workers = malloc(sizeof(struct vine_worker_info) * (q->temp_replica_count));
+
+	HASH_TABLE_ITERATE(q->worker_table, id, peer)
```
btovar, line 75: With this, workers listed at the start of the table will get all the replicas. Do you want to make this more uniform?

taskvine/src/manager/vine_manager_put.c
```diff
@@ -211,6 +211,36 @@ may be an estimate at this point and will be updated by return
...
+			w,
+			"puturl_now %s %s %lld %o %s\n",
+			source_encoded,
+			cached_name_encoded,
+			(long long)f->size,
+			0777,
+			transfer_id);
+
+	free(transfer_id);
+	return VINE_SUCCESS;
```
btovar, line 234: I think you want to return the return status of vine_manager_send.

taskvine/src/worker/vine_worker.c
```diff
@@ -1092,6 +1101,19 @@ static int handle_manager(struct link *manager)
...
+					   filename_encoded,
+					   &length,
+					   &mode,
+					   transfer_id) == 5) {
+			url_decode(filename_encoded, filename, sizeof(filename));
+			url_decode(source_encoded, source, sizeof(source));
+			r = do_put_url_now(filename, length, mode, source);
+			reset_idle_timer();
+			hash_table_insert(current_transfers, strdup(filename), strdup(transfer_id));
+			debug(D_VINE, "Insert ID-File pair into transfer table : %s :: %s", filename, transfer_id);
```
btovar, line 1116: Is this a debug statement for you?

btovar (changes requested):

taskvine/src/worker/vine_worker.c
colinthomas-z80: I suppose so, it is just mirroring the existing puturl handler. I will remove them

taskvine/src/manager/vine_file_replica_table.c
colinthomas-z80: Yes its not very sophisticated, it will only give the first worker q->worker_source_max_transfers before moving on to the next. If temp files are sporadically created it will only deliver to the first worker. I will think up some heuristic

taskvine/src/worker/vine_cache.c
```diff
@@ -250,6 +250,18 @@ int vine_cache_queue_transfer(struct vine_cache *c, const char *source, const ch
...
 
+/*
+Perform a remote file transfer to produce a file.
+*/
+
+int vine_cache_transfer_now(struct vine_cache *c, const char *source, const char *cachename, int64_t size, int mode)
+{
+	struct vine_cache_file *f = vine_cache_file_create(VINE_CACHE_TRANSFER, source, size, mode, 0);
+	hash_table_insert(c->table, cachename, f);
+	vine_cache_ensure(c, cachename);
```
dthain, line 261: Suggest that you just make this method a variant of `vine_cache_queue_transfer` and have an "immediate" flag of some kind.

taskvine/src/manager/vine_manager.c
```diff
@@ -380,6 +418,11 @@ static int handle_cache_update(struct vine_manager *q, struct vine_worker_info *
 		vine_current_transfers_remove(q, id);
 
 		vine_txn_log_write_cache_update(q, w, size, transfer_time, start_time, cachename);
+
+		// If the file was not a url, eg. 'X', and it is a temp
+		if (*id == 'X' && strncmp(cachename, "temp-rnd-", 9) == 0) {
```
dthain, line 423: This seems unavoidable for now, but I would rather have some direct information present in the `vine_file_replica` structure that tells you exactly what kind it is, whether it is a temporary file, etc, rather than inferring it from the name.

To make this work, the manager would have to send the relevant data to the worker, so that it could be returned back after failure recovery.

taskvine/src/manager/vine_manager.c
```diff
@@ -343,6 +343,44 @@ static vine_msg_code_t handle_info(struct vine_manager *q, struct vine_worker_in
...
+	workers = vine_file_replica_table_find_replication_targets(q, w, cachename, &found);
+
+	debug(D_VINE, "Found %d available workers to replicate %s", found, cachename);
+
+	for (int i = 0; i < found; i++) {
+		struct vine_worker_info *peer = workers[i];
+		char *worker_source = string_format("worker://%s:%d/%s", w->transfer_addr, w->transfer_port, cachename);
+		char *worker_addr = string_format("worker://%s:%d", w->transfer_addr, w->transfer_port);
+
+		char *transfer_id = vine_current_transfers_add(q, peer, worker_source);
```
dthain, line 367: For consistency, break this out into `vine_manager_put_url_now` next to `vine_manager_put_url`.
Also need to url encode the names.

dthain (commented):
This is on the right track, I know you are still looking at scheduling/placement issues.

dthain:
Status on this one?

colinthomas-z80:
File distribution is looking good. Still need to clean up the interface as was requested

taskvine/src/worker/vine_cache.h
```diff
@@ -45,8 +50,9 @@ void vine_cache_scan( struct vine_cache *c, struct link *manager );
 char *vine_cache_full_path( struct vine_cache *c, const char *cachename );
 
 int vine_cache_addfile( struct vine_cache *c, int64_t size, int mode, const char *cachename );
-int vine_cache_queue_transfer( struct vine_cache *c, const char *source, const char *cachename, int64_t size, int mode );
+int vine_cache_queue_transfer( struct vine_cache *c, const char *source, const char *cachename, int64_t size, int mode, int flags );
```
dthain, line 53: Typically, `flags` implies some bit values or-ed together.  Is that your intention here?  If so, then use the enumeration in the type, and call things consistently `VINE_CACHE_FLAGS_X` and `vine_cache_flags_t` with assigned values.  (And then check for `flags&value` instead of `flags==value`.)

If this is meant to be restricted to a single integer value indicating when to act, then call the parameter `vine_cache_when_t when` and make the values `VINE_CACHE_WHEN_X` etc so that it is unavoidably clear what goes where.

dthain (changes requested):
Last little bit -

colinthomas-z80:
I will go the flag route, since we have been discussing ensuring the cache on task startup and we may think of other ideas

dthain:
And finally make sure the setting is documented in `taskvine.h` and find the right place to briefly mention how temp files are duplicated/recovered in the main manual.

dthain (approved):

PR #3581, opened 2023-11-29
DaskVine Serverless

BarrySlyDelgado:
## Proposed changes

This enables serverless with DaskVine. A new class, FunctionCallDask, creates function calls instead of PythonTasks. These options set with the `task_mode` option.   

## Post-change actions

## Additional comments
I have tested it on a small scale locally and distributed. There are still some things to fix.


dthain:
Please merge or rebase on master.

Also note change needed to #3589 .

Also talk to @tphung3 about the bug he just found in scheduling!

dthain:
Taking a quick look, I see that an output file is declared in `submit_finalize`.  Is there a corresponding `remove_file` in the destructor?

BarrySlyDelgado:
No, there is no `remove_file` at the moment.

dthain:
I'm asking you if it needs one.  :)

BarrySlyDelgado:
It can be added to the destructor however, the DaskVine executor will need only to delete the FunctionCall tasks and remove the files once all the tasks that may consume its output are completed.

dthain:
@BarrySlyDelgado please rebase or merge on master to pick up the memory leak fix from @JinZhou5042 .

dthain:
@BarrySlyDelgado is this the code being used to run HEP applications?  If so, rebase on master and let's get this merged.

dthain:
Hello?

BarrySlyDelgado:
Yes, this is currently used for testing function calls for the HEP applications.

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
```diff
@@ -87,6 +94,12 @@ def get(self, dsk, keys, *,
             low_memory_mode=False,
             checkpoint_fn=None,
             resources=None,
+            lib_resources=None,
+            lib_command=None,
+            import_modules=None,
+            dumb_env=None,
```
dthain, line 100: Make this a permanent option with a good name, but just describe in a way that somebody won't use it by accident.

BarrySlyDelgado: Changed to env_per_task

dthain:
I'm ready to merge this once you have pulled in the necessary changes from master.

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
```diff
@@ -71,14 +74,22 @@ class DaskVine(Manager):
     #                      to the manager.
     # @param resources     A dictionary with optional keys of cores, memory and disk (MB)
     #                      to set maximum resource usage per task.
+    # @param lib_resources A dictionary with optional keys of cores, memory and disk (MB)
```
btovar, line 77: I think we want all these arguments to be in  `get`, and not in the definition of the DaskVine manager. Note that environment and resources for regular tasks are defined in `get`.

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
```diff
@@ -122,6 +145,12 @@ def get(self, dsk, keys, *,
             self.tune("immediate-recovery", 1)
             self.tune("max-retrievals", 10)
 
+            if self.env_per_task:
+                self.poncho_file = self.declare_file(shutil.which('ponho_package_run'))
```
btovar, line 149: Environments do not need poncho_package_run, and should be added with `add_environment` or similar.

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
```diff
@@ -137,6 +166,26 @@ def _dask_execute(self, dsk, keys):
         dag = DaskVineDag(dsk, low_memory_mode=self.low_memory_mode)
         tag = f"dag-{id(dag)}"
 
+        # create Library if using 'function_calls' task mode.
+        if self.task_mode == 'function_calls':
```
btovar, line 170: If this is part of the API, then either "function calls" or "function-calls"

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
btovar: Also, there is a typo in poncho_package_run.

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
```diff
@@ -182,26 +231,44 @@ def _enqueue_dask_calls(self, dag, tag, rs, retries, enqueued_calls):
...
+                                   dag, k, sexpr,
+                                   category=cat,
+                                   environment=self.environment,
+                                   extra_files=self.extra_files,
+                                   env_vars=self.env_vars,
+                                   retries=retries,
+                                   lazy_transfers=lazy)
+
+                if self.env_per_task:
+                    t.set_command('{} -e {} {}'.format('poncho_package_run', self.environment_name, t._command))
```
btovar, line 255: Why not use `add_environment`?

btovar (changes requested):

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
BarrySlyDelgado: This is a means to have each task unpack its environment and activate it. And is intended for testing runs. I was unsure if there was a better way to do so asI saw some caching with `declare_poncho` and `add_environment` even when `cache=False`

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
BarrySlyDelgado: For clarification, I think I put these arguments in the definition for the function `get` and not `DaskVine(manager)`

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
btovar: The problem is that not all environments are poncho. Some options:
- Add a flag to any of the `declare_*` that relate to environments that indicate whether they should be shared.
- Or, only share if the environment file was declared as cachable.

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
btovar: Sorry for the noise, I think I saw the class definition and blindly assumed that the first method was __init__.

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
btovar: See https://github.com/cooperative-computing-lab/cctools/issues/3643

btovar: Looking at this code, I think that if you have two tasks running at the worker using the same poncho tarball, then they still share the environment, right? The only way they don't share it is if only one task runs at any time. If the worker gets an uninterrupted overlapping series of tasks, then I think that the same environment is used over and over.

btovar: If that's true, and this code is only added for testing, then it might be much easier if the same tarball gets hard or symlinked with different names before declaring the tasks in your tests.  Otherwise we would need to modify how things are named according to the cache rules for little pay off.

Another option is to add an option to the run_in_env that poncho_package_create generates so that it re-expand the tarball every time, maybe with an environment variable or something like that.

BarrySlyDelgado: While testing I haven't had them share environments using this method (from looking at the plots). Each task would unpack and and activate the environment in its own sandbox. I think. I'll look into another method.

btovar (approved):

PR #3585, opened 2023-12-07
TaskVine: Fix problem where function calls print to stdout of the vine_worker process

tphung3:
## Proposed changes

Before, if a function call prints anything to stdout, then it is routed to the vine_worker process' stdout. This fixes that and makes the library task have the same stdin/out/err setup as the standard task.

## Post-change actions

### DIFF
--- a/poncho/src/poncho/library_network_code.py
+++ b/poncho/src/poncho/library_network_code.py
@@ -88,7 +88,7 @@ def library_network_code():
         
         # now read the buffer to get invocation details
         line = str(os.read(in_pipe_fd, buffer_len), encoding='utf-8')
-        function_id, function_name, function_sandbox = line.split(" ", maxsplit=2)
+        function_id, function_name, function_sandbox, function_stdout_file = line.split(" ", maxsplit=3)
         function_id = int(function_id)
         
         if function_name:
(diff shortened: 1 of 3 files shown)


dthain:
So, as far as I can tell:
- A library task will get two pipes created, which are noted on the command line.
- All tasks get the same stdin/stdout/stderr handling.
- A function call will get an output file created in the task sandbox, but it won't write anything there because any printfs in the library will go to the library's stdout instead.

Is that correct?

Now, conceivably, we could have the library main loop re-open its stdout/err to write to the function call tasks' stdout instead, after the fork in library_network_code.py.  That way, any output specific to the function call would come back with the function call itself.

Good idea or bad idea?

tphung3:
Yes to points 1 and 2. For point 3, the function call writes the result to its output file once it's done, but its stdout/err are not captured like other tasks.

tphung3:
The new change should capture stdout/err of the library task and the function call in two separate places.

dthain (changes requested):
This seems like a nice simplification of file handling in general.
1 - Please bump the VINE_PROTOCOL number, since this would break compatibility with prior managers.
2 - Do we need to update the corresponding library network code used in parsl, or is that generated automatically?

tphung3:
1) Do you mean this line? https://github.com/cooperative-computing-lab/cctools/blob/7d5cf463a4175330335c91a54b31860a9da5b444/taskvine/src/manager/vine_protocol.h#L16

2) The network code is generated automatically from poncho network code.

dthain (approved):

PR #3591, opened 2023-12-13
TaskVine: Fix bug in function call scheduling

tphung3:
## Proposed changes

The current implementation of scheduling function calls is buggy, as for a given task T1, it looks for a candidate worker W in the regular way as other tasks. It then checks W to see if W has the capacity to run T1. The problem is when the second check fails, the main loop in `send_one_task` rotates to the next task T2, T3, T4, etc., until W is available to run a function call again, instead of finding the next worker to run T1.

The fix moves the checking code inside `check_worker_against_task` in `vine_schedule.c`, effectively merging two checks into one. So all workers are checked for a task first before we loop through the task list.

## Post-change actions

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -3109,7 +3109,8 @@ static struct vine_task *find_library_on_worker_for_task(struct vine_worker_info
 /* Check if this worker can run the function task.
  * @return 1 if it can, 0 otherwise.
  */
-static int check_worker_can_run_function_task(struct vine_manager *q, struct vine_worker_info *w, struct vine_task *t)
+int vine_manager_check_worker_can_run_function_task(
+		struct vine_manager *q, struct vine_worker_info *w, struct vine_task *t)
 {
 	struct vine_task *library = find_library_on_worker_for_task(w, t->needs_library);
 	if (!library) {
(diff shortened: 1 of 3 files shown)


btovar (approved):

dthain:
This is surprisingly simple!  I was going to comment on a few matters, but it seems to handle them:
- `find_library_on_worker_for_task` identifies a library with a slot available
- `send_library_to_worker` will fail cleanly if the library isn't compatible with the worker.

The only thing that is odd about it is that the scheduler itself may end up dispatching library tasks in the middle of scheduling.  Is that a problem?

tphung3:
In terms of delaying the scheduling loop, the cost of sending the library to a worker is amortized by the upcoming function calls to that library. Anything else we should be worried about?

dthain:
I'm not worried about costs here, just correctness.  As far as I can tell `check_task_against_worker` is reentrant as long as the recursion doesn't occur inside of a data structure traversal.

tphung3:
I see. About this, I think `check_worker_against_task` doesn't retain any state before the recursion point, and all functions above the recursion point are self-contained in the sense that they delete what they modify or they use read-only operations to check things. Also the recursion is at most 2 levels so it's a bit simpler for me to imagine how this can go wrong.

PR #3599, opened 2024-01-10
wq: adds dask executor.

btovar:
Works pretty much as the dask executor for dask:

```python
from ndcctools.work_queue_dask import DaskWQ

m = DaskWQ(port=args.port, ssl=True)
m.specify_name(args.name)

...

with dask.config.set(scheduler=m.get):
    result = some.compute(resources={"cores": 1})

staging dir is set with the environment variable CCTOOLS_TEMP outside of the python interpreter (or before importing any wq related modules).
```

### DIFF
--- a/work_queue/src/bindings/python3/Makefile
+++ b/work_queue/src/bindings/python3/Makefile
@@ -29,7 +29,7 @@ clean:
 install: all
 	mkdir -p $(CCTOOLS_PYTHON3_PATH)/ndcctools
 	chmod 755 work_queue_example.py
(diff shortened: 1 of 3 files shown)


dthain (approved):
This is looking nice.  Since this is for performance eval rather than recommended usage, I will for my usual request for doc updates.

btovar:
Yes, I figured we probably don't want to advertise this one too much as dask+taskvine is the recommended path.

btovar:
I think it should be ready to merge. Something will come up for sure when trying the examples at scale, but the basic functionality is there.

Issue #3568, opened 2023-11-10
Vine: Generalize Futures to FunctionCall Tasks


JinZhou5042:
Currently, by using the serverless feature we can easily install a library on TaskVine and invoke functions in it whenever we need:
```python
def divide(dividend, divisor):
    import math
    return dividend / math.sqrt(divisor)

libtask = q.create_library_from_functions('test-library', divide)
s_task = vine.FunctionCall('test-library', 'divide', 2, 2**2)
q.submit(s_task)
```

We can also create a future function call in which the input arguments depend on the previously ones' outputs:
```python
def my_sum(x, y, negate=False):
    from operator import add, mul

    f = 1
    if negate:
        f = -1
    s = mul(f, add(x, y))
    return s

t1 = executor.task(my_sum, 3, 4)
t2 = executor.task(my_sum, 2, 5)
a = executor.submit(t1)
b = executor.submit(t2)
t3 = executor.task(my_sum, a, b)
c = executor.submit(t3)
```

I think it would be more helpful if we could integrate future function calls into the library (or implement a combination of both), utilizing both features for enhanced usability.

dthain:
This is a good idea, and the way to go is to change FutureTask so that it is not an extension of PythonTask, but rather is constructed with a pointer to either a FunctionCallTask or a PythonTask and forwards its methods as appropriate.

PR #3607, opened 2024-01-18
TaskVine: Future Function Call 

JinZhou5042:
## Proposed changes

An enhancement for #3568.

- Enabled FutureFunctionCall tasks, in accordance with the idea of FuturePythonTask by @BarrySlyDelgado   
- Modified the FunctionCall wrapper to add a auxiliary retriever function.
- Used a hash value as the name of input files in case an output file serves as input for multiple arguments.
- Performance evaluation is yet to be done.

## Post-change actions

## Additional comments


JinZhou5042:
Tasks can be submitted out of order (i.e., submit t2 or t3 before submitting t1) in some multithreaded environment, except the fact that this impacts the performance.

dthain:
@BarrySlyDelgado you are the expert on Futures, please review and comment so Jin can move forward.

taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
```diff
@@ -45,16 +49,34 @@ def __init__(self, port=9123, batch_type="local", manager=None, manager_host_por
...
             self.factory = None
 
-    def submit(self, fn, *args, **kwargs):
-        if isinstance(fn, FutureTask):
-            self.manager.submit(fn)
-            return fn._future
-        future_task = FutureTask(self.manager, False, fn, *args, **kwargs)
-        self.manager.submit(future_task)
-        return future_task._future
+    def submit(self, task):
```
BarrySlyDelgado, line 52: Here we want to keep the definition of submit to be (`(self, fn, *args, *kwargs)`). If the fn argument is actually of type `FuturePythonTask` or `FutureFunctionCall` then we just submit the task. if fn is a function we create a `FuturePythonTask` 

This is so that our executor conforms to [this](https://docs.python.org/3/library/concurrent.futures.html) so that we can use this in other applications that use futures.

taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
```diff
@@ -45,16 +49,34 @@ def __init__(self, port=9123, batch_type="local", manager=None, manager_host_por
...
+        if isinstance(task, FuturePythonTask):
+            self.manager.submit(task)
+            return task._future
+        if isinstance(task, FutureFunctionCall):
+            # submit the task
+            self.manager.submit(task)
+            # append it to the task table
+            self.task_table.append(task)
+            return
```
BarrySlyDelgado, line 61: When we submit a `FutureFunctionCall` it should return a `VineFuture`. So, a `FutureFunctionCall` should have a future field.

taskvine/test/vine_python_future_funcall.py
```diff
@@ -0,0 +1,52 @@
...
+# Create library task
+print("creating library from functions...")
+libtask = executor.create_library_from_functions('test-library', divide, import_modules=None, add_env=False)
+
+# Install library on executor.manager
+executor.install_library(libtask)
+
+# Submit several tasks for execution
+print("creating future function call tasks...")
+t1 = executor.future_funcall('test-library', 'divide', 128, 4)
```
BarrySlyDelgado, line 35: We want the argument here to just be `t1` rather than `t1.future`. some handling might have to be done to make that work.

BarrySlyDelgado (changes requested):
There should also be a test that tests both FutureFunctionCalls and FuturePythonTasks

dthain (approved):
Please make Barry's requested changes, and then we can go ahead and marge.  Looks like you need to merge in recent changes from master as well.

BarrySlyDelgado (approved):

JinZhou5042:
RTM if @dthain thinks it's good.

JinZhou5042:
Wait a moment. I tested the performance on Condor workers, the FutureFunctionCall does work, but tasks were seemingly sparsely run. 

Screenshot from 2024-02-03 14-41-31

Compared to submit task and retrieve output one by one:
Screenshot from 2024-02-03 14-51-15


I suspect this is because each task comes with an input depending on a prior task's output. One single task keeps idle until its dependent task puts generates output. As I submitted hundreds of tasks, 399 tasks kept idle until the first task was done, 398 tasks kept idle until the second task was done, and so on and so forth. Therefore, the worker is really taking heavy pressure to schedule hundreds of tasks. 

I think FuturePythonTask also has this effect, how do you think @BarrySlyDelgado?

BarrySlyDelgado:
What code did you use to run this test?

dthain:
For the record, the manager is not particularly optimized to deal with tasks whose dependencies are not satisfied.  They get considered every scheduling round, and tossed out when it is seen that their inputs are not ready.  (This is why we usually retain not-ready tasks outside of TaskVine and submit them when they are satisfied.)

FYI @colinthomas-z80

BarrySlyDelgado:
Also, you may want to look into testing a different workflow. From the plot, it looks sequential which isn't  conducive to good performance, but is not likely to solve this issue.

 I don't hypothesize that the manager scheduling tasks is the issue (at least at the order of 10s of seconds for 400 tasks)

JinZhou5042:
The posted one was run on FutureFunctionCall, I observed FuturePythonTask also has this effect.

This ran 300 tasks on regular PythonTask:
txn

Whereas the following one ran 300 tasks on FuturePythonTask:
txn

This experiment was run on the most recent cctools, without any modification in this PR. @BarrySlyDelgado 

I think at this moment it is enough to merge because it has been in accordance with the idea of Distant Future. Let's move on and take a deeper look at the performance issue later. @dthain

BarrySlyDelgado:
I was able to repeat these tests. It could be the time considering tasks in which inputs are not available. Currently, the time in `vine_manager_check_inputs_available` is not measured it may be worth measuring that.

dthain:
I am ok with proceeding to merge, if the performance problem existed prior to this PR.
(but then please make a new issue describing the outstanding problem.)
Please fix up the build problems so we can proceed...

PR #3612, opened 2024-01-23
Vine: Unlink Intermediate Files Automatically

btovar:
## Proposed changes

Adds the VINE_UNLINK_ON_RC0 flag to delete files from disk when the reference count goes to 0.
Intended only for internal taskvine usage.

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
@@ -168,7 +168,7 @@ class FutureTask(PythonTask):
             if not self._output_loaded and self._future_resolved:
                 if self.successful():
                     try:
-                        with open(os.path.join(self._tmpdir, self._out_name_file), "rb") as f:
+                        with open(os.path.join(self._tmpdir, "output"), "rb") as f:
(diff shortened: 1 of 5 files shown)


dthain (changes requested):
Yes, I think this is the right idea. 
1 - I am comfortable making this a user-visible feature, since I think this situation is come to come up again in the future.  Trying to think of names that make sense to the end user.  Maybe `UNLINK_WHEN_DONE`?  I'm open to ideas here.
2 - Add a `debug` when the file is actually unlinked, since that will have some non-trivial cost, and we want to know when it happens.

dthain:
I'm thinking `DONE` is nice because it's actually vague about when that happens, but makes it clear the user is giving the responsibility over to the manager...

dthain:
This makes sense to me, does it solve the problem with Connor's app?

dthain (approved):

btovar:
I'm testing it... Now it is a matter of getting the logic in the dag right when adding the checkpoints.

dthain:
RTM?

btovar:
Splitting into several prs.

PR #3634, opened 2024-02-01
vine: use a single file for wrappers, buffer for functions

btovar:
Previously TaskVine would generate wrapper and function file per task, even when the wrapper was the same for all the PythonTask, and the function was used by other tasks.

This pr makes it so that wrappers and functions are serialized once. Further, functions are not written to disc but kept as a declare_buffer.

## Proposed changes

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
@@ -182,13 +182,13 @@ class FutureTask(PythonTask):
     def add_future_dep(self, arg):
         self.add_input(arg._task._output_file, str('outfile-' + str(arg._task.id)))
 
-    def submit_finalize(self, manager):
+    def submit_finalize(self):
         func, args, kwargs = self._fn_def
         for arg in args:
             if isinstance(arg, VineFuture):
                 self.add_future_dep(arg)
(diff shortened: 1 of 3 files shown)


BarrySlyDelgado (approved):

dthain:
I think this is likely a win for most cases, provided that we are effective at GCing and removing the buffer.

However, I am wondering if there is any situation in Parsl or Dask where the state passed to a function is so large that we would prefer to keep it on disk.

dthain (commented):
I think this is likely a win for most cases, provided that we are effective at GCing and removing the buffer.

However, I am wondering if there is any situation in Parsl or Dask where the state passed to a function is so large that we would prefer to keep it on disk.

btovar:
@dthain Just added `m.undeclare_function(fn)` to remove the buffers created. Does that address your concern?

dthain:
Oh, I'm not concerned about accumulating a (small constant) number of files in the manager -- I don't think anyone will use `undeclared_function`.  I'm thinking more about whether there is any corner case in which using buffers is a negative relative to files.

I like this idea, but we need to press pause on it for a few reasons:
1 - @BarrySlyDelgado and @JinZhou5042 need to get #3581 and #3067 complete merged first, since they intersect.
2 - This will change the performance of our "base cases" on all the HEP applications.  Let's get those measurements completed first.

Once those are done, then I am ok moving ahead with this.

PR #3639, opened 2024-02-02
vine: force gc on undeclare

btovar:
Force outstanding references to the file (or tasks that use the file) to be collected. This to make unlink_when_done work when we expect it (or to some approximation).

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
@@ -44,6 +44,7 @@ import subprocess
 import sys
 import tempfile
 import time
+import gc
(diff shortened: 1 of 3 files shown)


dthain:
Pondering here a bit whether we want to have `gc.collect` inside one of our methods -- if the user is going to undeclare a lot of files, then they will want to do a gc once, not many times.  Perhaps the `gc.collect` belongs in the application?

btovar:
Yes, this might be a bit overkill. Another option I'm investigating is how to do the submit_finalize without the need of that cell object that gets created.

btovar:
Found better solution, submitting soon.

PR #3641, opened 2024-02-05
vine: avoid capturing variables in closurs with submit_finalize

btovar:
This avoids the creating of a cell object that increases the reference count to the task object, making the garbage collection (and therefore unlink_when_done) a little more predictable.

## Post-change actions

## Additional comments


dthain (approved):
Interesting, the refactoring seems unobjectionable to me, but I don't see how the reference is avoided.  I think we need a five minute tutorial on cell objects.  :)

Should this be sequenced before or after #3581 ?

btovar:
It would be easier if we merge it before, I think.

PR #3655, opened 2024-02-09
Vine Cache Metadata

dthain:
## Proposed changes

This is a rather extended cleanup of the manager-worker protocol to enable others to work on some future projects:

- At the worker, all transfers and minitasks now land in a transfer directory, where they are measured and evaluated, and if successful, moved into the cache directory.  The same code path is used for all cases, which should eliminate a number of potential failures and boundary cases that could result in corrupted data in the cache.
- Transfer failures are captured in a {cachename}.error file in the transfer directory, and then conveyed back to the manager through the existing `cache-invalid` message.
- The worker now stores a "meta" file alongside each cached object, containing metadata sent by the manager, including file type, caching mode, size, modification time, etc.  This allows the worker to make independent cleanup decisions such as "on shutdown, remove all files not marked as cache-forever"
- The worker sends this metadata back to the manager in cache-update messages.  Some of these fields were present before, but filled incorrectly on worker startup because the proper metadata was not recorded.
- Both the manager and the worker now evaluate cached objects based on clean information about file type and caching mode, rather than attempting to infer properties from the structure of the cached name.

One aspect of this that I am undecided on:

- In the TaskVine API, we convey caching information through the `vine_file_flags_t flags` parameter, which can have the following values:

```
typedef enum {
        VINE_CACHE_NEVER = 0,  /**< Do not cache file at execution site. (default) */
        VINE_CACHE = 1,        /**< File remains in cache until workflow ends. */
        VINE_CACHE_ALWAYS = 3, /**< File remains in cache until the worker teminates. **/
        VINE_PEER_NOSHARE = 4,  /**< Schedule this file to be shared between peers where available. See @ref vine_enable_peer_transfers **/
        VINE_UNLINK_WHEN_DONE = 8  /**< Whether to delete the file when its reference count is 0. (Warning: Only use on files produced by the application, and never on irreplaceable input files.) */
} vine_file_flags_t;
```

On one hand, this is nice and compact for the user to express.  Internally, it's a little clunky b/c `VINE_CACHE_ALWAYS` covers two bits, and so the check is a bit awkward.  And this makes it difficult to introduce new caching levels.  To simplify things, I wrote up a new type which is used internally between the manager and worker to express what (I think) are the various cache levels:

```
typedef enum {
        VINE_CACHE_LEVEL_TASK=0,
        VINE_CACHE_LEVEL_WORKFLOW=1,
        VINE_CACHE_LEVEL_WORKER=2,
        VINE_CACHE_LEVEL_FOREVER=3
} vine_cache_level_t;
```

This does make things a little easier, in that the manager can tell the worker "clean up all files with cache level less than 3".  But I am undecided whether this is worth the trouble to express in the taskvine API.   I see three options going forward:

- Keep the existing bit flags, just pass those from manager to worker, and live with the minor awkwardness.
- Keep the existing bit flags in the API, but convert them internally to `cache_level` and pass that between the manager and The worker.  (That's what this PR does now.)
- Throw out the existing bit flags and add a new explicit parameter `cache_level` to `vine_declare_file` and friends, and use it consistently end to end.  This would break compatibility at the C API, but it would not be hard to keep compatibility at the Python API.

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/manager/vine_file.c
+++ b/taskvine/src/manager/vine_file.c
@@ -75,7 +75,8 @@ int vine_file_delete(struct vine_file *f)
 
 /* Create a new file object with the given properties. */
 struct vine_file *vine_file_create(const char *source, const char *cached_name, const char *data, size_t size,
-		vine_file_type_t type, struct vine_task *mini_task, vine_cache_level_t cache_level, vine_file_flags_t flags)
+		vine_file_type_t type, struct vine_task *mini_task, vine_cache_level_t cache_level,
+		vine_file_flags_t flags)
 {
 	struct vine_file *f = xxmalloc(sizeof(*f));
 	memset(f, 0, sizeof(*f));
@@ -221,7 +222,8 @@ struct vine_file *vine_file_buffer(const char *data, size_t size, vine_cache_lev
 	return vine_file_create("buffer", 0, data, size, VINE_BUFFER, 0, cache, flags);
 }
 
-struct vine_file *vine_file_mini_task(struct vine_task *t, const char *name, vine_cache_level_t cache, vine_file_flags_t flags)
+struct vine_file *vine_file_mini_task(
+		struct vine_task *t, const char *name, vine_cache_level_t cache, vine_file_flags_t flags)
 {
 	flags |= VINE_PEER_NOSHARE; // we don't know how to share mini tasks yet.
 	return vine_file_create(name, 0, 0, 0, VINE_MINI_TASK, t, cache, flags);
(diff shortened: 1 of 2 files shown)


dthain:
This one is a monster PR, it should wait for the next release, and will require some testing with applications first.

dthain:
I think we are on the same page that the cache level is a distinct concept within the flags.  The question is really down to how to encode it in the API and pass it around internally.   Which of the following do you prefer?

1 - Keep the current API and keep all the information in one flags integer.
2 - Keep the current API, but behind the scenes, separate out flags and cache level into distinct fields.
3 - Change the API to separate flags and cache level from the very beginning.


1 and 2 would look like this:
```
vine_declare_file( source, VINE_CACHE|VINE_CACHE_ALWAYS|VINE_PEER_NOSHARE )
```

3 would look like this:
```
vine_declare_file( source, VINE_CACHE_LEVEL_WORKFLOW, VINE_PEER_NOSHARE )
```

dthain:
Ok, do you want to undertake that change?  API, Python, Examples, Manual, etc..

btovar:
Yes, I'll do it. It will be a good way to become more familiar with this changes.

btovar (approved):

Issue #3650, opened 2024-02-08
Vine: Buffer Asynchronous Messages


dthain:
(capturing a discussion with @BarrySlyDelgado from today.)

The current interaction between the manager and the worker is partially synchronous, so as to avoid a potential deadlock.

We do not allow the worker to send "large" streams of data to the manager, unless the manager has specifically requested it.  This is because the manager could simultaneously be sending data to the worker in the form of a task dispatch or a file transmission.  If the manager and worker start transmitting concurrently, the result is deadlock, because both would fill up their sides of the TCP stream, and not be reading the other side in response.   (This did happen with larger Work Queue applications until we figured it out and fixed it.)

As a result, the manager-worker protocol works as follows:
- Generally, the manager must initiate a conversation by sending a request to the worker, which may then reply with something large.  For example, the manager may send `get file` and the worker may respond with the data stream.
- The worker may send a handful of (small) unprompted asynchronous messages to the manager.  These can cause the manager to begin a synchronous conversation with the worker.  As long as the set of outstanding messages is smaller than the TCP buffer size (typically 64KB) then deadlock is impossible.

For example, the worker sends a brief `available_results` message to indicate that one or more tasks have completed.  Then the manager sends `get_results` and the worker responds with all of the details of each task, including its standard output, which could potentially be large.

Upside: Deadlock is impossible.

Downside: Fetching completed tasks is bounded by network latency, and shows up in gaps in execution when we have a large number of fast-running tasks.

Is there a way we can permit asynchronous updates without running the risk of deadlock?

dthain:
The way to approach this is to first imagine that the worker sends back complete updates asynchronously, and then consider the four requirements for deadlock, and attack one of them:
- Mutual Exclusion
- Hold and Wait
- Circular Wait
- No Preemption

dthain:
As Barry suggested, a way to explore this  idea would be to just make a "simple" task completion an asynchronous message, and drop stdout completely.  Evaluate the performance of that, and if there is a big improvement, then make stdout handling more "normal" as part of the task.

dthain:
Ok, a plan for how to proceed on this:

The manager worker protocol really has two components:
1 - Synchronous exchanges where the manager sends a request, and the worker responds to that request.  Things like `get` a file, `send_results` and so forth.  Because these can be large streams of data, the worker must not send them unprompted.
[...]

PR #3669, opened 2024-02-20
Vine: Buffer Asynchronous Messages From Worker for Async Task Completion

BarrySlyDelgado:
## Proposed changes

For #3650

This PR creates a buffer list of asynchronous messages to be sent to the manager. `deliver_async_messages` sends pending asynchronous messages when available (when a message does not overflow the output buffer). Currently once each time in the worker's main loop. Additionally, it is called whenever a call to send_async_message is called (Right now I think it may be better to do this so the worker doesn't sit on messages). Also, added asynchronous task completion messages but still need to work out how this should be handled.  Also, swapped some of the calls of `send_message` to `send_async_message`

## Post-change actions

## Additional comments


taskvine/src/worker/vine_worker.c
```diff
@@ -208,6 +211,66 @@ __attribute__((format(printf, 2, 3))) void send_message(struct link *l, const ch
 	va_end(va);
 }
 
+/* Send messages from list of asychronous messages available*/
+
+void deliver_async_messages(struct link *l)
+{
+	int bytes_in_buffer = get_link_buffer_bytes(l);
+	int bytes_available = VINE_MAX_BUFFER_SIZE - bytes_in_buffer;
```
dthain, line 219: IMHO, `VINE_MAX_BUFFER_SIZE` is not a protocol constant, but the known size of the TCP window size.  This is 64KB by default but can be set/get with `link_window_set/get`.

taskvine/src/worker/vine_worker.c
```diff
@@ -208,6 +211,66 @@ __attribute__((format(printf, 2, 3))) void send_message(struct link *l, const ch
...
+	int messages = list_size(pending_async_messages);
+	int visited;
+	char *message;
+
+	for (visited = 0; visited < messages; visited++) {
```
dthain, line 224: Hmm, I don't think you want to mix up the order of the messages.
Just peek the head and send it if there is space available.
If there isn't room for the first item in the queue, stop sending altogether.

taskvine/src/manager/vine_manager.c
```diff
@@ -530,6 +530,8 @@ static vine_msg_code_t vine_manager_recv_no_retry(
 		result = handle_transfer_address(q, w, line);
 	} else if (sscanf(line, "GET %s HTTP/%*d.%*d", path) == 1) {
 		result = handle_http_request(q, w, path, stoptime);
+	} else if (string_prefix_is(line, "complete")) {
```
dthain, line 533: Handle asynchronous task completion messages here.

dthain (commented):
This is on the right track!

dthain:
Sorry Barry, you will need to rebase on this one.

taskvine/src/worker/vine_worker.c
```diff
@@ -556,6 +626,20 @@ static void reap_process(struct vine_process *p, struct link *manager)
 
 	itable_remove(procs_running, p->task->task_id);
 	itable_insert(procs_complete, p->task->task_id, p);
+
+	int output_file = open(p->output_file_name, O_RDONLY);
+	if (output_file >= 0) {
```
dthain, line 631: formatting

taskvine/src/worker/vine_worker.c
```diff
@@ -602,9 +686,9 @@ static void report_tasks_complete(struct link *manager)
 {
 	struct vine_process *p;
 
-	while ((p = itable_pop(procs_complete))) {
-		report_task_complete(manager, p);
-	}
+	//while ((p = itable_pop(procs_complete))) {
```
dthain, line 689: be bold and just delete the code!

taskvine/src/manager/vine_manager_get.c
```diff
@@ -355,6 +356,134 @@ vine_result_code_t vine_manager_get_single_file(struct vine_manager *q, struct v
 	return vine_manager_get_buffer(q, w, 0, f, &total_bytes);
 }
 
+
```
dthain, line 359: Can we simplify fetching of stdout to just do a normal `get` in the usual way?
-

taskvine/src/manager/vine_manager.c
```diff
@@ -1084,6 +1277,122 @@ static void resource_monitor_compress_logs(struct vine_manager *q, struct vine_t
...
+
+	w->finished_tasks--;
+	w->total_tasks_complete++;
+
+	// At least one task has finished without triggering a slow worker disconnect, thus we
+	// now have evidence that worker is not slow (e.g., it was probably the
+	// previous task that was slow).
+	w->alarm_slow_worker = 0;
+	
+	/* print warnings if the task ran for a very short time (1s) and exited with common non-zero status */
```
dthain, line 1349: Blech, don't duplicate code!

dthain (changes requested):
Main thing: I don't want to have two cases of "light" tasks and "regular" tasks.  We should not be changing the meaning of tasks here, just changing the protocol by which the manager and worker communicate with each other.  In short, the worker sends `complete` messages immediately when a task is done, and then the manager retrieves the outputs in the normal way.

taskvine/src/manager/vine_manager.c
```diff
@@ -501,6 +505,167 @@ static int handle_transfer_hostport(struct vine_manager *q, struct vine_worker_i
...
+
+	if (task_status != VINE_RESULT_SUCCESS) {
+		w->last_failure_time = timestamp_get();
+		t->time_when_last_failure = w->last_failure_time;
+	}
+
+	/* If the task was forsaken by the worker or couldn't exeute, it didn't really complete, so short circuit. */
+	if (task_status == VINE_RESULT_FORSAKEN) {
+		t->forsaken_count++;
+		itable_remove(q->running_table, t->task_id);
```
dthain, line 609: It looks like this `itable_remove`, `vine_task_set_result` and `change_task_state` are a common code path with the final few lines of this function.  Is there a way to simplify or combine here, so that we don't have duplicated paths through this function?  (I'm thinking about the stability of future edits to this code.)

taskvine/src/manager/vine_manager.c
```diff
@@ -501,6 +505,167 @@ static int handle_transfer_hostport(struct vine_manager *q, struct vine_worker_i
...
+	t->time_workers_execute_last =
+			observed_execution_time > execution_time ? execution_time : observed_execution_time;
+	t->time_workers_execute_last_start = start_time;
+	t->time_workers_execute_last_end = end_time;
+	t->time_workers_execute_all += t->time_workers_execute_last;
+	t->output_length = output_length;
+	t->result = task_status;
+	t->exit_code = exit_status;
+
+	/* If output is less than 1KB stdout is sent along with completion msg. retrieve it from the link. */
```
dthain, line 627: The system will be more robust with an explicit indicator rather than an implied assumption of 1024 bytes.
Suggest you change the message so that one field indicates "how many bytes immediately follow this message" and another indicates "what is the total size of the stdout."  And then this code can act directly on that information.

taskvine/src/worker/vine_worker.c
```diff
@@ -208,6 +212,90 @@ __attribute__((format(printf, 2, 3))) void send_message(struct link *l, const ch
 	va_end(va);
 }
 
+/* Send messages from list of asychronous messages available*/
```
dthain, line 215: Please add here a big long explanation of why asynchronous messages are needed, how they work, and what the constraints are, so that the next person reading this will get what's going on.

taskvine/src/worker/vine_worker.c
```diff
@@ -1121,6 +1168,29 @@ static void disconnect_manager(struct link *manager)
 	}
 }
 
+void send_stdout(struct link *l, int64_t task_id)
+{
+	struct vine_process *p = itable_lookup(procs_table, task_id);
+	if (!p) {
+		send_message(l, "failure %" SCNd64 "\n", task_id);
```
dthain, line 1175: Please make this correspond in format to the `error` message used by `vine_transfer_put/get`.  Or even better, is it possible to use the same code?

taskvine/src/worker/vine_worker.c
```diff
@@ -552,62 +640,21 @@ static void reap_process(struct vine_process *p, struct link *manager)
...
-
-static void report_tasks_complete(struct link *manager)
-{
-	struct vine_process *p;
-
-	while ((p = itable_pop(procs_complete))) {
-		report_task_complete(manager, p);
-	}
-
-	vine_watcher_send_changes(watcher, manager, time(0) + options->active_timeout);
```
dthain, line 605: Hmm, I think this was lost without replacement.  The manager still needs a way to request updates re watched files.  Is `vine_watcher_send_changes` still called from somewhere?

dthain (changes requested):
Barry this is coming along nicely.  A few things to tighten up still.

taskvine/src/manager/vine_manager.c
BarrySlyDelgado: Moved calls to bottom of function

taskvine/src/worker/vine_worker.c
BarrySlyDelgado: changed format to correspond with other error messages.

taskvine/src/worker/vine_worker.c
BarrySlyDelgado: re-added updates for watched files.

taskvine/src/worker/vine_worker.c
```diff
@@ -552,62 +662,21 @@ static void reap_process(struct vine_process *p, struct link *manager)
...
-
-/*
-Transmit the results of the given process to the manager.
-*/
-
-static void report_task_complete(struct link *manager, struct vine_process *p)
-{
-	int64_t output_length;
 
 	int output_file = open(p->output_file_name, O_RDONLY);
```
dthain, line 666: No longer necessary to `open` the file, just use `stat` here.

taskvine/src/manager/vine_task.h
```diff
@@ -81,6 +81,8 @@ struct vine_task {
 
 	vine_result_t result;          /**< The result of the task (see @ref vine_result_t */
 	int exit_code;               /**< The exit code of the command line. */
+	int output_recieved;          /**< If the stdout of the task has been recieved. */
```
dthain, line 84: `received`

taskvine/src/worker/vine_worker.c
```diff
@@ -208,6 +212,112 @@ __attribute__((format(printf, 2, 3))) void send_message(struct link *l, const ch
 	va_end(va);
 }
 
+/*
```
dthain, line 215: Bump the number in `vine_protocol.h` since you are changing the on-the-wire protocol.

taskvine/src/manager/vine_manager.c
```diff
@@ -884,6 +1058,7 @@ void vine_manager_remove_worker(
 
 	hash_table_remove(q->worker_table, w->hashkey);
 	hash_table_remove(q->workers_with_available_results, w->hashkey);
```
dthain, line 1060: Now `workers_with_available_results` really means specifically `workers_with_watched_file_updates` so please call it that throughout.

dthain (changes requested):
Almost there, a few last fixups for clarity.

taskvine/src/worker/vine_worker.c
```diff
@@ -663,15 +663,9 @@ static void reap_process(struct vine_process *p, struct link *manager)
...
-	if (output_file >= 0) {
-		struct stat info;
-		fstat(output_file, &info);
-		p->output_length = info.st_size;
-		close(output_file);
-	} else {
-		p->output_length = 0;
-	}
+	struct stat info;
+	stat(p->output_file_name, &info);
```
dthain, line 667: Still have to check for success!

taskvine/src/manager/vine_task.c
```diff
@@ -99,6 +100,7 @@ void vine_task_clean(struct vine_task *t)
 
 	free(t->output);
 	t->output = NULL;
+	t->output_length = 0;
```
dthain, line 103: Is it necessary to reset output_received here as well?

BarrySlyDelgado:
Asynchronous completion works with DV3 though produces little benefit over normal execution.
Here is an execution of DV3 with 100 12-core workers using the current master branch.
master2
Here is an execution of DV3 with the same worker configurations using asynchronous task completion.
async2

For clarity, the light blue "task execution" depicts the time the task executes on the worker. "task running" depicts when the manager sends a task to a worker to when the manager acknowledges its completion.  We observe sizeable gaps between "actual" task execution at this scale. 

Here is the same workflow exciting with 50 workers:
async2_50

While this is completed more slowly, we observe less overhead.

dthain:
Per our discussion today, please plot the manager time distribution plot -- hypothesis is that the manager is spending time iterating over tasks even when there is nothing to be accomplished.

BarrySlyDelgado:
Not sure how much there is to garner from the manager time plot (TOP) synchronous (BOTTOM) asynchronous:
pf_dv3_100j_run-5 times-stacked
pf_dv3_100j_run-8 times-stacked

It seems most of the time is spent in the application. @btovar wad this a similar thing that you were seeing with Kelci's applications?

dthain:
Related to #3849?

dthain:
Based on our discussion so far, I am pretty confident that the asynchronous updates are working.

And now we are working on a separate issue to improve the task materialization time.

I am inclined to merge this one and start a new issue.

What do you think?

taskvine/src/manager/vine_manager.c
```diff
@@ -501,6 +505,174 @@ static int handle_transfer_hostport(struct vine_manager *q, struct vine_worker_i
...
+			warn(D_VINE, "This usually means that the task's command had a segmentation fault,\n");
+			warn(D_VINE, "either because it has a memory access error (segfault), or because\n");
+			warn(D_VINE, "it uses a version of a shared library different from the one at the worker.\n");
+			break;
+		default:
+			break;
+		}
+	}
+
+	vine_task_info_add(q, t);
```
btovar, line 543: Move this one just after `exit_alert` is called? It does not really fit in the `exit_alert` function.

taskvine/src/manager/vine_manager.c
```diff
@@ -1174,32 +1331,15 @@ static int fetch_output_from_worker(struct vine_manager *q, struct vine_worker_i
...
 
-	switch (t->result) {
-	case VINE_RESULT_INPUT_MISSING:
-	case VINE_RESULT_FORSAKEN:
-		/* do not count tasks that didn't execute as complete, or finished tasks */
-		break;
-	default:
-		w->finished_tasks--;
-		w->total_tasks_complete++;
+	w->finished_tasks--;
```
btovar, line 1336: Is this function ever called for `FORSAKEN` tasks? If so, they should not be counted as finished tasks.

taskvine/src/manager/vine_manager.c
```diff
@@ -1235,12 +1375,15 @@ static int fetch_output_from_worker(struct vine_manager *q, struct vine_worker_i
 		}
 	}
 
-	/* XXX: temp fix. Make this a tunable parameter and/or make it more sophisticated */
-	if (w->forsaken_tasks > VINE_DEFAULT_MAX_FORSAKEN_PER_WORKER && w->total_tasks_complete == 0) {
-		debug(D_VINE, "Disconnecting worker that keeps forsaking tasks %s (%s).", w->hostname, w->addrport);
-		handle_failure(q, w, t, VINE_WORKER_FAILURE);
-		return 0;
-	}
+	vine_task_info_add(q, t);
```
btovar, line 1378: Isn't this part of the `exit_alert` above?

taskvine/src/manager/vine_task.c
```diff
@@ -99,6 +100,8 @@ void vine_task_clean(struct vine_task *t)
 
 	free(t->output);
```
btovar, line 101: rename all `t->output` and `t->output_length` to `t->std_output`? It gets confusing when tasks have other outputs.

btovar (changes requested):

taskvine/src/manager/vine_task.c
dthain: Actually @JinZhou5042 just made a similar change in a different pending PR, I would recommend against also changing that here, just to avoid further conflicts.

JinZhou5042: Do we use `stdout`/`stdout_length` or `std_output`/`std_output_length`?

taskvine/src/manager/vine_manager.c
BarrySlyDelgado: `exit_alert` is removed as it was not called.

taskvine/src/manager/vine_manager.c
BarrySlyDelgado: this one remains as `exit_alert` is not called.

taskvine/src/manager/vine_manager.c
BarrySlyDelgado: Fixed to reflect previous handling.

taskvine/src/manager/vine_task.c
BarrySlyDelgado: Remains unchanged to avoid conflicts.

dthain (approved):

dthain:
@btovar any last thoughts on this one?

taskvine/src/manager/vine_manager.c
btovar: Right, I think the pr moved that code to the function, and then the function was never called.

btovar (approved):

dthain:
Ok, merge when ready!

PR #3690, opened 2024-03-05
vine: consider cached output buffers as replicas at workers

btovar:

## Additional comments

### DIFF
--- a/taskvine/src/manager/vine_manager_get.c
+++ b/taskvine/src/manager/vine_manager_get.c
@@ -426,7 +426,7 @@ vine_result_code_t vine_manager_get_output_file(struct vine_manager *q, struct v
 	// If the transfer was successful, make a record of it in the cache.
 	if (result == VINE_SUCCESS && (f->cache_level > VINE_CACHE_LEVEL_TASK)) {
 		struct stat local_info;
-		if (stat(f->source, &local_info) == 0) {
+		if (f->type == VINE_BUFFER || stat(f->source, &local_info) == 0) {
 			struct vine_file_replica *replica = vine_file_replica_create(
 					f->type, f->cache_level, local_info.st_size, local_info.st_mtime);
 			vine_file_replica_table_insert(w, f->cached_name, replica);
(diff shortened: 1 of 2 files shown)


taskvine/src/manager/vine_manager_get.c
```diff
@@ -426,12 +426,12 @@ vine_result_code_t vine_manager_get_output_file(struct vine_manager *q, struct v
 	// If the transfer was successful, make a record of it in the cache.
 	if (result == VINE_SUCCESS && (f->cache_level > VINE_CACHE_LEVEL_TASK)) {
 		struct stat local_info;
-		if (stat(f->source, &local_info) == 0) {
+		if (f->type == VINE_BUFFER || stat(f->source, &local_info) == 0) {
 			struct vine_file_replica *replica = vine_file_replica_create(
 					f->type, f->cache_level, local_info.st_size, local_info.st_mtime);
```
dthain, line 431: `local_info.st_size` will be uninitialized here in the case of a buffer.

However, `f->size` was already set in `vine_manager_get_buffer` and I'm sure the information is also already available in `vine_manager_get_file`.  Maybe the `stat` here is unnecessary if `f->size` has been set in all prior cases?

dthain (changes requested):

taskvine/src/manager/vine_manager_get.c
btovar: Good catch. Let met see about f->size.

btovar: Actually the st._size would be wrong for directories too.

dthain (approved):

btovar:
vine_file_has_changed uses  st_size.  This is an issue for directories because we would like to have f->size to be the total size of the files inside the directory for scheduling purposes, but we don't want to stat every time all its contents every time to check if it has changed.


One easy fix is to check if the payload is a directory, and if so, drop the check against st_mtime.

@dthain

dthain (approved):

Issue #3706, opened 2024-03-12
Vine: Track Last Failure Time


dthain:
In certain pathological circumstances, Vine will attempt to retry the same transfer of a file multiple times rapidly, even though it fails.  We see this situation when running RSTriPhoton:

- Worker A has the only available copy of file F.
- Manager tells Worker B to fetch F from A and run task T1.
- Worker A fails in someway, but the manager doesn't see it yet.
- Worker B reports back that the transfer failed, and task T1 fails with `INPUT_MISSING`
- Manager assigns another task to worker B and tells it to fetch from A.
- This process repeats for some time until the manager finally runs out of things to do, and notices that worker A fails.

So, the solution is this:
- Each worker should have a `last_transfer_failure` field.
- Upon receiving cache-invalid, the manager should set `last_transfer_failure` of the source worker to the current time.
- When selecting replicas, the manager should skip over replicas on workers that failed within the last N seconds.   (configurable).

The effect of this should be to "slow down" when failures are encountered, so that the manager will go do other things instead and "catch up" on failure information.

dthain:
@colinthomas-z80 can you take a crack at this one?

colinthomas-z80:
Sure!

colinthomas-z80:
#3745

PR #3709, opened 2024-03-12
Vine: record last transfer failure

colinthomas-z80:
## Proposed changes

Setting up work to count file transfer failures in order to avoid the stall issues being observed.

The first step is to remap the current transfer table from worker->source to worker->worker, since we want to attach a failed transfer to a specific worker we need to keep a reference to it when the cache-invalid comes back.

## Post-change actions

## Additional comments

### DIFF
--- a/dttools/src/s3_file_io.c
+++ b/dttools/src/s3_file_io.c
@@ -422,17 +422,17 @@ static int s3_do_check ( char * const signature,
 
 static char* __aws_sign ( char * const str )
 {
-  HMAC_CTX ctx;
+  HMAC_CTX *ctx;
   unsigned char MD[256];
   unsigned len;
 
   __debug("StrToSign:%s", str );
 
-  HMAC_CTX_init(&ctx);
-  HMAC_Init(&ctx, awsKey, strlen(awsKey), EVP_sha1());
-  HMAC_Update(&ctx,(unsigned char*)str, strlen(str));
-  HMAC_Final(&ctx,(unsigned char*)MD,&len);
-  HMAC_CTX_cleanup(&ctx);
+  ctx = HMAC_CTX_new();
+  HMAC_Init(ctx, awsKey, strlen(awsKey), EVP_sha1());
+  HMAC_Update(ctx,(unsigned char*)str, strlen(str));
+  HMAC_Final(ctx,(unsigned char*)MD,&len);
+  HMAC_CTX_free(ctx);
 
   char * b64 = __b64_encode (MD,len);
   __debug("Signature:  %s", b64 );


colinthomas-z80:
@dthain @btovar I wanted to put this out here in case it is conflicting with any of your current work. 

Right now I believe I made the necessary modifications to remap the current transfer table and keep everything working

taskvine/src/manager/vine_file_replica_table.c
```diff
@@ -135,18 +138,16 @@ struct vine_worker_info **vine_file_replica_table_find_replication_targets(
 		if (!peer->transfer_port_active)
 			continue;
 
-		char *peer_addr = string_format("worker://%s:%d", peer->transfer_addr, peer->transfer_port);
 		if (!(replica = hash_table_lookup(peer->current_files, cachename)) &&
 				(strcmp(w->hostname, peer->hostname))) {
-			if ((vine_current_transfers_worker_in_use(q, peer_addr) < q->worker_source_max_transfers) &&
+			if ((vine_current_transfers_source_in_use(q, peer) < q->worker_source_max_transfers) &&
```
btovar, line 143: shouldn't one of these refer to the source, and the other to the destination?

btovar: Shouldn't source_in_use use worker instead of peer?

In the previous code I was getting tons of transient errors with destinations not able to contact sources. If instead of peer_addr I use worker_addr, then those errors went away. I think that the sources were being overwhelmed.

dthain:
Linking to #3706

taskvine/src/manager/vine_file_replica_table.c
colinthomas-z80: yeah you are right. We should consider worker for dest_in_use(). I think that was an existing bug

taskvine/src/manager/vine_current_transfers.c
```diff
@@ -11,27 +11,27 @@ See the file COPYING for details.
 
 struct vine_transfer_pair {
 	struct vine_worker_info *to;
-	char *source;
+	void *source;
```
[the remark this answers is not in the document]
colinthomas-z80: I used void * so it could still point to a char * for a static url file. eg. not a worker.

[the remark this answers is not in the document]
colinthomas-z80: We then determine the type by seeing if the worker_source field has been set in the vine_file object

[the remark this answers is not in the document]
dthain: Too clever, I think it's better to clearly distinguish the use cases with proper types to avoid strange accidents, something like:
```
struct vine_worker_info *source_worker;
char *source_url;
```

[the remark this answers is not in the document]
btovar: Or we can have two different functions as we seem to know whether the transfer comes from a worker or a generic URL.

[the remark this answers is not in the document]
colinthomas-z80: The operations on the different sources are separate with the exception of adding the transfer to the table in vine_manager_put_url where we needed to switch on the type of source. I am ok with having two separate fields it shouldn't be too cumbersome

taskvine/src/manager/vine_manager.c
```diff
@@ -3674,6 +3674,7 @@ struct vine_manager *vine_ssl_create(int port, const char *key, const char *cert
 	q->minimum_transfer_timeout = 60;
 	q->transfer_outlier_factor = 10;
 	q->default_transfer_rate = 1 * MEGABYTE;
+	q->transfer_retry_interval = 5;
```
btovar, line 3677: `q->last_failure_retry_interval = DEFAULT_LAST_FAILURE_RETRY_INTERVAL;`  and `#define` accordingly.
We can use the same value for the task retries, etc.

taskvine/src/manager/vine_manager.c
```diff
@@ -5083,6 +5084,9 @@ int vine_tune(struct vine_manager *q, const char *name, double value)
 	} else if (!strcmp(name, "min-transfer-timeout")) {
 		q->minimum_transfer_timeout = (int)value;
 
+	} else if (!strcmp(name, "transfer-retry-interval")) {
```
btovar, line 5087: Rename as above.

doc/manuals/taskvine/index.md
```diff
@@ -764,6 +764,10 @@ If peer transfers have been disabled, they may be re-enabled accordingly:
     vine_enable_peer_transfers(m);
     ```
 
+Transfers between workers may be impacted by transient issues which may cause intermittent transfer failures. In these situations we take note of the 
+failure that occured, and avoid using the same worker as a source for a period of time. This time period has a default value of 10 seconds.
+It may be changed by the user using `vine_tune` with the parameter `last-failure-retry-interval`.
```
btovar, line 769: In the current master is `transient-error-interval`

doc/manuals/taskvine/index.md
```diff
@@ -2415,6 +2419,7 @@ change.
 | max-retrievals | Sets the max number of tasks to retrieve per manager wait(). If less than 1, the manager prefers to retrieve all completed tasks before dispatching new tasks to workers. | 1 |
 | prefer-dispatch | If 1, try to dispatch tasks even if there are retrieved tasks ready to be reportedas done. | 0 |
 | worker-retrievals | If 1, retrieve all completed tasks from a worker when retrieving results, even if going above the parameter max-retrievals . Otherwise, if 0, retrieve just one task before deciding to dispatch new tasks or connect new workers. | 1 |
+| last-failure-retry-interval | Time to wait in seconds after a resource failure before attempting to use it again | 5 |
```
btovar, line 2422: Already in the manual (as transient-error-interval), no need to add it again.

taskvine/src/manager/taskvine.h
```diff
@@ -1247,6 +1247,7 @@ void vine_set_manager_preferred_connection(struct vine_manager *m, const char *p
  - "monitor-interval" Parameter to change how frequently the resource monitor records resource consumption of a task in a times series, if this feature is enabled. See @ref vine_enable_monitoring.
  - "update_interval"  Seconds between updates to the catalog. (default=60)
  - "temp-replica-count" Degree of replication across workers for remote temp files (default=0)
+ - "last-failure-retry-interval" Time to wait in seconds after a resource failure before attempting to use it again (default=10)
```
btovar, line 1250: Rename to "transient-error-interval"

taskvine/src/manager/vine_current_transfers.c
```diff
@@ -11,33 +11,37 @@ See the file COPYING for details.
...
 
-static struct vine_transfer_pair *vine_transfer_pair_create(struct vine_worker_info *to, const char *source)
+static struct vine_transfer_pair *vine_transfer_pair_create(
+		struct vine_worker_info *to, struct vine_worker_info *source_worker, const char *source_url)
 {
 	struct vine_transfer_pair *t = malloc(sizeof(struct vine_transfer_pair));
 	t->to = to;
-	t->source = strdup(source);
+	t->source_worker = source_worker;
+	t->source_url = source_url ? strdup(source_url) : 0;
```
btovar, line 24: Let's use xxstrdup for consistency with the rest of the code.

taskvine/src/manager/vine_manager.c
```diff
@@ -99,6 +99,9 @@ See the file COPYING for details.
 /* Default value for how frequently to check for tasks that do not fit any worker. */
 #define VINE_LARGE_TASK_CHECK_INTERVAL 180000000 // 3 minutes in usecs
 
+/* Default value for how long to avoid a resource after a failure. */
+#define VINE_DEFAULT_LAST_FAILURE_RETRY_INTERVAL 10 // 10 seconds
```
btovar, line 103: Already as `VINE_DEFAULT_TRANSIENT_ERROR_INTERVAL`

btovar (changes requested):
Probably you will need to also modify `vine_file_replica_table_replicate` in current master.

colinthomas-z80:
@btovar I am a little worried this is getting stale with all the other retry_interval changes. Do you think this PR is still in line with the general idea? Should anything be changed?

PR #3728, opened 2024-03-19
vine: increment peer transfer initial timeout to 60s

btovar:
In some runs I am getting more than expected cache-invalid messages because the worker couldn't contact the peer. However, the peer is healthy as it subsequently executes tasks and is able to serve as source of files. It seems that the timeout of 15s is a bit harsh.

For duplication of files this is not such a big deal, but for input files (e.g. puturl), these failures are expensive as the rest of the input files of the respective tasks are cleaned up.

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/worker/vine_cache.c
+++ b/taskvine/src/worker/vine_cache.c
@@ -463,7 +463,7 @@ static int do_worker_transfer(
 	sscanf(f->source, "worker://%99[^:]:%d/%s", addr, &port_num, source_path);
 	debug(D_VINE, "cache: setting up worker transfer file %s", f->source);
 
-	stoptime = time(0) + 15;
+	stoptime = time(0) + 60;
 	worker_link = link_connect(addr, port_num, stoptime);
 
 	if (worker_link == NULL) {


dthain (commented):
This may be the solution, but let's look at the bigger picture to make sure that the pieces fit together.

Note that `worker/vine_transfer_server.c` accepts up to `VINE_TRANSFER_PROC_MAX_CHILD` (8) connections, after which requests will queue up.  That could be the cause of delays, although the manager should be limiting concurrency so that we don't hit that number.

Also note there that the worker expects the requester to send a request within 5 seconds after connecting, otherwise it disconnects.

The initial idea of the code was to "fail fast" rather than getting stuck in transfers.  But you make a good point that transfer failures can have large cascading effects.

btovar:
Got it. Many of the issues will become clear with @colinthomas-z80 pr. At least it will be easier to see why the manager is overloading workers with replica transfers.

dthain:
Isn't that the purpose of `vine_manager_transfer_capacity_available` ?

btovar:
Yes, you are correct.

btovar:
Actually, it doesn't check for temp files, only url and mini tasks.

dthain (approved):

dthain:
RTM or close?

PR #3729, opened 2024-03-19
vine: adds a transient result failure to tasks

btovar:
When an input file transfer between workers fails, the task is returned to the manager with VINE_RESULT_TRANSFER_MISSING.  Tasks with that result are considered to reschedule after VINE_DEFAULT_TRANSIENT_ERROR_INTERVAL (15 s).

With such a change, my DaskVine test runs had all task retries inside taskvine proper. DaskVine didn't need to resubmit a function for retry.

## Post-change actions

## Additional comments


btovar:
This one differs from #3709 as it deals with failures of tasks in general. #3709 deals with last time of failure for replicas.

dthain:
I actually got started writing very similar code but then got stuck with the following two thoughts:

1 - This doesn't distinguish between "couldn't transfer" and "doesn't exist".  And so it's going to be hard to get [this test case](https://github.com/cooperative-computing-lab/cctools/blob/686630b79ca6fe1fe609bdd0535edd67893557ae/taskvine/test/vine_python.py#L267) to fail correctly, unless we start digging down deeper into the transfer details.  Unless we are willing to allow curl transfers to be less robust than worker-worker transfers.

2 - I'm a little worried about introducing a new failure case into the manager, which already seems a bit brittle relative to the number of possible task result types.  One could argue that the task itself didn't fail, just the transfer (already communicated by a cache-invalid) and so the task should be returned to the manager as `FORSAKEN` because the conditions for its execution are no longer satisfied.  At some point, when we consider a task for scheduling, if its input file has been "tried enough" then the task can be returned with an appropriate error.  This would also cut down on unnecessary retries if the same input file is shared among a large number of tasks.

What do you think?

btovar:
Some ideas:
- I think we still want a state different than `FORSAKEN`, as `FORSAKEN` should be treated as a transient error. The application should never really deal with such state. Also `FORSAKEN` should never count as a retry for the task.
- cache-invalid messages do not have information about the tasks they affect (there might be more than one task affected). 
- Replicas refer to a single resource. We don't want to have a limit on `vine_file` because for popular files the chance of many replicas failing is large.
- On `FORSAKEN` we would need to check whether all replicas have exhausted their retries, etc., which I think may get messy.

With the current pr, once the task hits the maximum number of retries, if the last retry is `TRANSFER_MISSING` the return status can be changed to `INPUT_MISSING`. This would respect the spirit of the test with the url that does not exist.

btovar:
I added a pr to show how that would look like. The two issues I don't understand very well:

1. How do we blame a file for its failing replica attempts? It seems that an external url should be blamed more harshly than a worker subject to eviction.
2. How do we set limits on retries? It is easy to think that a task should be retried n times, or that a replica should be given up after n times. It is not so clear when to give up on a file if some of its replicas work some of the time. (If none of the replica attempts have ever worked, that may be easy.)

This pr takes the easy way and puts a limit on the tasks, and lets @colinthomas-z80 pr set a limit on particular replicas. (But it is the easy way, and probably there are more correct ways of doing it.)

dthain:
Ok, I see your point that the failure belongs to the **replica** not to the **file**.   Given that, I think it makes sense to count a failed transfer as a retryable failure, and then return the task after N failures.  (It does occur to me now that a task might experience different kinds of failures, and only the last one becomes the "reason" for the failure.)

dthain (changes requested):
The general approach makes sense to me, but please make `VINE_DEFAULT_TRANSIENT_ERROR_INTERVAL` a tunable parameter.

btovar:
An important issue with this pr is that if url is not accessible, then the task will be retried forever by default. I think that for puturl_now between workers we do want to retry indefinitely, but not for puturl. How about treating puturl as input missing?

dthain:
Hmm, I'm concerned about making the worker actions too tailored to a specific manager purpose.  Broadly speaking, the worker operations are pretty simple and general, and the manager puts them together in complex ways.  Both `puturl` and `puturl_now` really do the same thing, but only differ in the fact that the former is done lazily and the latter is done eagerly.  Both can be used to access things via `curl` or fetch from the worker.

So, it seems uncomfortable to me to give those two things different semantics.

You might convince me that both `puturl` and `puturl_now` should have different semantics for worker fetches versus http fetches, though.

btovar:
Ah yes, I keep thinking that `puturl_now` is only for temp file worker fetches. Maybe it is about responsibility? A worker transfer failure is taskvine responsibility, an http failure is the app responsibility.

PR #3740, opened 2024-03-22
copy manager addrs with DOMAIN_NAME_MAX limit

btovar:
## Proposed changes

## Post-change actions

## Additional comments

### DIFF
--- a/dttools/src/interfaces_address.c
+++ b/dttools/src/interfaces_address.c
@@ -29,7 +29,7 @@ struct jx *interfaces_of_host()
 {
 	struct ifaddrs *ifa, *head_if;
 
-	char host[IP_ADDRESS_MAX];
+	char address[IP_ADDRESS_MAX];
 
 	if (getifaddrs(&head_if) == -1) {
 		warn(D_NOTICE, "Could not get network interfaces information: %s", strerror(errno));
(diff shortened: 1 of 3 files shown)


dthain:
Something still doesn't add up there.
Maybe it's just the names of variables.
Anything `addr` should refer to an Internet address, IPV4 or IPV6.
Anything `host` or `name` should refer to a domain name.

```
struct manager_address {
        char host[DOMAIN_NAME_MAX];
        int port;
        char addr[DOMAIN_NAME_MAX];
};
```

The `host` field is correct but the `addr` field should be `LINK_ADDR_MAX`.

And then this line is confusing because it appears to be copying a interface address into a domain name:
```
strncpy(m->host, ifa_addr, DOMAIN_NAME_MAX - 1);
```

But `ifa_addr` comes from `jx_lookup_string(ifa,host)`

I think we are getting confused b/c variables names do not clearly reflect whether something is a domain name or an address.

btovar:
Right, we are copying ifa_addr to host in order to force a particular network interface. 

I agree that it should be `addr[LINK_ADDR_MAX]`, but that only wastes some bits.

dthain:
Sorry, I wasn't very clear, let me try again:

IMHO, the fundamental problem here isn't the length of the `addr` array. 

The problem is that the code doesn't make it clear which variables are addresses, and which ones are hostnames. The former can be passed to `link_connect` etc, and the latter must be passed to `domain_name_lookup`, and they cannot be trivially exchanged. I think if you take a few moments to clean up the names of those variables, then the correctness (or not) of the code will become obvious.

This is made worse by the fact that `interfaces_to_list` can take either a hostname or an address as its first argument, because it deals with the various connect options advertised by the manager:
https://github.com/cooperative-computing-lab/cctools/blob/ce4affa74399bb8301786dff4d263964b21fd12e/taskvine/src/worker/vine_worker.c#L1869

And so perhaps that argument should be `host_or_addr` in order to be abundantly clear.

PR #3742, opened 2024-03-22
Vine: Forsake functions when library exits.

dthain:
## Proposed changes

Previously, when a library would exit (success or otherwise), two problems resulted:
1 - The function calls sent subsequent to that library would simply sit on the worker doing nothing.
2 - The manager would quickly re-dispatch a library to that same worker, resulting in a tight loop of failures.

This PR fixes these as follows:
1 - The worker now has extended logic for `task_can_fit_eventually` so that functions will be forsaken if there is no library process at all that could execute them.
2 - Forsaken tasks are now handled correctly by deleting their sandboxes once they are retrieved.
3 - The most recent failure of a library is recorded in the library prototype object and used to throttle re-dispatch of failed library tasks.
4 - Fetching of task stdout is factored out of get_result to make the logic of fetching a result more clearly visible.

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -1029,6 +1029,9 @@ static int fetch_output_from_worker(struct vine_manager *q, struct vine_worker_i
 		break;
 	}
 
+	/* Regardless of the exit status, remove the task and sandbox from the worker. */
+	vine_manager_send(q, w, "kill %d\n", t->task_id);
+
 	if (result != VINE_SUCCESS) {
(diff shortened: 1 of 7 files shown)


btovar (approved):

dthain:
Still thinking about this one in two respects:
1 - The `FORSAKEN` code path in the manager is not overly trod, I want to make sure this behaves as expected.
2 - This will still cause the manger to retry libraries quickly, we need some kind of hysteresis in library attempts.

dthain:
Since this is an improvement to a corner case and not a critical bug, I think it can wait until after the pending release...

PR #3767, opened 2024-04-16
vine: Worker contact address

btovar:
Adds explicit --transfer-address to vine worker. If given, manager does nothing to the address. worker urls are now `workerip://` and `worker://`.  If `worker://`, then on get, the worker does a name resolution.

## Post-change actions

## Additional comments

### DIFF
--- a/doc/man/md/vine_worker.md
+++ b/doc/man/md/vine_worker.md
@@ -90,6 +90,7 @@ OPTION_LONG(keep-workspace) Do not delete the contents of the workspace on worke
 - **--volatility=_&lt;chance&gt;_**<br />Set the percent chance per minute that the worker will shut down (simulates worker failures, for testing only).
 - **--connection-mode=_&lt;mode&gt;_**<br />When using -M, override manager preference to resolve its address. One of by_ip, by_hostname, or by_apparent_ip. Default is set by manager.
 - **--transfer-port=_&lt;port&gt;_**<br /> Listening port for worker-worker transfers.  (default: any))
+- **--transfer-address=_&lt;addr&gt;_**<br /> Explicit contact address for worker-worker transfers. (default: :<transfer_port>)
 
 - **--ssl**<br />Enable tls connection to manager (manager should support it).
 - **--tls-sni=_&lt;&gt;_**<br />SNI domain name if different from manager hostname. Implies --ssl.
--- a/doc/man/md/vine_worker.md
+++ b/doc/man/md/vine_worker.md
@@ -90,7 +90,7 @@ OPTION_LONG(keep-workspace) Do not delete the contents of the workspace on worke
 - **--volatility=_&lt;chance&gt;_**<br />Set the percent chance per minute that the worker will shut down (simulates worker failures, for testing only).
 - **--connection-mode=_&lt;mode&gt;_**<br />When using -M, override manager preference to resolve its address. One of by_ip, by_hostname, or by_apparent_ip. Default is set by manager.
 - **--transfer-port=_&lt;port&gt;_**<br /> Listening port for worker-worker transfers.  (default: any))
-- **--transfer-address=_&lt;addr&gt;_**<br /> Explicit contact address for worker-worker transfers. (default: :<transfer_port>)
+- **--contact-hostport=_&lt;hostport&gt;_**<br /> Explicit contact host:port for worker-worker transfers, e.g., when routing is used. (default: :<transfer_port>)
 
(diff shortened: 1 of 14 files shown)


dthain:
Bump the protocol version, please

taskvine/src/manager/vine_manager.c
```diff
@@ -465,13 +466,25 @@ on its own port to receive get requests from other workers.
 static int handle_transfer_address(struct vine_manager *q, struct vine_worker_info *w, const char *line)
 {
 	int dummy_port;
-	if (sscanf(line, "transfer-address %s %d", w->transfer_addr, &w->transfer_port)) {
-		w->transfer_port_active = 1;
-		link_address_remote(w->link, w->transfer_addr, &dummy_port);
-		return VINE_MSG_PROCESSED;
-	} else {
+	int explicit;
```
dthain, line 469: Hmm, I'm concerned that this is tripping over some existing foot guns, and then also creating some new ones.  I had forgotten that the old worker code didn't really send the right worker address anyway (since normally it would listen on any interface.). And so now this overloads the `transfer-address` method with two or three strange subclasses.

I suggest that you break it apart doing something like this:
- When the worker listens normally on all interfaces on a specific port, have it send `transfer-port` with just the port number, and the manager should assume `link_address_remote` as before.
- When the worker sends a (potentially fake) host and port, then let's use a separate message `transfer-hostport` for that case.

doc/man/md/vine_worker.md
```diff
@@ -90,6 +90,7 @@ OPTION_LONG(keep-workspace) Do not delete the contents of the workspace on worke
 - **--volatility=_&lt;chance&gt;_**<br />Set the percent chance per minute that the worker will shut down (simulates worker failures, for testing only).
 - **--connection-mode=_&lt;mode&gt;_**<br />When using -M, override manager preference to resolve its address. One of by_ip, by_hostname, or by_apparent_ip. Default is set by manager.
 - **--transfer-port=_&lt;port&gt;_**<br /> Listening port for worker-worker transfers.  (default: any))
+- **--transfer-address=_&lt;addr&gt;_**<br /> Explicit contact address for worker-worker transfers. (default: :<transfer_port>)
```
dthain, line 93: Let's be very clear to distinguish between three things:
- `addr` to represent an IP address in string form.  (IPV4 or IPV6)
- `host` to represent a DNS domain name or address
- `hostport` to represent a DNS domain name/addr and port.

So I think this should be called `transfer-hostport`

taskvine/src/manager/vine_manager.c
dthain: And then in both cases it constructs `w->transfer_url` as appropriate.

taskvine/src/manager/vine_worker_info.h
```diff
@@ -42,9 +43,10 @@ struct vine_worker_info {
 	char *hashkey;
 
 	/* Address and port where this worker will accept transfers from peers. */
-	char transfer_addr[LINK_ADDRESS_MAX];
+	char transfer_addr[DOMAIN_NAME_MAX];
```
dthain, line 46: If this is going to store a domain name, it should be called `transfer_host` now.

taskvine/src/manager/vine_worker_info.h
```diff
@@ -42,9 +43,10 @@ struct vine_worker_info {
...
 	int  transfer_port;
 	int  transfer_port_active;
+	char *transfer_addr_url;       /* worker(ip)?://transfer_addr:transfer_port */
```
dthain, line 49: Likewise `transfer_url` because it could contain an address or a hostname.

taskvine/src/worker/vine_worker.c
```diff
@@ -421,8 +421,21 @@ static void send_transfer_address(struct link *manager)
 {
```
dthain, line 421: See above, send either `transfer-hostport` or `transfer-port`

taskvine/src/worker/vine_worker_options.c
```diff
@@ -69,6 +74,8 @@ void vine_worker_options_delete(struct vine_worker_options *self)
 		free(self->catalog_hosts);
 	if (self->factory_name)
 		free(self->factory_name);
+	if (self->contact_address)
```
dthain, line 77: `contact_host` if it could contain a dNS name.

dthain (changes requested):
A couple of thoughts on how to clearly distinguish between addresses and names.  We have been burned by that before, so I want to be more careful going forward...

PR #3780, opened 2024-04-24
TaskVine: Fix logging of library deployment

tphung3:
## Proposed changes

Previously whenever a function call is about to be dispatched, the manager writes `VINE_LIBRARY_SENT` to the transaction log which is incorrect. This is fixed so that the manager only writes it when a library is about to be dispatched.

Also finding the library in `commit_task_to_worker()` is redundant as `t->library_task` is set in vine_manager_check_worker_can_run_function_task()`, so it is removed.

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -2825,6 +2825,7 @@ static vine_result_code_t commit_task_to_worker(struct vine_manager *q, struct v
 	 * If the manager fails to send this function task to the worker however,
 	 * then the count will be decremented properly in @handle_failure() below. */
 	if (t->needs_library) {
+		t->library_task = find_library_on_worker_for_task(w, t->needs_library);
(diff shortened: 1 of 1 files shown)


dthain (changes requested):
Hmm, although this is a simple change, I think you should split it into two PRs.

Changing the library update message is obviously correct.

It's not so obvious to me that removing `find_library_on_worker_for_task` is correct.

BarrySlyDelgado:
I think `find_library_on_worker_for_task` was put there as `check_worker_can_run_function_task` would set the function call's worker for each worker we checked during scheduling. I.e. the worker with the library_task we last checked that could run the function call would be set as the function_call's worker rather than the worker we return during scheduling, like in `find_worker_by_files`.

dthain (approved):

dthain (approved):

PR #3787, opened 2024-04-26
vine: fixed_location task groups 

colinthomas-z80:
## Proposed changes

With task groups enabled, detect groups of sequential, temp-file dependent tasks, and send them in batches to a single worker, avoiding the overhead and complexity of maintaining fixed_location policy. 

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -2973,7 +2973,7 @@ static vine_result_code_t commit_task_to_worker(struct vine_manager *q, struct v
 	result = VINE_SUCCESS;
 	struct list *l = 0;
 	if (t->group_id) {
-		debug(D_VINE, "Task Has GroupID of %s. Should Send:", t->group_id);
+		debug(D_VINE, "Task %d has GroupID of %s. Should Send:", t->task_id, t->group_id);
 		l = hash_table_lookup(q->task_group_table, t->group_id);
 
 		struct vine_task *logt;
@@ -2986,17 +2986,34 @@ static vine_result_code_t commit_task_to_worker(struct vine_manager *q, struct v
 		// decrement refcount
 		vine_task_delete(t);
 	}
+
 	int counter = 0;
 	do {
-		if (counter) {
-			list_remove(q->ready_list, t);
-			// decrement refcount
-			vine_task_delete(t);
-		}
 
 		/* Kill empty libraries to reclaim resources. Match the assumption of
 		 * @vine.schedule.c:check_worker_have_enough_resources() */
 		kill_empty_libraries_on_worker(q, w, t);
+
+		/* If this is a function needing a library, dispatch the library. */
+		if (t->needs_library) {
+			/* Consider whether the library task is already on that machine. */
+			t->library_task = vine_schedule_find_library(q, w, t->needs_library);
+			if (!t->library_task) {
+				/* Otherwise send the library to the worker. */
+				/* Note that this call will re-enter commit_task_to_worker. */
+				t->library_task = send_library_to_worker(q, w, t->needs_library, &result);
+
+				/* Careful: if the above failed, then w may no longer be valid */
+				/* In that case return immediately without making further changes. */
+				if (!t->library_task) {
+					list_push_head(l, t);
+					return result;
+				}
+			}
+			/* If start_one_task_fails, this will be decremented in handle_failure below. */
+			t->library_task->function_slots_inuse++;
+		}
+
 		t->hostname = xxstrdup(w->hostname);
 		t->addrport = xxstrdup(w->addrport);
 
@@ -3006,15 +3023,6 @@ static vine_result_code_t commit_task_to_worker(struct vine_manager *q, struct v
 
 		itable_insert(w->current_tasks, t->task_id, t);
 		t->worker = w;
-
-		/* Increment the function count if this is a function task.
-		 * If the manager fails to send this function task to the worker however,
-		 * then the count will be decremented properly in @handle_failure() below. */
-		if (t->needs_library) {
-			t->library_task = vine_schedule_find_library(q, w, t->needs_library);
-			t->library_task->function_slots_inuse++;
-			vine_txn_log_write_library_update(q, w, t->task_id, VINE_LIBRARY_SENT);
-		}
 
 		change_task_state(q, t, VINE_TASK_RUNNING);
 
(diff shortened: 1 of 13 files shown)


colinthomas-z80:
This currently works (from some testing i have done). It does some pretty expensive searches and is a bit invasive to the code. I wanted to open it up at this stage in case anyone has thoughts on the idea and implementation.

dthain:
Colin, this has been sitting here for a while but based on our recent discussions I am getting more comfortable with the approach.  Do you think this is ready to merge?  Does it need more testing? Rebasing?

dthain:
Accidentally closed?

colinthomas-z80:
Yes I’m not sure why but I’ll fix it up so it doesn’t have all the outside commits.

Regarding merging in it should not create any issues with the feature turned off by default. While using the task groups there are still a few holes in the policy for error handling.

dthain:
Per our discussion today, waiting until this works smoothly with the Montage+Parsl example, then will be more confident in merging...

taskvine/src/manager/vine_manager.c
```diff
@@ -2861,49 +2862,82 @@ static vine_result_code_t commit_task_to_worker(struct vine_manager *q, struct v
 {
```
dthain, line 2862: While I think the logic here is correct, instead of adding a loop around the existing logic, do it like this:

Create new function `commit_task_group_to_worker` that calls `commit_task_to_worker` on the desired task, and then (if needed) also calls `commit_task_to_worker` on other tasks in the same group.

Change upstream calls to `commit_task_to_worker` to `commit_task_group_to_worker` as needed.

I think that will keep the code clearer and more maintainable down the road.

dthain: And also reduce the number of lines touched by this PR...

taskvine/src/manager/vine_task.h
```diff
@@ -127,6 +128,7 @@ struct vine_task {
 		
 	int has_fixed_locations;                               /**< Whether at least one file was added with the VINE_FIXED_LOCATION flag. Task fails immediately if no
 															 worker can satisfy all the strict inputs of the task. */
+	char *group_id;
```
dthain, line 131: Add comment.

taskvine/src/manager/vine_task_groups.c
```diff
@@ -0,0 +1,105 @@
...
+#include "vine_task_groups.h"
+#include "debug.h"
+#include "vine_mount.h"
+#include "vine_task.h"
+
+// create a new task group for this task based on the temp mount file
+static int vine_task_groups_create_group(struct vine_manager *q, struct vine_task *t, struct vine_mount *m)
+{
+	cctools_uuid_t uuid;
+	cctools_uuid_create(&uuid);
```
dthain, line 16: A uuid is a relatively expensive operation needed for ensuring uniqueness across hosts and invocations.  Do we really need that here, or would an integer counter suffice?

taskvine/src/manager/vine_task_groups.c
```diff
@@ -0,0 +1,105 @@
...
+	hash_table_insert(q->task_group_table, id, l);
+	return 1;
+}
+
+// locate the group with the task which outputs the desired file, and add the new task
+static int vine_task_groups_add_to_group(struct vine_manager *q, struct vine_task *t, struct vine_mount *m)
+{
+	struct list *l;
+	char *id;
+	HASH_TABLE_ITERATE(q->task_group_table, id, l)
```
dthain, line 34: Wow this is a deep iteration which may get expensive at scale. Since the goal is to find a task associated with a file, could this be represented with a hash mapping the address of the file object to the corresponding task object?

dthain (changes requested):

taskvine/src/manager/vine_task_groups.c
colinthomas-z80: It turns out the desired information was right there all along!

taskvine/src/worker/vine_sandbox.c
```diff
@@ -54,6 +54,10 @@ vine_cache_status_t vine_sandbox_ensure(struct vine_process *p, struct vine_cach
 			break;
 		case VINE_CACHE_STATUS_UNKNOWN:
 		case VINE_CACHE_STATUS_FAILED:
+			if (p->task->group_id) {
```
dthain, line 57: Re our discussion today, move this failure-handling logic to `task_can_run_eventually`.
If a task has an input file that is `UNKNOWN`, check it see if another known tasks produces it as an output file.
If so, then the task can stay until it becomes runnable.

dthain (changes requested):

btovar:
@colinthomas-z80 Please see conflicts above...

dthain:
@colinthomas-z80 this is looking pretty close, except for the task_can_run_eventually bit.

What do you think -- is there something else missing here before we are ready to merge?

colinthomas-z80:
Yes that part could be a bit more elegant. Before merging I want to clean up the worker debug logging and I am testing a change that fixes an issue with recovery tasks.

dthain:
ok, thanks for the update

colinthomas-z80:
RTM

taskvine/src/manager/vine_task_groups.c
```diff
@@ -0,0 +1,96 @@
...
+
+// create a new task group for this task based on the temp mount file
+static int vine_task_groups_create_group(struct vine_manager *q, struct vine_task *t, struct vine_mount *m)
+{
+	static int group_id_counter = 1;
+
+	char *id = string_format("%d", group_id_counter++);
+	struct list *l = list_create();
+
+	t->group_id = id;
```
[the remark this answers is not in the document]
colinthomas-z80: it is because the group_id is a key to a hash table. They used to be a full uuid string but that was deemed to be excessive since the number of groups does not get very large.

taskvine/src/manager/vine_manager.c
```diff
@@ -5472,6 +5543,14 @@ int vine_cancel_by_task_id(struct vine_manager *q, int task_id)
 		return 0;
 	}
 
+	if (task->group_id && (task->refcount > 1)) {
```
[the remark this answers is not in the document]
colinthomas-z80: I don't think so since the task is getting deleted afterwards

taskvine/src/manager/vine_task_groups.c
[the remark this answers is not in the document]
btovar: I suggest using `struct itable` then. It is like `struct hash_table` but the keys are ints.

taskvine/src/manager/vine_manager.c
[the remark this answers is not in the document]
btovar: I'd prefer to have the `if (task->group_id) {` only, set it to NULL inside the if, and do not check for the `refcount`. The only place where we should be checking for the refcount is when deleting the task, otherwise it gets tricky if we ever call functions in a different context.

btovar (approved):

PR #3791, opened 2024-04-27
vine: send library memleak

btovar:
If a worker had a library with all the slots full, the library would be sent again to the worker. 
Also, library tasks were cloned when submitted, but this is not needed as taskvine is the user of the task. For all other tasks we increase the ref count because is the responsibility of the user to delete them after wait.

These two conditions created a memleak just by checking whether tasks could be run.

I think this is the memleak observed in coffea.casa, but I have not tried yet.

## Proposed changes

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -4529,14 +4529,13 @@ struct vine_task *send_library_to_worker(struct vine_manager *q, struct vine_wor
 		return 0;
 	}
 
-	/* Duplicate the original task */
-	struct vine_task *t = vine_task_copy(original);
-
 	/* Check if this library task can fit in this worker. */
-	if (!check_worker_against_task(q, w, t)) {
-		vine_task_delete(t);
+	if (!check_worker_against_task(q, w, original)) {
 		return 0;
 	}
+
+	/* Duplicate the original task */
+	struct vine_task *t = vine_task_copy(original);
 
 	/* Give it a unique taskid if library fits the worker. */
 	t->task_id = q->next_task_id++;
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -4523,12 +4523,6 @@ int vine_submit(struct vine_manager *q, struct vine_task *t)
  */
 struct vine_task *send_library_to_worker(struct vine_manager *q, struct vine_worker_info *w, const char *name)
 {
-	struct vine_task *library = vine_schedule_find_library(w, name);
-	if (library) {
-		/* worker already has this library, not sending again. */
-		return library;
-	}
-
 	/* Find the original prototype library task by name, if it exists. */
 	struct vine_task *original = hash_table_lookup(q->libraries, name);
 	if (!original) {
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -3200,12 +3200,11 @@ int vine_manager_check_worker_can_run_function_task(
 	if (!library) {
 		library = send_library_to_worker(q, w, t->needs_library);
 		/* Careful: If this failed, then the worker object may longer be valid! */
-		if (!library)
+		if (!library) {
 			return 0;
+		}
 	}
 
-	// Mark that this function task will be run on this library.
-	t->library_task = library;
 	return 1;
 }
 
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -4553,13 +4553,15 @@ struct vine_task *send_library_to_worker(struct vine_manager *q, struct vine_wor
 	itable_insert(q->tasks, t->task_id, vine_task_clone(t));
 
 	/* Send the task to the worker in the usual way. */
+	/* Careful: If this failed, then the worker object or task object may no longer be valid! */
 	vine_result_code_t result = commit_task_to_worker(q, w, t);
 
-	/* Careful: If this failed, then the worker object may longer be valid! */
+	/* Careful again: If commit_task_to_worker failed the worker object or task object may no longer be valid! */
 	if (result == VINE_SUCCESS) {
 		vine_txn_log_write_library_update(q, w, t->task_id, VINE_LIBRARY_SENT);
 		return t;
 	} else {
+		/* if failure, tasks was handled by handle_failure(...) according to result. */
 		return 0;
 	}
 }
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -144,8 +144,6 @@ static void handle_library_update(struct vine_manager *q, struct vine_worker_inf
 static struct jx *manager_to_jx(struct vine_manager *q);
 static struct jx *manager_lean_to_jx(struct vine_manager *q);
 
-static struct vine_task *find_library_on_worker_for_task(struct vine_worker_info *w, const char *library_name);
-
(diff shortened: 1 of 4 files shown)


btovar:
There is still a memleak from the library tasks that I can't find. I think it happens at `vine_delete`. I'm able to execute large workflows with coffea.casa now, though, which is nice.

taskvine/src/manager/vine_manager.c
```diff
@@ -4048,6 +4005,24 @@ int vine_set_password_file(struct vine_manager *q, const char *file)
 	return copy_file_to_buffer(file, &q->password, NULL) > 0;
 }
 
+static void delete_task_at_exit(struct vine_task *t)
+{
+	if (!t) {
+		return;
+	}
+
+	if (t->type == VINE_TASK_TYPE_LIBRARY) {
```
dthain, line 4014: Hmm, this doesn't seem right to me.
I think our general principle should be: whoever added a reference is responsible for deleting that reference.
e.g. If a user created a task once, then they are responsible for deleting it once.  If the manager cloned it once, then the manager is responsible for deleting it once.  Just deleting all available references will cause trouble if there is some other entity that had a reason to clone the object.

taskvine/src/manager/vine_manager.c
```diff
@@ -4546,36 +4521,40 @@ int vine_submit(struct vine_manager *q, struct vine_task *t)
  * @param name The name of the library to be sent.
  * @return pointer to the library task if the operation succeeds, 0 otherwise.
  */
-
-static struct vine_task *send_library_to_worker(struct vine_manager *q, struct vine_worker_info *w, const char *name)
+struct vine_task *send_library_to_worker(struct vine_manager *q, struct vine_worker_info *w, const char *name)
 {
+	struct vine_task *library = vine_schedule_find_library(w, name);
```
dthain, line 4526: This assumes that there is exactly one library per worker, which isn't necessarily true.
e.g. an 8-core worker might run two instances of a 4-core library.

taskvine/src/manager/vine_schedule.c
```diff
@@ -211,9 +204,42 @@ int check_worker_against_task(struct vine_manager *q, struct vine_worker_info *w
...
 	}
 
+	if (t->needs_library) {
+		struct vine_task *library = vine_schedule_find_library(w, t->needs_library);
+		if (library) {
+			if (library->function_slots_inuse >= library->function_slots) {
+				return 0;
+			}
+		} else {
+			/* this is bad and we should feel bad. checking for matches should not modify the state of
```
dthain, line 214: I agree this is a code smell and a hazard to the unwary, but it seems unavoidable with the current framework.

taskvine/src/manager/vine_schedule.c
```diff
@@ -211,9 +204,42 @@ int check_worker_against_task(struct vine_manager *q, struct vine_worker_info *w
...
+			}
+		}
+	}
+
 	return 1;
 }
 
+/* Find a library task running on a specific worker that has an available slot.
+ * @return pointer to the library task if there's one, 0 otherwise. */
+struct vine_task *vine_schedule_find_library(struct vine_worker_info *w, const char *library_name)
```
dthain, line 229: Does this need to check for available slots on the library?
Otherwise it could return a library that is already fully occupied.

dthain (changes requested):

taskvine/src/manager/vine_manager.c
btovar: In this case the manager is the user. The manager `vine_task_copy` these tasks. This case is handling library tasks that are never DONE. If they were DONE, they would be deleted by the manager.

taskvine/src/manager/vine_manager.c
btovar: I agree, but I want to have the basic case working of one library without any bugs. We need to rethink how libraries are assigned to workers. This is an emergency fix, if you will.

taskvine/src/manager/vine_schedule.c
btovar: That's fine. We check for slots somewhere else. And that was the bug, a library occupied was 'not found', so the library was sent again.

taskvine/src/manager/vine_manager.c
btovar: One thing that could work is to re-add the slot check to that function, and just try to send the library at the last step. If the library does not fit, then it shouldn't be sent again.

btovar: That worked, tested on coffea-casa.

btovar: One alternative here is to not `vine_task_clone` library tasks when adding them to the task table, and then not `vine_task_delete` them when they are DONE.

dthain: Ah, I think this is the source of the confusion:

The `q->library` table contains *only* the prototype library tasks that are submitted by the user via `install_library`. So the user has one reference, and then the manager adds one when adding it to the `library` table. So, the manager should simple remove one reference when taking it out of the table.  (Maybe that table should have a more precise time.)

Now, when it is time to deploy a library to a worker, the manager **copies** the prototype task, and then puts it into the `tasks` table with an additional reference.  (Just like any other task.). So, when the library task is done, it should lose a reference for being taking out of the task table (just like any other task).  And then you are right, the manager (as creator) of that task should delete its final reference.

I think we are getting mixed up between the prototype library tasks and their copies.  Perhaps we need a better naming scheme to make it clear.

btovar: Since the original is copied, I think that the copies also get the `LIBRARY` type?
The originals were also mem-leaked, but that was fixed in another pr.

Since this is not fixing the mem leak anyway, I can remove it from the pr and merge the rest that we know works.

taskvine/src/manager/vine_manager.c
```diff
@@ -4586,30 +4555,14 @@ static struct vine_task *send_library_to_worker(struct vine_manager *q, struct v
 	vine_result_code_t result = commit_task_to_worker(q, w, t);
 
 	/* Careful: If this failed, then the worker object may longer be valid! */
-
 	if (result == VINE_SUCCESS) {
+		vine_txn_log_write_library_update(q, w, t->task_id, VINE_LIBRARY_SENT);
 		return t;
 	} else {
 		return 0;
```
dthain, line 4562: What happens on failure here?
Is it necessary to delete the task and back it out of the table, or does that happen naturally as part of the task state evolution from commit_task_to_worker?
(Either way some commentary may help here.)

taskvine/src/manager/vine_manager.c
```diff
@@ -4546,36 +4517,34 @@ int vine_submit(struct vine_manager *q, struct vine_task *t)
...
 	}
 
-	/* Duplicate the original task */
-	struct vine_task *t = vine_task_copy(original);
-
 	/* Check if this library task can fit in this worker. */
-	if (!check_worker_against_task(q, w, t)) {
-		vine_task_delete(t);
+	if (!check_worker_against_task(q, w, original)) {
```
dthain, line 4541: Hmm, the duplication of the task in the original version was inefficient, but it did provide some insurance against the possibility that `check_worker_against_task` might modify the original library object.    (For example, does `check_worker` do anything with regard to resource boxes?).   If you are 100% sure that this is the case, then we need some commentary (or maybe even use of `const`?) to ensure that tasks are not modified by `check_worker_against_task`.

dthain (changes requested):

PR #3802, opened 2024-05-01
vine: manager check temps

btovar:
Manager checks that temp files were created when task is done. If not, task is marked missing output.

Fixes issue where function calls and tasks may lie about an exit code 0 without generating required temp files.

## Post-change actions

## Additional comments


poncho/src/poncho/library_network_code.py
```diff
@@ -123,8 +123,20 @@ def start_function(in_pipe_fd, thread_limit=1):
...
-                    with open('outfile', 'wb') as f:
-                        cloudpickle.dump(globals()[function_name](event), f)
+                    result = globals()[function_name](event)
+                    try:
+                        with open('outfile', 'wb') as f:
+                            cloudpickle.dump(result, f)
+                    except Exception:
+                        if os.path.exits('outfile'):
+                            os.remove('outfile')
+                        raise
```
tphung3, line 133: I'm not sure about raising an exception here, because this means if a function fails to produce an output file the library dies with it too.

poncho/src/poncho/library_network_code.py
```diff
@@ -123,8 +123,20 @@ def start_function(in_pipe_fd, thread_limit=1):
...
+
+                    try:
+                        if not result["Success"]:
+                            raise Exception(result["Reason"])
+                    except Exception:
+                        raise
```
tphung3, line 139: Same as above.

poncho/src/poncho/library_network_code.py
```diff
@@ -145,10 +157,27 @@ def start_function(in_pipe_fd, thread_limit=1):
...
+                    exit_status = 0
+                    try:
+                        result = globals()[function_name](event)
+                        try:
+                            with open('outfile', 'wb') as f:
+                                cloudpickle.dump(result, f)
+                        except Exception:
+                            if os.path.exits('outfile'):
+                                os.remove('outfile')
+                            raise
```
tphung3, line 169: Here as well.

poncho/src/poncho/library_network_code.py
```diff
@@ -145,10 +157,27 @@ def start_function(in_pipe_fd, thread_limit=1):
...
+                    except Exception:
+                        exit_status = 1
+
+                    try:
+                        if not result["Success"]:
+                            exit_status = 1
+                    except Exception:
+                        exit_status = 1
```
tphung3, line 177: Maybe a different exit code here?

tphung3 (changes requested):
PR looks good overall. I don't think a library should die because a function fails however.

poncho/src/poncho/library_network_code.py
btovar: This is the direct method, and even without these changes, the library dies if function fails.

poncho/src/poncho/library_network_code.py
btovar: Exception is caught and child exits. This child exits anyway when call finishes, this just changes the exit call on failure.

poncho/src/poncho/library_network_code.py
btovar: See above...

poncho/src/poncho/library_network_code.py
btovar: Exit code only means call failed. The actual error is in event if it was created at all. This exit code is just checked by the manager to be nonzero. In regular tasks the exit code is checked by the user, but here there is no user...

btovar:
The only library that dies on exception is the one using the direct method, but that's the way it works. In the other case, it is only the forked child that terminates, and does so even when successful.

PR #3815, opened 2024-05-07
vine: Wait no wait

btovar:
Return any completed task to the application without doing any work.

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
@@ -239,7 +239,7 @@ class DaskVine(Manager):
                     submitted += 1
                     pending += 1
 
-                t = self.wait(tag, timeout)
+                t = self.wait_for_tag(tag, timeout)
                 if t:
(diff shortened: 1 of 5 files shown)


dthain:
I see the need for this and it makes sense, but a few things about the API:
- Is `wait_no_wait` somehow different from `wait_for_tag` with a timeout of zero?
- Would it make sense to have `wait_no_wait` without a tag?
- Let's brainstorm some additional ideas for the name.
-

btovar:
Time zero has meant to wait at least one second, so I didn't want to mess with that. But yes, if not to preserve backwards compatibility I would have preferred not to add a call and use `timeout=0`.  If you think it is ok, we can use timeout=0, in that way we don't have to come up with a new name.

btovar:
It has to have a tag, just in case the daskvine manager is managing two dags at the same time.  (Currently not possible as it waits for the queues to be empty, but we will need that for notebooks in the future.)

dthain:
That makes sense to me.

dthain (approved):
This approach looks good.

Issue #3763, opened 2024-04-12
vine: run gets stuck if worker with temp files can't communicate with peers


btovar:
Manger only creates recovery tasks when the files are not available somewhere, but it never checks if the worker has been unable to serve the files.

Some options:
- Create recovery tasks anyway
- Disconnect the offending worker immediately to create recovery tasks
- Transfer files back to manager and continue.

Observed with coffea-casa. @dthain, @colinthomas-z80 ?

PR #3818, opened 2024-05-08
vine: temporary, release worker that failed to xfer input file

btovar:
Temporary fix for #3763. If worker can't transfer input file, it is released.
Without this change, coffea casa workflows get stuck, as there is always a worker that can't communicate with other workers.

Maybe we can add a counter so that the worker is released after 3 consecutive failed tasks, or something like that. 

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/manager/vine_blocklist.c
+++ b/taskvine/src/manager/vine_blocklist.c
@@ -111,3 +111,13 @@ int vine_blocklist_is_blocked(struct vine_manager *q, const char *hostname)
 	struct vine_blocklist_info *info = hash_table_lookup(q->worker_blocklist, hostname);
 	return info && info->blocked;
 }
+
+int vine_blocklist_times_blocked(struct vine_manager *q, const char *hostname)
+{
+	struct vine_blocklist_info *info = hash_table_lookup(q->worker_blocklist, hostname);
+	if (info) {
+		return info->blocked;
+	}
+
+	return 0;
+}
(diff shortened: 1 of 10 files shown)


dthain:
@colinthomas-z80 what do you think about this one?

colinthomas-z80 (approved):

colinthomas-z80:
Looks good for this particular situation. If we had a particular worker that cannot upload then we would have another problem

btovar:
That's my fear. A worker that can't send files will kill all the peers, but I think this is the simplest thing we can do for now.

dthain:
I am concerned that this fix addresses a very immediate problem but could easily have unexpected effects, even within coffee-casa.  Let's think through it a little more.

So if a worker reports a transfer failure, who is at fault here?  The sender?  The receiver?  Both?

Is one transfer failure sufficient to indicate a broken worker?  Do we see transient worker failures in any other situation?

btovar:
I was thinking of a scheme like this:
- As a destination, forgive n task failures because of fail transfers. On failure, skip worker for  2t seconds.  After n failures, block worker for T seconds. If it has unique files, they are recomputed. After being blocked, any other failure causes it to be released.
- As a source, forgive m task failures because of fail transfers. m > n. On failure, skip worker for t seconds. After m failures, block worker for T seconds,. If it has unique files, they are recomputed.

Any successful transfer resets n and m.

dthain:
Once a worker is determined to be "bad", what should the consequences be?
- Disconnect the worker?  (It will likely reconnect)
- Blocklist the worker? (Does the blocklisting apply to file transfers?)
- Keep the worker, but penalize it?  (Your example of delays above.)

btovar:
I'd like it to be blocked, but in coffea.casa all workers seem to come from the same host.  We could try by worker-id, though.

btovar:
@dthain Added temporary block/disable transfer of bad workers.

colinthomas-z80 (approved):

PR #3832, opened 2024-05-13
vine: check if the library exists when submitting FunctionCall tasks

JinZhou5042:
## Proposed changes

While constructing and submitting a FunctionCall task, for example, `task = vine.FunctionCall('test-library', 'my_func', args.size)`, this PR checks one thing:
- `test-library` should exist in current libraries

It also comes with some renaming just for better readability 

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
@@ -890,14 +890,12 @@ class Manager(object):
     # @param task    A FunctionCall task
     def check_funcall_validity(self, task):
         library_name = task.get_library_name()
-        available_library_names = [library_name for library_name in self._library_table.keys()]
-        if library_name not in available_library_names:
-            raise ValueError(f"invalid library name \'{library_name}\', available libraries are {available_library_names}")
+        if library_name not in [library_name for library_name in self._library_table.keys()]:
+            raise ValueError(f"invalid library name \'{library_name}\'")
         library_task = self._library_table[library_name]
-        function_names_on_library = library_task.get_function_names()
         function_name_to_be_called = task.get_function_name()
-        if function_name_to_be_called not in function_names_on_library:
-            raise ValueError(f"invalid function name \'{function_name_to_be_called}\', available libraries are {function_names_on_library}")
+        if function_name_to_be_called not in library_task.get_function_names():
+            raise ValueError(f"invalid function name \'{function_name_to_be_called}\'")
 
     ##
     # Submit a library to install on all connected workers
@@ -977,11 +975,9 @@ class Manager(object):
 
         # Create Task to execute the Library and prepend it with some setup code if needed.
         if init_command:
-            t = LibraryTask(f"{init_command} python ./library_code.py", library_name)
+            t = LibraryTask(f"{init_command} python ./library_code.py", library_name, function_list=function_list, library_code_path=library_code_path)
         else:
-            t = LibraryTask("python ./library_code.py", library_name)
-        function_names = [function.__name__ for function in function_list]
-        t.set_function_names(function_names)
+            t = LibraryTask("python ./library_code.py", library_name, function_list, function_list=function_list, library_code_path=library_code_path)
 
         # Declare the environment if needed.
         if add_env:
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
@@ -906,6 +906,8 @@ class Manager(object):
     # @param self           Reference to the current manager object.
     # @param library_name   Name of the library to be checked
     def check_library_exists(self, library_name):
+        if not isinstance(library_name, str):
+            raise TypeError(f"library_name should be str, not {type(library_name)}")
         return cvine.vine_manager_get_library(self._taskvine, library_name) is not None
 
     ##
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
@@ -977,7 +977,7 @@ class Manager(object):
         if init_command:
             t = LibraryTask(f"{init_command} python ./library_code.py", library_name, function_list=function_list, library_code_path=library_code_path)
         else:
-            t = LibraryTask("python ./library_code.py", library_name, function_list, function_list=function_list, library_code_path=library_code_path)
+            t = LibraryTask("python ./library_code.py", library_name, function_list=function_list, library_code_path=library_code_path)
 
         # Declare the environment if needed.
         if add_env:
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
@@ -908,7 +908,7 @@ class Manager(object):
     def check_library_exists(self, library_name):
         if not isinstance(library_name, str):
             raise TypeError(f"library_name should be str, not {type(library_name)}")
-        return cvine.vine_manager_get_library(self._taskvine, library_name) is not None
+        return cvine.vine_manager_find_library_template(self._taskvine, library_name) is not None
 
     ##
     # Turn a list of python functions into a Library Task.
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
@@ -869,6 +869,10 @@ class Manager(object):
     # @param self   Reference to the current manager object.
     # @param task   A task description created from @ref ndcctools.taskvine.task.Task.
     def submit(self, task):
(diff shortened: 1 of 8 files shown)


dthain:
@BarrySlyDelgado what do you think?  Is there any reason a submit time check wouldn't work?

BarrySlyDelgado (approved):

BarrySlyDelgado:
Looks fine to me. Can't think of reason it wouldn't work.

BarrySlyDelgado (approved):

taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
```diff
@@ -878,6 +882,23 @@ def submit(self, task):
...
 
+    ##
+    # Check the validity of a FunctionCall task
+    # The name of library should exist, and the function to be executed should exist on that library, too
+    #
+    # @param self    Reference to the current manager object.
+    # @param task    A FunctionCall task
+    def check_funcall_validity(self, task):
+        library_name = task.get_library_name()
+        available_library_names = [library_name for library_name in self._library_table.keys()]
```
btovar, line 893: `available_library_names` is not needed. Do:
```
if library_name not in self._library_table:
   raise ...
```

taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
```diff
@@ -949,16 +970,18 @@ def create_library_from_functions(self, name, *function_list, poncho_env=None, i
...
             os.chmod(library_code_path, 0o775)
 
         # Create Task to execute the Library and prepend it with some setup code if needed.
         if init_command:
-            t = LibraryTask(f"{init_command} python ./library_code.py", name)
+            t = LibraryTask(f"{init_command} python ./library_code.py", library_name)
         else:
-            t = LibraryTask("python ./library_code.py", name)
+            t = LibraryTask("python ./library_code.py", library_name)
+        function_names = [function.__name__ for function in function_list]
```
btovar, line 983: Functions may be passed already as strings.  Also, function names should probably be part of the `LibraryTask` constructor, as it is fundamental for its purpose (`LibraryTask(cmd, *functions)`). In this way you do not need the function to set library names, which if called after the construction of the library will create inconsistencies.

btovar (changes requested):

taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
JinZhou5042: solved, thank you!

taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
JinZhou5042: I now pass new arguments `function_list` as well as `library_code_path` for the sake of consistency. Thank you!

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
```diff
@@ -1162,12 +1172,22 @@ class LibraryTask(Task):
...
+        self.library_name = library_name
+        self.function_list = function_list
+        self.library_code_path = library_code_path
         self._manager_will_free = True
-        self.provides_library(name)
+        self.provides_library(self.library_name)
+
+    def get_function_names(self):
+        if self.function_list:
+            return [f.__name__ for f in self.function_list]
```
btovar, line 1190: I'm not sure this works in general.  A library is not required to declare its functions up-front. The function names are only known when using `poncho_serverize`, but this is not the most general case.

This pr should only check if the library with a name exists.  If the function does not exist at the library should be an error reported by the worker (in another pr).

btovar (changes requested):

JinZhou5042:
@btovar Thanks for your constructive suggestions! Please check and let me know if this one makes sense to you.

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
```diff
@@ -1059,12 +1047,34 @@ def __init__(self, library, fn, *args, **kwargs):
...
+
         self._event = {}
         self._event["fn_args"] = args
         self._event["fn_kwargs"] = kwargs
 
         self._saved_output = None
-        self.set_library_required(library)
+        self.set_library_required(self._library_name)
+
+    def get_library_name(self):
```
btovar, line 1061: This should simply return `self._library_name`. In general we use `get_` to obtain the value of attributes only.

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
```diff
@@ -1059,12 +1047,34 @@ def __init__(self, library, fn, *args, **kwargs):
         # function calls at worker only need the name of the function.
         self.set_command(fn)
 
+        self._library = library
```
btovar, line 1050: This attribute should not be needed, specially when it could get out of sync with _library_name, and because it could be a string or a function, which make bookkepping hard.

taskvine/src/bindings/python3/ndcctools/taskvine/task.py
btovar: Instead, convert to `_compute_library_name(...)`, which should set `self._library_name`.

taskvine/src/manager/taskvine.h
```diff
@@ -889,6 +889,12 @@ void vine_manager_install_library( struct vine_manager *m, struct vine_task *t,
 */
 void vine_manager_remove_library( struct vine_manager *m, const char *name );
 
+/** Check if a given library name exists on the manager.
+@param m A manager object
+@param library_name The library name to be checked
+*/
+int vine_manager_check_library_exists(struct vine_manager *m, const char *library_name);
```
btovar, line 896: `struct vine_task *vine_manager_get_library(...)` and in python check for `None`.

taskvine/src/manager/vine_manager.c
```diff
@@ -4603,6 +4603,14 @@ void vine_manager_remove_library(struct vine_manager *q, const char *name)
 	hash_table_remove(q->library_templates, name);
 }
 
+int vine_manager_check_library_exists(struct vine_manager *q, const char *library_name)
+{
+	if (hash_table_lookup(q->library_templates, library_name)) {
```
btovar, line 4608: simply: `return hash_table_lookup(q->library_templates, library_name);`

taskvine/src/manager/vine_manager.h
```diff
@@ -271,6 +271,8 @@ struct vine_task *vine_manager_no_wait(struct vine_manager *q, const char *tag,
 
 void vine_manager_remove_worker(struct vine_manager *q, struct vine_worker_info *w, vine_worker_disconnect_reason_t reason);
 
+/* Check if a given library name exists on the manager */
+int vine_manager_check_library_exists(struct vine_manager *q, const char *library_name);
```
btovar, line 275: Rename as above.

btovar (changes requested):

JinZhou5042:
Thanks for the comments! What do you think of this version?

taskvine/src/manager/taskvine.h
```diff
@@ -889,6 +889,12 @@ void vine_manager_install_library( struct vine_manager *m, struct vine_task *t,
 */
 void vine_manager_remove_library( struct vine_manager *m, const char *name );
 
+/** Get the library task with the given name, return null if not found.
+@param m A manager object
+@param library_name The library name
+*/
+struct vine_task *vine_manager_get_library(struct vine_manager *m, const char *library_name);
```
btovar, line 896: Move to `vine_manager.h`. This is not part of the official API.

btovar: Actually you already have it in `vine_manager.h`.  Delete from here.

JinZhou5042: I am not sure how could we invoke it from `cvine.py` if I delete it from `vine_manager.h`?

JinZhou5042: Since I have 
``` Python
def check_library_exists(self, library_name):
    return cvine.vine_manager_get_library(self._taskvine, library_name) is not None
``` 
in `manager.py`, it seems that this API is supposed to be exposed?

btovar: It is exposed in the bindings when you add it to `vine_manager.h`. `vine_manager.h` declares functions related to the manager that are not exposed in the API.  API functions go to `taskvine.h`. This declaration should be in `vine_manager.h` and not `taskvine.h`.  Look at `taskvine.i` somewhere in the bindings source directory to see which `.h` files are included to construct the bindings.

JinZhou5042: It seems that we didn't include `vine_manager.h` to construct bindings, I added it so that we can invoke in python. Does that look good?

btovar: Good catch!

btovar (approved):

JinZhou5042:
RTM

dthain (approved):

PR #3834, opened 2024-05-14
vine: remove redundant library duplications

JinZhou5042:
## Proposed changes

When checking if a task is compatible with a given worker (in `check_worker_against_task`), the last step is to check whether the required library exists on that particular worker or not. Previously, there were two cases:
1. the library exists, with available function slots
2. the library doesn't exist, or exists but no available function slots (all are busy)

However, it comes with a problem: If the total cores are 64, but the `function_slots` of a library is 16, each consuming 1 core. Though as users we want there to be 16 tasks running concurrently as we specifically assigned that much, the actual concurrent tasks are 64, with 4 libraries running simultaneously.

To mitigate this redundant duplication, we can extract a status from the second one and it becomes:

1. the library exists, with available function slots
2. the library exists, no available function slots
3. the library doesn't exist

By carefully handle each case, we can make sure that there is only one library running on a single worker and no redundant libraries are duplicated.

## Post-change actions

## Additional comments

### DIFF
--- a/taskvine/src/manager/vine_schedule.c
+++ b/taskvine/src/manager/vine_schedule.c
@@ -215,7 +215,7 @@ int check_worker_against_task(struct vine_manager *q, struct vine_worker_info *w
 			/* If the library is already running on this worker, then it's ok. */
 			return 1;
 		} else if (status == LIBRARY_TASK_FOUND_NO_AVAILABLE_SLOT) {
-			/* If the library is already running on this worker, but there are no available slots, then it's not ok. */
+			/* If the library exists, but there are no available slots, then it's not ok. */
 			return 0;
 		} else if (status == LIBRARY_TASK_NOT_FOUND) {
 			/* XXX: checking for matches should not modify the state of workers. */
--- a/taskvine/src/manager/vine_schedule.c
+++ b/taskvine/src/manager/vine_schedule.c
@@ -209,8 +209,15 @@ int check_worker_against_task(struct vine_manager *q, struct vine_worker_info *w
 
 	/* do this check last! otherwise the library is sent to workers that can't fit it. */
 	if (t->needs_library) {
-		struct vine_task *library = vine_schedule_find_library(w, t->needs_library);
-		if (!library) {
+		int status;
+		struct vine_task *library = vine_schedule_find_library(w, t->needs_library, &status);
+		if (status == LIBRARY_TASK_FOUND_WITH_AVAILABLE_SLOT) {
+			/* If the library is already running on this worker, then it's ok. */
+			return 1;
+		} else if (status == LIBRARY_TASK_FOUND_NO_AVAILABLE_SLOT) {
+			/* If the library is already running on this worker, but there are no available slots, then it's not ok. */
+			return 0;
+		} else if (status == LIBRARY_TASK_NOT_FOUND) {
 			/* XXX: checking for matches should not modify the state of workers. */
 			library = send_library_to_worker(q, w, t->needs_library);
 			/* Careful: If this failed, then the worker object may longer be valid! */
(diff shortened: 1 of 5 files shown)


taskvine/src/manager/vine_schedule.c
```diff
@@ -209,8 +209,15 @@ int check_worker_against_task(struct vine_manager *q, struct vine_worker_info *w
...
 	/* do this check last! otherwise the library is sent to workers that can't fit it. */
 	if (t->needs_library) {
-		struct vine_task *library = vine_schedule_find_library(w, t->needs_library);
-		if (!library) {
+		int status;
+		struct vine_task *library = vine_schedule_find_library(w, t->needs_library, &status);
+		if (status == LIBRARY_TASK_FOUND_WITH_AVAILABLE_SLOT) {
+			/* If the library is already running on this worker, then it's ok. */
+			return 1;
+		} else if (status == LIBRARY_TASK_FOUND_NO_AVAILABLE_SLOT) {
```
btovar, line 217: If the library has no slots it should do the same as library not found. The same library may run more than once in a worker.

taskvine/src/manager/vine_schedule.c
```diff
@@ -225,19 +232,26 @@ int check_worker_against_task(struct vine_manager *q, struct vine_worker_info *w
...
 {
 	uint64_t task_id;
 	struct vine_task *task;
 	ITABLE_ITERATE(w->current_tasks, task_id, task)
 	{
 		if (task->type == VINE_TASK_TYPE_LIBRARY_INSTANCE && task->provides_library &&
-				!strcmp(task->provides_library, library_name) &&
-				(task->function_slots_inuse < task->function_slots)) {
-			return task;
+				!strcmp(task->provides_library, library_name)) {
```
btovar, line 242: See above. This distinction is not needed.

taskvine/src/manager/vine_schedule.h
```diff
@@ -18,10 +18,14 @@ This module is private to the manager and should not be invoked by the end user.
 #include "vine_task.h"
 #include "vine_worker_info.h"
 
+#define LIBRARY_TASK_FOUND_WITH_AVAILABLE_SLOT 1
```
btovar, line 21: Use an enum.

taskvine/src/worker/vine_worker_options.c
```diff
@@ -28,6 +28,7 @@ struct vine_worker_options *vine_worker_options_create()
 	struct vine_worker_options *self = malloc(sizeof(*self));
 	memset(self, 0, sizeof(*self));
 
+	self->cores_total = 0;
```
btovar, line 31: This does not seem to belong to this pr.

btovar (changes requested):

taskvine/src/manager/vine_schedule.c
JinZhou5042: Just be confused. In this case, for example, if a user installs 1 library with 16 cores and 16 slots, and the total cores are 64, how many concurrent tasks are expected to be?

btovar: At least 16 for that library. However the worker would fit 3 more libraries, so it 4 libraries are installed, then 64. That is 64 slots across 4 libraries.

JinZhou5042: Oh, I thought that was an issue... As my experience, sometimes I want to over-allocate the cores on a worker, so that I can easily modify the usable library cores without removing and collecting a new worker. I think sometimes the users may want to have the option to not to use out all available cores on a worker, that was the intention of this pr. What do you think?

btovar: We want to precisely account for the resources that each task is using. Instead, I suggest having libraries that fill the whole worker, and add an option to the function calls that specifies how many slots they use.  Such an option may be needed anyway as functions from a library may have different resources needs.

JinZhou5042: So for a 64-core worker, this library setting:
``` Python
libtask.set_cores(1)
libtask.set_function_slots(1)
```
has the same effect with:
``` Python
libtask.set_cores(64)
libtask.set_function_slots(64)
```
except that the first one has only 1 single process but the second one has 64 independent processes ... If we would like to fill the worker with libraries, `set_cores` seems to not really matter. How about automatically setting the library cores to the maximum of total cores on the worker? 

Or, can we also encourage the users to leverage as many cores as they can, but also leaves them an option to use partial cores on the worker?

btovar: It does not have the same effect... One has one library with 64 slots, and the other has 64 libraries, each with one slot. That overhead may become important.

```
libtask.set_cores(60)
libtask.set_function_slots(60)
```
This would be 61 threads? 60 slots + the thread of the library?

btovar: A user might want to run one concurrent function per library and just take advantage of the fact that the interpreter is not started every time. Also, they may want to run different types of libraries, so a library should not take the whole worker.

btovar: Given the above, is it ok to close this pr?  We need to allow multiple libraries and library types in a worker, and this pr seems to disallow that.  This is something I often get confused and I forget that a worker may run more than one library.

btovar (changes requested):

dthain:
That's right, it is incorrect to assume that there is only once instance of a library per worker.

dthain:
@JinZhou5042 because this is such a tricky area of code with many constraints, let's back up and discuss a solution before coding.  Please identify the appropriate issue and start a discussion there, outlining the nature of the problem and potential solutions.

JinZhou5042:
Oh, now I understand, that clarifies my confusion and enhances my understanding regarding the library duplication, thank you! @btovar @dthain

Issue #3569, opened 2023-11-13
Vine: Function Hash Calculation Excluding Preamble Imports


JinZhou5042:
Currently, the way to create the hash of a library is:
```python
functions_hash = package_serverize.generate_functions_hash(function_list)
```
Which is inside of function `create_library_from_functions(...)`

The unique hash is generated based on the content of functions, excluding preamble import statements. 

Therefore, if we want to update a previously installed library by altering only the preamble import statements, it actually results in an identical hash and does not lead to the creation of a new library.

Should this be a problem?

dthain:
@BarrySlyDelgado what do you think?

BarrySlyDelgado:
This is an issue. I have to familiarize myself with the library however, with this change, it is probably best to create a hash using the whole network code with the addition of the preamble imports now.

dthain:
Just FYI this particular hash is not the TaskVine hash for caching objects at remote workers.

Instead, it is a hash to compute the uniqueness of the library itself, so that Poncho does not have to constantly rebuild the library (and environment) when its constituents have not changed.  That way, the library can be cached locally at the manager and not constantly rebuilt.

JinZhou5042:
Resolved in #3837

PR #3837, opened 2024-05-15
vine: combine imports when calculating function hash

JinZhou5042:
## Proposed changes

Fix for #3569, while calculating the hash for library functions, this pr does two things to avoid creating a new environment with minor modifications that do not affect the execution of functions:

- consider the module names in `import_modules`
- remove comments, blank lines and extraneous newlines 

## Post-change actions

## Additional comments

### DIFF
--- a/poncho/src/poncho/package_serverize.py
+++ b/poncho/src/poncho/package_serverize.py
@@ -103,11 +103,6 @@ def create_library_code(path, funcs, dest, version, import_modules=None):
             output_file.write("@remote_execute\n")
             output_file.write(function_code)
             output_file.write("\n")
-        
-        # write a retriever function for retriever tasks used in FutureFunctionCall tasks
-        output_file.write("@remote_execute\n")
-        output_file.write("def retrieve_output(arg):\n")
-        output_file.write("    return arg\n")
 
         output_file.write(init_function)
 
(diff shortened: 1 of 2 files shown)


poncho/src/poncho/package_serverize.py
```diff
@@ -172,18 +167,29 @@ def pack_library_code(path, envpath):
...
+
     source_code = ""
+    if import_modules:
+        source_code += " ".join(["import " + module.__name__ + "\n" for module in import_modules])
+
     for fnc in functions:
         try:
-            source_code += fnc.__name__
-            source_code += inspect.getsource(fnc)
+            # get function source code, remove single-line comments, blank lines and excessive newlines
```
btovar, line 180: Why are you removing the comments, etc?  They don't hurt the hash computation below, and instead have the potential of introducing bugs as it is easy to get regular expressions wrong. For example, your expressions don't take into account that # may be inside a string.  I recommend against this change.

poncho/src/poncho/package_serverize.py
```diff
@@ -172,18 +167,29 @@ def pack_library_code(path, envpath):
 # Python's exec or Jupyter Notebooks won't work here.
 # @param functions  A list of functions to generate the hash value from.
 # @return           a string of hex characters resulted from hashing the contents and names of functions.
-def generate_functions_hash(functions: list) -> str:
+def generate_functions_hash(functions: list, import_modules=None) -> str:
+    import re
+    import sys
+
     source_code = ""
```
btovar, line 174: A better implementation would be to have `source_code = []` as a list to which you `source.append(some_string)`. Then before computing the hash do a `" ".join(source_code)`.  Modules can added with `source_code.extend(["import " + module.__name___ etc.])`

btovar (changes requested):

poncho/src/poncho/package_serverize.py
JinZhou5042: That's a fair point. I thought sometimes the users may somehow enter a few random blanks or newlines, but they might don't want to re-install the environment as that takes several minutes but has the same effect with the previous one. You are right that the regular expressions are easy to introduce bugs, I will go ahead and remove it!

btovar: If they modify the function in any way, then that's ok. It is better to be safe and recompute the hash rather than try to interpret if two functions are the same.

JinZhou5042: Copy that!

JinZhou5042:
RTM

dthain:
@BarrySlyDelgado does this fit your plan for how hashes should be constructed?

BarrySlyDelgado:
This looks good.

btovar (approved):

btovar:
Nice job @JinZhou5042!

PR #3843, opened 2024-05-21
shepherd - shepherd workflow manager

saifulislampi:
**Naming Update:** "As of now, 'shepherd' has been provisionally selected as the name for this tool." - May 30, 2024

----
## servicexd - Service Execution Daemon [old name]

This is a very early prototype for servicexd - the service execution daemon. The name is currently a placeholder. Please note that I am actively pushing changes to this project, so it may be unstable or broken at times.


dthain:
Yep, you are on the right track here!

dthain:
FYI, builds are failing though not your fault -- please rebase in order to pick up some recent fixes to the test infrastructure.

dthain:
Based on our conversation so far, I am liking Shepherd or some variation on that for the name of the tool.
There are some systems out there using that name, but it's a pretty broad generic term.
I suggest that you use "shepherd" for the short name of the Unix command, directory structure, etc.
But then call it something more elaborate like "Shepherd Workflow Manager" as the full name for the webpage, documentation ,etc.

dthain:
You have the right directory structure for integrating with the rest of the cctools.

One addition: In the `test` directory, please add a script `TR_shepherd.sh` with functions like `prepare()` `run()` and `clean()` that will execute the self-contained test example.  See `dttools/test/TR_jx.sh` as an example of how to do it.  (Obviously, we can't easily run the SADE example on every integration test, but's it nice to have that complete example there.)

saifulislampi:
I will add the SADE workflow code here. But you are right, it would be difficult to run SADE workflow outside the whole prepared container. Let's storm some other test cases together that we can run on every integration test. Right now, I am using some bash scripts that echo specific lines and transition into different states.

shepherd/diagram/dot/diagram1.dot
```diff
@@ -0,0 +1,49 @@
+// Workflow Visualization
+digraph {
+	subgraph cluster_service1 {
+		node [color=lightblue style=filled]
+		color=lightgrey style=filled
+// 		service1_initialized [label="initialized"]
```
btovar, line 6: I'd remove entries rather than just comment them.

shepherd/environment.yml
```diff
@@ -0,0 +1,13 @@
...
+  - conda-forge
+  - defaults
+dependencies:
+  - python=3.11
+  - pip
+  - pyyaml
+  - graphviz
+  - matplotlib
+  - seaborn
+#  - pip:
```
btovar, line 12: Remove, not comment.

shepherd/py_source_combiner.py
```diff
@@ -0,0 +1,94 @@
+import os
+import re
+import sys
+
+
+def extract_imports(file_content, source_dir):
```
btovar, line 6: You may want to use https://docs.python.org/3/library/ast.html  rather than grep to get the imports, as regexps are  not very robust. For example, I don't think the following case is handled correctly:

```python
if something:
   import ...
```

shepherd/readme.md
```diff
@@ -0,0 +1,371 @@
+# Shepherd - Service Orchestration and Monitoring Tool
```
btovar, line 1: We usually use capital letters, like README.md.

shepherd/shepherd.py
```diff
@@ -0,0 +1,438 @@
...
+
+
+def load_and_preprocess_config(filepath):
+    """Loads and preprocesses configuration from a YAML file."""
+    with open(filepath, 'r') as file:
+        config = yaml.safe_load(file)
+
+    preprocess_config(config, filepath)
+
+    print(f"DEBUG: Loaded and preprocessed config from {filepath}")
```
btovar, line 22: Since this is an independent module, you may want to use the standard python logging module.

shepherd/shepherd.py
```diff
@@ -0,0 +1,438 @@
...
+def preprocess_config(config, config_path):
+    """Automatically fills in missing stdout_path and stderr_path paths."""
+    services = config.get('services', {})
+    stdout_dir = config.get('output', {}).get('stdout_dir', '')
+    working_dir = os.path.dirname(os.path.abspath(config_path))
+
+    for service_name, details in services.items():
+        # Auto-fill log and error files if not specified
+        if 'stdout_path' not in details:
+            details['stdout_path'] = f"{service_name}_stdout.log"
```
btovar, line 35: https://www.w3schools.com/python/ref_dictionary_setdefault.asp

shepherd/shepherd.py
```diff
@@ -0,0 +1,438 @@
...
+
+        if state_file_path:
+            details['state']['file']['path'] = os.path.join(working_dir, state_file_path)
+
+
+def validate_and_sort_programs(config):
+    print("DEBUG: Validating and sorting programs")
+    required_keys = ['services']
+
+    for key in required_keys:
```
btovar, line 56: Any reason why are you checking here and not letting the exception be raised in whatever code uses it?
Also, if you are going to raise it, it is a `KeyError`.

shepherd/shepherd.py
```diff
@@ -0,0 +1,438 @@
...
+import multiprocessing
+import os
+import signal
+import subprocess
+import sys
+import threading
+import time
+import yaml
+
+# --- config_loader.py ---
```
btovar, line 12: should these type of comments be removed?

shepherd/shepherd.py
```diff
@@ -0,0 +1,438 @@
+from multiprocessing import Process
```
btovar, line 1: Do you really want to commit this file? It seems to be a concatenation of all the others.

btovar (changes requested):

shepherd/diagram/dot/diagram1.dot
saifulislampi: Yes. I agree. I will remove all comented code.

shepherd/py_source_combiner.py
saifulislampi: I used this as a temporary combiner for this project. I will remove this file entirely.

shepherd/shepherd.py
saifulislampi: No. Thanks for pointing it out. I will remove all unncessary files and code update the PR soon.

dthain:
@saifulislampi is it ok to close this PR in favor of the new Shepherd repository?

dthain:
Ping?

saifulislampi:
> @saifulislampi is it ok to close this PR in favor of the new Shepherd repository?

Yes. Please close this one.

PR #3853, opened 2024-05-31
vine: compute hash only when cache level requires it

btovar:
## Proposed Changes

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_cached_name.c
+++ b/taskvine/src/manager/vine_cached_name.c
@@ -338,17 +338,22 @@ Returns a string that must be freed with free().
 char *vine_cached_name(const struct vine_file *f, ssize_t *totalsize)
 {
 	unsigned char digest[MD5_DIGEST_LENGTH];
-	char *hash, *name;
+	char *hash = NULL;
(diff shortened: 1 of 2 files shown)


dthain:
Hmm, how does this interact with logic in `vine_file_create` prior to calling `vine_cached_name`?

(maybe we need to clearly distinguish "compute the name by some means" from "compute a hash")

BarrySlyDelgado (approved):

btovar:
Aaaah, yes, this is repeating what `vine_file_create` does.  The only thing to correct is to also compute names for the worker cache level.

btovar:
When I looked at vine checksum name functions I thought we were computing the hashes in all instances. From your comment, I saw that I was incorrect, and that the logic of which name to compute was in `vine_file_create`. Looking at the function I saw that we were not computing the name for the worker level cache.

So yes, the original purpose was to compute less hashes because of a misunderstanding, and then I saw that in fact we need to compute more caches to be consistent to what the difference cache levels mean.

dthain (approved):

PR #3857, opened 2024-06-11
Vine: Keep Scheduler Stateless for Libraries/Functions

### LINKED ISSUE #3825: vine: send library task when sending first function call that uses it
Currently libraries are send from `check_worker_against_task`.  This modifies the worker even when the task won't get to run there because a better worker is found. Instead the library should be sent from `start_one_task`  `check_worker_against_task` should simply check whether the library template fits the worker.

dthain:
## Proposed Changes

The scheduler should *not* modify the state of the system in check_worker_against_task. Instead, when scheduling a function call, look for compatibility and capacity for library tasks, and then dispatch them in commit_task_to_worker.

Fixes #3825

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -169,6 +169,9 @@ static void delete_uncacheable_files(struct vine_manager *q, struct vine_worker_
 static int delete_worker_file(struct vine_manager *q, struct vine_worker_info *w, const char *filename,
 		vine_cache_level_t cache_level, vine_cache_level_t delete_upto_level);
 
+struct vine_task *send_library_to_worker(struct vine_manager *q, struct vine_worker_info *w, const char *name, vine_result_code_t *result );
+
+
 /* Return the number of workers matching a given type: WORKER, STATUS, etc */
 
 static int count_workers(struct vine_manager *q, vine_worker_type_t type)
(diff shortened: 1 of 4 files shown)


dthain:
So the idea here is that the scheduler should never change the state of the system while considering a task-worker match.  This is because `check_worker_against_task(q,w,t)` could be called for many prospective matches (i.e. find the worker with the most data matched).  And the worker so selected is not necessarily the first (or the last) worker for which this function returns true.

@btovar @JinZhou5042 I think the idea is straightforward, but there are lots of tricky corner cases with libraries and functions, so I would like your feedback.

taskvine/src/manager/vine_manager.c
```diff
@@ -2873,27 +2876,40 @@ assignment and the new task state.
 
 static vine_result_code_t commit_task_to_worker(struct vine_manager *q, struct vine_worker_info *w, struct vine_task *t)
 {
-	/* Kill empty libraries to reclaim resources. Match the assumption of
-	 * @vine.schedule.c:check_worker_have_enough_resources() */
+	vine_result_code_t result = VINE_SUCCESS;
+
+	/* Kill unused libraries on this worker to reclaim resources. */
+	/* Matches assumption in vine_schedule.c:check_available_resources() */
 	kill_empty_libraries_on_worker(q, w, t);
```
btovar, line 2883: I think this kill should be done unconditionally only for tasks that do not require a library. Then, for tasks that require a library it should be done only if the check at line 2889 fails. Otherwise for workers running one-slot libraries, wouldn't the libraries get kill and re-send every time?

dthain: The expression is a bit twisted, but I think the effect of line 2865 is to *not* kill the library if the incoming function needs it.  Am I missing something?

The logic is a bit opaque, perhaps the expression should be rewritten to make it more clear.

btovar: You are right, now I see that kill_empty_libraries does the right thing.

btovar (approved):

Issue #3862, opened 2024-06-13
vine: unable to pack environment using poncho_package_create


JinZhou5042:
I just downloaded the ndcctools package using conda
```
conda install ndcctools
```
Then used `poncho_package_create` to pack my current python environment
```
poncho_package_create $CONDA_PREFIX coffea2024.tar.gz
```
Then tried to launch a worker using `vine_factory` with the following command
```
vine_factory -T condor -C factory.json --scratch-dir /scratch365/jzhou24 --python-env coffea2024.tar.
gz
```
and json file
```
{
    "manager-name": "jzhou24-hgg",
    "max-workers": 1,
    "min-workers": 1,
    "workers-per-cycle": 1,
    "cores": 12,
    "memory": 200000,
    "disk": 200000,
    "condor_requirements": "((has_vast))",
    "timeout": 36000
}
```
However, I got an error message showing that the factory is unable to copy the tarball into my scratch directory

```
(coffea2024) [jzhou24@crcfe02 jzhou24]$ vine_factory -T condor -C factory.json --scratch-dir /scratch365/jzhou24 --python-env coffea2024.tar.gz
Configuration file '/project01/ndcms/jzhou24/factory.json' has been loaded.manager-name: jzhou24-hgg
max-workers: 1
min-workers: 1
workers-per-cycle: 1
tasks-per-worker: 12
timeout: 36000 s
cores: 12.000
memory: 200000 MB
disk: 200000 MB

vine_factory: Cannot copy wrapper input file coffea2024.tar.gz to factory scratch directory /scratch365/jzhou24/coffea2024.tar.gz:
Invalid argument
```

I am pretty sure the `coffea2024.tar.gz` had been copied to the target place as expected. I reinstalled the environment twice but didn't help.

I used to be able to add environment with the above procedure, is there something missing?

dthain:
I don't think we have any special insight into this problem, so you will have to debug it.
I would start with the given error message:
```
vine_factory: Cannot copy wrapper input file coffea2024.tar.gz to factory scratch directory /scratch365/jzhou24/coffea2024.tar.gz:
```

And then start digging in from top to bottom:
- Do those files/directories exist?
- Can you read and write them?
- What is the function that does the file copy? (Start with `vine_factory.c`)
- Where does it fail?  (Use strace or gdb to find out.)

dthain:
[...]

PR #3863, opened 2024-06-15
vine: consistently use int64 in copy_file_to_file

JinZhou5042:
## Proposed Changes

Fix for #3862

The current returning value of `copy_file_to_file` is `int64` if success. However, there are some places receiving the returned value with `int`, which might be overflowing and causing errors. 

In this PR I simply modified some variable types when using `copy_file_to_file` so that it consistently uses `int64` to avoid overflow.

## Merge Checklist

### DIFF
--- a/batch_job/src/batch_job.c
+++ b/batch_job/src/batch_job.c
@@ -287,7 +287,7 @@ int batch_fs_mkdir (struct batch_queue *q, const char *path, mode_t mode, int re
 	return q->module->fs.mkdir(q, path, mode, recursive);
 }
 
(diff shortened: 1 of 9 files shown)


dthain (changes requested):
Good catch! 

While you are there, take a little closer look at `makeflow_module_archive.c`. It appears to (incorrectly) assume that `copy_file_to_file` returns zero on failure.  it should instead be checking for less-than-zero to indicate an error.

JinZhou5042:
RTM

Issue #3864, opened 2024-06-17
vine: zero values in time areas in transactions log


JinZhou5042:
I found in the transactions log some `CACHE_UPDATE` records the `wall_time` and `start_time` are `0`, I am guessing there is a bug in the logging function.

<img width="1182" alt="image" src="https://github.com/cooperative-computing-lab/cctools/assets/142265839/a4711418-ad04-461a-a16d-c563170b0967">

JinZhou5042:
I think this issue had been addressed in a more recent update (I extracted those lines from Connor's log file, and he was using an older version). Now I think it's a different problem:

I matched the lines in `debug` file with `cache-update .* X` where 'X' indicates a newly created temp file. It turns out that all newly created temp files have `0` in their size fields (the third field after the file name), for example
```
2024/06/15 15:31:30.33 vine_manager[4141059] vine: rx from d12chas310.crc.nd.edu (10.32.84.127:44562): cache-update temp-rnd-tuswduszjzfjxbm 1 0 0 33188 10269273 1718479837323798 X
2024/06/15 15:31:30.33 vine_manager[4141059] vine: rx from d12chas300.crc.nd.edu (10.32.84.107:36810): cache-update temp-rnd-itjtcdufkwhhjqx 1 0 0 33188 10198761 1718479836672333 X
2024/06/15 15:31:30.33 vine_manager[4141059] vine: rx from d12chas179.crc.nd.edu (10.32.83.121:43724): cache-update temp-rnd-ofdjmychbnvzzjw 1 0 0 33188 10081782 1718479837556866 X
2024/06/15 15:31:30.33 vine_manager[4141059] vine: rx from d12chas117.crc.nd.edu (10.32.82.253:52712): cache-update temp-rnd-hncnylwtftjwubs 1 0 0 33188 9853472 1718479836955804 X
2024/06/15 15:31:30.33 vine_manager[4141059] vine: rx from d12chas185.crc.nd.edu (10.32.83.133:49704): cache-update temp-rnd-ervgivalnekcwyr 1 0 0 33188 9643875 1718479837137254 X
2024/06/15 15:31:30.33 vine_manager[4141059] vine: rx from d12chas163.crc.nd.edu (10.32.83.89:51000): cache-update temp-rnd-pnowkagcgtsumdl 1 0 0 33188 10366510 1718479837350099 X
[...]

PR #3866, opened 2024-06-17
vine: pass f->size instead of size in stage_output_file

JinZhou5042:
## Proposed Changes

Fix for #3864 

## Merge Checklist


dthain:
Hmm, no I don't think that's the problem, because the size of an output file is not initially known and `f->size` should be zero.

`vine_cache_file_measure_metadata` should be measuring the size of the file, and then filling in the `size` variable appropriately.

Ah, I see now.  Take a careful look at `vine_cache_measure_metadata`.  It doesn't always fill in `size` appropriately.  Do you see why?

JinZhou5042:
I think `vine_cache_file_measure_metadata ` is properly measuring the size of the file, but that wasn't passed to the `f-size` variable...

PR #3867, opened 2024-06-18
vine: assign size to t->size in stage_output_file

JinZhou5042:
## Proposed Changes

Fix for #3864. 

Here the issue is that in the function `stage_output_file`, the parameter size is correctly passed to `vine_cache_add_file` to obtain the file size. However, this does not update the `f->size` field. In `vine_cache_add_file`, a new `f` object is created using `vine_hwache_file_create`, which is not the same as the `vine_file` object from `stage_output_file`. Therefore, despite the local `f` object using this size within `vine_cache_add_file`, it is discarded after the function ends, leaving the original `f->size` in `stage_output_file` unchanged. The correct approach would be to update the original `f->size` with `f->size = size` after the call to `vine_cache_add_file`.

## Merge Checklist

### DIFF
--- a/taskvine/src/worker/vine_sandbox.c
+++ b/taskvine/src/worker/vine_sandbox.c
@@ -191,6 +191,7 @@ static int stage_output_file(struct vine_process *p, struct vine_mount *m, struc
 				    size,
 				    mtime,
 				    transfer_time)) {
+            f->size = size;
 			vine_worker_send_cache_update(manager,
 					f->cached_name,
 					f->type,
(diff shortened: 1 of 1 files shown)


dthain:
This still doesn't make sense to make.  Take a look at what's happening in `vine_cache_add_file`.

dthain:
I see now, the confusion here is that one `f` is a `vine_cache_file` while the other is a `vine_file`.

This makes sense to me now.

But put a big fat comment in explaining why this is necessary..

JinZhou5042:
RTM if you think it looks good!

dthain (approved):

Issue #3835, opened 2024-05-14
vine: `FuturesExecutor` examples do not work


gpauloski:
I am trying to use the `FuturesExecutor` and have run into a few problems. I've enumerated them in the order that I came across them. I have opened a PR (#3836) for issues 1, 2, 4, and 5.

Issues 3 and 6 are more nuanced so hopefully someone can take a look at those.

## Installation
I've installed the latest `ndcctools` as follows:
```bash
conda create -p $PWD/test-env python=3.11
conda install -y -c conda-forge ndcctools
```

### 1. `None` return type in `FuturesExecutor.submit()`
The first example in the docs does not work (https://cctools.readthedocs.io/en/stable/taskvine/#futures).

```python
import ndcctools.taskvine as vine

def my_sum(x, y):
    return x + y

m = vine.FuturesExecutor(manager_name='my_manager')

a = m.submit(my_sum, 3, 4)
b = m.submit(my_sum, 5, 2)
c = m.submit(my_sum, a, b)  # note that the futures a and b are
                            # a passed as any other argument.

print(c.result())
```

```bash
$ python t.py
Traceback (most recent call last):
  File "/home/jgpaul/workspace/taskvine/t.py", line 13, in <module>
    print(c.result())
          ^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'result'
```
This is because `Futures.Executor.submit()` is missing a return statement if the two isinstance checks fail: https://github.com/cooperative-computing-lab/cctools/blob/930509871b6a0cf85194c2561612e6c442c035fa/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py#L68-L69

The second example works correctly because a `FuturePythonTask` is created manually, *but* this (and another) of the examples in the docs have a typo and use `FutureExecutor` instead of `FuturesExecutor`.

### 2. Adding a callback to `VineFuture` fails

This should be `_callback_fns`:
https://github.com/cooperative-computing-lab/cctools/blob/930509871b6a0cf85194c2561612e6c442c035fa/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py#L137-L138

### 3. Printing a `VineFuture` fails

While trying to print the futures returned by `FuturesExecutor.submit()`, you get the following error:
```
  File ".../lib/python3.11/concurrent/futures/_base.py", line 345, in __repr__
    with self._condition:
         ^^^^^^^^^^^^^^^
[...]

PR #3876, opened 2024-06-25
Vine: add wait and as_completed for futures

BarrySlyDelgado:
## Proposed Changes

This adds module functions `wait` and `as_completed` to the vine futures. in regard to #3835
## Merge Checklist 

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
@@ -3,6 +3,11 @@ from . import cvine
 import hashlib
 from concurrent.futures import Executor
 from concurrent.futures import Future
+from concurrent.futures import FIRST_COMPLETED
+from concurrent.futures import FIRST_EXCEPTION
+from concurrent.futures import ALL_COMPLETED
+from concurrent.futures import TimeoutError
+from collections import namedtuple
 from .task import (
     PythonTask,
     FunctionCall,
@@ -15,6 +20,7 @@ from .manager import (
 )
 
 import os
+import time
 import textwrap
 
 try:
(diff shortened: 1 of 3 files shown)


btovar:
@BarrySlyDelgado Is there an easy way to add a test for this one?

BarrySlyDelgado:
I'll create a test that tests these methods.

gpauloski:
Thanks, @BarrySlyDelgado! Is it also possible to either call `super().__init__()` in `VineFuture` or have `VineFuture` not inherit from `Future`?

I'm not sure which option is best, but I think both `repr(future)` and `future.exception()` will fail because both those methods are getting inherited. I've only briefly looked so do correct me if I'm wrong.

BarrySlyDelgado:
I think i'll have it call `super().__init()` since we're leaning into having our executor conform more to `concurrent.futures`. I'll do some testing to incorporate it.

dthain:
@BarrySlyDelgado what's the status on this one?

BarrySlyDelgado:
The branch has been tested and should work.

gpauloski:
Hi @BarrySlyDelgado, I've been testing out the branch and noticed a couple of issues.

Invoking future callbacks in `VineFuture.result()` causes infinite recursion if you access `future.result()` in the callback. (Maybe this is why the result, rather than the future, was passed to the callback originally?) It's safe to access `future.result()` in the callback of a `concurrent.futures.Future` because the callbacks are invoked in `Future.set_result()`.
```python
import time
import ndcctools.taskvine as vine

def foo() -> int:
    time.sleep(1)
    return 42

def check(future: vine.VineFuture) -> None:
    assert future.result() == 42

m = vine.FuturesExecutor(manager_name='my_manager')

t = m.submit(foo)
t.add_done_callback(check)

print(t.result())
```

Exceptions don't get set on nor raised by `VineFuture`.
```python
import ndcctools.taskvine as vine

def foo():
    raise RuntimeError('oops')

m = vine.FuturesExecutor(manager_name='my_manager')

t = m.submit(foo)

try:
    t.result()
except RuntimeError:
    print('Correctly raised exception')
else:
    raise AssertionError('Did not raise exception')
```

BarrySlyDelgado:
Thanks for letting me know, after looking at the source code the infinite recursion is due to the order in which callback functions are executed. Essentially, what is happening is that a call to `f.result()` executes callback functions if `self._ran_functions` is False and it dose not get set to True until the callbacks execute. So, a call back that also calls `f.result()` will also end up running callbacks. This can be fixed by setting `Self._ran_functions` before running the callbacks. I'll add to this PR with that change and one that raises exceptions from futures.

BarrySlyDelgado:
@gpauloski The recursion bug should be fixed. Tasks that raise an exception now raise an exception when `f.result()`is called.

gpauloski:
Thanks, @BarrySlyDelgado. The recursion is fixed for me, but I think this only raises exceptions if loading the output fails. I had to check if the output was an instance of `Exception` in `VineFuture.result()` and raise it there.

Also, I opened issue #3892 about poor performance. But it's possible the problem is not specific to the `FuturesExecutor` as I also observe it with Parsl's `TaskVineExecutor`.

Issue #3851, opened 2024-05-30
Vine: Prune Files in Dask Graph during Evaluation


cmoore24-24:
Issue: I am currently running into an issue when trying to run a process with Task/Dask Vine, wherein the process inevitably fails when the input dataset is too large. The symptoms are:
   * The process crashing because a dask task has failed 4 times, exiting the process
   * The process gets caught in a loop of recovery tasks-- producing, and presumably completing, recovery tasks indefinitely, never finishing the main process and requiring an manual interrupt

Use Case: Using Dask Vine to perform large calculations on QCD sample root files, and output the results into parquet files. The total size of the dataset(s) that fail is ~170 GB. I have independently worked around this issue by manually cutting the datasets into batches; doing one-quarter of the dataset at a time has been a successful workaround thus far.

Manager Options: 
   * Min Workers = Max Workers = 280
   * Cores = 12
   * Memory = cores*19000 = 228000 MB
   * Disk = cores*20000 = 240000 MB

I will work on recreating this scenario on my end so I can provide the vine-run-info logs.

dthain:
@BarrySlyDelgado I think you mentioned that the distributed storage plot may come in handy here, is that present in our repo so that Connor can try it out?

BarrySlyDelgado:
Once the logs are generated we should see what tasks are failing and what workers are failing (If any). I currently suspect that the the accumulation of intermediate results may cause workers to fail. This is somewhat of a knock on affect of TaskVine scheduling tasks to workers that have the most bytes of data dependencies without checking if data transfers will cause that worker to exceed its allocation. 
The tool that generates the distributed storage plot is Its not currently present but I'll add it.

cmoore24-24:
Apologies for taking so long to get logs-- for the past few days, I haven't been able to get more than ~20 workers at a time, probably a combination my bad priority and heavy use by other users. I will keep trying to make a faithful recreation of the issue for the purposes of logs.

dthain:
Hmm, does the crash not occur when using the large dataset with a small number of workers.  (Our first rough hypothesis would suggest that a crash is *more* likely with fewer workers...)

cmoore24-24:
I haven't let it go long enough to get to a crash, but I can tell you the behavior where essentially countless recovery tasks are produced occurs almost immediately.

cmoore24-24:
[...]

PR #3880, opened 2024-07-16
vine: prune stale files in daskvine

JinZhou5042:
## Proposed Changes

The first solution for #3851 

This pr implements the file pruning feature in DaskVine executor. Once the DaskVine executor figures out that an output file is no longer needed in the future, it prunes it right away so that the stale files are not piling up on the workers and the workers have more storage for subsequent tasks. 

I tested the performance before and after the pruning feature is enabled on the following task graph (a compact version of the [skimmer](https://github.com/cmoore24-24/hgg/blob/main/notebooks/skims/skimmer/skimmer.py) application)

subgraph_1

This is the per-worker disk usage(MB) over time(s) initially
worker_disk_usage
And this is after the pruning is active
worker_disk_usage (1)

As the application proceeds, files that are no longer needed will be pruned right away, rather than remaining on the worker until the final task in the dag is completed.

But there is a problem.

If a file is unexpectedly lost due to a worker crash and there is no replica in the cluster, a recovery task is submitted to restore that file. However, the input files for the recovery task might have been pruned, leading to a situation where we need to recover all file dependencies from the bottom up. This process can be time-consuming, especially when there are extremely long tasks in the recovery path.

For example, in this partial graph, where the pink represents recovery tasks, a worker crashes and we lost file `temp-rnd-euugxunbpganuad`, task `1830` is submitted. However, the input file for task 1830 had been pruned and we have to submit task `1831` and `1834` to recover the two files. Likewise, task `1832`, `1833` are submitted to recover the previous input files. 

image

Task `1834` is submitted later when the first input file is ready, which is another issue we can improve. Instead, we can submit all the recovery tasks for the input files simultaneously, allowing for more parallel processing. This will be addressed in a separate PR.

## Merge Checklist

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
@@ -275,11 +275,7 @@ class DaskVine(Manager):
                         if t.key in dsk:
                             bar_update(advance=1)
 
-                        children = dag.get_children(t.key)
-                        for c in children:
-                            if len(dag.get_pending_parents(c)) == 0:
-                                c_result = dag.get_result(c)
-                                self.prune_file(c_result._file)
+                        self._prune_cluster(dag, t.key)
                     else:
                         retries_left = t.decrement_retry()
                         print(f"task id {t.id} key {t.key} failed: {t.result}. {retries_left} attempts left.\n{t.std_output}")
@@ -394,6 +390,13 @@ class DaskVine(Manager):
         else:
             return raw
 
+    def _prune_cluster(self, dag, key):
+        children = dag.get_children(key)
+        for c in children:
+            if len(dag.get_pending_parents(c)) == 0:
+                c_result = dag.get_result(c)
+                self.prune_file(c_result._file)
+
 ##
 # @class ndcctools.taskvine.dask_executor.DaskVineFile
 #
@@ -494,8 +497,6 @@ class PythonTaskDask(PythonTask):
         self._wrapper_output_file = None
         self._wrapper_output = None
 
-        self.inputs = []
-
         args_raw = {k: dag.get_result(k) for k in dag.get_children(key)}
         args = {
             k: f"{uuid4()}.p"
@@ -516,7 +517,6 @@ class PythonTaskDask(PythonTask):
 
         for k, f in args_raw.items():
             if isinstance(f, DaskVineFile):
-                self.inputs.append(f.file)
                 self.add_input(f.file, args[k])
 
         if category:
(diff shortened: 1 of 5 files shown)


JinZhou5042:
I don't quite get the reason why I fail on the ` build-conda-linux (pull_request)` part...

JinZhou5042:
evaluation at a larger scale

with 200 workers, before pruning
worker_disk_usage

with 117 workers, after pruning
worker_disk_usage (1)

btovar:
This looks awesome!

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
```diff
@@ -272,10 +272,14 @@ def _dask_execute(self, dsk, keys):
                         if self.wrapper:
                             self.wrapper_proc(t.load_wrapper_output(self))
 
-                        result_file.garbage_collect_children(self)
-
                         if t.key in dsk:
                             bar_update(advance=1)
+
+                        children = dag.get_children(t.key)
```
btovar, line 278: Move this code to a method, like `DaskVine.prune_cluster` or something like that.

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
```diff
@@ -511,6 +494,8 @@ def __init__(self, m,
         self._wrapper_output_file = None
         self._wrapper_output = None
 
+        self.inputs = []
```
btovar, line 497: I don't think this is needed.

btovar (changes requested):

btovar:
The failure in the CI is unrelated to your pr. Submitting a fix shortly.

btovar (approved):

btovar:
Ready to merge?

JinZhou5042:
RTM

btovar:
Actually before merging, could you add an option to `DaskVine.get` called `prune_files` so that users can turn on/off this facility? The default should be False (not to prune), as that is the current behavior.

JinZhou5042:
Absoulutely!

JinZhou5042:
RTM if you think it's good.

taskvine/src/bindings/python3/ndcctools/taskvine/dask_executor.py
```diff
@@ -127,6 +128,7 @@ def get(self, dsk, keys, *,
             progress_label="[green]tasks",
             wrapper=None,
             wrapper_proc=print,
+            prune_cluster=False,
```
btovar, line 131: Let's call it `prune_files`.  I don't think that we explain the term `cluster` in the taskvine documentation, etc.

btovar (changes requested):

btovar (approved):
nice, thanks!

PR #3885, opened 2024-07-18
vine: submit all potential recovery tasks for one task in one go

JinZhou5042:
## Proposed Changes

The second solution for #3851 

While checking for available inputs, we return 0 as soon as an input file is missing and submit a relevant recovery task. For Connor's application, there are a large number of parallel tasks in the final stages, and most of them are lengthy. The final task can only run when all inputs are available. However, when considering the first several inputs' availability, some at the end may be missing and are not properly accounted for. Given that some tasks are extremely long, this increases the likelihood of losing other files during recovery. Recovering them one by one results in an endless loop.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -3145,7 +3145,7 @@ and should consider re-creating it via a recovery task.
 static int vine_manager_check_inputs_available(struct vine_manager *q, struct vine_task *t)
 {
 	struct vine_mount *m;
-	// check if all input files are available, if not, consider recovery tasks for them at once
+	/* all input files are available, if not, consider recovery tasks for them at once */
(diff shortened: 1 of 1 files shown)


JinZhou5042:
With this change, that application has a much higher chance to run to completion.

The evaluation is on the way.

btovar:
I think this change makes sense.

JinZhou5042:
For this run I had 120 workers and with the pruning feature enabled, here is the execution details
execution_details

Due to the reason that lost files can be recovered simultaneously, the final task can be executed earlier and thus the total execution time is reduced.

JinZhou5042:
RTM if @btovar thinks it's good.

btovar:
sounds good!

PR #3887, opened 2024-07-22
vine: recover lost temporary files on worker removal

JinZhou5042:
## Proposed Changes

The third solution for #3851

Currently, with tuning `m.tune("transfer-temps-recovery", 1)`, the manager is able to replicate the lost temp files to reach the threshold, for better failure tolerance. 

The way to replicate temp files on worker removal is to invoke `recall_worker_lost_temp_files`, where every temp file is enqueued into the `temp_files_to_replicate`. Files in this queue will be considered to be replicated to reach the `temp-replica-count` threshold.  

However, there are two problems:

1. If the `temp-replica-count` is not tuned, the default replica count is `0`, a temp file will be recovered when a future task claims it as an input.
2. If the `temp-replica-count` is tuned, and all workers containing a temp file are lost at the same time, as no transfer sources are available, a lost temp file will also not be recovered until a task needs it as an input.

If a lost file can be re-created earlier than when it is actually needed, we may be able to reduce the overall execution time. 

This PR addresses the second problem. I modified the logic for how the manager handles files in the `temp_files_to_replicate` queue. 

Specifically, if there are available sources, then transfer immediately from an appropriate source; if there are no sources in the cluster, we must consider a recovery task. 

A file should not appear in the recovery queue if it has already been pruned because it was deleted from the recovery queue in `vine_prune_file`.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -157,8 +157,9 @@ static struct vine_task *vine_wait_internal(struct vine_manager *q, int timeout,
 static void release_all_workers(struct vine_manager *q);
 
 static int vine_manager_check_inputs_available(struct vine_manager *q, struct vine_task *t);
-static void delete_uncacheable_files(struct vine_manager *q, struct vine_worker_info *w, struct vine_task *t);
+static void vine_manager_consider_recovery_task(struct vine_manager *q, struct vine_file *lost_file, struct vine_task *rt);
 
+static void delete_uncacheable_files(struct vine_manager *q, struct vine_worker_info *w, struct vine_task *t);
 static int delete_worker_file(struct vine_manager *q, struct vine_worker_info *w, const char *filename, vine_cache_level_t cache_level, vine_cache_level_t delete_upto_level);
 
 struct vine_task *send_library_to_worker(struct vine_manager *q, struct vine_worker_info *w, const char *name, vine_result_code_t *result);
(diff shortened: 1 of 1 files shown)


JinZhou5042:
For the first problem, I am not so sure about the meaning of `temp-replica-count`. As the default value is `0`, I'm assuming an original file has 0 replica in the cluster. If this parameter is tuned to `1`, should there be 2 copies in the cluster (1 original + 1 replica)? If it is this case, we may have some logic issue in the `vine_file_replica_table_replicate` function:

```
struct set *sources = hash_table_lookup(m->file_worker_table, f->cached_name);
if (!sources) {
return found;
}

int nsources = set_size(sources);
int to_find = MIN(m->temp_replica_count - nsources, m->transfer_replica_per_cycle);
```

When a file is newly created, the `nsources` is 1, `temp_replica_count` is also 1, and thus `to_find` is 0, no replicas will be created. Therefore, I think the default value of `temp-replica-count` should be 1?

dthain:
IMHO, the value `temp-replica-count` would make the most sense if interpreted to be "the total number of desired replicas" and **not** "number of replicas to create after a failure".

JinZhou5042:
I agree with that. That way, should we set the default value of `temp-replica-count` to 1? @dthain

btovar:
@JinZhou5042 I think so. When you do, please check all the places where it is used so we don't create two copies instead of 1.

taskvine/src/manager/vine_manager.c
```diff
@@ -5897,6 +5910,12 @@ void vine_prune_file(struct vine_manager *m, struct vine_file *f)
 			}
 		}
 	}
+
+	/*
+	Also, a file can be added into the temp_files_to_replicate
```
btovar, line 5915: /* Pruned files do not need to be scheduled for replication anymore. */

btovar (approved):
@JinZhou5042 looks good! The only change would be the replica count to start at 1.

btovar (approved):
looks good! rtm?

JinZhou5042:
RTM

PR #3889, opened 2024-07-23
vine: measure and report worker sandbox used

colinthomas-z80:
## Proposed Changes

At the worker, measure sandbox on completed processes and send this info back to manager.

This also included work on the method the worker uses to measure the sandbox. Before it was including hard linked input files, which do not belong in the sandbox measurement. 

## Merge Checklist


btovar:
Isn't the measurement already done in the function `enforce_process_limits`?

colinthomas-z80:
@btovar It is not guaranteed that the measurements from the call in the loop will be up to date or present at all.

I tested with some of the examples before the change and the sandbox_size field was never populated before the task completes. I'm not sure if it was a sequence issue or if the process spent very little time in the procs_running list

btovar:
I'd caution against measuring more often. Measuring the disk greatly increases the overhead. We even added a check for that in `enforce_process_limits`.

Further, by the time you are doing this measurement the task is already ready to be sent back. I worry about regular python tasks that ran for a few seconds that then have to wait  `max_time_on_measurement` for their sandboxes to be measured.

Another thing to consider is that there `vine_process_measure_disk(p, options->max_time_on_measurement);` will only give you correct results for sandboxes that are small in the number of files. If there are many files, the measurement will quit before `max_time_on_measurement`  This function was meant to be used for tasks that ran long when the disk could be progressively measured.

Given all that, I'm not sure what benefit we get by measuring the sandboxes in this way. Once the task is done, we only care about the size of the cache, right?

btovar:
I'm concerned that this measurement is going to create a train effect where tasks that finish around the same time have to wait for the measurement of others. Is there a way you can move the measurement for when the process finishes so that it can be done in parallel?   Also, I would have it disabled by default. My strong hunch is that this will increase the time the tasks wait to be retrieved.

If the problem is that tasks get little disk at the end, what about dividing the disk of the worker?  Say, reserve x% of the disk for task sandboxes?

colinthomas-z80:
It appears there was a bug in _enforce_process_limits_ where we were looking at a disk field attached to the process rather than the individual task. I am not sure of the usage of process->disk, however it doesn't seem to get populated.

With this change, we seem to measure disk often enough in the loop that it will report accurate values back to the manager.

colinthomas-z80:
There is a function _vine_process_compute_disk_needed_ that populates it but is never called

btovar:
nice catch!

dthain:
So does the sandbox measurement match our current understanding by skipping over linked input items?

colinthomas-z80:
This is one possible solution to accurately measuring the sandbox avoiding recursive stats of hard linked input files/directories.

I have tested this to successfully measure only task outputs in the sandbox (declared or undeclared) and reporting the disk consumed back to the manager in the complete message.

Since the sandbox measurement is only periodic there can be inaccurate reporting for very short running tasks or some other specific scenarios. I am relying on the assumption that tasks in danger of filling up their sandbox are probably long running.

taskvine/src/worker/vine_process.c
```diff
@@ -600,4 +605,160 @@ int vine_process_measure_disk(struct vine_process *p, int max_time_on_measuremen
 	return result;
 }
 
+static int vine_process_sandbox_disk_measure(struct vine_process *p, int max_secs, struct path_disk_size_info **state)
```
btovar, line 608: Since this function is similar to those in `dttools`, would it be better to add an extra argument to the `dttools` one with a list of paths to skip?

btovar (changes requested):

taskvine/src/worker/vine_process.c
colinthomas-z80: I can give that a try

colinthomas-z80: I ended up duplicating the code in dttools since I felt there are enough usages of the function elsewhere where the extra arguments would not be relevant.

dthain:
If this is RTM, please check the appropriate boxes in the PR...

colinthomas-z80:
@btovar does that change work for you?

taskvine/src/manager/vine_manager.c
```diff
@@ -590,6 +591,13 @@ static vine_result_code_t get_completion_result(struct vine_manager *q, struct v
 			}
 		}
 
+		/* Update category disk info */
```
dthain, line 594: So I think you have the right idea here of tracking the max disk space used.

But isn't that the purpose of `category->max_resource_use->disk` ?

(And if not, why not?)

dthain:
Ok, per our discussion today, go ahead and add controls for disk management separately so it makes sense to you.

Once that's all working, we will ask Ben for some advice on how to roll it into the existing resource management framework.

I think what we have learned is that disk is a critical resource (things fail if it is exhausted) and so managing cannot be optional.  The worker has to be measuring the cache and the sandboxes even if `q->monitoring_enabled` is false.

colinthomas-z80:
I confirmed that this is working to update the minimum allocation described in the category as tasks complete. 

So the proportional resources are still operating as normal. But if the disk is a constrained environment, it will keep allocating portions of the disk until it reaches a minimum which is largest sandbox size we have seen for the category. This is all in the default case, where no resources are requested by the user and the tasks are not assigned a category. 

It would be interesting to have a setting to try and pack the allocations as close to the minimum as possible, but I'm not sure how useful it would be to try and budget sandbox space on running tasks, unless there is a use case for a lot of overcommit on cores.

colinthomas-z80:
One small amendment to the policy you described:

> 2 - Always measure disk usage at task completion.

The frequency of disk measurement has not changed. We currently do it at most every 5 seconds. So short running tasks may finish before we measure them, or we may get one measurement from a task before it generates some output file and we therefore under-report.

We hold onto the max value seen with the hope that we will eventually measure at the right time. It is imperfect, but in a previous conversation we ruled out the idea of a disk measurement occurring at the end of each task. Perhaps we could bump up the frequency without too much trouble.

btovar:
I think this one is ready. Colin, does it work for you?

colinthomas-z80:
Looks good to me!

PR #3891, opened 2024-07-30
vine: fix useless recovery tasks submission

### LINKED ISSUE #3890: vine: useless recovery tasks submitted
Last week, @cmoore24-24 had an observation in which he found `m.tune("transfer-temps-recovery", 1)`  does not play well at all with `m.tune("temp-replica-count", 3)`, having them both active at the same time resulted in a run that generated almost 1000 recovery tasks and took much longer than it should do. Commenting out the latter solved the problem immediately.

The dag primarily looked like (the pink are recovery tasks (also linked by pink dashed line):
original_dag

I highlighted the **useful** tasks in orange, where their outputs are the actual files used to produce the leaves of the task graph.

highlighted_dag

In general, a file can be produced by multiple tasks, but only one of them was submitted by the user and the remaining are all recovery tasks. Among all the producers of a single file, only the one with the largest task id is useful in order to complete the whole graph.

If a recovery task is in orange, and its task id is the largest among all the producers, then it's clear that the output file was lost at a certain time and the function of recovery task is working as expected. For example:

QQ_1722287582216

However, I see many examples in this graph where strangely the original task is useful, producing a file that remains in the cluster and can be used by the final task. All other outputs produced by recovery tasks are merely redundant copies. For example:

QQ_1722287880304

That said, there might be a bug in the logic of considering recovery tasks, causing such amount of useless recovery tasks submitted...

JinZhou5042:
## Proposed Changes

Fix #3890 

I previously misunderstood the meaning of `curr_file_replication_cnt`, which should be *the count of replicated files in the current round*, rather than *the total count of replicas for the current file*. 

The `curr_file_replication_cnt` can be `0` because the only worker holding it is busy replicating or the manager reaches the transfer limit. But that doesn't mean the file was lost.

It should be working now.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_file_replica_table.c
+++ b/taskvine/src/manager/vine_file_replica_table.c
@@ -112,7 +112,7 @@ struct vine_worker_info *vine_file_replica_table_find_worker(struct vine_manager
 // trigger replications of file to satisfy temp_replica_count
 int vine_file_replica_table_replicate(struct vine_manager *m, struct vine_file *f)
 {
-	/* the number of replicas in this round */
+	/* the number of replicated copies in this round */
 	int round_replication_count = 0;
 
 	if (vine_current_transfers_get_table_size(m) >= hash_table_size(m->worker_table) * m->worker_source_max_transfers) {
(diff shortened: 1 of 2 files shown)


btovar (approved):

dthain (approved):

JinZhou5042:
RTM

PR #3898, opened 2024-07-31
Batch: Rework Interface Into Queue, Task, File Objects

dthain:
## Proposed Changes

The goal here is to rework the batch interface to use the (existing) batch_task and batch_file structures natively.  This will allow us to pass more rich information about files through the batch interface, eventually allowing makeflow to make proper use of taskvine facilities.  It will also eliminate a lot of icky string manipulations from input and output file lists and result in more robust code.

- Now uses `batch_queue`, `batch_job`, and `batch_file` objects to cleanly specify complete jobs with files specs.
- Dropped support for experimental systems: lambda, amazon-batch, mpi, mesos, blue waters.
- Dropped support for filesystem operations on batch systems, which only ever made sense on Hadoop.
- Makeflow now uses `batch_queue_submit( queue, job )` cleanly.
- Factory now uses `batch_queue_submit( queue, job )` cleanly.
- Update docs and man pages to reflect currently supported systems.

Along the way, I have decided to drop support (and docs) for several batch modules that were useful for producing some papers, but have not seen any active use and therefore have low confidence that they work correctly.  While there is some hazard to changing our old friends like `batch_queue_condor`, I am also confident that we use these frequently enough to sort out any new bugs quickly.

(While I would consider the amazon and k8s modules quite a bit less tested than the others, it's easy to imagine continuing use cases for deploying workers through them, so I would like to keep them around.)

I would like to apply code formatting to the batch library particularly, but I'll hold off on doing that for a while, to make the intended changes clear.

## Merge Checklist


dthain:
@btovar could you take a high-level view at this PR?  I realize that it is a monster, and I'm not expected you to look at every line.  But I would particularly like your feedback on the improved batch API in `batch_queue` `batch_job` and `batch_file`.

Looking forward, I am seeing some good reasons to get Makeflow back in the action running on top of Vine.  To get benefit from that, Makeflow needs to pass some file-level information down through the batch interface.  So this rework gives us the opportunity to present files cleanly to the batch interface and add more details in the future.  It also cleans up some of the unused mess, to make it easier to get some new people involved in fixing up the Amazon and K8S modules as places to run workers.

btovar (approved):
Seems like the right approach!

Issue #3892, opened 2024-07-30
TaskVine Futures: High latency on chained tasks.


gpauloski:
This is related to PR #3876, but different enough that I figured I'd open it as a separate issue to track.

I was using the `vine.FuturesExecutor` and getting pretty poor task latency (> 1s) with no-op tasks (using a compute-zen3 node at CHI@TACC). To figure out if it was my `vine.FuturesExecutor` executor, I also tested with the Parsl `TaskVineExecutor` and got similarly bad performance.

I wrote a script to test round-trip task time, and it seems the `vine.FuturesExecutor` and `TaskVineExecutor` are much slower than just creating `PythonTask` directly. Is there a reason for this performance difference or have I configured something wrong? I also see quite some variance in the results for the "executor" and "parsl" configurations (as low as 0.25 and as high as 10). The "basic" configuration is a pretty consistent 25 tasks/s.
```bash
$ python taskvine.py --tasks 100 basic
Basic -- 25.752 tasks/s
$ python taskvine.py --tasks 100 executor
Executor -- 4.396 tasks/s
$ python taskvine.py --tasks 100 parsl
Parsl -- 1.394 tasks/s
```

```python
import argparse
import time

import ndcctools.taskvine as vine

from parsl.concurrent import ParslPoolExecutor
from parsl.config import Config
from parsl.executors.taskvine import (
    TaskVineExecutor,
    TaskVineFactoryConfig,
    TaskVineManagerConfig,
)


def noop() -> None:
    pass


def basic(tasks: int) -> None:
    manager = vine.Manager(9123)
    factory = vine.Factory(manager=manager)
    factory.min_workers = 1
    factory.start()

    start = time.perf_counter()
    for i in range(tasks):
        task = vine.PythonTask(noop)
        manager.submit(task)

        while not manager.empty():
            result = manager.wait(5)
            if result:
                if isinstance(result.output, vine.PythonTaskNoResult):
                    print(f'Task {result.id} failed.')
    total = time.perf_counter() - start
    print(f'Basic -- {tasks/total:.3f} tasks/s')


def executor(tasks: int) -> None:
    with vine.FuturesExecutor(
        manager_name='taskvine-manager',
        opts={'min_workers': 1},
    ) as executor:
        start = time.perf_counter()
        for i in range(tasks):
            future = executor.submit(noop)
            future.result()
        total = time.perf_counter() - start
        print(f'Executor -- {tasks/total:.3f} tasks/s')


[...]

PR #3903, opened 2024-08-07
vine: include functions and classes in hoisting_modules

JinZhou5042:
## Proposed Changes

This PR is in part to support #3892 to run on a serverless mode. It essentially does two things:

1. Rename the `import_modules` to `hoisting_modules` to expand it's definition, so as to accept not only python packages, but also classes and functions

2. Accept class and function declarations and hoist in the library code. For example:

``` Python
import uuid
import random

class test_class():
    def __init__(self):
        print("test")

def test_func():
    return 100

libtask = engine.executor.create_library_from_functions(
    "test-library", noop_task, hoisting_modules=[uuid, random, test_class, test_func], add_env=False
)
```

This will hoist the declarations of `test_class` and `test_func` so that function calls can invoke directly.

## Merge Checklist

### DIFF
--- a/doc/manuals/taskvine/index.md
+++ b/doc/manuals/taskvine/index.md
@@ -1556,7 +1556,7 @@ You can certainly embed `import` statements within the function and install any 
     libtask = m.create_library_from_functions("my_library", divide)
     ```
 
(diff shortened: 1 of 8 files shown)


JinZhou5042:
I will propose another PR to update the doc and test examples, it's good to be merged for now so that they can try some experiments.

JinZhou5042:
RTM

PR #3877, opened 2024-07-03
vine: sum disk usage to avoid overfilling worker allocation

colinthomas-z80:
## Proposed Changes

Before we only considered the size of active task data when calculating worker->disk.inuse. 

This additionally sums the cached files currently on the worker. This presents a more accurate picture of worker disk usage.

In combination with the default proportional resource allocation, provided the estimated allocation is large enough, we will not schedule a final task to fill up the disk and make the worker exit.

However if the proportional resource allocation is not large enough we may still overfill the worker and cause it to exit.  

## Merge Checklist

The following items must be completed before PRs can be merge.
Check these off to verify you have completed all steps.

- [ ] `make test`       Run local tests prior to pushing.
- [ ] `make format`     Format source code to comply with lint policies. Note that some lint errors can only be resolved manually (e.g., Python)
- [ ] `make lint`       Run lint on source code prior to pushing.
- [ ] Manual Update     Update the manual to reflect user-visible changes.
- [ ] Type Labels       Select a github label for the type: bugfix, enhancement, etc.
- [ ] Product Labels    Select a github label for the product: TaskVine, Makeflow, etc.
- [ ] PR RTM            Mark your PR as ready to merge.

PR #3909, opened 2024-08-07
vine: remove disk check for library task

colinthomas-z80:
## Proposed Changes

In a local setting (and perhaps remote) the total worker disk will grow smaller as files are generated and will quickly go below the library task allocation if it is has been given the whole disk.

Currently the resources given to a library task by the manager are fixed, where ideally we would adjust the allocation at runtime based on available remaining disk in the same way we schedule normal tasks. However the process of doing that is somewhat of a question. 

## Merge Checklist

### DIFF
--- a/taskvine/src/worker/vine_worker.c
+++ b/taskvine/src/worker/vine_worker.c
@@ -1328,7 +1328,7 @@ static int task_resources_fit_now(struct vine_task *t)
 	return (cores_allocated + t->resources_requested->cores <= total_resources->cores.total) &&
 	       (memory_allocated + t->resources_requested->memory <= total_resources->memory.total) &&
 	       ((t->needs_library || disk_allocated + t->resources_requested->disk <= total_resources->disk.total)) && (gpus_allocated + t->resources_requested->gpus <= total_resources->gpus.total);
-		// XXX Disk is constantly shrinking, and library disk requests are currently static. Once we generate some files things will hang. 
+	// XXX Disk is constantly shrinking, and library disk requests are currently static. Once we generate some files things will hang.
 }
 
 /*
--- a/taskvine/src/worker/vine_worker.c
+++ b/taskvine/src/worker/vine_worker.c
@@ -1327,7 +1327,8 @@ static int task_resources_fit_now(struct vine_task *t)
 {
 	return (cores_allocated + t->resources_requested->cores <= total_resources->cores.total) &&
 	       (memory_allocated + t->resources_requested->memory <= total_resources->memory.total) &&
-	       (disk_allocated + t->resources_requested->disk <= total_resources->disk.total) && (gpus_allocated + t->resources_requested->gpus <= total_resources->gpus.total);
+	       ((t->needs_library || disk_allocated + t->resources_requested->disk <= total_resources->disk.total)) && (gpus_allocated + t->resources_requested->gpus <= total_resources->gpus.total);
+		// XXX Disk is constantly shrinking, and library disk requests are currently static. Once we generate some files things will hang. 
 }
 
 /*


dthain:
Sorry, I'm not sure I understand what's going on here.  Why are function calls somehow different?  They still have a sandbox directly and stage files in and out just like normal tasks.

colinthomas-z80:
I observed this while running the parsl benchmark with function invocations. It causes a silent hanging issue.

When the library task is sent over to start at the worker it has a resource specification of c,m,d, gpu. When the library task starts all of these are allocated at the worker and considered “inuse”. 

When a function invocation shows up, it has no resource specification of its own. The worker will refer to the original library resource specification when starting an invocation. Instead of limiting overcommit by resource constraints, it keeps track of function_slots and functions_running.

If the whole disk was given to the library, once the total disk available gets smaller over the course of an application, the worker will fail to invoke functions because the disk quantity given to the library task no longer exists.

colinthomas-z80:
I'm not sure we are yet that deliberate to consider cache and sandbox space and how they fit together. That should be following the work in #3889

When the worker attempts to run a task or invocation it will call the _task_resources_fit_now_. It checks the disk request against disk.total, which gets updated with something like a _df_ periodically. 

Especially evident on a local worker where things are getting written to /tmp and the manager log is filling up space the total remaining disk will shrink consistently.

This is sort of analogous to #3877, where a similar problem would occur when scheduling regular tasks and giving them the whole disk. A task was given the "whole disk", but the manager's information was stale and ended up requesting more disk than was actually remaining, causing forsaken tasks.

The difference here is that the library task is scheduled once in the beginning with a fixed allocation and is never updated. Once the output of _df_ at the worker shows less disk available than before, it cannot meet the allocation requested by the library.

colinthomas-z80:
Or formally:
The worker starts up and reports a disk size of `D`

The manager schedules the library task to request resources, including `T` amount of disk which is equal to `D`

The library starts up at the worker and starts running invocations. Periodically running a _df_ and realizes its available disk space is smaller. So it resets `D` accordingly and reports back to the manager. 

Even though `D` has grown smaller, `T` remains the same as before. The next time the worker attempts and invocation it will find `T > D` and fail to invoke the function. The manager does not realize this because invocations do not have resource requirements to be scheduled.

dthain:
Here is the way I see it:

The disk space allocation means exactly what it says: the **quantity of disk needed by the task**.  If the task requires X GB of space, then the manager should ask the worker for that.  And the worker should attempt to deliver exactly that.  If there isn't enough space for that task, then it should not be placed on the worker.  (Or it should be forsaken if already there.)

The problem is that (when the user provides no information) the manager is assuming that a task can and should consume all the remaining available space on the worker.  That's too aggressive, because the task doesn't really need that space, and we end up getting tied into knots when it is all allocated.

If the user states `task.set_disk(1000)` for a library task, then everything will work as desired.

What we need here is a different strategy for assigning a disk allocation when the user has not provided any information.  Any suggestions on how to proceed?

The current code for assigning resources to a task at the point where it is placed on a worker is here:
https://github.com/cooperative-computing-lab/cctools/blob/3a96addbd6acd28871f24ef80b1e9f07b68834e2/taskvine/src/manager/vine_manager.c#L2534

colinthomas-z80:
I think this is quite a fundamental question with no definite answer. Sort of why I made an issue along with this PR for the broader topic, since this is a bug fix for a relatively urgent issue, especially now that I see that Jin is affected with a real application. When we give the whole disk we are effectively admitting that we don't know how much disk is required. Therefore I don't see any fundamental change in this PR, since we aren't enforcing any disk minimum anyway. This will just let it keep going until we run out. 

I think allocating the whole disk is the right idea, especially by default for tasks which use all the cores. For library tasks which occupy the whole worker I think we should be able to effectively give it the whole disk without some race condition getting in our way. We just need to be able to update the library allocation as the worker available disk shrinks.

JinZhou5042:
I just tried `libtask.set_disk(1000)` and that doesn't seem to work, the library task still consumes the whole disk.

colinthomas-z80:
That is because by default if the task uses all of the worker cores it will get the whole disk, regardless of user specification.

dthain:
Right, the resource selection is a bit complicated.  Once one resource on a worker is exhausted, the rest are consumed up proportionally.  So perhaps that is also something that needs to be reconsidered.  (At least for disk.)

dthain (approved):
This is a temporary hack to get functions running while @colinthomas-z80 sorts out some more fundamental issues with storage management.

PR #3914, opened 2024-08-12
vine: do not override user resource specifications

colinthomas-z80:
## Proposed Changes

We have observed in a few different scenarios where our proportional resource calculation and other default values are not effective in running some applications, sometimes even detrimental to their function. 

If a user specifies resources for a task we should give them exactly what they asked for. Some may find it confusing to see tasks are receiving more resources than were specified. This will also give the user the control to get their application running in the event our calculation is causing some issue. 

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -2663,6 +2663,16 @@ struct rmsummary *vine_manager_choose_resources_for_task(struct vine_manager *q,
 	/* never go below specified min resources. */
 	rmsummary_merge_max(limits, min);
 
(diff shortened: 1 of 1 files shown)


dthain:
Yes, I think this makes sense -- giving more than explicitly asked for makes the user wonder if they did something wrong.  (BenC has already mentioned this.)

dthain (approved):

btovar:
I'm going to revert this pr.  If an override is needed, it should be  done through tune parameters. The way this pr did it bypasses all the checks for first allocations, etc.

PR #3916, opened 2024-08-14
vine: default worker disk to half avaialble

colinthomas-z80:
## Proposed Changes

When other processes alongside the worker are consuming disk space on the machine, tasks which request the whole worker disk will be forsaken periodically because the worker is offering the entire available disk space, and the manager does not know about the unrelated disk consumption. 

This can be mitigated by either assigning a disk value to a task, or starting the worker with a finite disk value. If the user does none of these things the problem will persist. This will set a default worker disk value of half the available disk at the time the worker starts.

## Merge Checklist

### DIFF
--- a/taskvine/src/worker/vine_worker.c
+++ b/taskvine/src/worker/vine_worker.c
@@ -386,6 +386,7 @@ and apply any local options that override it.
 
 static void measure_worker_resources()
 {
+	static int disk_set = 0;
 	static time_t last_resources_measurement = 0;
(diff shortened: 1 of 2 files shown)


dthain (commented):
Seems like `disk_set` is unnecessary?

colinthomas-z80:
In order to keep a fixed value based on available disk we need to use disk_total only once. If something else is consuming disk then that calculation will continue to return a smaller number.

dthain:
Oh, I see, it's static.  Better add a big fat comment.

dthain (approved):

PR #3919, opened 2024-08-20
vine: priority queue implementation

JinZhou5042:
## Proposed Changes

## Merge Checklist


taskvine/src/manager/vine_manager.c
```diff
@@ -4250,28 +4276,10 @@ static double vine_task_priority(void *item)
 
 /* Put a given task on the ready list, taking into account the task priority and the manager schedule. */
 
-static void push_task_to_ready_list(struct vine_manager *q, struct vine_task *t)
+static void push_task_to_ready_tasks(struct vine_manager *q, struct vine_task *t)
 {
-	int by_priority = 1;
-
-	if (t->result == VINE_RESULT_RESOURCE_EXHAUSTION) {
```
dthain, line 4257: This is an important case that needs to be addressed.
An exhausted task should run prior to those of the same priority, so that we gain accurate resource info.

taskvine/src/manager/vine_manager.c
```diff
@@ -1369,19 +1372,14 @@ static int enforce_waiting_fixed_locations(struct vine_manager *q)
...
 	struct vine_task *t;
 	int terminated = 0;
-	int count;
-
-	count = task_state_count(q, NULL, VINE_TASK_READY);
-	while (count > 0) {
-		count--;
 
-		t = list_pop_head(q->ready_list);
+	PRIORITY_QUEUE_ITERATE(q->ready_tasks, t)
```
dthain, line 1376: This looks like a problem to me:
- PRIORITY_QUEUE_ITERATE is a linear search.
- inside that iteration priority_queue_remove does another linear search to find t and remove it.

Put those together and yout have a O(n^2) algorithm.

Can you design this so that the item (already found) can be removed without a linear search?

taskvine/src/manager/vine_manager.c
```diff
@@ -3259,13 +3284,14 @@ static int send_one_task(struct vine_manager *q)
...
 
-		// Otherwise, remove it from the ready list and start it:
-		list_pop_tail(q->ready_list);
+		// Otherwise, remove it from the ready queue and start it:
+		priority_queue_remove(q->ready_tasks, t);
+
 		commit_task_to_worker(q, w, t);
 		return 1;
-	}
+	} while ((priority_queue_update_priority(q->ready_tasks, t, --t->priority)) && (t = priority_queue_scheduling_next(q->ready_tasks)));
```
dthain, line 3292: I don't think the scheduler should be changing the priority of tasks.
That information comes from the outside, and the scheduler should respect it.

taskvine/src/manager/vine_manager.c
colinthomas-z80: This routine should also break out after q->attempt_schedule_depth iterations and set a cursor to return to. It causes just as much if not overhead as send_one_task when it iterates through the whole ready list.

colinthomas-z80: However the overhead of this function is only of concern if we are using fixed_location temp files. During our previous optimization effort we conceded that fixed_location will generally be inefficient.

JinZhou5042: This logic is integrated into `priority_queue_scheduling_next`. Every time trying to check if a task is able to run, that function advances the `scheduling_cursor`, which is an attribute in the `struct priority_queue`. If it doesn't find any tasks in the current round, the scheduling cursor is automatically remembered and will iterate starting from the current state in the next round. If something interesting happens (`events > 0` in the main loop), `priority_queue_scheduling_reset` is called and the scheduling cursor is reset.

However I am not very sure if `events > 0` is enough for the purpose of restoring the scheduling cursor. Is this too strict or too lenient, or just enough? @colinthomas-z80

colinthomas-z80: I think `events > 0` is the right idea. From what I understand most of the things that increment the event counter are things that we are interested in, such as a task retrieval or new worker connecting. Is there anything that increments the event counter that would not have an impact on being able to schedule a new task?

JinZhou5042: We have these events that cause the event counter to increment
- a new worker is connected
- a worker is removed
- a task is dispatched to a worker
- a task is removed from the ready queue for whatever reasons
- a task is retrieved
- a file is replicated
- `process_pending` returns 1

JinZhou5042: I think that 4, 6 and 7 don't seem to have an impact on enabling a blocking task to be scheduled. I might be wrong because I'm not very familiar with when `process_pending` would return 1.

dthain:
Yes, this is the right idea, but several clarifications and questions:

1 - Insert has two cases:
a - insert before nodes of equal priority (used for failure of resources)
b - insert after nodes of equal priority (user at submit time)

2 - Remove a task using what key?  The current cursor position?  The integer taskid? A pointer to the task object?

3 - Get a task really has two parts:
a - Examine the task at the current cursor position, without removing it.
b - Remove a task at the current cursor position.

Now, suppose that you submit 1M tasks and want to schedule them to 1K workers.

- Which of the operations above will be performed O(1M) times?
- Which will be performed O(1M * 1K) times?
- Which ones will only be performed a small number of times?

(Just edit your comment above to elaborate on each of these.)

JinZhou5042:
Essentially there are two operations causing the reconstruction of the priority queue, the deletion and insert. 

**Delete an element**
Deleting an element from the queue essentially triggers the reconstruction of the queue from the deletion point to the end. The elements prior to the deletion point will keep as what it is. 

How to handle the cursor depends on whether its position is greater than, less than, or equal to the deletion point.

For example, in the figure below elements are arranged by their priorities (array is not sorted by priority), the 0th element is taken permanently by a sentinel:

QQ_1724354964105

1. cursor == deletion point (idx 4)
QQ_1724355127632

The current element is removed and the cursor points to a new element. In this case, the previously declined tasks won't have a better chance to be able to run. So, instead of reseting the cursor to the head, we simply decrease it by 1, to examine the new element next time.

2. cursor > deletion point (idx 2)
QQ_1724355205834

The elements before the current cursor have changed. Some tasks after the current position might move to an earlier position, breaking the assumption that front tasks were already considered. In this case, the cursor needs to be reset to the position before the deletion point.

3. cursor < deletion point (idx 5)
QQ_1724355273053

This doesn’t affect any assumption or the tasks we already iterated. So don’t do anything in this case.

From the manager’s perspective, it is simply removing an element and no need to take care of reseting the cursor. The priority queue itself should be responsible of adjusting the cursor.


**Insert an element**
Inserting a new element also triggers the reconstruction of the entire priority queue, as the new element could be positioned before, after, or exactly at the current position.

Similarly, we have three cases to think about if to reset the cursor.

1. cursor == insert point
    
Decrease the cursor in order to give respect to the new element in the next step.
    
2. cursor > insert point
    
Reset the cursor to point to the element just before the insert point.
    
3. cursor < insert point
    
Don’t do anything, because it doesn’t break any our assumptions.
    
Also, we let the interface to adjust the cursor and hide details to the manager.

JinZhou5042:
There are cases that the manager should reset the cursor manually:

- A new worker is connected - more resources available
- A task is done - more resources available
- A missing file is recovered - inputs available

JinZhou5042:
To sum up, the manager resets the cursor in three conditions:
- worker connection
- task done
- file recovery

And the data structure resets the cursor in four scenarios:
- delete/insert an element prior to the cursor
- delete/insert an element exactly at the cursor

The manager always resets the cursor to the head, while the data structure always resets to the insert/deletion point - 1.

dttools/src/priority_queue.c
```diff
@@ -0,0 +1,342 @@
...
+
+int priority_queue_size(struct priority_queue *pq)
+{
+	if (!pq)
+		return -1;
+
+	return pq->size;
+}
+
+void swap_elements(struct priority_queue *pq, int i, int j)
```
dthain, line 77: static

dttools/src/priority_queue.c
```diff
@@ -0,0 +1,342 @@
...
+{
+	struct element *temp = pq->elements[i];
+	pq->elements[i] = pq->elements[j];
+	pq->elements[j] = temp;
+}
+
+int swim(struct priority_queue *pq, int k)
```
dthain, line 84: static

dttools/src/priority_queue.c
```diff
@@ -0,0 +1,342 @@
...
+
+	while (k > 1 && pq->elements[k / 2]->priority < pq->elements[k]->priority) {
+		swap_elements(pq, k, k / 2);
+		k /= 2;
+	}
+
+	return k;
+}
+
+int sink(struct priority_queue *pq, int k)
```
dthain, line 97: static

taskvine/src/manager/vine_manager.c
```diff
@@ -3269,13 +3280,15 @@ static int send_one_task(struct vine_manager *q)
...
 
-		// Otherwise, remove it from the ready list and start it:
-		list_pop_tail(q->ready_list);
+		// Otherwise, remove it from the ready queue and start it:
+		priority_queue_remove(q->ready_tasks, t_idx);
+
 		commit_task_to_worker(q, w, t);
 		return 1;
 	}
+	PRIORITY_QUEUE_ROTATE_ITERATE(q->ready_tasks, t_idx, t);
```
dthain, line 3289: Yikes, this code is really confusing because the macro evaluates to a `while` that modifies the loop above it.  I think every person who looks at this code is going to be confused.  Can you find a way to make its purpose more obvious?  For example, can you express the same code like this:

```
SOME_MACRO(queue, index,task) {
   /* body of loop */
}
```

taskvine/src/manager/vine_manager.c
```diff
@@ -4251,36 +4265,18 @@ char *vine_monitor_wrap(struct vine_manager *q, struct vine_worker_info *w, stru
...
-	}
-
-	if (by_priority) {
-		/*If a task has a priority of 0 it gets added to the end of the ready list.*/
-		if (vine_task_priority(t) != 0) {
-			list_push_priority(q->ready_list, vine_task_priority, t);
-		} else {
-			list_push_tail(q->ready_list, t);
-		}
+		priority_queue_push(q->ready_tasks, t, DBL_MAX);
```
dthain, line 4277: No, this isn't right.  You don't want to change the priority of the task to "best possible".   You just want it to be considered first among tasks of equal priority.  Perhaps you need an operation to that effect on the queue.

dthain (changes requested):
- Is it really necessary to have three iterators in priority_queue.h?    Why not just one?
- Add a test `dttools/test/TR_priority_queue.sh` that exercises the priority queue independently of taskvine.  The data structure must be absolutely bulletproof.
- Add a big enormous comment at the beginning of `start_one_task` that explains why we are iterating over the queue in this way, and where/why the iterator should be reset.

JinZhou5042:
@benjaminlyons @colinthomas-z80 Could you please take a look at this PR and see if there is anything that is potentially causing any issues? Your suggestions would be greatly helpful as I am modifying the underlying data structure, and there may be areas where I am not very careful.

dthain:
@btovar please review

btovar (changes requested):

dttools/src/priority_queue.h
```diff
@@ -0,0 +1,235 @@
...
+
+/** Push an element into a priority queue.
+New elements are placed upper than tasks of the same priority.
+Used when a task is resubmitted given resource exhaustion, we want it to get to run as soon as possible among tasks of the same priority.
+@param pq A pointer to a priority queue.
+@param data A pointer to store in the queue.
+@param priority The specified priority with the given object.
+@return The idex of data if the push succeeded, -1 on failure.
+*/
+int priority_queue_push_upward(struct priority_queue *pq, void *data, double priority);
```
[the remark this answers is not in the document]
JinZhou5042: @btovar This version of push was pointed out by @dthain, that when [a task is resubmitted given previous resource exhaustion](https://github.com/cooperative-computing-lab/cctools/blob/a9785efa61520d1acb1a73bbb705068d5bc714d9/taskvine/src/manager/vine_manager.c#L4346C13-L4346C36), we want it to be dispatched as soon as possible in order to get more resources allocated to it. For those big tasks, we don't want to change their assigned priorities, but pushing them upward of those with the same priority is likely to get a bigger chance of scheduling earlier.

dttools/src/priority_queue.h
```diff
@@ -0,0 +1,235 @@
...
+*/
+void *priority_queue_get_head(struct priority_queue *pq);
+
+/** Get an element from a priority queue by a specified index.
+The first accessible element is at index 1.
+@param pq A pointer to a priority queue.
+@param index The index of the element to get.
+@return The pointer to the element if any, failure otherwise
+*/
+void *priority_queue_get_element(struct priority_queue *pq, int index);
```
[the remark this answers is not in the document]
JinZhou5042: @btovar The index 0 is taken by a [sentinel](https://github.com/JinZhou5042/cctools/blob/4455ee4e77f8be53048668d6b0563c10c5ef1b61/dttools/src/priority_queue.c#L129) element, I think its ambiguous if both 0 and 1 index are allowed and return the same value?

dttools/src/priority_queue.h
```diff
@@ -0,0 +1,235 @@
...
+	printf("Data idx: %d\n", idx);
+}
+
+</pre>
+*/
+
+/* Iterate from begining every time starts, ends when manually set a depth or reach the end. */
+#define PRIORITY_QUEUE_BASE_ITERATE( pq, idx, data ) priority_queue_base_reset(pq); while ((idx = priority_queue_base_next(pq)) && (data = priority_queue_get_element(pq, idx)))
+
+/* Iterate from last position, never reset. MUST munually break the iteration, otherwise it never ends. */
```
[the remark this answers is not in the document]
JinZhou5042: @btovar This is a very nice suggestion!

dttools/src/priority_queue.h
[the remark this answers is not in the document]
btovar: @JinZhou5042, shouldn't this be the way to push tasks every time then? I don't like that the priority queue abstraction is leaking. If we ever need to change the priority queue because of a future discovered bug, etc., then this will affect the way taskvine schedules tasks, and it will be difficult to figure out why.

Thus, I would remove the other push function and use only this one (after removing the _upward suffix), or increase the priority of elements that need to be pushed again. One option would be to have two priority fields, one as you have it here and another that counts how many times an element has been pushed (e.g. populated with the task try_count from taskvine.)

btovar (changes requested):
See added comments.

dttools/src/priority_queue.h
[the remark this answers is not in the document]
JinZhou5042: @btovar This makes sense, removing another push function is a better way for abstraction.

dttools/src/priority_queue.h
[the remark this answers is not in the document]
JinZhou5042: @btovar Copy that!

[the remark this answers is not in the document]
JinZhou5042: @btovar Now the pq starts with the index 0!

btovar (approved):

Issue #3920, opened 2024-08-20
Vine: errors are not correctly propagated in `FuturesExecutor`


gpauloski:
I mentioned last week that I noted some small errors I ran into the `FuturesExecutor`. I was able to patch most of them when running my experiments, so I'd be happy to spend some time today/tomorrow on a PR. I'm not quite sure about (2) though, as it's an error in the `Manager` and maybe unrelated to the `FuturesExecutor`.

1. Exceptions are not raised by `Future.result()` nor is `Future.exception()` set.

```python
from ndcctools.taskvine import FuturesExecutor

class TestException(Exception):
    pass

def success() -> bool:
    return True

def failure() -> None:
    raise TestException()

with FuturesExecutor(manager_name='exceptions-example') as executor:
    future = executor.submit(success)
    assert future.result()

    future = executor.submit(failure)
    try:
        future.result()
    except TestException:
        pass
    else:
        raise RuntimeError('Future did not raise exception')

    assert isinstance(future.exception(), TestException)
```

2. `Future.done()` does not work.

```python
from ndcctools.taskvine import FuturesExecutor

def success() -> bool:
    return True

with FuturesExecutor(manager_name='done-example') as executor:
    future = executor.submit(success)
    assert future.result()
    assert future.done()
```
```
Traceback (most recent call last):
  File "/Users/jgpaul/workspace/cctools/scratch/done.py", line 9, in <module>
    assert future.done()
           ^^^^^^^^^^^^^
  File "/Users/jgpaul/miniconda3/envs/cctools-build/lib/python3.12/site-packages/ndcctools/taskvine/futures.py", line 258, in done
    state = self._task._module_manager.task_state(self._task.id)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jgpaul/miniconda3/envs/cctools-build/lib/python3.12/site-packages/ndcctools/taskvine/manager.py", line 393, in task_state
    return cvine.vine_task_state(self._taskvine, task_id)
           ^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'ndcctools.taskvine.cvine' has no attribute 'vine_task_state'. Did you mean: 'vine_task_create'?
```

3. `Future.result(timeout=TIMEOUT)` returns `'result_pending'` instead of raising `TimeoutError`.

```python
from ndcctools.taskvine import FuturesExecutor
[...]

PR #3922, opened 2024-08-21
Vine: Improve how errors are propagated with `FuturesExecutor`

gpauloski:
## Proposed Changes

Improves how task errors and timeouts are propagated in `Future`, `wait()`, and `as_completed()` when using the `FuturesExecutor` (see #3920). The commits are in chronological order, each addressing a single class/function, if you prefer to review it that way.

A couple notes:
- In addition to fixing how `TimeoutErrors` are raised in `as_completed`, I changed it to return a generator which will reduce latency on the consumer side compared to waiting to construct an iterator from a fully constructed list of futures.
- `FutureFunctionCall.output()`/`FuturePythonTask.output()` were inconsistent in when they raised vs returned task exceptions. I changed those methods to always return task exceptions, and then left the responsibility of raising exceptions to the `Future.result()`.
- `FutureFunctionCall.output()` seems to only have access to the string representation of the exception (not the pickled exception like in `FuturePythonTask`). I construct a `FunctionCallNoResult` to return containing the string representation of the task's exception. This has the downside of not letting the user catch specific types of task exceptions (e.g., exceptions they might expect to occur), but they will at least see the true exception message and not just an empty `FunctionCallNoResult`.

**Warning:** This PR only fixes four of the five items in #3920. I didn't immediately see how to fix `Future.done()`. It calls `Manager.task_state()` which then calls `cvine.vine_task_state` but this function was removed in c1f1cd7beeb4cad8a29b3025067d67daa04ff136. Could `Future.done()` be implemented by calling `Future.result(timeout=0)`? It's probably not very efficient to do it that way---maybe there's an alternative to the remove `cvine.vine_task_state`?

Please feel free to push to my branch if you want to alter anything yourself.

## Merge Checklist

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
@@ -333,11 +333,10 @@ class FutureFunctionCall(FunctionCall):
                     if output['Success']:
                         self._saved_output = output['Result']
                     else:
-                        self._saved_output = output['Reason']
+                        self._saved_output = FunctionCallNoResult(output['Reason'])
 
                 except Exception as e:
                     self._saved_output = e
-                    raise e
             else:
                 self._saved_output = FunctionCallNoResult()
             self._output_loaded = True
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/futures.py
@@ -14,7 +14,6 @@ from collections import namedtuple
 from .task import (
     PythonTask,
     FunctionCall,
-    PythonTaskNoResult,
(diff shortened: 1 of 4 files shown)


dthain:
@BarrySlyDelgado what do you think?

BarrySlyDelgado (commented):
These changes look good to me. Thanks @gpauloski!

gpauloski:
Thanks! What should be done about the missing `cvine.vine_task_state`? I'm happy to push a fix if there is one, but maybe it should become a separate issue?

gpauloski:
Great! FWIW none of my applications use task cancelling so that won't block me (many other executors have limited support for task cancelling).

BarrySlyDelgado (approved):

dthain (approved):

PR #3928, opened 2024-09-05
Revert "user resource specifications are absolute (#3914)"

btovar:
This reverts commit c4da6557541c06ee48245d6361451b1d11fb1538.

If an override is needed, use:

m.tune("proportional-resources", 0)
m.tune("proportional-whole-tasks", 0)

(See taskvine manual for more info.)

## Proposed Changes

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -2663,16 +2663,6 @@ struct rmsummary *vine_manager_choose_resources_for_task(struct vine_manager *q,
 	/* never go below specified min resources. */
 	rmsummary_merge_max(limits, min);
 
-	/* If the user specified resources, override proportional calculation */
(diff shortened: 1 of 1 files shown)


dthain:
Let's talk to make sure that we are all on the same page here.  Adding @colinthomas-z80

dthain:
I believe that there are two distinct problems that this is meant to solve:
1 - Users (like BenC) set up tasks with specific allocations to run on a worker, then look at the logs and outcome and are baffled that the actual allocations don't match what was requested, thinking there is a bug of some kind.
2 - Even with a single task, the entire storage of a worker should not be used, because we still need space for output files, function task files, etc...

@colinthomas-z80 is that correct, or are there other reasons?

btovar:
Integrating this one to #3889.

PR #3941, opened 2024-09-25
vine: shut down workers properly

JinZhou5042:
## Proposed Changes

Workers are not appropriately shut down because they never go into the if statement.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -5197,7 +5197,7 @@ int vine_workers_shutdown(struct vine_manager *q, int n)
 	{
 		if (i >= n)
 			break;
-		if (itable_size(w->current_tasks) == 0) {
+		if (itable_size(w->current_tasks) != 0) {
 			vine_manager_shut_down_worker(q, w);
 
 			/* vine_manager_shut_down_worker alters the table, so we reset it here. */


JinZhou5042:
Uh, I found I can bring back the worker logs by manually  calling `m.workers_shutdown(0)`, which has the following definition:
``` python
##
# Shutdown workers connected to manager.
#
# Gives a best effort and then returns the number of workers given the shutdown order.
#
# @param self   Reference to the current manager object.
# @param n      The number to shutdown.  0 shutdowns all workers
def workers_shutdown(self, n=0):
    return cvine.vine_workers_shutdown(self._taskvine, n)
```

It calls the `vine_workers_shutdown` in C, which I though the parameter `n` indicates the number of workers we want to shut down, but not a try to shut down empty workers.

dthain:
Your change reverses the sense of the function:

Before, it would prefer to shutdown empty workers and skip over busy workers.  This is the desired behavior when downsizing so that you don't lose work in progress.

Now, if prefers to shutdown busy workers and skip over empty workers.  I can't think of a reason why the user would want to do that.

What are you trying to accomplish?

JinZhou5042:
Whoops, you are right, I was out of my right mind.

PR #3964, opened 2024-10-30
vine: handling temp files when peer transfers disabled

colinthomas-z80:
## Proposed Changes

If peer transfers are disabled change temp file behavior so they are returned on creation like normal files. Taskvine will still manage naming and cleanup after the workflow completes.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_file.c
+++ b/taskvine/src/manager/vine_file.c
@@ -249,8 +249,11 @@ struct vine_file *vine_file_pseudo_temp()
 {
 	// temp files are always cached at workers until explicitely removed.
 	vine_cache_level_t cache = VINE_CACHE_LEVEL_WORKFLOW;
+	uid_t uid = getuid();
 
-	return vine_file_create("temp", 0, 0, 0, VINE_FILE, 0, cache, 0);
+	char *name = string_format("temp-local-%u", uid);
+	return vine_file_create(name, 0, 0, 0, VINE_FILE, 0, cache, 0);
+	free(name);
 }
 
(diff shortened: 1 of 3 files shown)


taskvine/src/manager/vine_file.c
```diff
@@ -245,6 +245,14 @@ struct vine_file *vine_file_temp()
 	return vine_file_create("temp", 0, 0, 0, VINE_TEMP, 0, cache, 0);
 }
 
+struct vine_file *vine_file_pseudo_temp()
+{
+	// temp files are always cached at workers until explicitely removed.
+	vine_cache_level_t cache = VINE_CACHE_LEVEL_WORKFLOW;
+
+	return vine_file_create("temp", 0, 0, 0, VINE_FILE, 0, cache, 0);
```
btovar, line 253: I'm not quite sure this one will work. Wouldn't all the temp files get the same name locally at the manager? Also I suggest changing the name to `vine_file_temp_no_peers` or something like that.

btovar (changes requested):

colinthomas-z80:
I am thinking about how to add a mechanism to clean up these local temp files when the workflow completes, to make them similar in function to real temp files

dthain:
https://cctools.readthedocs.io/en/latest/api/html/taskvine_8h.html#a2d4d0d4621e65083488ccef6464f334ca898d527e82997c21744baf54409cce36

dthain:
This would seem to give all temp files from the same user the same name.
```
char *name = string_format("temp-local-%u", uid);
```
Am I missing something?

colinthomas-z80:
You are right. I was mistaken about uid vs. uuid behavior

dthain (approved):

PR #3972, opened 2024-11-06
vine: fix local-temp scheduling failure 

colinthomas-z80:
## Proposed Changes

Add a dedicated VINE_LOCAL_TEMP file flag to distinguish local temp files. 

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/taskvine.h
+++ b/taskvine/src/manager/taskvine.h
@@ -64,10 +64,11 @@ typedef enum {
 
 /** Control sharing and garbage collection behavior of file objects. **/
 typedef enum {
-	VINE_PEER_NOSHARE = 1,	  /**< Schedule this file to be shared between peers where available. See @ref
+	VINE_PEER_NOSHARE = 1,	   /**< Schedule this file to be shared between peers where available. See @ref
 				     vine_enable_peer_transfers **/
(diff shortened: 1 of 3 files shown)


dthain:
Hmm, this seems odd to me, can you explain this one in more detail?

I understand that you are substituting VINE_FILE objects for VINE_TEMP in the case where peer transfers can't be done.

And also that a task cannot run unless its input files have been materialized.

But if there is a bug for files-as-temp-substitutes, then isn't it also a bug for plain files used as intermediates in the normal way?

Also, `f->state == VINE_FILE_STATE_CREATED` should be a cheaper check than the system call `lstat`

Maybe I am missing something important here.

colinthomas-z80:
Yes it seems that disabling peer transfers breaks the ability to use regular files as intermediates. As we will try to schedule them before the files have been created. I will fix that first.

dthain:
Ah, so that problem isn't really disabling peer transfers per se, it has to do with checking input file states.

For a given task, *all* of its input files should be in `VINE_FILE_STATE_CREATED`.   And additionally, temporary files in that state should be checked to see if replicas still exist.

colinthomas-z80:
Actually disregard the peer transfer causing the problem bit, but yes there is a bug

PR #3975, opened 2024-11-07
vine: assume file created on declaration until specified as output

colinthomas-z80:
## Proposed Changes

Assuming a declared file is created until it is specified as an output allows us to do two things:
- Fail a task if its input file is missing and no source for the input file has been given
- Use regular files as intermediate data without failing in the same way, since we specify the file as an input and revert it to the pending state

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_file.c
+++ b/taskvine/src/manager/vine_file.c
@@ -87,7 +87,7 @@ struct vine_file *vine_file_create(const char *source, const char *cached_name, 
 	f->size = size;
 	f->mini_task = mini_task;
 	f->recovery_task = 0;
-	f->state = VINE_FILE_STATE_PENDING;
+	f->state = VINE_FILE_STATE_CREATED; /* Assume state created until told otherwise */
 	f->cache_level = cache_level;
 	f->flags = flags;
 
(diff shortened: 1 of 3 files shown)


colinthomas-z80:
@dthain I am not certain regular files ever worked as intermediate data, and whether or not our implementation would allow it. 

My stat check works as anticipated, so it will only send tasks where the regular file inputs are present at the manager. But now we fail one of the tests which is supposed to return "input missing", since the manager does not give up on the task when it finds the file does not exist yet. 

If we want to use regular files as intermediate data we need to tolerate these kinds of tasks. But I think we want to warn the user if their inputs are missing rather than hang waiting for more tasks to show up.

dthain:
Ah, this is interesting,  please stop by today and let's talk about it.
I think there is a distinction to be made between the existence of the file in the filesystem and the state of the file in the DAG.

colinthomas-z80:
Discussion notes:

- On file declaration assume state "created"
- when a declared file is added as an output to a task, change to state "pending"

This way we will fail (properly) if we run into a task where the input file is "created" but missing. And we will also hold off on failing if we run into a pending file that will be the output of another task.

dthain (approved):

dthain:
Looks nice and straightforward.  Anything else needed?

colinthomas-z80:
I think it’s good to go

PR #3991, opened 2024-11-25
vine: submit more tasks when daskvine manager is hungry

JinZhou5042:
## Merge Checklist


btovar:
@JinZhou5042 One issue with this solution is that `vine_hungry()` is potentially an expensive operation. #3990 makes it a little more efficient by considering only some of the waiting tasks, but it may still be slow if there are many workers connected. I got an idea on how to make this more efficient as the number of workers does not change much every time we call `vine_wait`, but I have not added it to #3990.

`vine_hungry` is supposed to return the number of tasks that would make the managernot hungry. So rather than calling `vine_hungry` every time, I suggest you call it once like `min_to_sate = manager.hungry()`, and decrease `min_to_sate`  until it gets to less than 1.

PR #3997, opened 2024-12-02
add vine_hungry wrapper

btovar:
`vine_hungry` is a potentially expensive operation if there are many workers connected or many waiting tasks. However, the hunger value only changes significantly when the number of connected workers changes, and this does not happen very often. This pr makes it so that we only call the hungry computation every 5s (by default), and in between these calls we simply keep track how many tasks have been added/removed to the ready queue since last time. 

This should make it safer to call `vine_hungry` inside a loop, like in #3991. 

## Merge Checklist


dthain:
Add a big fat comment!

JinZhou5042 (approved):

PR #4000, opened 2024-12-06
fix double reap of tasks on app failure

btovar:
@JinZhou5042, this is related to the fix you did for `&=`.
@colinthomas-z80, this is fix I was talking about that might be hiding why you tasks  get stuck.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -1297,17 +1297,29 @@ static int fetch_outputs_from_worker(struct vine_manager *q, struct vine_worker_
 		break;
 	}
 
-	if (result != VINE_SUCCESS) {
-		debug(D_VINE, "Failed to receive output from worker %s (%s).", w->hostname, w->addrport);
-		handle_failure(q, w, t, result);
-	}
+	/* reap task, delete output files, or do nothing if task succeeded. */
+	handle_failure(q, w, t, result);
 
-	if (result == VINE_WORKER_FAILURE) {
+	switch (result) {
+	case VINE_MGR_FAILURE:
+	case VINE_WORKER_FAILURE:
+		debug(D_VINE, "Failed to receive output because of worker failure at %s (%s).", w->hostname, w->addrport);
 		t->time_when_done = timestamp_get();
 		return 0;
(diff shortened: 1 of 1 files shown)


dthain (approved):
This looks good, we should be using switch-on-enum like this more consistently throughout the code.

JinZhou5042 (approved):

dthain:
Odd build failures on this one, is it just a transient issue?

The bug is a serious one, so it would be good to get the fix deployed soon.

btovar:
I agree! A simpler hot fix has been merged already (#4002). I'm trying to figure out why it fails in the docker CI but not locally.

dthain:
Ok, I figured out a few things here.

First, there is a minor deficiency in our test scripts -- if the vine manager does not run to completion, then `wait_for_file_creation` causes the whole script to exit.  And so it never gets to the point of displaying the manager and worker logs, which it would normally do if the manager exits with a failure indication.  So I just added a quick hack to display manager and worker logs regardless.

Now, the actual symptom is that the manager gets as far as running Task 14 to completion, then Task 15 is submitted and marked as READY.... and then does [nothing](https://github.com/cooperative-computing-lab/cctools/actions/runs/12713497790/job/35441528907?pr=4017#step:6:3417).  The worker, as designed, notices that nothing happens for ten seconds, and then bails out.

Task 14 is `ls -lR cctools | wc -l` and Task 15 is then `wc -l infile` where `infile` is generated by a `curl` minidisk.

dthain:
So that makes me think that either the task does not match to the worker in scheduling, or some prior action manipulated the data structures improperly.

One thing that does stand out to me as a potential problem is that the (newly modified) `handle_failure` treats `VINE_MGR_FAILURE` as a reason to call `remove_worker`.  I don't think this is right, because a manager failure encompasses things like "Couldn't open an input file locally" and does not imply something is wrong with the worker.

I am not entirely sure that is the source of the problem, but it looks suspicious to me.

@btovar I hope that is helpful and I'll pass the baton back to you on this one.

btovar:
Related to #4109. Some conditionals just consider worker vs app failures.

PR #4001, opened 2024-12-09
vine factory:always use manager.name as fallback when no host:port is available

### LINKED ISSUE #3999: vine: in-context factory uses wrong hosturl for remote workers when given manager object as argument.
Depending on how you create a factory in your application, if using condor/sge as the batch_type it may create the worker command pointing to localhost, which will not reach the manager when it is on a remote worker.

If the factory is created like this, as in some of our examples: `f = vine.Factory(manager=m)` the problem occurs.

If the factory is created like this: `f = vine.Factory(manager_name="x")` where x is the name used when creating the manager, it will work properly.

btovar:
Fix #3999 

## Proposed Changes

## Merge Checklist

### DIFF
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
@@ -1851,6 +1851,12 @@ class Factory(object):
         "condor-requirements",
     ]
 
+    # subset of command line options that need special handling once the factory object has been created.
+    _config_init_options = [
+        "batch-type",
+        "manager-name",
+    ]
+
     ##
     # Create a factory for the given batch_type and manager name.
     #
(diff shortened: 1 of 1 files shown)


colinthomas-z80:
Make sure this works even when both manager name and object are specified. like:

`Factory(manager=m, manager_name="xyz")`

btovar:
localhost:port should only be set when using batch_type local.  Is this not the case?

colinthomas-z80:
Yeah the problem I brought up in the issue was that when the manager object is passed to the factory it always invokes the worker with `localhost:port` regardless of batch type or other kwargs. So condor workers would not connect

colinthomas-z80:
It might be that I was creating the factory giving only the manager object as an argument, then modifying the factory object afterwards like `f.batch_type = "condor"`. So the batch type gets changed but it does not modify the manager_hostport

colinthomas-z80:
Could we stop using localhost and always use a proper hostname so that local and remote workers can resolve if batch_type gets changed on the fly?

btovar:
I'm more inclined to forbid changing the batch type and manager host port after __init__. This would reflect what the `vine_factory` command line does.  We should only update parameters that correspond to fields in the config file once the factory is created. The localhost:port default is meant to be user friendly for the introductory examples.

colinthomas-z80:
Sounds good to me.

colinthomas-z80:
yes!

colinthomas-z80 (approved):

PR #4004, opened 2024-12-10
vine: properly replicate temp files

### LINKED ISSUE #3996: vine: too strict worker p2p transfer limitation
`VINE_WORKER_SOURCE_MAX_TRANSFERS` defines the maximum number of ports on a worker that can be used to transfer temp files to other workers. The default value is 3.

If each temp file is assigned to replicate for 5 times, these are what will happen:
- Initially, when the file was created, all workers are free on transfer, it can immediately replicate to 4 other workers.
- When more files are gradually created, most workers are busy with transferring, and most files remain unreplicated, because the number of tasks is substantially larger than the number of workers.
- The needed number of ports is very large, say we have `n` workers, each worker is allowed to open `k` ports to replicate, and we have `m` tasks. The required ports in total is `(n-1)*k*m` throughout the lifetime.
- There is a delay from when a file is successfully replicated to when the cache update message arrives at the manager, resulting the inefficient replicating process.
- It seems when a file is successfully replicated, the ports are not freed on that worker? (Maybe I just didn't find the code)

I posted the following figure two weeks ago and thought it was because the manager prefers task dispatching and outputs retrieving, so most files don't reach their replication threshold. But after I prioritized the replication step, I got the same distribution and thought there might be something more complex.

QQ_1732742763541

JinZhou5042:
## Proposed Changes

Fix #3996

- Update the default values for a set of arguments regarding temp file replication:
    - `q->worker_source_max_transfers` from 3 to 10
    - `max_transfer_procs` from 5 to 10
    - I keep `q->transfer_replica_per_cycle` as is (which is 10), as it limits the number of replications to request per temp file per iteration, and we are not likely to replicate each temp file more than 10 times.

- Improve the logic in the replication process, which helps the manager replicate more efficiently, add more comments
- Provide an API for Python to query the replica count for a file

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_current_transfers.h
+++ b/taskvine/src/manager/vine_current_transfers.h
@@ -24,7 +24,6 @@ int vine_current_transfers_url_in_use(struct vine_manager *q, const char *source
 
 int vine_current_transfers_dest_in_use(struct vine_manager *q,struct vine_worker_info *w);
 
-
 int vine_current_transfers_wipe_worker(struct vine_manager *q, struct vine_worker_info *w);
 
 void vine_current_transfers_print_table(struct vine_manager *q);
--- a/taskvine/src/manager/vine_current_transfers.h
+++ b/taskvine/src/manager/vine_current_transfers.h
@@ -8,7 +8,7 @@ See the file COPYING for details.
 #include "uuid.h"
 
 #define VINE_FILE_SOURCE_MAX_TRANSFERS 1
-#define VINE_WORKER_SOURCE_MAX_TRANSFERS 10 // static 10 until if/when multiple transfer ports are opened up on worker transfer server
+#define VINE_WORKER_SOURCE_MAX_TRANSFERS 10
 
 char *vine_current_transfers_add(struct vine_manager *q, struct vine_worker_info *to, struct vine_worker_info *source_worker, const char *source_url);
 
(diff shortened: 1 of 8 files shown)


taskvine/src/manager/vine_current_transfers.h
```diff
@@ -8,7 +8,7 @@ See the file COPYING for details.
 #include "uuid.h"
 
 #define VINE_FILE_SOURCE_MAX_TRANSFERS 1
-#define VINE_WORKER_SOURCE_MAX_TRANSFERS 3 // static 1 until if/when multiple transfer ports are opened up on worker transfer server
+#define VINE_WORKER_SOURCE_MAX_TRANSFERS 10 // static 10 until if/when multiple transfer ports are opened up on worker transfer server
```
dthain, line 11: Remove the comment, it no longer makes sense.

Also note that these are not really "max" values but default values that can be changed with `vine_tune`

JinZhou5042:
I believe there were some logic bugs or careless `if` statements in the previous implementation. I've made the logic clearer, and with these changes, the replica count for temp files is now having the right number in every run.

Before:
image


After:

image


The arg values is definitely not the actual issue. The file replication runs very fast, even though I set the default values to be very small, I was able to have the right replica numbers at the end of the run.

dthain (approved):
Looks good!

JinZhou5042:
RTM

btovar (approved):

PR #4006, opened 2024-12-11
vine: efficient resource allocation

### LINKED ISSUE #3995: vine: scheduling inefficiency b/c task resource pre-allocation usually fails
In `check_worker_against_task`, we check if a task is able to run on a worker, this includes two essential steps:
1. Estimate the resources to be allocated for this task, in `vine_manager_choose_resources_for_task`. By default, we use a proportional technique, which chooses the max proportional cpu/memory/disk/gpu resources for the task.
2. Check if the chosen resources fit the actual worker, in `check_worker_have_enough_resources`. This compares the chosen cpu/memory/disk/gpu against the actual or current resource usage state on the worker.

However, the chosen resources are usually larger than the available resources, which results in the second step constantly failing until some tasks complete on the worker and release more resources. The first step tends to choose a larger portion of resources than the available, so the scheduling overhead seems to dominate the latency. I observed this by printing some debug messages in terminal. 

To be more clear, I ran a small version of DV5 with the current version (without any change) and my test version (select a much smaller portion of resources in `vine_manager_choose_resources_for_task`), separately, the below shows their differences:

- current version: the resource allocation is expensive and tasks run very sparse
QQ_1732740902599

- test version: the resource allocation mostly succeed and task concurrency seems good
QQ_1732740965414

In total there are 8801 tasks, 99% of them finish in 10s. Factory configuration:
```
{
    "manager-name": "jzhou24-hgg2",
    "max-workers": 32,
    "min-workers": 16,
    "workers-per-cycle": 16,
    "condor-requirements":"((has_vast))",
    "cores": 4,
    "memory": 10*1024,
    "disk": 50*1024,
    "timeout": 36000
}
```

That said, I think there is a room to reduce the scheduling latency by combining a set of techniques:
- Maybe allocate the resources for a task based on the available resources on a worker (combine the first and the second step)? 
- Track the global usable cores on all workers and one task can be considered only if there is at least one available core in the cluster.
- Improve the proportional resource allocation.

JinZhou5042:
## Proposed Changes

Fix #3995  

## Problem

Main issue: resource allocation for tasks mostly fail, which harms the concurrency and extends the workflow execution time

Tentative resource allocation policy:

- The number of cores is the first-class citizen – if there are no available cores, never attempt to consider a task or even select a worker for the task
- With proportional resource allocation, if there is one core usable, there will always be enough memory to use. 
- The overuse of worker cache is the only reason for the failure of task resource allocation if there are available cores 
- Given that we the overuse of worker cache is something not likely to happen (we usually have large enough disk), the number cores has a decisive impact on whether the resource allocation will be succeed.
- Therefore, we want to keep track of the global usable cores (or function slots) and allow task resource allocation if there is any. This ensures that most task resource allocations succeed, and that we don't waste time on considering a depth of tasks if no cores available at all.

Here is an analysis on why the allocated disk tends to be bigger than the available disk (I excluded the input size for tasks as they don't seem to matter):

<div align="center">
image

</div>

In `vine_manager_choose_resources_for_task`, using proportional resource allocation, this is how we choose disk resource for a task:

<div align="center">
image
</div>

In `check_worker_have_enough_resources`, this is how we calculate the available disk on a worker:
<div align="center">
image
</div>

Initially, there are no tasks running, `c` is totally free, so the first several tasks get larger allocated disk. As more tasks are assigned on that worker, their outputs are brought to the cache, so `c` becomes smaller. 

Say the size of cache increases `delta_c` on task `t_i` completion, now task `t_(i+1)` gets scheduled. Compared to `t_i`, we have a decrease in both disk allocation and disk available:

<div align="center">
image
</div>

Apparently, as cache being more used, the available disk tends to shrink more than the allocated disk. `disk_allocate > disk_available` happens when:

<div align="center">
image
</div>

Which is:

<div align="center">
image
</div>

When we have more tasks running, we use more cache space, `c` becomes larger, the right hand side gets smaller; we use more task sandboxes, `s` becomes larger, the left hand side gets bigger. Therefore, such inequality becomes more likely to be satisfied and that's why more disk allocations fail as we have more tasks running.

Here is an example that aligns with the analysis. I requested for 16 cores, but at each time, there are at most 15 tasks running concurrently:

image

The [csv](https://drive.google.com/file/d/1slOhOMnRsQEBFKxKtorGTxvi5GNagYEu/view?usp=drive_link) file that records the resource allocation history

## Solutions
1. The proportional resource allocation should not exceed the available resources, force a hard limit on it.
2. Keep track of the global usable cores / function slots, one task can be considered only if there are available cores or slots.

## Results

A dramatic improvement of task concurrency! 

- Original:

image

image

- With the proposed solutions

image

image

## Merge Checklist


JinZhou5042:
Before @colinthomas-z80 's comment, I didn't fully understand the problem from the perspective of algorithm design, the math was straightforward but didn't identify the underlying problem. 

What's going wrong in the code is that we are using `disk_total - cache_inuse` as the available disk to allocate. However, that's not the actual amount of available disk to use. Instead, the total size of task sandboxes should be accounted as well, which means the available disk should be: `disk_total - cache_inuse - sandboxes`.

That said, we should use `task_disk_estimate = (disk_total - disk_inuse) / proportion` to guess the disk allocation. @dthain further suggested that for the disk estimate, leaving half of it being freed would provide more disk space for incoming tasks and allow for potential cache expansion, so we have `task_disk_estimate /= 2`.

With these changes, results are very encouraging! Both concurrency and success rate for resource estimate improved significantly, and the policy is very self-consistent!

btovar:
Jin,  I don't think that `disk_total - cache_inuse - sandboxes` is correct. One way to see this is that if the tasks do not have any input files, and I want to schedule two tasks to the worker, the second task will get a smaller proportion than the first one.

I believe that cache_inuse and sandboxes are not important here and are just confusing the main issue, which is that by design, the proportional computation gives conservative allocations. This is true for all the resources, and it is my hunch that this is not an accounting problem.

I'd be more comfortable with a solution that includes all the resources. For example, check at the end if an allocation from a computed proportion would not fit in the worker, then modify it in an easy-to-explain way (e.g., divide it by 2), and let the scheduler decide if it would fit, and maybe reject the allocation. I do not think we want to allocate whatever is left, as we want automated allocations for similar tasks to be similar (about the same order of magnitude). For example, we do not want one task to be 1GB and another to be 10MB just because that's what was left.

taskvine/src/manager/vine_manager.c
```diff
@@ -2409,6 +2409,7 @@ static vine_msg_code_t handle_resources(struct vine_manager *q, struct vine_work
 		debug(D_VINE, "%s", line);
 
 		if (sscanf(line, "cores %" PRId64, &total)) {
+			q->total_available_cores += (total - w->resources->cores.total);
```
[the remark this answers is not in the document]
btovar: Or we could update the count of cores every time a worker sends a resource update / disconnects.

taskvine/src/manager/vine_manager.c
```diff
@@ -2610,14 +2611,18 @@ struct rmsummary *vine_manager_choose_resources_for_task(struct vine_manager *q,
...
 
-	/* available disk for all sandboxes */
-	int64_t available_disk = w->resources->disk.total - BYTES_TO_MEGABYTES(w->inuse_cache);
+	/* total disk for all sandboxes */
+	int64_t disk_total = w->resources->disk.total - BYTES_TO_MEGABYTES(w->inuse_cache);
 
-	/* do not count the size of input files as available.
+	/* do not count the size of input files as the total.
 	 * TODO: efficiently discount the size of files already at worker. */
-	if (t->input_files_size > 0) {
```
btovar, line 2618: We need this `if`.  `input_files_size` is -1 when unset, so you do not want to -(-1) = +1 a MB to the available disk.

taskvine/src/manager/vine_manager.c
```diff
@@ -2698,7 +2703,13 @@ struct rmsummary *vine_manager_choose_resources_for_task(struct vine_manager *q,
 			/* worker's disk is shared evenly among tasks that are not running,
 			 * thus the proportion is modified by the current overcommit
 			 * multiplier */
-			limits->disk = MAX(1, MAX(limits->disk, floor(available_disk * max_proportion / q->resource_submit_multiplier)));
+			limits->disk = MAX(1, MAX(limits->disk, floor(disk_total * max_proportion / q->resource_submit_multiplier)));
+
+			/* when the proportional resource allocation is larger than the available, use the available instead */
+			limits->cores = MIN(limits->cores, cores_available);
```
btovar, line 2709: As discussed, we may only want to update the proportion when it is larger than the *_available, then only update it to  allocated/2.  If that would not fit in *_available, that's ok.  We don't want a tasks to end up with 1MB of memory/disk.

JinZhou5042:
What is the purpose for `min` and `min_proportion`? I think `max` and `max_proportion` are used throughout the way.

btovar:
The are two kinds of resource specifications for a task. “max resources” specifies a hard limit on the resources that the task can use. “min resources” indicates the minimum of resources that a task needs to even fit the worker. In almost every case, “max resources” is the resource specification used to determine how many resources to allocate.
“min resources” comes into play when:

-	They are specified explicitly by the user through the task categories.
-	There is an observed sandbox size for a task category.
-	The automatic allocation modes have not determined yet a “max resource” allocation, but they have enough information to know that an allocation size won’t be enough for the task.

 In the code, max_proportion computes the proportional resources using “max resources,” while min_proportion does the same for “min resources.” Then there is a check that forces max_proportion to be at least min_proportion. Here is a case when this check comes into play when using the resource monitor: (It is similar for disk without the resource monitor, but you have to take into account the cache, which makes the example less clear.)

-	The worker has  10 cores and 10GB of memory.
-	The task states that it uses 1 core, either through explicit value by user, or automatic allocation by the category code.
-	Because of proportion, tasks run with 1 core and 1 GB of memory.
-	The resource monitor kills the task as the usage went to 3GB of memory.
-	When the task is to be scheduled again, max_proportion will say to use 1 core and 1 GB of memory, but min_proportion will say at least  ceiling(10/3) = 4GB of memory, which turns into 5GB if the “fit an integer number of tasks” switch is activated (this is the default). Thus, the minimum allocation would be 5 cores and 5GB of memory. This is larger than the max proportion, so it is used instead. This proportion is used for disk too, as appropriate.
-	Finally, the 5 cores allocated may be revised again. If the 1 core value was set explicitly by the user, then the limit is set to 1 core. If the limit came from automatic allocations, then we use the 5 cores value.

JinZhou5042:
@btovar Thanks Ben, that was very clear!

JinZhou5042:
Just a reminder, there was a temporary hack in #3913 in order to let Greg's application to run, but I forgot about it after that. However, that may cause potential issues for the library/function task executions. This PR mainly improves resource allocations for regular python tasks. We will discuss about the case for library/function tasks after the break and I will make another PR to fix that hack!

taskvine/src/manager/vine_manager.c
```diff
@@ -554,11 +554,13 @@ static vine_result_code_t get_completion_result(struct vine_manager *q, struct v
 			original->library_failed_count++;
 			original->time_when_last_failure = timestamp_get();
 		}
+		itable_remove(q->running_library_instances, t->task_id);
```
dthain, line 557: It makes sense to have a running library table.

But let's put this down at line 617 to keep all the core data structure changes together.

(`itable_remove` will be a no-op if it is not a library task.)

taskvine/src/manager/vine_manager.c
```diff
@@ -554,11 +554,13 @@ static vine_result_code_t get_completion_result(struct vine_manager *q, struct v
...
 		printf("Library %s failed on worker %s (%s)", t->provides_library, w->hostname, w->addrport);
```
dthain, line 558: Use `debug(D_VINE|D_NOTICE)` to have consistent handling of messages to the user.  (It will also go to the log file)

taskvine/src/manager/vine_manager.c
```diff
@@ -2602,154 +2603,117 @@ struct rmsummary *vine_manager_choose_resources_for_task(struct vine_manager *q,
 		return limits;
 	}
 
-	if (t->input_files_size < 0) {
-		vine_manager_compute_input_size(q, t);
+	/* Shortcut if the task explicitly requests all resources */
```
dthain, line 2606: This is a large rework of a critical function, can you explain in more detail what this accomplishes?

taskvine/src/manager/vine_manager.c
```diff
@@ -4682,7 +4687,7 @@ struct vine_task *send_library_to_worker(struct vine_manager *q, struct vine_wor
 	/* If watch-library-logfiles is tuned, watch the output file of every duplicated library instance */
 	if (q->watch_library_logfiles) {
 		char *remote_stdout_filename = string_format(".taskvine.stdout");
-		char *local_library_log_filename = string_format("library-%d.debug.log", q->duplicated_libraries);
+		char *local_library_log_filename = string_format("library-%d.debug.log", q->next_library_id++);
```
dthain, line 4690: Hmm, I think this will be confusing because the library ids don't look any different than task ids.

How about just use the task_id in the name of the debug file?

That will make it easy to align library debug file names with other events about that library task in the manager's debug log.

dthain:
Come on over today and let's talk through a few things that I would like to understand better.

taskvine/src/manager/vine_schedule.c
```diff
@@ -78,84 +119,77 @@ int vine_schedule_in_ramp_down(struct vine_manager *q)
...
-	This is a temporary hack in order to get Greg's application running.
-	The problem is in line with the problem pointed out in PR #3909 but a special case in the manager's end.
-
-	The problem here is that the manager always allocates the whole disk the the library task,
-	so that libtask->current_resource_box->disk is always equal to worker_net_resources->disk.total.
-	For the first task, the ti->current_resource_box->disk == 0, it is able to run because the expr in line 114 doesn't satisfy.
-	However, if the first task causes any increase in the worker's disk, worker_net_resources->disk.inuse will go above the total,
-	which causes the consistent true of the expr in line 114.
-	*/
+	/* Skip if it is a function task */
```
dthain, line 124: More detailed comment here that compatibility was checked elsewhere.

dthain:
Per our discussion today, the disk allocation should be:
```
disk = ((total - cache)*frac)/proportion
```
Where `frac` is a tunable value with a default of 0.75

taskvine/src/manager/vine_manager.c
```diff
@@ -4002,6 +4002,7 @@ struct vine_manager *vine_ssl_create(int port, const char *key, const char *cert
 	q->large_task_check_interval = VINE_LARGE_TASK_CHECK_INTERVAL;
 	q->hungry_check_interval = VINE_HUNGRY_CHECK_INTERVAL;
 	q->option_blocklist_slow_workers_timeout = vine_option_blocklist_slow_workers_timeout;
+	q->disk_allocation_throttle_factor = 0.75;
```
btovar, line 4005: `disk_proportion_available_to_task = 0.75` ?

taskvine/src/manager/vine_manager.c
```diff
@@ -5694,6 +5667,9 @@ int vine_tune(struct vine_manager *q, const char *name, double value)
 	} else if (!strcmp(name, "sandbox-grow-factor")) {
 		q->sandbox_grow_factor = MAX(1.1, value);
 
+	} else if (!strcmp(name, "disk-allocation-throttle-factor")) {
```
dthain, line 5670: Fix name to match variable!

dthain (changes requested):
One last thing, almost there!

dthain (approved):
Hooray!  Good work on a long and complex PR!

taskvine/src/manager/vine_manager.c
```diff
@@ -2625,154 +2626,118 @@ struct rmsummary *vine_manager_choose_resources_for_task(struct vine_manager *q,
...
-	}
+	/* Never go below min resources. */
+	rmsummary_merge_max(limits, min);
 
-	/* Proportionally assign the worker's resources to the task if configured. */
 	if (q->proportional_resources) {
-
 		/* Compute the proportion of the worker the task shall have across resource types. */
 		double max_proportion = -1;
-		double min_proportion = -1;
```
btovar, line 2662: Min proportion is needed when using automatic resource allocation via categories, please do not remove it.

btovar (changes requested):
There seem to be a lot of changes in this pr that are unrelated.
I suggest to close this pr and resubmit with only the following changes:
- The multiplying factor to the disk_available.
- The corresponding tune parameter.
- Updates to the tune parameter documentation in the manual.

dthain:
Hmm, are you able to run a test by going back to an earlier version of Coffea and/or DV5?

JinZhou5042:
Just in case something unexpected happens since we had a few PRs getting merged. I can test with an earlier version of Dask + DaskVine in my end, but if try to merge that PR then resource allocations for the latest cctools will remain untested.

JinZhou5042:
All features in this PR have been separated into a sequence of PRs.

PR #4035, opened 2025-01-24
vine: reserve a factor of disk when allocating resources

JinZhou5042:
## Proposed Changes

Part of #4006

Testing results using 32 4-core workers

- Before
image

- After 
image

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -5779,9 +5779,7 @@ int vine_tune(struct vine_manager *q, const char *name, double value)
 	} else if (!strcmp(name, "max-library-retries")) {
 		q->max_library_retries = MIN(1, value);
 	} else if (!strcmp(name, "disk-proportion-available-to-task")) {
-		if (value > 1 || value <= 0) {
-			q->disk_proportion_available_to_task = 0.75;
-		} else {
+		if (value < 1 && value > 0) {
 			q->disk_proportion_available_to_task = value;
 		}
 	} else {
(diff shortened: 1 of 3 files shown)


taskvine/src/manager/vine_manager.c
```diff
@@ -5772,7 +5779,12 @@ int vine_tune(struct vine_manager *q, const char *name, double value)
 
 	} else if (!strcmp(name, "max-library-retries")) {
 		q->max_library_retries = MIN(1, value);
-
+	} else if (!strcmp(name, "disk-proportion-available-to-task")) {
+		if (value > 1 || value < 0) {
+			q->disk_proportion_available_to_task = 0.75;
```
btovar, line 5784: Let's make it so that > 1 or <= 0 there is no change.

btovar (changes requested):

taskvine/src/manager/vine_manager.c
```diff
@@ -5772,7 +5779,12 @@ int vine_tune(struct vine_manager *q, const char *name, double value)
 
 	} else if (!strcmp(name, "max-library-retries")) {
 		q->max_library_retries = MIN(1, value);
-
+	} else if (!strcmp(name, "disk-proportion-available-to-task")) {
+		if (value > 1 || value <= 0) {
```
btovar, line 5783: If this condition is true, the value should not change. Do not reset to the default.

btovar (changes requested):

btovar (changes requested):

dthain (approved):

JinZhou5042:
We are alomost there. @btovar, could you confirm if factoring the disk just before `rmsummary_merge_max(limits, min);` is the right way to do it?

Are the position and the condition on proportional resources correct?

```
if (q->proportional_resources) {
	limits->disk *= q->disk_proportion_available_to_task;
}

rmsummary_merge_max(limits, min);
rmsummary_merge_override_basic(limits, t->resources_requested);

return limits;
```

btovar:
I think you want `limits->disk *= q->disk_proportion_available_to_task;` always, not only with proportional resources? The reasoning is that it is a policy for the cache?

btovar (approved):

PR #4037, opened 2025-01-25
vine: count the available resources for a task on a worker

JinZhou5042:
## Proposed Changes

Calculate the usable resources for a task on a worker and check it earlier. The avoids resource allocations if the worker doesn't have usable resources for a task at all.

This speeds up my workflows slightly, for about 1.5% 

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_schedule.c
+++ b/taskvine/src/manager/vine_schedule.c
@@ -153,7 +153,7 @@ static struct rmsummary *count_worker_available_resources(struct vine_manager *q
 	worker_available_resources->gpus = overcommitted_resource_total(q, worker_net_resources->gpus.total) - (double)worker_net_resources->gpus.inuse;
 	worker_available_resources->memory = overcommitted_resource_total(q, worker_net_resources->memory.total) - (double)worker_net_resources->memory.inuse;
 	/* No overcommit disk */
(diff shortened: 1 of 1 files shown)


btovar (changes requested):
This change is fragile to memory leaks. There are some returns that do not free the struct rmsummary created.  Unless there is a strong reason, I'd advise against it.

JinZhou5042:
The reason I do this is that even though most allocations work fine in #4035 and tasks run pretty densely, the manager still wastes a lot of time checking tasks when there aren’t any usable cores. 

The most common issue in `check_worker_against_task` is that the worker doesn’t have a free core. But it still goes through the whole resource allocation process, and this keeps happening for a bunch of tasks each time, based on the depth set by `q->attempt_schedule_depth`.

Task scheduling seems to be an expensive part of the main loop, we could save a lot of time by making this more efficient, and then the manager has more time to do other stuff like output retrieving and temp file replication.

btovar:
But this change is not reducing the number of computations? It is just moving code around and creates mem leaks?

What about adding a flag to the worker like `has_free_resources`. On reaping a task, `has_free_resources` is set to 1. On scheduling a task to the worker is set to 0 if any resource in use matches the total * overcommitment (unless it starts at 0, like gpus).

JinZhou5042:
An alternative in #4039

PR #4039, opened 2025-01-27
vine: check worker has committable resources

JinZhou5042:
## Proposed Changes

An improved version suggested by @btovar in #4037

The issue is that even though most allocations work fine in #4035 and tasks run pretty densely, the manager still wastes a lot of time checking tasks when there aren’t any usable cores.

Task scheduling seems to be an expensive part of the main loop, we could save a lot of time by making this more efficient, and then the manager has more time to do other jobs like output retrieving and temp file replication.

The most common issue in `check_worker_against_task` is that the worker doesn’t have a free core. But it still goes through the whole resource allocation process, and this keeps happening for a bunch of tasks each time, based on the depth set by `q->attempt_schedule_depth`.

This PR adds a tag for workers to note whether they have any available resources. This allows the checking process to terminate earlier.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -4991,14 +4991,17 @@ static int poll_active_workers(struct vine_manager *q, int stoptime)
 
 	BEGIN_ACCUM_TIME(q, time_status_msgs);
 
-	int i, j = 1;
+	int i;
 	int workers_failed = 0;
 	// Then consider all existing active workers
-	for (i = j; i < n; i++) {
+	for (i = 1; i < n; i++) {
 		if (q->poll_table[i].revents) {
-			if (handle_worker(q, q->poll_table[i].link) == VINE_WORKER_FAILURE) {
-				workers_failed++;
-			}
+			do {
+				if (handle_worker(q, q->poll_table[i].link) == VINE_WORKER_FAILURE) {
+					workers_failed++;
+					break;
+				}
+			} while (!link_buffer_empty(q->poll_table[i].link));
 		}
 	}
 
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
@@ -65,7 +65,8 @@ class Manager(object):
     # @param port       The port number to listen on. If zero, then a random port is chosen. A range of possible ports (low, hight) can be also specified instead of a single integer. Default is 9123
     # @param name       The project name to use.
     # @param shutdown   Automatically shutdown workers when manager is finished. Disabled by default.
-    # @param run_info_path Directory to write log (and staging if staging_path not given) files per run. If None, defaults to "vine-run-info"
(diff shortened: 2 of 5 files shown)


btovar:
Looks good. Are you still testing it?

JinZhou5042:
Yes, I am going to do some large tests to make sure they work well.

taskvine/src/manager/vine_manager.c
```diff
@@ -2829,6 +2829,39 @@ static vine_result_code_t start_one_task(struct vine_manager *q, struct vine_wor
 	return result;
 }
 
+/* Check if any of the resources is overcommitted on a given worker. */
+static int is_resource_overcommitted(struct vine_manager *q, struct vine_resource resource)
```
dthain, line 2833: Let's call this function `is_resource_fully_allocated`.

The reason is that `overcommit` already has a meaning (used by overcommitted_resource_total).  When an overcommit factor of 2 is enabled, the manager will seek to allocate 2x the resources actually available, in the interest of keeping the worker fully busy.

dthain (changes requested):

taskvine/src/manager/vine_manager.c
JinZhou5042: @dthain That makes sense!

taskvine/src/manager/vine_schedule.c
```diff
@@ -149,6 +149,46 @@ int check_worker_have_enough_disk_with_inputs(struct vine_manager *q, struct vin
 	return ok;
 }
 
+/* Check if any of the resources is defined and committable on a given worker.
+ * @param q         Manager info structure
+ * @param resource  @vine_resources.h:struct vine_resources
+ * @return 1 if the resource type is defined and can be allocated to a task, 0 otherwise.
+ */
+static int is_resource_committable(struct vine_manager *q, struct vine_resource resource)
```
btovar, line 157: You are unnecessarily copying the `struct vine_resource` data. In this case, I think it will be clearer if you do the ` resource.total > 0 && resource.inuse < overcommitted_resource_total(q, resource.total);` per resource inside `check_worker_have_committable_resources`.

taskvine/src/manager/vine_schedule.c
```diff
@@ -149,6 +149,46 @@ int check_worker_have_enough_disk_with_inputs(struct vine_manager *q, struct vin
...
+ * If it returns true, the worker has either free slots for function calls or sufficient resources for regular tasks.
+ * @param q         Manager info structure
+ * @param w The worker info structure.
+ */
+static int check_worker_have_committable_resources(struct vine_manager *q, struct vine_worker_info *w)
+{
+	/* If any slots are committable */
+	uint64_t task_id;
+	struct vine_task *t;
+	ITABLE_ITERATE(w->current_libraries, task_id, t)
```
btovar, line 173: Maybe first check that `itable_size > 0` before setting the iteration?

btovar (changes requested):

btovar (approved):

btovar:
ready to merge?

JinZhou5042:
RTM

PR #4041, opened 2025-01-27
vine: a manager method to rename runtime directory

### LINKED ISSUE #3942: vine: rename logs
In TaskVine, we can assign a path as the top-level entry for all logs, like
```
m = Manager(run_info_path="vine-run-info")
```

But there is no method to rename each `yyyy-mm-dd` formatted entry under this directory, such as:

```
drwxr-xr-x 4 jzhou24 campus 4.0K Sep 26 13:22 2024-09-26T132240
drwxr-xr-x 4 jzhou24 campus 4.0K Sep 26 13:23 2024-09-26T132301
drwxr-xr-x 4 jzhou24 campus 4.0K Sep 26 13:23 2024-09-26T132315
drwxr-xr-x 5 jzhou24 campus 4.0K Sep 26 13:23 2024-09-26T132357
lrwxrwxrwx 1 jzhou24 campus   51 Sep 26 13:23 most-recent -> /scratch365/jzhou24/vine-run-info/2024-09-26T132357
drwxr-xr-x 2 jzhou24 campus 4.0K Sep 26 13:22 vine-cache
```

Adding a method to rename these subdirectories directly in the code would make it easier to label each entry based on specific settings, and can relieve users from distinguishing logs on checking.

JinZhou5042:
## Proposed Changes

Fix #3942

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_runtime_dir.c
+++ b/taskvine/src/manager/vine_runtime_dir.c
@@ -163,3 +163,9 @@ void vine_set_runtime_info_path(const char *path)
 	assert(path);
 	vine_runtime_info_path = xxstrdup(path);
 }
+
+void vine_set_runtime_info_dir(const char *dir)
+{
+	assert(dir);
+	setenv("VINE_RUNTIME_INFO_DIR", dir, 1);
+}
--- a/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
+++ b/taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
@@ -65,7 +65,8 @@ class Manager(object):
     # @param port       The port number to listen on. If zero, then a random port is chosen. A range of possible ports (low, hight) can be also specified instead of a single integer. Default is 9123
     # @param name       The project name to use.
     # @param shutdown   Automatically shutdown workers when manager is finished. Disabled by default.
-    # @param run_info_path Directory to write log (and staging if staging_path not given) files per run. If None, defaults to "vine-run-info"
+    # @param run_info_path Directory to archive workflow log directories, it is the upper level directory to run_info_dir. If None, defaults to "vine-run-info"
+    # @param run_info_dir  Directory to write log (and staging if staging_path not given) files per run. If None, defaults by a %Y-%m-%dT%H%M%S format.
     # @param staging_path Directory to write temporary files. Defaults to run_info_path if not given.
     # @param ssl        A tuple of filenames (ssl_key, ssl_cert) in pem format, or True.
     #                   If not given, then TSL is not activated. If True, a self-signed temporary key and cert are generated.
@@ -78,6 +79,7 @@ class Manager(object):
                  name=None,
                  shutdown=False,
                  run_info_path="vine-run-info",
+                 run_info_dir=None,
                  staging_path=None,
                  ssl=None,
                  init_fn=None,
@@ -113,6 +115,9 @@ class Manager(object):
         try:
             if run_info_path:
                 self.set_runtime_info_path(run_info_path)
+            if run_info_dir:
+                os.environ["VINE_RUNTIME_INFO_DIR"] = run_info_dir
+                # self.set_runtime_info_dir(run_info_dir)
 
             self._stats = cvine.vine_stats()
             self._stats_hierarchy = cvine.vine_stats()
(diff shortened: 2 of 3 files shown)


JinZhou5042:
The problem I can foresee is that if the users forget to change the argument `run_info_dir` between different runs, their previous files might be appended with the new log text.

Do we want to add a time-formatted suffix if there has been a directory with the specified `run_info_dir` name?

JinZhou5042:
Yeah, I agree with you, so we can do it in two ways

- Set `os.environ["VINE_RUNTIME_INFO_DIR"] = xxx` in Python before creating the manager
- Specify `run_info_dir` as one of the manager's arguments, as in this pr

@btovar Which one do you prefer? Is the conflict caused by this env var a problem that needs to be fixed?

btovar:
That's what `run_info_path` is for, right?

btovar:
I'm afraid that the distinction between the names/functions of these two is very small and it will cause confusion.

JinZhou5042:
My directory tree is like this:
```
vine-run-info
	├── 2025-01-27T163604
	│   ├── library-logs
	│   └── vine-logs
	│       ├── debug
	│       ├── performance
	│       ├── taskgraph
	│       ├── transactions
	│       └── workflow.json
	├── 2025-01-28T095429
	│   ├── library-logs
	│   └── vine-logs
	│       ├── debug
	│       ├── performance
	│       ├── taskgraph
	│       ├── transactions
	│       └── workflow.json
	├── 2025-01-28T095515
		├── library-logs
		└── vine-logs
			├── debug
			├── performance
			├── taskgraph
			├── transactions
			└── workflow.json
```

Where `run_info_path` represents `vine-run-info` and `run_info_dir` stands for `2025-01-27T163604`, `2025-01-28T095429` and `2025-01-28T095515`.

dthain:
I would prefer to avoid the use of environment variables.  In my experience they lead to user confusion and troubleshooting difficulties.  Better to make options explicit as command-line arguments or arguments to functions, so that users can clearly see where things are going and why.

(Limited exceptions for things like `TCP_PORT_RANGE` that have a common effect across many programs.)

btovar:
@JinZhou5042 I see that the mess of names it's actually my fault and that you are just copying the names from C.  We should look for a better name for runtime_dir so that it is clear is just for this manager, and that it is relative to the other path.

JinZhou5042:
Oh, now I'm understanding it deeper! @btovar Do you want me to proceed this PR with a better name, or address the underlying naming policies in a separate PR?

btovar:
let's keep this pr and find better names.

JinZhou5042:
Sounds good!

dthain:
As a general rule, I would prefer to keep the current C API which (mostly) separates object creation from configuration:
```
m = vine_create(port);
vine_set_password(m,password);
vine_set_scheduler(q,SCHEDULE_TO_FILES);
```
The idea here is that `vine_create` does very little beyond just creating the data structures, and then we have the freedom to add new features and capabilities without constantly re-creating the data structures.

In this particular case, is there a compelling reason why we can't do this?

```
m = vine_create(port);
vine_set_runtime_info(m, path);
```
With that said, I have no objection if the Python constructor looks like this:
```
m = taskvine.Manager(port,password=p,scheduler=s,runtime_info=path)
```
Because that easily translates into the structure above.

btovar:
The reason is that changing the runtime dir once the manager is up it is not really supported. We would need to check that no file or task refers to the old directory. I rather have it so that it is a parameter that can't be changed once set (like we do with the manager's port).

dthain:
So then let's bite the bullet and make it a proper argument to the constructor.

(But we can't do that for everything!)

JinZhou5042:
Sounds like the right way to do it! How do we want to name the two parameters about the runtime path and the runtime directory?

btovar:
`run_info_path` and `run_info_template`?

btovar:
`run_info_path` remains the same. `run_info_dir` is the one that becomes `run_info_template`.

JinZhou5042:
The main changes are:

- `run_info_path` specifies the directory where log entries across different runs are archived
- `run_info_template` specifies the workflow related directory where debug logs and staging directory are written into
  - If `run_info_template` has alreay exists, append a suffix with the current time, by a `%Y-%m-%dT%H%M%S` format, This avoids logs from different runs from being written to the same file when users forget to update the argument for a new run
  - If `run_info_template` is not set, explicitly unset that environmental variable, to avoid potential issues caused by leftover env vars from the previous runs
- `set_runtime_info_path`is removed from the Python function, as it doesn't seem to be callable once the manager is initialized

taskvine/src/manager/vine_runtime_dir.c
```diff
@@ -163,3 +163,38 @@ void vine_set_runtime_info_path(const char *path)
 	assert(path);
 	vine_runtime_info_path = xxstrdup(path);
 }
+
+void vine_set_runtime_info_template(const char *dir)
```
btovar, line 167: Setting the template should only modify the value from `"%Y-%m-%dT%H%M%S"`.  `vine_set_runtime_info_template` should be called inside the other function to actually create the directories.

btovar (changes requested):

JinZhou5042:
Looks great!

taskvine/src/manager/vine_runtime_dir.c
btovar: Correct, we are setting the variables just before the manager creation.

PR #4042, opened 2025-01-28
dttools: priority queue data structure fix

JinZhou5042:
## Proposed Changes

Some clarities about the data structure:

- This data structure only holds references to the data—it doesn’t actually delete them when the structure is destroyed.
- When a function fails, it returns either -1 if it normally provides an index, or 0 otherwise.
- In `priority_queue_remove`, the data structure is adjusted via `sink` or `swim`, depending on whether the new priority is higher than the old one.
- Enqueue the data and the corresponding priority if called `priority_queue_update_priority` but didn't find the data.
- Change name `peak` to `peek`

## Merge Checklist

### DIFF
--- a/dttools/src/priority_queue.c
+++ b/dttools/src/priority_queue.c
@@ -13,8 +13,6 @@ See the file COPYING for details.
 #include <float.h>
 
 #define DEFAULT_CAPACITY 127
-#define MAX_PRIORITY DBL_MAX
-#define MIN_PRIORITY DBL_MIN
 
 struct element {
 	void *data;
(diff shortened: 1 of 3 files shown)


dthain:
What's the severity of the change to remove()?    Is this fixing a bug, improving the performance, or just clarifying?

JinZhou5042:
This change to remove is a bug fixing.

The pq maintains a binary tree structure where every node is strictly bigger than its two child nodes.

When an element is removed from the tree, that node is replaced with the last node. But it might be either bigger or smaller than the removed one. 

If it is bigger, the lower part of the tree is satisfied, we should call `swim` to format the upper part; If it is smaller, the upper part of the tree is satisfied, we should call `sink` to adjust the lower part.

btovar (approved):

Issue #4043, opened 2025-01-28
vine: library functions in remote execution cannot be composed


btovar:
This issue came up in #4011.  Say a user defines a library with function `fn` and wants to use it in a `map`. The auxiliary function for chunks in map would also need to run remotely, which means that either it should be added to the library, or be included in the library support code from poncho.  When running library functions, their arguments are wrapped in an event json rpc, thus it becomes difficult to compose them.

One solution would be to make the auxiliary functions understand the json rpcs, but that would make further changes to code difficult. Another option is to modify the `@remote_execution` decorator so that by default it calls the original function, and only calls the version with the rpc json arguments when called like `fn.wrapped`.  With this, only the top call in `library_code` would use `fn.wrapped`.

PR #4044, opened 2025-01-28
vine: Compose remote function calls

btovar:
For #4043.

## Merge Checklist

### DIFF
--- a/poncho/src/poncho/library_network_code.py
+++ b/poncho/src/poncho/library_network_code.py
@@ -50,7 +50,7 @@ class LibraryResponse:
 
 # A wrapper around functions in library to extract arguments and formulate responses.
 def remote_execute(func):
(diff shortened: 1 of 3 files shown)


dthain (commented):
I have no objection, but I also haven't deeply understood the problem yet.  Perhaps ask @tphung3 who has been thinking about functions and context?

tphung3:
I just tested the new test code and the serverless_composed test runs with the current master (without the proposed changes). I believe this is because cloudpickle inspects each function's possible call stack and also pickles the functions that the original function depends on. So composed functions should not be a problem. @btovar Can you confirm this on your end? If true then perhaps the problem is something else?

btovar:
Thanks @tphung3! I'm trying to figure out what not worked before.

PR #4045, opened 2025-01-28
(WIP): find worker by files

JinZhou5042:
## Proposed Changes

As Connor has been complained about for several months, there was a significant slowdown towards the end of the workflow that has more than 100K tasks and terabytes of data.

@dthain and I discussed the inefficiency in the algorithm that iterates through all workers to select the one with the largest inputs, which could be very time consuming when each worker holds numerous files.

Although the manager uses `hash_table_lookup` to locate files, which is about `O(1)`, it still traverses every worker and calls `check_worker_against_task` for each one, which could be improved be first sorting workers by the task's input sizes and then calling `check_worker_against_task` in descending order of likelihood.

As for the the dramatic slowdown towards the workflow's final phases, my impression was that the problem was solved by simply switching to the `find_worker_by_random`, but as the `find_worker_by_files` is not actually iterating the file table I need to investigate it further...

## Merge Checklist

### DIFF
--- a/dttools/src/debug.c
+++ b/dttools/src/debug.c
@@ -34,9 +34,11 @@ extern int debug_file_reopen(void);
 extern int debug_file_close(void);
 
 static void (*debug_write)(int64_t flags, const char *str) = debug_stderr_write;
-static pid_t (*debug_getpid)(void) = getpid;
+static pid_t (*debug_child_getpid)(void) = 0;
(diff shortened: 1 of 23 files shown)


JinZhou5042:
We may no longer need this as the issue is not that traversing files is expensive, but transfers get stuck, and was addressed in #4076

Issue #4038, opened 2025-01-26
vine: occasionally the last reduction task is retained


JinZhou5042:
I've occasionally run into the case where most of the tasks were scheduled and retrieved normally, while the last reduction task gets stuck in the waiting list.

<img width="1578" alt="Image" src="https://github.com/user-attachments/assets/72ff6fe4-3e5c-4e5f-aa59-f6da9aeefab0" />

dthain:
What else do you know about the final task?  Does the log show that it executed and was evicted, or never ran at all?

JinZhou5042:
A few tasks are getting blocked in the waiting queue

![Image](https://github.com/user-attachments/assets/ac862ee3-bb47-4491-b516-1fe63ba4831d)

I am using `tail -f` to monitor the debug file, it is weird that task 5604 is repeating the state circle `READY` -> `RUNNING` -> `WAITING_RETRIEVAL` -> `READY` over and over again

```
2025/01/30 09:47:31.76 vine_manager[569695] vine: Task 5604 state change: READY (1) to RUNNING (2)
2025/01/30 09:47:31.76 vine_manager[569695] vine: rx from d12chas054.crc.nd.edu (10.32.81.233:52918): alive
2025/01/30 09:47:31.76 vine_manager[569695] vine: rx from d12chas054.crc.nd.edu (10.32.81.233:52918): resources
2025/01/30 09:47:31.76 vine_manager[569695] vine: cores 20
2025/01/30 09:47:31.76 vine_manager[569695] vine: memory 20480
2025/01/30 09:47:31.76 vine_manager[569695] vine: disk 307201
2025/01/30 09:47:31.76 vine_manager[569695] vine: gpus 0
2025/01/30 09:47:31.76 vine_manager[569695] vine: workers 1
2025/01/30 09:47:31.76 vine_manager[569695] vine: tag 8225
2025/01/30 09:47:31.76 vine_manager[569695] vine: end
2025/01/30 09:47:31.76 vine_manager[569695] vine: rx from d12chas054.crc.nd.edu (10.32.81.233:52918): complete 40 0 0 0 0 0 0 8225
2025/01/30 09:47:31.76 vine_manager[569695] vine: Task 8225 state change: RUNNING (2) to WAITING_RETRIEVAL (3)
[...]

PR #4050, opened 2025-01-30
Vine: Worker transfer limit must be larger than manager limit.

JinZhou5042:
## Proposed Changes

For part of #4038

## Merge Checklist

### DIFF
--- a/taskvine/src/worker/vine_transfer_server.h
+++ b/taskvine/src/worker/vine_transfer_server.h
@@ -10,7 +10,11 @@ See the file COPYING for details.
 #include "vine_cache.h"
 #include "link.h"
 
-#define VINE_TRANSFER_PROC_MAX_CHILD 8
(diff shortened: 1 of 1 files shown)


dthain:
The point here is that VINE_WORKER_TRANSFER_LIMIT cannot be smaller than q->worker_source_max_transfers.  (Otherwise transfers fail and tasks get forsaken even when the manager is doing the right thing.)

btovar:
I think that this value should be defined in `vine_manager.h` instead. In that way, it will be available to ensure a correct bound when using `vine_manager_tune()`.

dthain:
I like the idea of connecting the two together clearly in the code.

dthain (approved):
I think this is the right change to solve the immediate problem.  We can contemplate ways to make this more configurable from the manager down the road.

PR #4052, opened 2025-02-02
Debug: Remove Excess System Calls

dthain:
## Proposed Changes

On Linux, every single debug() event was resulting in the following system calls:
```
stat("/etc/localtime")
getpid()
getpid()
getpid()
write(fd,event)
```
With the sort of extensive debugging output emitted by taskvine, the cost of debug messages was becoming surprisingly high.  This PR reduces the cost to:
```
getpid()
write(fd,event)
```
Note that getpid() was previously cached and the system call optimized away, but this is no longer true:
https://sourceware.org/glibc/wiki/Release/2.25#pid_cache_removal

It turns out that it is surprisingly hard to catch all variations of fork() in order to invalidate such a cached value.

Maybe we don't need to write it to every debug line?

## Merge Checklist

### DIFF
--- a/dttools/src/debug.c
+++ b/dttools/src/debug.c
@@ -190,9 +190,9 @@ static void do_debug(int64_t flags, const char *fmt, va_list args)
 		the minimum latency of a debug event.
 		*/
 
-		if(!debug_time_zone_cached) {
-			if(!getenv("TZ")) {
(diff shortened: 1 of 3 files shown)


btovar (approved):

dthain:
Merge when you are ready...

PR #4083, opened 2025-03-04
vine: remove items after iterating hash table

JinZhou5042:
## Proposed Changes

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_current_transfers.c
+++ b/taskvine/src/manager/vine_current_transfers.c
@@ -214,15 +214,25 @@ int vine_current_transfers_wipe_worker(struct vine_manager *q, struct vine_worke
 		return removed;
 	}
 
+	char **ids_to_remove = malloc(hash_table_size(q->current_transfer_table) * sizeof(char *));
+	int n_ids = 0;
+
 	char *id;
 	struct vine_transfer_pair *t;
-	HASH_TABLE_ITERATE(q->current_transfer_table, id, t)
-	{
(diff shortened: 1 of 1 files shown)


taskvine/src/manager/vine_current_transfers.c
```diff
@@ -214,16 +214,26 @@ int vine_current_transfers_wipe_worker(struct vine_manager *q, struct vine_worke
 		return removed;
 	}
 
+	char **ids_to_remove = malloc(hash_table_size(q->current_transfer_table) * sizeof(char *));
```
btovar, line 217: I'd suggest using a `struct list` instead. It is what we do in the rest of code and will ease the life of future developers.

btovar (changes requested):

dthain:
Removing things from a hash table inside of an iteration seems like a common operation that should be supported more directly.  Is there a good abstraction that we can provide so that it's easy for the caller to get it right?

Ian2327:
After running blast.py using this fix, the number of valgrind errors are reduced from 411572 to 244766. The 2 major sources of errors are still the same, those involving either `vine_manager_choose_resources_for_task` or `vine_file_replica_table_remove`

btovar:
> Removing things from a hash table inside of an iteration seems like a common operation that should be supported more directly. Is there a good abstraction that we can provide so that it's easy for the caller to get it right?

We could add a delayed delete and then a commit deletes. I'm not sure we can do much better than what @JinZhou5042 did here, though.

PR #4099, opened 2025-04-01
vine: dangling pointer, crashes worker

JinZhou5042:
## Proposed Changes

I met a consistent segfault error when running a worker locally which performs frequent temp file staging in/out.

It turned out that there was a bug in the `copy_file_to_buffer` function when handling error cases. The function would free the buffer on read failures but not set it to `NULL`, creating a dangling pointer. Specifically, when `full_read` failed, the function would free `*buffer` without nullifying it, then return `-1`. 

``` C
	int64_t total = full_read(fd, *buffer, *len);
	if (total == -1) {
		printf("copy_file_to_buffer: total == -1\n");
		free(*buffer);
	}
```

The calling code's condition `if (copy_file_to_buffer(...))` was problematic because in C, any non-zero value (including -1) evaluates to true. This meant that with a return value of `-1`, the code would incorrectly enter a branch that assumed a valid buffer, eventually calling `strlen()` on the dangling pointer and later attempting to `free` it again, resulting in the crash.

## Merge Checklist

### DIFF
--- a/dttools/src/copy_stream.c
+++ b/dttools/src/copy_stream.c
@@ -124,7 +124,6 @@ int64_t copy_file_to_buffer(const char *filename, char **buffer, size_t *len)
 	}
 
 	*len = info.st_size;
-	*buffer = NULL;
 	*buffer = malloc(*len + 1);
 	if (*buffer == NULL) {
 		close(fd);
--- a/dttools/src/copy_stream.c
+++ b/dttools/src/copy_stream.c
@@ -124,6 +124,7 @@ int64_t copy_file_to_buffer(const char *filename, char **buffer, size_t *len)
 	}
 
 	*len = info.st_size;
+	*buffer = NULL;
 	*buffer = malloc(*len + 1);
 	if (*buffer == NULL) {
 		close(fd);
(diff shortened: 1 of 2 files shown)


dttools/src/copy_stream.c
```diff
@@ -124,6 +124,7 @@ int64_t copy_file_to_buffer(const char *filename, char **buffer, size_t *len)
 	}
 
 	*len = info.st_size;
+	*buffer = NULL;
```
btovar, line 127: This one is not needed.

btovar (changes requested):

dthain:
No harm in zeroing the buffer.  But the real problem here is that the caller assumed that `copy_file_to_buffer` returned zero on failure.  It (apparently) returns -1.

JinZhou5042:
Yes, I agree. Though the function `copy_file_to_buffer` is used in many places, and the return value is checked inconsistently ( < 0, > 0, < 1). I am concerned that modifying the original implementation risks creating a chain of downstream effects.

For example:

1. image

2. image

3. image

4. image

JinZhou5042:
That makes sense to me. In this case, [this line](https://github.com/cooperative-computing-lab/cctools/blob/c4664874c2be40cd347d08699c5a8c7e93e4f737/taskvine/src/manager/vine_manager.c#L6477) also looks problamatic?
image

btovar (approved):

PR #4104, opened 2025-04-04
Valgrind error

Ian2327:
## Proposed Changes

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_file_replica_table.c
+++ b/taskvine/src/manager/vine_file_replica_table.c
@@ -51,6 +51,11 @@ int vine_file_replica_table_insert(struct vine_manager *m, struct vine_worker_in
 // remove a file from the remote file table.
 struct vine_file_replica *vine_file_replica_table_remove(struct vine_manager *m, struct vine_worker_info *w, const char *cachename)
 {
+	if (!w->current_files || !m->file_worker_table) {
+	    // Handle error: invalid pointer
+	    return 0;
+	}
+
 	if (!hash_table_lookup(w->current_files, cachename)) {
 		return 0;
 	}
--- a/taskvine/src/manager/vine_file_replica_table.c
+++ b/taskvine/src/manager/vine_file_replica_table.c
@@ -51,7 +51,7 @@ int vine_file_replica_table_insert(struct vine_manager *m, struct vine_worker_in
 // remove a file from the remote file table.
 struct vine_file_replica *vine_file_replica_table_remove(struct vine_manager *m, struct vine_worker_info *w, const char *cachename)
 {
-	if (!w->current_files || !m->file_worker_table) {
+	if (!w) {
 	    // Handle error: invalid pointer
 	    return 0;
 	}
@@ -59,6 +59,8 @@ struct vine_file_replica *vine_file_replica_table_remove(struct vine_manager *m,
 	if (!hash_table_lookup(w->current_files, cachename)) {
 		return 0;
 	}
+
+	char *cachename2 = strdup(cachename);	
 
 	struct vine_file_replica *replica = hash_table_remove(w->current_files, cachename);
 	if (replica) {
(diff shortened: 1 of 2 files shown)


taskvine/src/manager/vine_file_replica_table.c
```diff
@@ -51,10 +51,17 @@ int vine_file_replica_table_insert(struct vine_manager *m, struct vine_worker_in
...
+	if (!w) {
+	    // Handle error: invalid pointer
+	    return 0;
+	}
+
 	if (!hash_table_lookup(w->current_files, cachename)) {
 		return 0;
 	}
 
+	char *cachename2 = strdup(cachename);
```
btovar, line 63: See below, I think the issue is that cachename as an actual pointer to the internal structure of the hash_table, which gets freed after the remove. Rather than using `cachename2`, I think we should ensure that cachename is is independent of the hash_table.

taskvine/src/manager/vine_manager.c
```diff
@@ -898,10 +898,13 @@ static void cleanup_worker_files(struct vine_manager *q, struct vine_worker_info
 		// may correspond to a cache-update of a file that has not been declared
 		// yet.
 		if (!f || !delete_worker_file(q, w, f->cached_name, f->cache_level, VINE_CACHE_LEVEL_WORKFLOW)) {
-			replica = vine_file_replica_table_remove(q, w, cached_name);
+			if(cached_name){
```
btovar, line 901: I think the issue is in line 889. `cached_names` is keeping pointers to the actual table keys, which may may free at any moment. A `cached_names[i] = xxstrdup(cached_name);` and then a free on those should fix it, but we should look for a cleaner way.

btovar (changes requested):

btovar:
@Ian2327 can you confirm that the changes to this pr still work for your tests?

JinZhou5042:
I am testing the system under a high churn rate of worker evictions and encountered a segfault caused by the `cleanup_worker_files` function. I attempted to fix it with the updated version in this pr, but the issue didn't seem to be resolved. I suspect there might be intricate problems with how we manage the `file_replica_table` and the `file_table`.

btovar:
Ah nice, I have been trying to debug that segfault. It doesn't happen in my runs, but it has happened to one of our collaborators at Florida. Do you think it is a high rate of evictions that is triggering it?

JinZhou5042:
Yes, I've found a couple of bugs that result in seg faults during the last two weeks, and this function is one of the causes. My temporary solution is to add the following code inside of `vine_file_replica_table_insert`:

```
struct vine_file *f = hash_table_lookup(m->file_table, cachename);
if (!f) {
    return 0;
}
vine_file_addref(f);
``` 

My `cleanup_worker_files` looks like:
```
static void cleanup_worker_files(struct vine_manager *q, struct vine_worker_info *w)
{
    if (!w || !w->current_files) return;

    int nfiles = hash_table_size(w->current_files);
    if (nfiles < 1) return;

    char **cached_names = malloc(nfiles * sizeof(char *));
    struct vine_file_replica **replicas = malloc(nfiles * sizeof(struct vine_file_replica *));

    int i = 0;
    char *cached_name = NULL;
    struct vine_file_replica *replica = NULL;

    HASH_TABLE_ITERATE(w->current_files, cached_name, replica) {
        cached_names[i] = xxstrdup(cached_name);
        replicas[i] = replica;
        i++;
    }

    for (i = 0; i < nfiles; i++) {
        cached_name = cached_names[i];
        replica = replicas[i];

        if (!cached_name) {
			continue;
		}

		delete_worker_file(q, w, cached_name, 0, 0);

        free(cached_names[i]);
    }

    free(cached_names);
    free(replicas);
}
```

Though there might be some other places that can be part of it, adding a ref to the original file structure seems quite critical for solving it.

JinZhou5042:
This bug manifests particularly when the following conditions are met:
- peer transfer is enabled
- a level of file replication
- DAG structured workflow
- high rate of worker evictions

Ian2327:
> @Ian2327 can you confirm that the changes to this pr still work for your tests?
I just ran blast.py on your changes and it reported back 0 errors from valgrind, so moving the strdup to the outer function doesn't cause any further issues. The only thing is that of 10,000 tasks run, 13,334 tasks came back as completed.

btovar:
@JinZhou5042  Do you think `w->current_files` has files that are no longer in `m->file_table`?

JinZhou5042:
@btovar Yes, I've been used gdb to debug it for multiple times and that is definitely the symptom!

Issue #4108, opened 2025-04-11
vine: segfault when task is marked as waiting retrieval and worker disconnects


btovar:
See discussion at #4104.

btovar:
The issue seems to be that fetching outputs may return `WORKER_FAILURE`, which removes the worker. However, right after that we try to delete uncachable outputs from the now unavailable worker using an out-of-date pointer.

btovar:
Mmm, or maybe not. We seem to be doing the correct thing in fetch output from worker.

btovar:
The issue seems to be related to an inconsistent handling of app vs worker failure.

btovar:
Fixed by #4109

PR #4109, opened 2025-04-11
vine: hotfix, use consistent results in handle_failure/fetch_output

btovar:
@JinZhou5042 @Ian2327 could you test this pr and confirm it fixes the sefgaults observed?

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -1343,7 +1343,7 @@ static int fetch_outputs_from_worker(struct vine_manager *q, struct vine_worker_
 		handle_failure(q, w, t, result);
 	}
 
-	if (result == VINE_WORKER_FAILURE) {
+	if (result != VINE_APP_FAILURE ) {
 		t->time_when_done = timestamp_get();
 		return 0;
 	}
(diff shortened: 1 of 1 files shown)


btovar:
For #4108

btovar:
Interesting... Why is fetch being called on a null worker?

btovar:
That case seems to happen only if the task is in `q->waiting_retrival_list`, but it's worker was set to 0 by a previous reap.

btovar:
I think it is related to the fix in this pr. The state of the task is never updated because a reap sets the worker to 0 and never deleted from the waiting retrieval list.

JinZhou5042:
I just ran DV5 with the change in this pr, but unfortunately I didn't see the progress bar returning any completed tasks and the application eventually failed

image


Once reverted this change the application finishes as expected

btovar:
Ah of course, I'm treating successes as failures...

btovar:
can you give it a try again?

JinZhou5042:
It works fine with the scale of 10K tasks, running with 250K tasks...

JinZhou5042:
Having another segfault...

image

image

JinZhou5042:
The `task->input_mounts` is valid but some elements in it were freed unexpectedly

image

btovar:
Is this using any kind of pruning, or is it regular operation?

btovar:
I'm merging this one as it fixes the original segfault.

JinZhou5042:
> Is this using any kind of pruning, or is it regular operation?

It was using pruning with depth 2 and replicating for 3 times.

JinZhou5042:
Yes, I think that could be the case. However, I believe the underlying problem is that the file deletion operation is not atomic and can sometimes have severe side effects. In other words, the deleted files may be permanently removed from the `q->file_table`, yet somehow they remain in other data structures.

Issue #4122, opened 2025-04-15
vine:


JinZhou5042:
One segfault I met looks like this:

![Image](https://github.com/user-attachments/assets/e9528e70-57f2-4f62-bba4-b5f660b2d0cc)

Looks like it was caused by incorrect use of `malloc` in `rmsummary_create`.

Here is the implementation of this function:

![Image](https://github.com/user-attachments/assets/af6f22d2-3648-45ad-8b98-eba897ea74c8)

The problem is that, in `rmsummary_create`, the struct is initialized with `memset(s, default_value, sizeof(struct rmsummary))`. If `default_value` is `-1` (this is how we initialize the structure), this causes all pointer fields to be filled with 0xFF..., later usage of that pointer may results in segfaults.

This can be fixed by using `memset(s, 0, ...)` followed by explicit initialization of numeric fields with `default_value`.

I didn't observe this segfault again with this fix.

btovar:
No, that's not it. That would only matter for actual pointer attributes in the struct, and thus we do set to NULL. The -1 is only applied to actual values.

dthain:
Actually, I think there is a problem because `memset` fills memory with *byte* values equal to -1, and the values we are intending to fill that way are doubles!

btovar:
But even then, this value is never assigned to pointers. What @JinZhou5042 observes is more likely coming from somewhere else.

JinZhou5042:
Similar to #4138, maybe `malloc` causes the error because the memory was already corrupted by some other dangling pointers?

btovar:
That's my feeling too. If malloc failed, the corruption happened somewhere else.

dthain:
valgrind is your friend here

PR #4126, opened 2025-04-16
vine: memset with 0

JinZhou5042:
## Proposed Changes

For #4122

## Merge Checklist

### DIFF
--- a/dttools/src/rmsummary.c
+++ b/dttools/src/rmsummary.c
@@ -609,7 +609,7 @@ all string fields with NULL. Usual values are 0, or -1. */
 struct rmsummary *rmsummary_create(double default_value)
 {
 	struct rmsummary *s = malloc(sizeof(struct rmsummary));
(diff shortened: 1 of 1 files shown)


btovar (changes requested):
This won't work. Code depends on checking negative values. Also, the default values are not applied to pointers, but to actual numbers.

Note that this worked fine with work queue, and with taskvine until recently. I'd suggest we look for recent changes that modify how the requested resources are set.

dthain:
Ah, that's a good point.  I overlooked the fact that setting bytes to -1 sets all bits to one, which give you -1 for any integer size, or even for a double:

```
(gdb) print /x (double)-1
$4 = 0xffffffffffffffff
```

Something else is going on...

JinZhou5042:
Closing as this isn't it, it was caused by other memory issues, see #4138 and #4140

Issue #4118, opened 2025-04-15
vine: make killing workers periodically an argument


JinZhou5042:
I found two settings useful when conducting experiments on the system's fault tolerance and stability:

- Control the worker failure rate using an argument to the manager. For example, we may want to kill workers periodically, say, every 60 seconds, regardless of whether the failures are unexpected or intentional. We calculate the actual failure rate since the beginning of the run and compare it to the target, and decide how many workers to kill in order to maintain the desired failure rate.
- Control the maximum number of workers we want to maintain. We already have an argument `wait-for-workers` which delays the task scheduling until the desired number of workers are connected. But in some senarios we may want to avoid using more workers than we want.

It would be great if the manager could natively support these two arguments, as that would make experiments more convenient.

PR #4127, opened 2025-04-16
vine: add testing parameters

JinZhou5042:
## Proposed Changes

For #4118

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/taskvine.h
+++ b/taskvine/src/manager/taskvine.h
@@ -213,6 +213,9 @@ struct vine_stats {
 	timestamp_t time_workers_execute_exhaustion; /**< Total time workers spent executing tasks that exhausted
 							resources. */
 
+	/* Workflow time statistics: */
+	timestamp_t time_first_task_started;          /**< Time the first task started. */
+
 	/* BW statistics */
 	int64_t bytes_sent; /**< Total number of file bytes (not including protocol control msg bytes) sent out to the
 			       workers by the manager. */
--- a/taskvine/src/manager/taskvine.h
+++ b/taskvine/src/manager/taskvine.h
@@ -214,7 +214,7 @@ struct vine_stats {
 							resources. */
 
 	/* Workflow time statistics: */
-	timestamp_t time_first_task_started;          /**< Time the first task started. */
+	timestamp_t time_first_task_dispatched;          /**< Time the first task started. */
 
 	/* BW statistics */
 	int64_t bytes_sent; /**< Total number of file bytes (not including protocol control msg bytes) sent out to the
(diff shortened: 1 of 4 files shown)


taskvine/src/manager/vine_manager.h
```diff
@@ -237,6 +237,9 @@ struct vine_manager {
 
 	timestamp_t large_task_check_interval;	/* How frequently to check for tasks that do not fit any worker. */
 	double option_blocklist_slow_workers_timeout;	/* Default timeout for slow workers to come back to the pool, can be set prior to creating a manager. */
+
+	int enforce_worker_eviction_interval;   /* Used in experimental mode to enforce worker eviction interval in seconds */
```
btovar, line 241: Let's say "testing" instead of "experimental mode".

btovar (changes requested):

taskvine/src/manager/taskvine.h
```diff
@@ -213,6 +213,9 @@ struct vine_stats {
 	timestamp_t time_workers_execute_exhaustion; /**< Total time workers spent executing tasks that exhausted
 							resources. */
 
+	/* Workflow time statistics: */
+	timestamp_t time_first_task_started;          /**< Time the first task started. */
```
btovar, line 217: I don't see this stat being used anywhere, and it can be deduced from info in the logs. I suggest removing it. Also, it doesn't record 'started' but 'dispatched'.

btovar (changes requested):

btovar:
Also, thinking more about this pr, let's hold it for now. An unintended effect of the max_workers as implemented here is that the unconnected workers will hang waiting for the manager to read the initial connection.

dthain:
Consider using the "milestone" feature to mark which of these PRs are for the imminent release, or the next one down the road...

taskvine/src/manager/taskvine.h
JinZhou5042: @btovar  Thanks for pointing that out! This parameter is used in the `enforce_worker_eviction_interval` function. Before we begin evicting workers intentionally, we want to wait until the workflow actually starts. That timing is determined by when the first task is dispatched. Otherwise, if we set `wait-for-workers` to a high number, nothing happens in the first 10 minutes, the manager still evict workers while connecting new ones simultaneously, which slows things down early on. Do you think there’s a smarter way to handle this?

btovar: Then I think you simply want to check `q->stats->tasks_dispatched > 0`

JinZhou5042: @btovar That is indeed smart!!

dthain:
An additional consideration here is that the `vine_stats` interface, for better or worse, uses a fixed size C structure for output.  And so any changes to that structure necessarily break the ABI.

dthain:
What we need to do here -- instead of declining TCP connections, we should check for too-many-workers when a `taskvine` message is received, and then disconnect that worker.

btovar:
ready to merge?

Issue #4077, opened 2025-02-26
vine: delay


JinZhou5042:
This was originally discussed in #4061 

`VINE_FILE_REPLICA_STATE_READY` is set under two circumstances:
- when the manager receives a `cache-update` message of a replica
- when the manager starts to `put` an input file  using `vine_manager_put_input_file_if_needed`, if the file is in a type of `VINE_FILE`, `VINE_MINI_TASK`, or `VINE_BUFFER`, it will assume the file is ready there:
```
	/* If the send succeeded, then record it in the worker */
	if (result == VINE_SUCCESS) {
		struct vine_file_replica *replica = vine_file_replica_create(f->type, f->cache_level, f->size, f->mtime);
		vine_file_replica_table_insert(q, w, f->cached_name, replica);

		switch (file_to_send->type) {
		case VINE_URL:
		case VINE_TEMP:
			/* For these types, a cache-update will arrive when the replica actually exists. */
			replica->state = VINE_FILE_REPLICA_STATE_PENDING;
			break;
		case VINE_FILE:
		case VINE_MINI_TASK:
		case VINE_BUFFER:
			/* For these types, we sent the data, so we know it exists. */
			replica->state = VINE_FILE_REPLICA_STATE_READY;
			f->state = VINE_FILE_STATE_CREATED;
			break;
		}
	}
```

However, in the second condition, since each successful file stage-in will come along with a `cache-update` message, we may want to delay setting the replica's `READY` status until the corresponding `cache-update` message arrives. This may improve the robustness while dispatching tasks.

dthain:
That's not quite correct -- in the second case, the worker does *not* send cache updates for things provided by `vine_cache_add_file`.  That needs to be changed too.

btovar:
Fixed.

PR #4128, opened 2025-04-17
Vine: Send cache-update on all file types.

dthain:
## Proposed Changes

@JinZhou5042 observed that (ordinary) files sent and then quickly replicated could result in failures due to a race condition between the source file arriving, and the replication request being sent.  This PR eliminates this race condition by requiring all file types to remain in the PENDING state until a cache-update message is received from the worker.

## Merge Checklist


dthain:
Fixes #4077

taskvine/src/worker/vine_cache.c
```diff
@@ -118,7 +118,7 @@ void vine_cache_scan(struct vine_cache *c, struct link *manager)
 	char *cachename;
 	HASH_TABLE_ITERATE(c->table, cachename, f)
 	{
-		vine_worker_send_cache_update(manager, cachename, f->original_type, f->cache_level, f->size, f->mode, f->transfer_time, f->start_time);
```
btovar, line 121: Could be make it so that the same info is sent back in the ready message?  In that way we don't have to change the protocol. Changing the protocol is always a bit of a headache when workers can't connect because of it.

dthain: So the `cache-update` message itself didn't change, it still has the same format. I just cleaned up the function signature a bit.

What did change is that the manager now expects the worker to send a cache-update for all file types, where it didn't send them before.   That's the protocol change.

btovar: Ah got it. Thanks for explaining it!

btovar (approved):

btovar:
ready to merge?

dthain:
I think maybe hold off on merging this until Jin's sequence of PRs is accepted.  I'm not entirely sure of the impact on a large workflow  yet.

btovar (approved):

btovar:
Resolving conflicts...

btovar:
@JinZhou5042 thanks!

PR #4133, opened 2025-04-22
vine: remove invalid files on worker

JinZhou5042:
## Proposed Changes

For #4134

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_current_transfers.c
+++ b/taskvine/src/manager/vine_current_transfers.c
@@ -35,6 +35,18 @@ static void vine_transfer_pair_delete(struct vine_transfer_pair *p)
 	}
 }
 
+struct vine_worker_info *vine_current_transfers_get_source_worker(struct vine_manager *q, const char *id)
+{
+	struct vine_transfer_pair *p = hash_table_lookup(q->current_transfer_table, id);
+	return p ? p->source_worker : NULL;
+}
+
+struct vine_worker_info *vine_current_transfers_get_dest_worker(struct vine_manager *q, const char *id)
+{
+	struct vine_transfer_pair *p = hash_table_lookup(q->current_transfer_table, id);
+	return p ? p->to : NULL;
+}
+
 // add a current transaction to the transfer table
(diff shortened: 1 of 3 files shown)


JinZhou5042:
Combining with this [pr](https://github.com/cooperative-computing-lab/cctools/pull/4130):

The long tails and execution sparsity seem gone:
image

All file transfers can complete:
image

dthain:
@JinZhou5042 I don't doubt that this fixes your immediate problem.  But it's pretty obvious that this fix is not a good general solution.  If you delete a replica at the source every time a transfer files, then problems at the receiving end are going to destroy good data aggressively.

We need to better understand the source of a problem in order to design a good solution.

Can you explain why these transfers are failing in the first place?

JinZhou5042:
@dthain  It turned out that when we receive the `cache-invalid` message, the `source` is mostly NULL, and the code really played an effect is this:

```
if (to_worker) {
    delete_worker_file(q, to_worker, cachename, 0, 0);
} 
```

I believe the sequence is as follows:

1. Worker A is fetching many files from worker B.
2. Worker B suddenly crashes.
3. All pending transfers on A fail with a series of `cache-invalid` messages.
4. Since B has already crashed, its data structures have been released, so `vine_current_transfers_get_source_worker(q, transfer_id)` consistently returns NULL.
5. So, we end up deleting files on the destination worker instead of the source.

Even if we don't want to agreesively delete files from the source worker, at least deleting them from the destination worker that returns the `cache-invalid` seems reasonable?

dthain:
@JinZhou5042 you are making a lot of hypotheses that could be answered by reading the code.  :)

Look at `handle_cache_invalid` It removes the replica, and then removes the transfer record.  An `unlink` is not necessary when the transfer fails, because the failed transfer is not committed to the cache on the worker side.  Also note that it calls `vine_current_transfers_set_failure` to keep track of the number of related failures, and throttle future replications.

It's ok to hypothesize potential bugs, but then you need to show a path through the code that is clearly incorrect.  Don't just guess!

JinZhou5042:
See #4152

PR #4153, opened 2025-05-01
vine: separate pending and ready tasks

JinZhou5042:
## Proposed Changes

## Merge Checklist


btovar (changes requested):
I'm not completely sure about this approach. If the problem is starvation because of the priority queue, then I think it would be better to implement the priority queue with a more adequate data structure.

dthain:
Main issue here: this is adding complexity, which may or may not be worth the improvement.

We need to see a before/after test that quantifies the improvement.

PR #4156, opened 2025-05-02
vine: Update cache on invalid xfer

btovar:
## Proposed Changes

## Merge Checklist

### DIFF
--- a/taskvine/src/worker/vine_cache.c
+++ b/taskvine/src/worker/vine_cache.c
@@ -342,10 +342,10 @@ int vine_cache_add_transfer(struct vine_cache *c, const char *cachename, const c
 	f->transfer_time = 0;
 
 	int status = hash_table_insert(c->table, cachename, f);
-	if(!status) {
+	if (!status) {
 		int bad_cache = 0;
 		struct vine_cache_file *previous = hash_table_remove(c->table, cachename);
-		if(!previous) {
+		if (!previous) {
 			bad_cache = 1;
 		} else {
 			switch (previous->status) {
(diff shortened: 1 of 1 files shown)


JinZhou5042 (approved):

dthain (commented):
That single change makes sense to me.

But I am uncertain b/c there are a number of other places where `f->status = VINE_CACHE_STATUS_FAILED`.

Do those also need to get updated?

Are there any callers that expect to see `FAILED` as distinct from `UNKNOWN`?

btovar:
Actually this simple solution does not work, as it deletes an entry from a table that's being iterated.
I think you might be right about other status checks. The coffea.casa example works better with this, but still some tasks get stuck, but now with debug messages like: file-meta-3d37ad0d807c1b31abf8f6fe7dfab5f6 is unknown, perhaps it failed to transfer earlier?

JinZhou5042:
An alternative is to call `delete_worker_file` on the destination worker upon `cache-invalid`, this way we don't need to change the worker code.

dthain:
In some sense, the `FAILED` state should be persistent -- if several tasks need the file in short order, then a single attempt to fetch (or create it) should be done, and then the later attempts should fail, at least until the manager gets to decide whether it is worth doing again.  (For example, a mini-task that fails to generate the desired file should not be endlessly retried.)

So the question is, when does the `FAILED` state reset to some other state?  Some possible approaches:
1 - Immediately dispose of `FAILED` entries as soon as they are consumed.
2 - Keep `FAILED` entries until some point later in the main loop that clears them out.
3 - Keep `FAILED` entries until the manager explicitly deletes them.
4 - Reset `FAILED` entries back to `PENDING` when the manager sends the file again.

PR #4158, opened 2025-05-07
Mini task fail on input missing

btovar:
Now mini tasks are allowed to fork if their sandbox cannot be staged. In this way the output file of the mini task can generate the corresponding cache-invalid message.

## Merge Checklist

### DIFF
--- a/taskvine/src/worker/vine_cache.c
+++ b/taskvine/src/worker/vine_cache.c
@@ -700,7 +700,7 @@ vine_cache_status_t vine_cache_ensure(struct vine_cache *c, const char *cachenam
 			LIST_ITERATE(f->mini_task->input_mounts, m)
 			{
 				result = vine_cache_ensure(c, m->file->cached_name);
-				if (result == VINE_CACHE_STATUS_FAILED) {
+				if (result == VINE_CACHE_STATUS_FAILED || result == VINE_CACHE_STATUS_UNKNOWN) {
 					/* an input file for the mini task failed to transfer,
 					 * thus the mini task is marked also as failed */
 					mini_task_inputs_permanent_failure = 1;
--- a/taskvine/src/worker/vine_cache.c
+++ b/taskvine/src/worker/vine_cache.c
@@ -159,7 +159,9 @@ static void vine_cache_kill(struct vine_cache *c, struct vine_cache_file *f, con
 	while (f->status == VINE_CACHE_STATUS_PROCESSING) {
 		debug(D_VINE, "cache: killing pending transfer process %d...", f->pid);
 		kill(f->pid, SIGKILL);
-		vine_cache_wait_for_file(c, f, cachename, manager);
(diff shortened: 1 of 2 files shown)


taskvine/src/worker/vine_cache.c
```diff
@@ -422,7 +407,9 @@ int vine_cache_remove(struct vine_cache *c, const char *cachename, struct link *
 		return 0;
 
 	/* Ensure that any child process associated with the entry is stopped. */
-	vine_cache_kill(c, f, cachename, manager);
```
dthain, line 425: I think that you want to still call `vine_cache_kill` even if `manager` is null, since that does other work.  Just check for manager null at the point where it actually gets used (sending a cache-update/invalid message, I think).

taskvine/src/worker/vine_cache.c
```diff
@@ -711,8 +699,14 @@ vine_cache_status_t vine_cache_ensure(struct vine_cache *c, const char *cachenam
 			LIST_ITERATE(f->mini_task->input_mounts, m)
 			{
 				result = vine_cache_ensure(c, m->file->cached_name);
-				if (result != VINE_CACHE_STATUS_READY)
+				if (result == VINE_CACHE_STATUS_FAILED) {
```
dthain, line 702: This may not come into play right away, but based on our prior discussion, it seems that `FAILED` may quickly transition to `UNKNOWN` when cleaned up, and so they should be treated as equivalent.

dthain (changes requested):
This looks like the right idea, a few minor items noted.

I wonder if `FAILED` and `UNKNOWN` should be merged into a common state?

btovar:
> This looks like the right idea, a few minor items noted.
> 
> I wonder if `FAILED` and `UNKNOWN` should be merged into a common state?

The is only one place where they are treated differently, although it looks like dead code. I wonder if https://github.com/cooperative-computing-lab/cctools/blob/06302b84413fe73da52b38d7fdf603483614d889/taskvine/src/worker/vine_sandbox.c#L78 should be `break;` instead of `return VINE_CACHE_STATUS_FAILED;`?

taskvine/src/worker/vine_cache.c
```diff
@@ -159,7 +159,9 @@ static void vine_cache_kill(struct vine_cache *c, struct vine_cache_file *f, con
 	while (f->status == VINE_CACHE_STATUS_PROCESSING) {
 		debug(D_VINE, "cache: killing pending transfer process %d...", f->pid);
 		kill(f->pid, SIGKILL);
-		vine_cache_wait_for_file(c, f, cachename, manager);
+		if (manager) {
```
dthain, line 162: vine_cache_wait_for_file needs to be called unconditionally.
It does a variety of things, in addition to sending updates to the manager.

dthain (changes requested):

PR #4159, opened 2025-05-07
vine: Fix waiting retrieval

btovar:
When retrieving tasks results, we want to do one worker at a time. The previous mechanism would maintain two data structures: one hash table with workers that have results, and a list of tasks with results ready to be retrieved. Looking at workers one at a time in the table, tasks would be retrieved and deleted from the list. Keeping the hash table was inefficient, and as @JinZhou5042 observed, the implementation lead to O(n^2) iterations.

This pr uses only the list of tasks to be retrieved and removes the hash table. Now we look at one task in the list, see which worker has it, and get all the tasks from that worker. Note that the first task we looked at may not be actually retrieved if there are more tasks to be retrieved in that worker, and the retrieval limit has been reached.

I like this approach better as we only need to maintain one data structure. Further this pr simplifies the code, as when not all the tasks of a worker were retrieved, the worker would be deleted from the retrieval table anyway. Later, tasks would be retrieved using the list. With this change, all task are retrieved per worker.

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -3570,20 +3570,6 @@ static int send_one_task(struct vine_manager *q)
 }
 
 /*
-get available results from a worker. This is typically used for signaling watched files.
-*/
-int get_results_from_worker(struct vine_manager *q, struct vine_worker_info *w)
-{
-	/* get available results from the worker, bail out if that also fails. */
-	vine_result_code_t r = get_available_results(q, w);
-	if (r != VINE_SUCCESS) {
-		handle_worker_failure(q, w);
-		return 0;
-	}
-	return 1;
-}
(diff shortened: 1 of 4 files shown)


dthain:
Can you explain more in the PR what the assumptions and intentions are here?  The `workers_with_complete_tasks` was added for a reason; if our understanding has changed, then that deserves some explanation...

btovar:
Will do! The idea is that we just need to know the workers with results, and each task with a result to be retrieved has a pointer to its worker. Therefore we only need the list of task ready to be retrieved and not the hash table with the workers.

dthain (changes requested):
Ok, the overall idea looks good, but see missing call to vine_factory_manager_prune

dthain (approved):

PR #4176, opened 2025-06-18
vine: worker unlinks file if invalid cache

btovar:
When the worker sends an invalid cache message to the manager, it now also removes the file from its cache as if the manager had sent an unlink message. This cleans up the state of the replicas at the worker so that further puturl request succeed.

This fixes a bug where puturls are ignored because of previous failures to transfer, which caused tasks to be forsaken forever.

## Proposed Changes

## Merge Checklist


dthain:
Hmm, there are a lot of nested cases inside of `vine_cache_check_outputs`, along with references to the object `f` that gets removed.

How about doing this check and remove after `vine_cache_check_outputs` in `vine_cache_wait_for_file` ?

I think this means we are committing to the idea that files do *not* stay in the cache in the `FAILED` state, but are immediately removed and will be considered in the `UNKNOWN` state.

btovar:
Sounds good!

dthain (approved):

Issue #4170, opened 2025-05-30
Vine: Pruning Files leading to worker failures.


BarrySlyDelgado:
As @JinZhou5042  noted, the pruning of files can lead to workers failing. In this scenario, the manager will direct a worker to remove a file while a pending transfer is taking place. This transfer will then fail. If this happens multiple times for the same worker, this worker will be treated as faulty and removed. As we have knowledge of pending transfers and which files are being pruned, we should not have these types of failures condemn a worker as we should know that the transfer is failing anyway. 

Working on replicating this issue.

PR #4187, opened 2025-06-25
vine: don't condemn workers due to expected cache-invalid

JinZhou5042:
## Proposed Changes

For #4170

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_current_transfers.c
+++ b/taskvine/src/manager/vine_current_transfers.c
@@ -6,6 +6,8 @@ See the file COPYING for details.
 
 #include "vine_current_transfers.h"
 #include "macros.h"
+#include "vine_file_replica.h"
+#include "vine_file_replica_table.h"
 #include "vine_blocklist.h"
 #include "vine_manager.h"
 #include "xxmalloc.h"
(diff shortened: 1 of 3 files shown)


taskvine/src/manager/vine_current_transfers.c
```diff
@@ -119,35 +121,36 @@ void set_throttle(struct vine_manager *m, struct vine_worker_info *w, int is_des
 	}
 }
 
-int vine_current_transfers_set_failure(struct vine_manager *q, char *id)
+int vine_current_transfers_set_failure(struct vine_manager *q, char *id, const char *cachename)
 {
 	struct vine_transfer_pair *p = hash_table_lookup(q->current_transfer_table, id);
+	if (!p) {
+		return 0;
```
dthain, line 128: Comment here: if there is no matching transfer record, then it means a worker failed and the record was removed b/c of `wipe_worker`.

taskvine/src/manager/vine_current_transfers.c
```diff
@@ -119,35 +121,36 @@ void set_throttle(struct vine_manager *m, struct vine_worker_info *w, int is_des
...
+	}
 
-	int throttled = 0;
-
-	if (!p)
-		return throttled;
-
+	/* if any of the source or to workers don't exist, we shouldn't be surprised by this transfer failure */
```
dthain, line 131: I don't think this comment is right.  As far as I can tell, the elements of the vine_transfer_pair should always be valid, and I don't see any code that sets them to null.  A failed worker causes the transfer record to be removed, not nulled out.

btovar (approved):

btovar:
Ready to merge?

dthain (changes requested):
The code makes sense to me, but the comments seem to offer the wrong explanation.

JinZhou5042:
RTM

PR #4219, opened 2025-07-17
Catalog: Abort faster on SIGALRM

dthain:
## Proposed Changes

Observed problem: web crawlers from google and msn are connecting to the catalog server, but are sloobow in completing SSL authentication. The catalog server (as designed) kills these connections after 3 seconds, because it doesn't want them hanging around and tar pitting real users.  However, when the SIGALRM is delivered, resulting in an `exit`, the SSL library attempts to clean up, attempting to acquire a lock that it already holds.  Deadlock.

Solution: Delivery of SIGALRM should result in `_exit` for immediate shutdown, rather than `exit` for an orderly cleanup.  `exit` is not generally signal safe.

## Merge Checklist


btovar (approved):

btovar:
ready to merge? I can deploy today if needed.

PR #4224, opened 2025-07-18
Vine: Fix cache removal error in worker.

### LINKED ISSUE #4205: Vine: Worker Cache Removal Inconsistency
We have a design problem in the waiting/removal of files in `taskvine/src/worker/vine_cache.c` resulting from the combination of some recent changes.

- `vine_cache_wait_for_file` checks to see if a transfer process has completed.  If so, it marks the file as either `TRANSFERRED` or `FAILED`.  If `FAILED`, it calls `vine_cache_remove` to remove the file from the cache.  So, following a call to this function, the file object may no longer be valid.
- `vine_cache_remove` calls `vine_cache_kill` which calls `vine_cache_wait_file`.  Both `vine_cache_kill` and `vine_cache_remove` assume that f is valid after that call.  (which it might not be)

I believe we got here, because the intention was to remove failed file transfers from the table aggressively, rather than having them stick around in the `FAILED` state, since they could very well be re-added by the manager.

I think we have two basic ways forward here:

1 - Rework (and heavily comment) callers of `vine_cache_wait_for_file` to take care that `f` may no longer be valid.
2 - Allow `vine_cache_wait_for_file` to leave the file in the `FAILED` state, and ensure that future `add` `remove` and `ensure` operations do the right thing when encountering that state.

@btovar can you sort this one out?

dthain:
## Proposed Changes

Fix removal sequence to avoid some rare memory errors:
- Rename vine_cache_wait -> vine_cache_check_files for clarity.  (It doesn't block!)
- vine_cache_check_file should *not* remove the object, because some callers expect it to be valid.
- vine_cache_check_files *should* remove because it has no further use for the object.

Fixes #4205

## Merge Checklist

### DIFF
--- a/taskvine/src/worker/vine_cache.c
+++ b/taskvine/src/worker/vine_cache.c
@@ -44,7 +44,7 @@ struct vine_cache {
 	int max_transfer_procs;
 };
 
-static void vine_cache_wait_for_file(struct vine_cache *c, struct vine_cache_file *f, const char *cachename, struct link *manager);
+static void vine_cache_check_file(struct vine_cache *c, struct vine_cache_file *f, const char *cachename, struct link *manager);
 
 /*
 Create the cache manager structure for a given cache directory.
@@ -151,6 +151,7 @@ void vine_cache_prune(struct vine_cache *c, vine_cache_level_t level)
 
 /*
 Kill off any process associated with this file object.
+The cache_file object is still valid afterwards.
 Used by both vine_cache_remove and vine_cache_delete.
 */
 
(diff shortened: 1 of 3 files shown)


JinZhou5042 (approved):

btovar (approved):

btovar:
Makes sense!, ready to merge?

dthain:
It works for me, but I would prefer if someone can test this out with a larger app.

JinZhou5042:
Ran this PR on DV5 with 250K tasks, configuring `temp-replica-count` to 3 and `--enforce-worker-eviction-interval` to 30. All tasks completed successfully without issues.

PR #4233, opened 2025-08-06
vine: explicitly specify file type for put/puturl/puturl_now

JinZhou5042:
## Proposed Changes

The issue was that `cache-update` returns file type value of 0, which should be in the range of 1 to 5. And because of the incorrect type, temp file replication cannot be proceeded.
image
image
image

To fix this, we should  let the manager control the file type and append it whenever it sends a file to a worker, rather than letting the worker infer the file type from the message.

For whatever file the manager sends to any worker, a `file_type` field must be explicitly specified in that message.

If the file was a `VINE_URL` (e.g., a file that needs to be downloaded from another worker), we will resolve its actual type once it is materialized and the manager recveives the related `cache-update`.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_protocol.h
+++ b/taskvine/src/manager/vine_protocol.h
@@ -13,7 +13,7 @@ worker, and catalog, but should not be visible to the public user API.
 #ifndef VINE_PROTOCOL_H
 #define VINE_PROTOCOL_H
 
-#define VINE_PROTOCOL_VERSION 13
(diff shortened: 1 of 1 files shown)


dthain:
Ok, it makes sense to pass the file type with `put` and `puturl`.

But it's not clear to me why the type has been added to `infile` and `outfile`.  Those messages describe `vine_mounts` that connect tasks to files.  The file and its details are already known at that point.  Am I missing something?

Issue #4097, opened 2025-03-28
vine: factory does not correctly quote manager's name regex


btovar:
For example, if the manager's name specified in the factory contains parentheses, then the worker command fails as the shell gets confused. This is also probability true for vine_submit_workers, but I have not confirmed it.

LaxminarayanaV7416:
please check the PR for this bug fix https://github.com/cooperative-computing-lab/cctools/pull/4235

PR #4235, opened 2025-08-27
Fixed the vine_factory bug during command line submission

LaxminarayanaV7416:
## Proposed Changes
This PR is for [BUG](https://github.com/cooperative-computing-lab/cctools/issues/4097), where the problem is that from the factory, we are submitting the vine_worker command using the command line. When we include any special character, the POSIX command line cannot accept it, so we need to escape them and include a double quote, something like hello'world -> "hello\'world" so then the POSIX will accept it. I am attaching with printf command while submitting the one without fix and with fix below for reference. NOTE that the printf is for dev and is not present in PR.

One without any fix
Screenshot 2025-08-27 at 7 01 56 PM

After the fix when we are able to submit the job
Screenshot 2025-08-27 at 7 02 54 PM

## Merge Checklist

- [X] `make test`       Run local tests prior to pushing.
- [X] `make format`     Format source code to comply with lint policies. Note that some lint errors can only be resolved manually (e.g., Python)
- [X] `make lint`       Run lint on source code prior to pushing.

### DIFF
--- a/batch_job/src/vine_factory.c
+++ b/batch_job/src/vine_factory.c
@@ -219,6 +219,34 @@ int manager_workers_capacity(struct jx *j) {
 	return capacity;
 }
 
+char *shell_escape(const char *input) {
+    size_t len = strlen(input);
+
+    // Worst case: every char becomes 2 chars + quotes
+    char *output = malloc(len * 2 + 3);
+    if (!output) return NULL;
+
+    char *dst = output;
+    *dst++ = '"';  // open double quote
+
+    for (const char *src = input; *src; src++) {
+        switch (*src) {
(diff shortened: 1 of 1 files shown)


batch_job/src/vine_factory.c
```diff
@@ -219,6 +219,34 @@ int manager_workers_capacity(struct jx *j) {
 	return capacity;
 }
 
+char *shell_escape(const char *input) {
```
btovar, line 222: Could we make it work with `string_escape_shell` in `dttools/src/stringtools.c`?

btovar:
Is there progress in this pr?

PR #4246, opened 2025-10-13
vine: function infile load mode

JinZhou5042:
## Proposed Changes

We had only one mode (`cloudpickle`) to wrap function arguments and send to workers.

But sometimes arguments can also be passed of a `json` format and we should support that.

## Merge Checklist

### DIFF
--- a/poncho/src/poncho/library_network_code.py
+++ b/poncho/src/poncho/library_network_code.py
@@ -175,7 +175,7 @@ def start_function(in_pipe_fd, thread_limit=1):
                 infile_path = os.path.join(function_sandbox, "infile")
                 event = load_function_infile(infile_path)
             except Exception:
-                stdout_timed_message(f"TASK {function_id} error: can't load the arguments from {arg_infile}")
+                stdout_timed_message(f"TASK {function_id} error: can't load the arguments from {infile_path}")
                 return -1, function_id
             p = os.fork()
             if p == 0:


tphung3:
If you want to support this feature you also need to add a flag somewhere around [here](https://github.com/cooperative-computing-lab/cctools/blob/2c8f80de5c6a296be8932ee7a1f011354a97a3cd/taskvine/src/bindings/python3/ndcctools/taskvine/task.py#L1113-L1125) to serialize Python objects by JSON too, as arguments via "infile" are always serialized with cloudpickle at the moment (see https://github.com/cooperative-computing-lab/cctools/blob/2c8f80de5c6a296be8932ee7a1f011354a97a3cd/taskvine/src/bindings/python3/ndcctools/taskvine/task.py#L1113-L1125)

poncho/src/poncho/library_network_code.py
```diff
@@ -84,6 +87,18 @@ def sigchld_handler(signum, frame):
     os.writev(w, [b"a"])
 
 
+# Load the infile for a function task inside this library
```
btovar: @JinZhou5042 I'm referring in particular to what this comment implies. Why is not the function part of the library in the first place? Why not generate the library with the functions needed instead of loading them from a file which may create difficulties when debugging, create new paths for errors, etc.?

PR #4263, opened 2025-11-03
vine: scale task priority multiplicatively

JinZhou5042:
## Proposed Changes

Task priorities can differ by several orders of magnitude, so increasing them by a fixed number internally would be meaningless. Instead, we should scale the priorities multiplicatively to preserve their relative magnitude.

## Merge Checklist


btovar:
I think this is the wrong approach. The objective of the increase in priority is to run the task sooner, and the `* 1.05` does not guarantee it. It is easy to set priorities far apart that the issue does not go away. I suggest coming up with a different policy (like having two tasks priorities: user and taskvine and sort first by the taskvine priorities. The priority queue would look at two priorities.) 

Because of this, I'm reluctant to merge this one.

btovar:
I can give the two priorities idea a try tomorrow.

PR #4264, opened 2025-11-03
vine: don't print `Submitted recovery task xxx` to terminal

JinZhou5042:
## Proposed Changes

For now, whenever a recovery task is submitted, a message is printed to the user's terminal, which can be distracting.

Users don't need to know about individual recovery tasks or their operations, instead, they only care about the overall progress. If we do want to inform them, only the total number of recovery tasks is relavant

Therefore, we should stop printing `Submitted recovery task xxx` to their terminal. Instead, the manager should track the number of recovery tasks submitted, and expose an API for the upper-level graph executor to decide whether or not to display that information.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -4841,7 +4841,7 @@ int vine_submit(struct vine_manager *q, struct vine_task *t)
 
 	/* Issue warnings if the files are set up strangely. */
 	vine_task_check_consistency(t);
-	
+
 	if (t->type == VINE_TASK_TYPE_RECOVERY) {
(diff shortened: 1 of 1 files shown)


dthain (commented):
I'm thinking that we shouldn't introduce a new API entry point for every little statistic that we would like to return.  How about modifying `vine_stats` instead?

In the past, I would be every loath to break ABI by changing the structure size, but I think everyone is coming through Python now, and it's easier to ensure a match between Python and C.

JinZhou5042:
That is a good point!

btovar (approved):

btovar (approved):

PR #4265, opened 2025-11-03
vine: tuning param return-recovery-tasks

JinZhou5042:
## Proposed Changes

A new paramer to let users choose whether to handle recovery tasks.

False is default. When set to true, the user explicitly opts to manage the completion recovery tasks.

Use case: We want to prune files created by recovery tasks. File pruning must be handled at the graph executor, which decides what files to delete when a task completes. The current implementation does not respect data produced by recovery tasks because they are implicitly skipped when complete.

## Merge Checklist

### DIFF
--- a/doc/manuals/taskvine/index.md
+++ b/doc/manuals/taskvine/index.md
@@ -2779,7 +2779,6 @@ change.
 | watch-library-logfiles | If 1, watch the output files produced by each of the library processes running on the remote workers, take
 them back the current logging directory. | 0 |
 
-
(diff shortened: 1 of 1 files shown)


btovar (changes requested):
Please update the taskvine manual, with a note that this is not expected to be used in regular applications.

btovar:
If I understand correctly, the recovery task return serves as an indicator that a file was lost and has been recreated. Is this what you want?

PR #4269, opened 2025-11-04
vine: vine_temp.c

JinZhou5042:
## Proposed Changes

This PR includes the following features:

- temp files related functions and logics are brought to a new file `vine_temp.c`, which controls file replication, redundancy cleanup, and disk load shifting.
- `temp_files_to_replicate` is turned into a priority queue, files with least replicas are considerred first to replicate.
- `vine_prune` will not trigger the removal of a record in `temp_files_to_replicate`, which makes the data pruning faster, the replication will fail silently if no available replicas.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -952,7 +952,7 @@ static void cleanup_worker_files(struct vine_manager *q, struct vine_worker_info
 			vine_file_replica_delete(removed_replica);
 		}
 		/* consider if this replica needs recovery because of worker removal */
-		if (q->immediate_recovery && f && f->type == VINE_TEMP && !vine_temp_exists_somewhere(q, f)) {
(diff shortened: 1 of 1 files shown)


btovar:
I see you included the cleanup of redundant replicas. We should find a way so that the redundant tasks are not created.

btovar:
Sorry, I misunderstood the source! Once you fix the conflicts I can merge this one.

JinZhou5042:
> I think it makes sense to factor out distinct bits of code out of the manager like this.
> 
> It seems to me that this module is on top of `vine_file_replica_table` and so I'm trying to think of a more descriptive name than `vine_temp`.
> 
> But I don't have a better idea, so I think we should go with this!

Sounds great! Just pushed an empty commit to trigger the rebuild as it somehow failed.

btovar (approved):

PR #4273, opened 2025-11-05
Pqueue priorities

btovar:
Use an array of priorities instead of a single priority per element. Priorities are compared in lexicographical order.

The idea is to be able to do:

```C
        double priority = VINE_PRIORITY_DEFAULT;
	if (t->type == VINE_TASK_TYPE_RECOVERY) {
		priority = VINE_PRIORITY_RECOVERY;
	} else if (t->result == VINE_RESULT_RESOURCE_EXHAUSTION) {
		priority = VINE_PRIORITY_EXHAUSTION;
	}

	priority_queue_push_varargs(q->ready_tasks, t, priority, t->priority);
```

where all `RECOVERY` come before all `EXHAUSTION`, etc.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_task.h
+++ b/taskvine/src/manager/vine_task.h
@@ -45,11 +45,11 @@ typedef enum {
 } vine_task_func_exec_mode_t;
 
 typedef enum {
-	VINE_PRIORITY_BLOCKED = -2 << 30,
+	VINE_PRIORITY_BLOCKED = -1000000,
 	VINE_PRIORITY_DEFAULT = 0,
-	VINE_PRIORITY_EXHAUSTION = 2 << 4,
-	VINE_PRIORITY_RECOVERY = 2 << 8,
-	VINE_PRIORITY_HIGHEST = 2 << 30,
+	VINE_PRIORITY_EXHAUSTION = 100,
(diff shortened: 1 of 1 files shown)


dthain (commented):
I appreciate the elegance of this approach, allowing for priority vectors of arbitrary size.  But I do wonder if the flexibility is worth the cost at the heart of the scheduler.   Does this materially change the performance of the data structure?  (due to extra mallocs). Would we be satisfied with simply having two values: a priority and a sub-priority?

JinZhou5042:
One concern I'm having is that the priority data structure is widely used in other scenarios, to which only one priority value would make sense. As multiple priority values only apply to task scheduling, would it be easier to store the two priority values in `struct task` and switch between them when needed?

btovar:
With only one priority, the only thing that changes is the call to priority_queue_create.

I think we still want the option for more than two, and it is likely that we will find situations where we want to have more control in the way tasks are ordered.  I'm going to change it so that the two (or maybe three) priority values are stored in the struct element, so that only one malloc per push is needed.

dthain:
That sounds good to me.

btovar:
No. With the taskid we know it's priorities and can look for it directly.

btovar:
Well directly I mean logn

PR #4281, opened 2025-11-12
vine: minumum blocked task queue implementation

JinZhou5042:
## Proposed Changes

This PR covers just the bare-bones structure for the `q->blocked_tasks`, excluding any of the other additional features.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager.c
+++ b/taskvine/src/manager/vine_manager.c
@@ -1533,7 +1533,7 @@ static int expire_waiting_tasks(struct vine_manager *q)
 
 	/* Only work through the queue up to iter_depth. */
 	int iter_count = 0;
-	int iter_depth = MIN(priority_queue_size(q->ready_tasks), q->attempt_schedule_depth);
+	int iter_depth = MIN(list_size(q->blocked_tasks), q->attempt_schedule_depth);
 
 	/* expire tasks after iteration */
 	struct list *tasks_to_expire = list_create();


taskvine/src/manager/vine_manager.c
```diff
@@ -3624,6 +3625,35 @@ int consider_task(struct vine_manager *q, struct vine_task *t)
...
+	if (list_size(q->blocked_tasks) == 0) {
+		return 0;
+	}
+
+	int unblocked_tasks = 0;
+	int tasks_considered = 0;
+	int tasks_to_consider = MIN(list_size(q->blocked_tasks), q->attempt_schedule_depth);
+
+	while (tasks_considered++ < tasks_to_consider) {
+		struct vine_task *t = list_pop_head(q->blocked_tasks);
```
btovar: Why not use the  `list_rotate` pattern instead?

taskvine/src/manager/vine_manager.c
```diff
@@ -3665,45 +3697,55 @@ static int send_one_task(struct vine_manager *q, int *tasks_ready_left_to_consid
 		struct vine_worker_info *w = vine_schedule_task_to_worker(q, t);
 		q->stats->time_scheduling += timestamp_get() - q->stats_measure->time_scheduling;
 
-		if (w) {
-			priority_queue_remove(q->ready_tasks, t_idx);
+		/* task is runnable but no worker is fit, demote it and consider later */
```
btovar: I don't think we want this as it impacts greatly priority. When resources become available, tasks with lower priority may be scheduled first because higher priority tasks are in the blocked list.

btovar (changes requested):

btovar:
Also there are some edits to code unrelated to the pr, please remove them as possible!

taskvine/src/manager/vine_manager.c
JinZhou5042: @btovar I used `list_pop_head` and `list_push_tail` in conjunction because they align well. Using `list_rotate` is equivalent to calling `list_pop_head` followed by `list_push_tail`, but would require an additional call to `list_pop_tail` when promoting a task.

taskvine/src/manager/vine_manager.c
```diff
@@ -3656,7 +3686,9 @@ static int send_one_task(struct vine_manager *q, int *tasks_ready_left_to_consid
 	{
 		*tasks_ready_left_to_consider -= 1;
 
+		/* this task is not runnable at all, demote it to the blocked queue */
 		if (!consider_task(q, t)) {
+			list_push_tail(q->blocked_tasks, t);
```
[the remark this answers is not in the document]
JinZhou5042: @btovar I think tasks that can't run for any reason should be blocked; it's unnecessary to keep them in the ready queue. In the next round, all blocked tasks that become eligible will be pushed back in order of priority, so we can not only simplify the demoting rule, but also ensure that their priority values are respected. What do you think?

JinZhou5042:
We need to return early in the `send_one_task` function when there are no committable workers, otherwise eligible tasks will be incorrectly blocked because the workers lack free cores.

JinZhou5042:
> Also there are some edits to code unrelated to the pr, please remove them as possible!

Could you point me to the places where the edits should be reverted?

JinZhou5042:
We will implement the idea of blocked queue in the upper-level DAG executor.

PR #4287, opened 2025-12-02
Skip list priority implementation

btovar:
This implementation has only one cursor. Let's see if it is enough as compared to work queue.

## Merge Checklist

### DIFF
--- a/dttools/src/skip_list.c
+++ b/dttools/src/skip_list.c
@@ -163,6 +163,7 @@ static struct skip_list_node *create_node(int level, void *data, double *priorit
 	struct skip_list_node *node = xxmalloc(sizeof(*node));
 	memset(node, 0, sizeof(*node));
 
+	// levels are 0-based, so we need to allocate one more than the level number
 	node->forward = xxmalloc((level + 1) * sizeof(struct skip_list_node *));
 	node->backward = xxmalloc((level + 1) * sizeof(struct skip_list_node *));
 	node->priority = xxmalloc(priority_size * sizeof(double));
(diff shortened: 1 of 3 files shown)


dttools/src/skip_list.h
```diff
@@ -0,0 +1,254 @@
...
+struct skip_list *skip_list_create(unsigned priority_size, double probability);
+
+/** Delete a skip list.
+ * The caller is responsible for removing all items before deleting.
+ * If the skip list is non-empty or there are living cursors, this call fails.
+ * @param sl The skip list to delete.
+ * @returns true if the skip list was deleted.
+ * @returns false if the skip list is non-empty or there are live cursors.
+ */
+bool skip_list_destroy(struct skip_list *sl);
```
dthain: `skip_list_delete` for consistency with cctools

dttools/src/skip_list.h
```diff
@@ -0,0 +1,254 @@
...
+
+/** Get the number of items in a skip list.
+ * @param sl The skip list to examine.
+ * @returns The number of items in the skip list.
+ */
+int skip_list_length(struct skip_list *sl);
```
dthain: `skip_list_size` for consistency

dthain (changes requested):
Overall, this is looking good, and I like how you have simplified several taskvine issues while maintaining basically the same interface to the list.  I have a couple of persnickety naming fixups, but no major concerns.

btovar:
@dthain Sounds good!

@JinZhou5042 It is so that we can have more than one cursor pointing to the same node. We do not want to delete a node to which a cursor is pointing to as that will cause a segfault. Additionally, this allows us to 'delete' the item a cursor is pointing to during an iteration, which simplifies the places where we have to call 'next'.

JinZhou5042:
@btovar Got it! I think this is a nice and very maintanable implementation, especially the design in `send_one_task` that allows us to retrieve ready tasks while evaluating their eligibility. It also makes it simple to expire tasks with missing inputs in the future PR.

PR #4304, opened 2025-12-09
vine: a couple of memory leaks

btovar:
I saw these leaks only with recovery tasks.

## Merge Checklist


dthain (commented):
Oh good catch!  But it seems to me that an insert of a duplicate key is a bigger problem than the leak of the table entry, it is also possibly the leak of the contained object too.

Do we need a notice() here to indicate the implied loss of an object at that key?  Or maybe check to see if it is the same object and key?

PR #4343, opened 2026-02-05
Vine: Waiting Tasks Consume Only Disk

dthain:
## Proposed Changes

Under the current accounting scheme,  tasks consume all resource types, even when they are in the WAITING_RETRIEVAL state.  This results in taskvine leaving cores/memory/gpus idle until it can clear out those waiting tasks.

This (experiment) changes the accounting of tasks, so that only RUNNING tasks consume cores/memory/gpus, and WAITING tasks only consume disk.  Accounting has been change to look at specific counters rather than `itable_size(w->current_tasks)` which captures both types.

The intended result is that we should see more tasks running and more overlap between task execution and result retrieval.

## Merge Checklist


btovar:
Solving conflicts.

btovar (approved):

PR #4391, opened 2026-04-11
implementation of libnvidia-ml.so library (minimal implementation) just for the GPU name and will progress from here to get GPU stats into Resource Monitoring

LaxminarayanaV7416:
## Proposed Changes

## Merge Checklist

- [V] `make test`       Run local tests prior to pushing.
- [V] `make format`     Format source code to comply with lint policies. Note that some lint errors can only be resolved manually (e.g., Python)
- [V] `make lint`       Run lint on source code prior to pushing.
- [V] Manual Update:     Update the manual to reflect user-visible changes.
- [V] Type Labels:       Select a github label for the type: bugfix, enhancement, etc.
- [V] Product Labels:    Select a github label for the product: TaskVine, Makeflow, etc.
- [V] PR RTM:            Mark your PR as ready to merge.

### DIFF
--- a/dttools/src/Makefile
+++ b/dttools/src/Makefile
@@ -124,6 +124,10 @@ SOURCES = \
 	username.c \
 	uuid.c \
 	xxmalloc.c \
+	gpu_commons.c \
+	nvidia_nvml_library.c \
+	gpu.c \
+
 
 HEADERS_PUBLIC = \
 	auth.h \
@@ -175,6 +179,9 @@ HEADERS_PUBLIC = \
 	timestamp.h \
 	unlink_recursive.h \
 	xxmalloc.h \
+	gpu_commons.h \
+	nvidia_nvml_library.h \
+	gpu.h \
 
 LIBRARIES = libdttools.a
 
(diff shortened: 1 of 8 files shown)


tphung3:
Please update the proposed changes section at the top. It is unclear what this PR is trying to achieve.

If this PR introduces a new feature, please include at least a test so we know the feature works.

btovar:
what is needed to move this pr forward?

dthain (changes requested):
Top level issue - this PR creates two distinct entry points for querying the GPU: the existing `gpu_info.h` and the new `gpu.h`.  This is going to create confusion because now the user has to decide which one to use, and how to deal with success of one and failure of the other.

I do like that your proposed interface has more of an open/close semantic to it, so that the caller gets into from the GPU once.

And the `nvidia_nvml_library.h` interface is fine as an internal object.

I suggest that you address this as follows.  Just have a single interface for querying all information about the GPU:

```
struct gpu_info {
  int count;
  const char *name;
  struct nvidia_nvml_library *nvml_library;
  ...
}:

info = gpu_info_open(...)
gpu_info_close(info)
```

PR #4427, opened 2026-06-23
vine: set output missing if file not found

JinZhou5042:
## Proposed Changes

In `vine_manager_get_output_files`, if the manager does not find a created `TEMP` output file of a task, the result of that task should be set to `VINE_RESULT_OUTPUT_MISSING`.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/vine_manager_get.c
+++ b/taskvine/src/manager/vine_manager_get.c
@@ -427,6 +427,7 @@ vine_result_code_t vine_manager_get_output_files(struct vine_manager *q, struct 
 				// if temp, check that we got a cache update message.
 				struct vine_file *f = hash_table_lookup(q->file_table, m->file->cached_name);
 				if (!f || f->state != VINE_FILE_STATE_CREATED) {
+					vine_task_set_result(t, VINE_RESULT_OUTPUT_MISSING);
 					result_single_file = VINE_APP_FAILURE;
 				}
 			} else {
--- a/taskvine/src/manager/vine_manager_get.c
+++ b/taskvine/src/manager/vine_manager_get.c
@@ -427,7 +427,6 @@ vine_result_code_t vine_manager_get_output_files(struct vine_manager *q, struct 
 				// if temp, check that we got a cache update message.
 				struct vine_file *f = hash_table_lookup(q->file_table, m->file->cached_name);
 				if (!f || f->state != VINE_FILE_STATE_CREATED) {
-					vine_task_set_result(t, VINE_RESULT_OUTPUT_MISSING);
 					result_single_file = VINE_APP_FAILURE;
 				}
 			} else {


taskvine/src/manager/vine_manager_get.c
```diff
@@ -427,6 +427,7 @@ vine_result_code_t vine_manager_get_output_files(struct vine_manager *q, struct
 				// if temp, check that we got a cache update message.
 				struct vine_file *f = hash_table_lookup(q->file_table, m->file->cached_name);
 				if (!f || f->state != VINE_FILE_STATE_CREATED) {
+					vine_task_set_result(t, VINE_RESULT_OUTPUT_MISSING);
```
btovar, line 430: I'm not sure this goes here? We set `VINE_RESULT_OUTPUT_MISSING` in `vine_manager_get.c`. I would not like to have several points in the code where that result is set. Further, it could be that we overwrite some other more specific result?

btovar (changes requested):

taskvine/src/manager/vine_manager_get.c
JinZhou5042: The case I was trying to handle is when a task has a `TEMP` output, exits successfully, but the manager somehow doesn't receive the cache update for that temp file. In that case the file state stays in `VINE_FILE_STATE_PENDING` and the task result stays in `VINE_RESULT_SUCCESS`. Do you think it is a problem or did I miss something?

btovar: Would that be a bug with the retrieval of files in `vine_manager_get_output_files`? I see in https://github.com/cooperative-computing-lab/cctools/blob/da88fd766bb04073fd7ec8a06f6443615cf2a0fd/taskvine/src/manager/vine_manager_get.c#L426 that we do check for temp creation?

btovar: That to say, I wouldn't expect the task to have `VINE_RESULT_SUCCESS`.

JinZhou5042: I agree with you. Any thoughts on how to handle this problem or this is not a big problem right now?

btovar: If it is true that the task is incorrectly marked as successful then please investigate why `vine_manager_get_output_files` is not working as it should, or why the files are not being marked correctly.

JinZhou5042: I looked into this a bit more, and for now I don't think this is causing an actual issue in practice, so I'd prefer to leave it as-is. If we later observe an actual issue then we can revisit this in more detail.

JinZhou5042:
We'll leave it as-is for now. If we later observe an actual issue then we can revisit this in more detail.

Issue #4168, opened 2025-05-28
vine: race condition in file recovery


JinZhou5042:
A task produces two types of outputs: its standard output and its declared output files. These outputs may be fetched in different orders, and a race condition can occur if the worker running the task crashes midway. The following sequence of events can lead to such inconsistency:

- Task `T` creates File `F` on worker `W` but hasn't notified the manager yet.

- `W` sends a `complete T` message to the manager. The manager marks `T` as completed, updates its state from `RUNNING` to `WAITING_RETRIEVAL`, and appends it to `q->waiting_retrieval_list`. The task's stdout will be fetched later.

- Next, a `cache-update` regarding `F` arrives, and we set the replica's state from `PENDING` to `READY`.

- Before the task's stdout is retrieved, worker `W` crashes. We call `vine_manager_remove_worker` to disconnect the worker. This function first iterates over `w->current_tasks` and sets their state to `READY`. It then deletes the existing replicas on that worker. If a replica is completely lost, a recovery task is submitted to regenerate it.

- As a result, both the original and the recovery task may run concurrently, each trying to produce the same output file. The file recovery logic is unaware that the original task has been rescheduled, and the manager cannot correlate that both tasks are producing the same data.

Here is the example sequenece from a log:

```
20:32:06.38  Task 238859: INITIAL -> READY 
20:32:11.87  Task 238859: READY -> RUNNING  
20:32:20.43  Task 238859: RUNNING -> WAITING_RETRIEVAL  
20:32:20.50  Task 238859: WAITING_RETRIEVAL -> READY  
20:32:20.51  Submitted recovery task 239138  
20:32:20.51  Task 239138: INITIAL -> READY  
20:32:20.91  Task 238859: READY -> RUNNING  
20:32:20.91  Task 239138: READY -> RUNNING  
20:32:41.89  Task 238859: RUNNING -> WAITING_RETRIEVAL  
20:32:41.89  Task 239138 RUNNING -> WAITING_RETRIEVAL  
20:32:41.99  Task 238859: WAITING_RETRIEVAL -> RETRIEVED  
20:32:42.00  Task 239138: WAITING_RETRIEVAL -> RETRIEVED  
20:32:59.00  Task 238859: RETRIEVED -> DONE  
20:32:59.00  Task 239138: RETRIEVED -> DONE
```

Possible solutions:

- Maintain two pointers to both the original task and the recovery task respectively, and check the status of the original task when considering the recovery task
-

[...]

PR #4428, opened 2026-06-23
vine: return recovery tasks

JinZhou5042:
## Proposed Changes

Goals:
- The executor wants the manager to return recovery tasks on completion so that it knows pruned files are present again, and then it can prune recovered files when their consumers finish.
- When taskvine returns a recovery task to the executor, it wants to know what the original task is so that it can target which node it belongs to. This is because each original task corresponds to a node in the graph, which recovery tasks can be dynamic.

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/taskvine.h
+++ b/taskvine/src/manager/taskvine.h
@@ -487,6 +487,12 @@ int vine_task_set_monitor_output(struct vine_task *t, const char *monitor_output
 
 const char *vine_task_get_state(struct vine_task *t);
 
+/** Return the original task id that a recovery task is restoring.
+@param t A task object.
+@return The source task id for a recovery task, or zero for non-recovery tasks.
+*/
+int vine_task_get_recovery_source_task_id(struct vine_task *t);
+
 /** Get the command line of the task.
 @param t A task object.
 @return The command line set by @ref vine_task_create.
@@ -1118,6 +1124,14 @@ int vine_enable_peer_transfers(struct vine_manager *m);
 
 /** Disable taskvine peer transfers to be scheduled by the manager **/
 int vine_disable_peer_transfers(struct vine_manager *m);
+
+/** Enable recovery tasks to be returned by vine_wait.
+By default, recovery tasks are handled internally by the manager. **/
+int vine_enable_return_recovery_tasks(struct vine_manager *m);
+
+/** Disable recovery tasks from being returned by vine_wait.
+Recovery tasks will be handled internally by the manager. **/
+int vine_disable_return_recovery_tasks(struct vine_manager *m);
 
 /** When enabled, resources to tasks in are assigned in proportion to the size
 of the worker. If a resource is specified (e.g. with @ref vine_task_set_cores),
--- a/taskvine/src/manager/taskvine.h
+++ b/taskvine/src/manager/taskvine.h
@@ -1125,13 +1125,13 @@ int vine_enable_peer_transfers(struct vine_manager *m);
 /** Disable taskvine peer transfers to be scheduled by the manager **/
 int vine_disable_peer_transfers(struct vine_manager *m);
 
-/** Enable recovery tasks to be returned by vine_wait.
+/** Enable external recovery handling by returning recovery tasks from vine_wait.
 By default, recovery tasks are handled internally by the manager. **/
-int vine_enable_recovery_tasks(struct vine_manager *m);
+int vine_enable_external_recovery_handling(struct vine_manager *m);
 
-/** Disable recovery tasks from being returned by vine_wait.
+/** Disable external recovery handling.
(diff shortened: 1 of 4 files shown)


dthain:
Let's talk about this a bit.  What does the DAG scheduler actually need to know?

1 - That a file was removed after creation?
2 - That a file was re-created after it was lost?
3 - That a recovery task has been created following a loss?
4 - That a recovery task has completed?

dthain:
And does the answer change if the DAG scheduler (and not the task scheduler) have responsibility for recovering from failures?

JinZhou5042:
> Let's talk about this a bit. What does the DAG scheduler actually need to know?
> 
> 1 - That a file was removed after creation? 2 - That a file was re-created after it was lost? 3 - That a recovery task has been created following a loss? 4 - That a recovery task has completed?

4. The executor needs to know the completion of a recovery task, so that it can mark the corresponding output file(s) as available, and then proceed with re-pruning.

JinZhou5042:
> And does the answer change if the DAG scheduler (and not the task scheduler) have responsibility for recovering from failures?

The answer changes to 1 if the DAG scheduler is responsible for recovery from failures.

dthain:
Sorry for mixing my questions.

Back to the first questions:

Are you saying that the DAG scheduler actually cares about the *files* and not the *task* per se?

JinZhou5042:
Yes. The DAG scheduler cares about the completion of a recovery task only because it tells the DAG scheduler that a file was lost/pruned before but now it is available again.

taskvine/src/manager/taskvine.h
```diff
@@ -1119,6 +1125,14 @@ int vine_enable_peer_transfers(struct vine_manager *m);
 /** Disable taskvine peer transfers to be scheduled by the manager **/
 int vine_disable_peer_transfers(struct vine_manager *m);
 
+/** Enable recovery tasks to be returned by vine_wait.
+By default, recovery tasks are handled internally by the manager. **/
+int vine_enable_return_recovery_tasks(struct vine_manager *m);
```
btovar: I'd suggest using:  `vine_enable_recovery_tasks` instead.  Less to explain what the "return" means. The behaviour is then to return a would be recovery task that would produced the files lost.

taskvine/src/manager/vine_manager.h
```diff
@@ -222,6 +222,7 @@ struct vine_manager {
 	int temp_replica_count;       /* Number of replicas per temp file */
 	int clean_redundant_replicas; /* If true, remove redundant replicas of temp files to save disk space. */
 	int shift_disk_load;          /* If true, shift storage burden to more available workers to minimize disk usage peaks. */
+	int return_recovery_tasks;    /* If true, recovery tasks are returned by vine_wait to the user. */
```
btovar: as above, `int use_recovery_tasks`?

taskvine/src/manager/vine_task.c
```diff
@@ -829,6 +829,27 @@ const char *vine_task_get_state(struct vine_task *t)
...
+
+	/*
+	Recovery tasks are copies of original tasks and keep the same output file objects.
+	The file records the original producer so the task does not need duplicate state.
+	*/
+	struct vine_mount *m;
+	LIST_ITERATE(t->output_mounts, m)
+	{
+		if (m && m->file && m->file->original_producer_task_id > 0) {
+			return m->file->original_producer_task_id;
```
btovar, line 846: Should the original producer task id be an attribute of the `struct vine_task` instead?

btovar (changes requested):

taskvine/src/manager/taskvine.h
JinZhou5042: Good point!

taskvine/src/manager/vine_manager.h
JinZhou5042: Will fix that!

taskvine/src/manager/vine_task.c
JinZhou5042: `original_producer_task_id` should be an attribute of `vine_file` and the original dicussion is in #4168. This variable is mainly used for tracking which original task produced a `TEMP` file, so it fits better in `vine_file` because recovery is driven by the lost file.

taskvine/src/manager/taskvine.h
dthain: Hmm, I think that name would be confusing too.  Recovery tasks are always *enabled* and used internally whenever temporary files are lost.  This API is changing the behavior so that recovery tasks are given back to the caller rather than consumed internally.  Which is (currently) a fringe use case.  I think it's better to have a forbidding name for this feature...

JinZhou5042: What about `vine_enable_external_recovery_handling()`?

PR #4431, opened 2026-06-23
vine: release random worker api

JinZhou5042:
## Proposed Changes

Provide an API for the DAG scheduler that releases a random worker

## Merge Checklist

### DIFF
--- a/taskvine/src/manager/taskvine.h
+++ b/taskvine/src/manager/taskvine.h
@@ -1172,6 +1172,13 @@ void vine_unblock_host(struct vine_manager *m, const char *hostname);
 */
 void vine_unblock_all(struct vine_manager *m);
 
+/** Release a random worker for failure-injection tests.
+This is a testing support hook, not a normal manager control operation.
+@param m A manager object.
+@return Non-zero if a worker was released.
+*/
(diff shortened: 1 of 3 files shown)


btovar:
Should this be really part of the public API, or something that should be implemented in your testing scripts?

JinZhou5042:
We need this to be part of the public API because the DAG executor may want to release workers based on the current graph completion progress. For example, it may release 50 workers during the run, which is 1 worker eviction at every 2% of workflow completion progress.

btovar:
In such cases releasing a random worker may be counterproductive, right? Wouldn't you want to release a worker you know it does not have any useful files? If this is for testing I think it should not be here, and if it is for policy then you need something more targeted to workers you want to release?

JinZhou5042:
Oh yeah, I forgot to mention that this is just for testing purposes. Some experiments I'm doing requires to release workers randomly and evaluate the system's resislience and fault-tolerance. Across different runs the number of evicted workers should be a fixed number so the results are comparable. The way I achieve that is to let the DAG executor release workers based on how many tasks in the graph have completed, and this matters because only the DAG executor knows about the completion progress while the task scheduler itself does not.

Issue #3827, opened 2024-05-09
vine: taskvine manager support for prometheus instrumentation


btovar:
Prometheus works by reading statistics served via http. A taskvine manager could serve statistics that are usually reported to the catalog. This would allow using autoscalers for taskvine workers in k8s.

PR #4433, opened 2026-06-24
vine: add prometheus instrumentation to python bindings.

btovar:
For #3827, coffea-casa.

## Merge Checklist

### DIFF
--- a/doc/manuals/taskvine/index.md
+++ b/doc/manuals/taskvine/index.md
@@ -1464,6 +1464,9 @@ to make a progress bar or other user-visible information:
     printf("%d\n", stats->workers_connected);
     ```
 
+In Python, you can also expose these statistics to a Prometheus server for
+external monitoring. See [Prometheus Metrics](#prometheus-metrics).
+
 ## Python Programming Models
 
 When writing a manager in Python, you have access to several types of tasks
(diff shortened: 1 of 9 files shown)


doc/manuals/taskvine/index.md
```diff
@@ -2569,6 +2572,48 @@ produces the following graphs:
...
+
+with vine.Manager(port=0, prometheus=True) as m:
+    print(f"metrics at http://127.0.0.1:{m.prometheus_port}/metrics")
+    ...
+```
+
+This feature requires the optional `prometheus_client` package:
+
+```sh
+pip install prometheus_client
```
dthain: A slight hesitation about extra manual installs.  Another potential hiccup for the user, and also a dependency that we (probably) won't test since it doesn't come along for the ride in testing.

Is Prometheus available via Conda?
How much does the package weigh?
Does it make sense to make it a default dependency?

doc/manuals/taskvine/index.md
```diff
@@ -2569,6 +2572,48 @@ produces the following graphs:
...
+```
+
+If `prometheus_client` is not installed, manager creation fails with an
+`ImportError` when metrics are enabled.
+
+Each scrape returns gauge metrics derived from the manager status API. Manager
+statistics are exported as `vine_<field>` (for example, `vine_tasks_waiting`
```
dthain: Does it make sense to reuse the status report that the manager is generating for catalog updates?
Or is that one too heavyweight?

taskvine/src/bindings/python3/ndcctools/taskvine/manager.py
```diff
@@ -74,6 +74,8 @@ class Manager(object):
     #                   If not given, then TSL is not activated. If True, a self-signed temporary key and cert are generated.
     # @param init_fn    Function applied to the newly created manager at initialization.
     # @param status_display_interval Number of seconds between updates to the jupyter status display. None, or less than 1 disables it.
+    # @param prometheus If True, serve Prometheus metrics on port 9090.
```
dthain: Mixing boolean and integer in the same parameter seems likely to cause headaches down the road.

Let's make it easy on the AI and call this `prometheus_port` and just make it an integer, such that zero means "any port" and 9090 can be chosen explicitly.

There is no avoiding the fact that multiple running taskvine instances will require some port planning/management by the end user...

dthain (changes requested):
Sorry, I didn't see this one earlier.  Integrating prometheus will add a lot of value.  See a few quibbles below:

doc/manuals/taskvine/index.md
btovar: If you mean the report as is, we cannot. Prometheus expects this type of gauges to be single values in a time series. Both the catalog status report and Prometheus here are using the same status function (vine_status), only that they present differently the information.

dthain: Ok, makes sense.

